PR #3, opened 2019-12-26
This PR adds initial improvements that allows developer easier to start

sleshchenko:
### What does this PR do?
This PR adds initial improvements that allow developer easier to start:
1. There is a bug in fsnotify transitive dependency that breaks Go Dep because of the empty versions list. As a workaround, direct dependency on the source archive should be added.
2. Latest plugin broker has additional methods that were not implemented in workspace CRD before, so it's fixed.
3. WorkspaceCRD controller requires some scripts that operator sdk generate during initializing, now they are committed.
4. For some reason, dockerimage build fails on Ubuntu when bazel is installer with yum, so, alternative implementation way is used - `dnf` works on Ubuntu and Fedora as well.
5. It was quite difficult to understand where I need to take a look to understand overall controller architecture, I believe it should be easier to navigate and learn overall architecture with separate packages.
6. Added initial instructions on how to run and test Workspace CRD controller. It also includes fixes for default values.
7. Format the code with `gofmt -w ./..`

It was done during https://github.com/eclipse/che/issues/15522


build/bin/user_setup
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -x
+
+# ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
+mkdir -p ${HOME}
+chown ${USER_UID}:0 ${HOME}
+chmod ug+rwx ${HOME}
```
amisevsk: Modifying the user-level permissions here is kind of a no-op, since this happens at build time and user ID is set at runtime. For clarity, I'd suggest we only do a `chgrp -R 0 /home`. 

The full command we use for the arbitrary user ID patch is [here](https://github.com/eclipse/che-devfile-registry/blob/master/arbitrary-users-patch/Dockerfile#L6) and I think it covers all our bases.

build/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM registry.access.redhat.com/ubi8/ubi
-RUN curl https://copr.fedorainfracloud.org/coprs/vbatts/bazel/repo/epel-7/vbatts-bazel-epel-7.repo > /etc/yum.repos.d/bazel.repo && \
-    yum install -y bazel && \
-    yum clean all
+
+RUN dnf install -y dnf-plugins-core && \
+       dnf copr -y enable vbatts/bazel && \
+       dnf install -y bazel
```
amisevsk: What's the effect on image size with doing it in this way? Also, should we `dnf clean all` as well?

deploy/controller_config.yaml
```diff
@@ -2,9 +2,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: che-workspace-crd-controller
+  namespace: operators
 data:
-  ingress.global.domain: 192.168.39.101.nip.io
-  plugin.registry: http://che-plugin-registry-crd-poc.192.168.39.101.nip.io/v3
+  ingress.global.domain: 192.168.99.100.nip.io
```
amisevsk: Looks like an accidental change?

pkg/controller/workspace/config/config.go
```diff
@@ -1,6 +1,6 @@
 // MY LICENSEdep
```
amisevsk: I know this isn't part of the changes in this PR, but this looks strange.

pkg/controller/workspace/log/log.go
```diff
@@ -0,0 +1,5 @@
+package workspace
+
+import logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+var Log = logf.Log.WithName("controller_workspace")
```
amisevsk: Trailing newline

pkg/controller/workspace/k8s/utilities.go
```diff
@@ -0,0 +1,38 @@
...
+	"strings"
+
+	workspaceApi "github.com/che-incubator/che-workspace-crd-operator/pkg/apis/workspace/v1alpha1"
+	. "github.com/che-incubator/che-workspace-crd-operator/pkg/controller/workspace/config"
+	. "github.com/che-incubator/che-workspace-crd-operator/pkg/controller/workspace/model"
+)
+
+func join(sep string, parts ...string) string {
+	return strings.Join(parts, sep)
+}
```
amisevsk: The purpose of this function isn't clear to me.

pkg/controller/workspace/status.go
```diff
@@ -35,8 +35,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	//	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	//	logf "sigs.k8s.io/controller-runtime/pkg/runtime/Log"
```
amisevsk: Find/replace accidental change?

pkg/controller/workspace/utils/utilities.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"strings"
+	"k8s.io/client-go/discovery"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func Join(sep string, parts ...string) string {
+	return strings.Join(parts, sep)
+}
```
amisevsk: Why do we need this function?

build/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM registry.access.redhat.com/ubi8/ubi
-RUN curl https://copr.fedorainfracloud.org/coprs/vbatts/bazel/repo/epel-7/vbatts-bazel-epel-7.repo > /etc/yum.repos.d/bazel.repo && \
```
davidfestal: In fact I assume that the bazel installation could be removed or commented here.
It was mainly when I wanted to experiment the on-the-fly creation of docker images with all the plugin artifacts injected inside.
But at this time I didn't have time to finish this experiment, and it's not currently prioritized. 
So this bazel-related code should probably be moved to a branch.

build/bin/entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+# This is documented here:
+# https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines
+
+if ! whoami &>/dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-devworkspace-api}:x:$(id -u):$(id -g):${USER_NAME:-devworkspace-api} user:${HOME}:/sbin/nologin" >> /etc/passwd
```
davidfestal: The default user name was `che-workspace-crd-operator` when it was initially generated (the file I forgot to commit).

`devworkspace-api` is the name of the package of the new Workspace Spec API.

build/bin/user_setup
davidfestal: This file is bootstrapped by the `operator-sdk`, and I'm not sure it would be a good idea to change it.

davidfestal: > user ID is set at runtime

The user ID referred to here is the one explicitly defined in the Dockerfile afaict.

deploy/controller_config.yaml
davidfestal: Probably a change made to have it running locally :-)

deploy/registry/local/deployment.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
   labels:
     app: che-plugin-registry
   name: che-plugin-registry
+  namespace: operators
```
davidfestal: Why do you force installing the plugin registry in the `operators` namespace ?

Afair this namespace doesn't exist on OpenShift.

deploy/registry/local/ingress.yaml
```diff
@@ -11,11 +11,12 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: che-plugin-registry
+  namespace: operators
```
davidfestal: Why do you force installing the plugin registry in the `operators` namespace ?

Afair this namespace doesn't exist on OpenShift.

deploy/registry/local/service.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
   labels:
     app: che-plugin-registry
   name: che-plugin-registry
+  namespace: operators
```
davidfestal: Why do you force installing the plugin registry in the `operators` namespace ?

Afair this namespace doesn't exist on OpenShift.

pkg/controller/workspace/config/config.go
davidfestal: This was probably created when the project files were initially bootstrapped by the `operator-sdk`. Afair there was an option to provide a license header, which I didn't do at first, which produced this sort-of default license comment.

We should probably just choose a license and add the license comments manually in files.

pkg/controller/workspace/k8s/utilities.go
davidfestal: Afair the purpose was to make string joining easier.
Moving the separator as first argument allows using variadic params for the string parts to join, and increases readability.

pkg/controller/workspace/k8s/convertToLowLevelObjects.go
```diff
@@ -1,4 +1,4 @@
-package workspace
+package k8s
```
davidfestal: I understand the idea of moving this one package down. But I don't really understand the choice of the `k8s` package name. Seems a bit confusing to me.

sleshchenko: Agree, it's just the first word that came to my mind, the package convert Che specific objects to K8s native. I will think more and try to find more suitable word here.

build/Dockerfile
davidfestal: It might not even be necessary to install bazel. See my comment here: https://github.com/che-incubator/che-workspace-crd-operator/pull/3/files#r363647973

build/bin/entrypoint
davidfestal: I assume it might be put here the same user name as the one used in the Dockerfile

README.md
```diff
@@ -1 +1,40 @@
...
+
+```
+dep ensure
+```
+
+Note that dep ensure doesn't automatically change Gopkg.toml which contains dependencies constraints. So, when a dependency is introduced or changed it should be reflected in Gopkg.toml.
+
+### Build docker image
+
+Since it uses Go Dep, you need operator-sdk 0.12.0 where Go Modules is not a default vendor.
```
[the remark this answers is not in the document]
sleshchenko: thanks, fixed

build/bin/entrypoint
sleshchenko: changed `devworkspace-api` to `che-workspace-crd-operator`

deploy/controller_config.yaml
sleshchenko: > Probably a change made to have it running locally :-)

Exactly. And `192.168.39.101` seems to be some generated value, I assume that kvm was used as VMDriver, while `192.168.99.100` should work for every in most cases when VirtualBox VMDriver is used.
So, I would leave `192.168.99.100` but I'm OK with reverting these changes as well.

P.S. I plan to add a command to README that would replace this value with real `minikube ip`

pkg/controller/workspace/k8s/utilities.go
sleshchenko: Well, it's used for message constructing only. I thought about replacing it with `fmt.Sprintf`, like:
```go
Log.Error(err, Join("", "Cannot find the '", ConfigMapReference.Name, "' ConfigMap in namespace '", ConfigMapReference.Namespace, "'"))
```
vs
```go
Log.Error(err, fmt.Sprintf("Cannot find the '%s' ConfigMap in namespace '%s'", ConfigMapReference.Name, ConfigMapReference.Namespace))
```
WDYT?

pkg/controller/workspace/status.go
sleshchenko: Good catch. I believe such commented import can be removed now.

pkg/controller/workspace/utils/utilities.go
sleshchenko: I think we don't. See https://github.com/che-incubator/che-workspace-crd-operator/pull/3/files#r363515566

deploy/registry/local/service.yaml
sleshchenko: it's just the same namespace where DevWorkspace Controller is deployed.

Do you think it's better to deploy Plugin Registry in it's own namespace or default one?
Also, where do you think DevWorkspace Controller should be deployed? We can precreate some namespace like `devworkspace-controller` for controller and plugin registry, WDYT?

P.S. I considered an option where user specify namespace for `kubectl apply -f ... -n ` BUT seems it's not possible since we need to know namespace before executing that command  for ClusterRoleBinding

build/Dockerfile
sleshchenko: bazel installation is removed now.

deploy/registry/local/ingress.yaml
sleshchenko: Let's discuss alternatives in https://github.com/che-incubator/che-workspace-crd-operator/pull/3#discussion_r363653290

deploy/registry/local/deployment.yaml
sleshchenko: Let's discuss alternatives in the following thread https://github.com/che-incubator/che-workspace-crd-operator/pull/3#discussion_r363653290

pkg/controller/workspace/k8s/convertToLowLevelObjects.go
sleshchenko: do you think `component` would be good enough name for this package?

deploy/controller_config.yaml
amisevsk: I'm +1 for `192.168.99.100` if we're including an ip address here, though I use kvm and it doesn't allow setting an address.

build/bin/user_setup
amisevsk: Do we know of any side effects in case the Dockerfile defines e.g. user 0?

pkg/controller/workspace/config/config.go
amisevsk: :+1: I think some other files already have a license header.

pkg/controller/workspace/k8s/convertToLowLevelObjects.go
amisevsk: I like `component`, or maybe `infrastructure`/`cluster`?

build/bin/user_setup
davidfestal: Don't know, but Dockerfile and scripts in the `/bin` are generated together by the `operator-sdk` in a way that is expected to be consistent afaict

pkg/controller/workspace/k8s/convertToLowLevelObjects.go
davidfestal: I like `component` the best I think. Or even `components` (with a `s`) ?

deploy/registry/local/service.yaml
davidfestal: > it's just the same namespace where DevWorkspace Controller is deployed.

Only in the raw `Kubernetes` case no ?

> Do you think it's better to deploy Plugin Registry in it's own namespace or default one?

Not necessarily

> Also, where do you think DevWorkspace Controller should be deployed? We can precreate some namespace like devworkspace-controller for controller and plugin registry, WDYT?

In fact when you deploy manually with raw yamls, you can depoy in any namespace, and this will only watch CRs in this namespace.

But I had also started setting up OperatorHub files (CSV) so that it can be installed through OperatorHub in `AllNamespaces` mode (which should be the preferred way in the future I assume)

> P.S. I considered an option where user specify namespace for kubectl apply -f ... -n BUT seems it's not possible since we need to know namespace before executing that command for ClusterRoleBinding

In such a case we could provide a script file that deploys all the file withing the namespace provided as an argument no ?

pkg/controller/workspace/k8s/convertToLowLevelObjects.go
sleshchenko: Seems it's more common to use singular word in go package name, but I do not have strong opinion

pkg/controller/workspace/config/config.go
sleshchenko: Added Eclipse Public License - v 2.0 - the same as we have for Che Plugin Broker
https://github.com/che-incubator/che-workspace-crd-operator/pull/3/commits/33352e3188079cd86ec8071ed4da64664bdbb7d5

Let me know if someone has another suggestion

pkg/controller/workspace/k8s/utilities.go
sleshchenko: Well, with Logr that is used here we should not use Sprintf, since it's supposed to print structured logs 
https://github.com/go-logr/logr#why-structured-logging
And then it would look like 
```
Log.Error(err, "ConfigMap does not exist'", "namespace", ConfigMapReference.Namespace, "name", ConfigMapReference.Name)
```

@davidfestal @amisevsk I haven't make myself familiar with that approach but in parallel to do it, I would like ask your opinions - is it better to move to another logger tool and using structured logs is fine.

pkg/controller/workspace/utils/utilities.go
sleshchenko: Already removed

pkg/controller/workspace/k8s/utilities.go
amisevsk: I'm fine with either option, though from what I can tell we're not really making use of the structured aspect of `Logr` very effectively.

sleshchenko: I leave Logr but used `fmt.Sprintf` for time being.

deploy/registry/local/service.yaml
sleshchenko: > In such a case we could provide a script file that deploys all the file withing the namespace provided as an argument no ?

I would consider it as a further improvement.

build/bin/user_setup
sleshchenko: From https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines
```
RUN chgrp -R 0 /some/directory && \
    chmod -R g=u /some/directory
```
@amisevsk Seems doc suggests to add chmod as well as chgrp. I may don't understand it in details, could you provide complete suggestion we should apply? Or let's do it later as a separate PR.

README.md
```diff
@@ -1 +1,41 @@
...
+### Run controller locally
+1. `kubectl apply -f ./deploy/crds`
+2. Make sure right ingress domain is set in `./deploy/registry/local/ingress.yaml` and execute `kubectl apply -f ./deploy/registry/local`
+3. `operator-sdk up local --namespace <your namespace>`
+
+### Test run controller
+
+1. Take a look samples workspace configuration in `.samples` folder.
+2. Apply any of them by executing `kubectl apply -f ./samples/workspace_java_mysql.yaml -n crd-poc`
+3. As soon as workspace is stared you're able to get IDE url by executing `kubectl get workspace -n <namespace>`
```
l0rd: You call the namespace `crd-poc` above and then `<namespace>` here.

README.md
```diff
@@ -1 +1,41 @@
...
+
+```
+dep ensure
+```
+
+Note that dep ensure doesn't automatically change Gopkg.toml which contains dependencies constraints. So, when a dependency is introduced or changed it should be reflected in Gopkg.toml.
+
+### Build docker image
+
+Since it uses Go Dep, you need operator-sdk 0.10.0 where Go Modules is not a default vendor.
```
l0rd: You mean version "0.10.0 or later" or that versions other than 0.10.0 won't work? In either case you should be more specific. And I guess that golang version is important here as well.

README.md
```diff
@@ -1 +1,41 @@
...
+To build docker image run the following command in the project's root
+
+```
+operator-sdk build quay.io/che-incubator/che-workspace-crd-controller:7.1.0
+```
+
+### Run controller within K8s/OS cluster
+1. `kubectl apply -f ./deploy/crds`
+2. `kubectl create namespace devworkspace-controller`
+3. For K8s cluster you must to make sure that right domain is set in `./deploy/controller_config.yaml` and `./deploy/registry/local/ingress.yaml`
```
l0rd: Why are you specifying "For K8s cluster" here? This section is about K8s/OS, if this step is for K8s only (not OS) that's ok to be explicit but I think that's for both so I would only mention "Make sure that the right domain is set in....".

README.md
```diff
@@ -1 +1,41 @@
...
+4. `kubectl apply -f ./deploy/registry/local`
+5. You may want to modify ./deploy/operator.yaml and put your docker image and pull policy there. Then do: `kubectl apply -f ./deploy`
```
l0rd:
```suggestion
5. [optional] Modify ./deploy/operator.yaml and put your docker image and pull policy. 
6. `kubectl apply -f ./deploy`
```

README.md
davidfestal: In fact it should *not* be 0.12.0 since the support for `dep` was removed in favor of go modules.
And it was started / implemented with the `0.10.0` version. But maybe using the `0.11.0` can work.
This is why recommending using the `0.10.0` was chosen.

README.md
davidfestal: For the OpenShift case we don't need to provide this information. It is retrieved automatically at start of the controller.

README.md
sleshchenko: Good catch. Now `-n <namespace>`  is used above as well.

README.md
sleshchenko: Good catch. Actually, I did not try to run it on OpenShift cluster, maybe Route instead of Ingress is needed. For time being I've renamed section to `Run controller within K8s cluster` and then it's needed to test and fill in instructions for OpenShift Cluster

build/bin/user_setup
amisevsk: :+1: for separate PR or leaving as-is and trusting the operator-sdk bootstrap.

README.md
sleshchenko: Rephrased as `you need exact operator-sdk 0.10.0 version`

I believe a bit later we will move to modules and will be able to change it to `0.12.0 or later` and for time being it's OK.

README.md
```diff
@@ -1 +1,42 @@
...
+### Run controller locally
+1. `kubectl apply -f ./deploy/crds`
+2. Make sure right ingress domain is set in `./deploy/registry/local/ingress.yaml` and execute `kubectl apply -f ./deploy/registry/local`
+3. `operator-sdk up local --namespace <your namespace>`
+
+### Test run controller
+
+1. Take a look samples workspace configuration in `.samples` folder.
+2. Apply any of them by executing `kubectl apply -f ./samples/workspace_java_mysql.yaml -n <namespace>`
+3. As soon as workspace is stared you're able to get IDE url by executing `kubectl get workspace -n <namespace>`
```
amisevsk: Typo: star**t**ed

README.md
```diff
@@ -1 +1,42 @@
...
+6. `kubectl apply -f ./deploy`
+
+### Run controller locally
+1. `kubectl apply -f ./deploy/crds`
+2. Make sure right ingress domain is set in `./deploy/registry/local/ingress.yaml` and execute `kubectl apply -f ./deploy/registry/local`
+3. `operator-sdk up local --namespace <your namespace>`
+
+### Test run controller
+
+1. Take a look samples workspace configuration in `.samples` folder.
```
amisevsk: Typo: `./samples`

PR #5, opened 2020-01-15
Update operator framework to v0.12.0

sleshchenko:
### What does this PR do?

This PR includes moving workspace-crd-operator to Go Modules https://github.com/che-incubator/che-workspace-crd-operator/pull/4
And updating the operator framework to v0.12.0. Each of the commits should self-describe its purpose pretty well.

### How are changes tested?
The following scenario is tested on the specified platforms:
1. dockerimage is rebuilt
2. A newer controller is deployed
3. java-mysql workspace is created
4. Theia is successfully opened

- [x] minikube v1.6.2 (commit: 54f28ac5d3a815d1196cd5d57d707439ee4bb392)
- [x] minishift v1.34.1+c2ff9cb (oc v3.11.0+0cbc58b | kubernetes v1.11.0+d4cacc0)


go.mod
```diff
@@ -0,0 +1,49 @@
+module github.com/che-incubator/che-workspace-crd-operator
+
+go 1.12
```
amisevsk: Note: the current PR is incompatible with Go 1.13, and apparently setting the go version to 1.12 does nothing to ensure compatibility. To build the project, I had to add 
```
replace github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible => github.com/openshift/api v0.0.0-20190424152011-77b8897ec79a
```

(Note this is a no-op change in Go 1.12, as both defined versions resolve to an identical snapshot of the package).

### Background
Go 1.13 changes the requirements for semantic import versioning; See section "Version validation" in the [release doc](https://golang.org/doc/go1.13#modules). Attempting to build this PR with Go 1.13.5 gives the error:
```
go: github.com/openshift/api@v3.9.1-0.20190424152011-77b8897ec79a+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```
This dependency comes in from the tree
```
github.com/operator-framework/operator-sdk v0.12.1-0.20191112211508-82fc57de5e5b
  github.com/operator-framework/operator-lifecycle-manager v0.0.0-20190605231540-b8a4faf68e36
    github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible
```
which is also built off Go 1.12, and so does not encounter the issue (building any of those snapshots with Go 1.13 fails for the same reason). 

To resolve the issue, I added the line
```
replace github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible => github.com/openshift/api 77b8897ec79a
```
(as suggested in the doc) and ran `go mod tidy`.

go.mod
```diff
@@ -0,0 +1,49 @@
...
+	github.com/operator-framework/operator-sdk v0.12.1-0.20191112211508-82fc57de5e5b
+	github.com/spf13/pflag v1.0.3
+	golang.org/x/net v0.0.0-20190812203447-cdfb69ac37fc
+	k8s.io/api v0.0.0
+	k8s.io/apimachinery v0.0.0
+	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d
+	sigs.k8s.io/controller-runtime v0.3.0
+)
+
```
amisevsk:
```suggestion
replace github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible => github.com/openshift/api v0.0.0-20190424152011-77b8897ec79a
```
(see above)

pkg/apis/workspace/v1alpha1/workspaceexposure_types.go
```diff
@@ -55,8 +55,13 @@ const (
...
 	// Workspace Exposure status
-	Phase            WorkspaceExposurePhase       `json:"phase,omitempty"`
-	ExposedEndpoints map[string][]ExposedEndpoint `json:"exposedEndpoints,omitempty"`
+	Phase            WorkspaceExposurePhase         `json:"phase,omitempty"`
+	ExposedEndpoints map[string]ExposedEndpointList `json:"exposedEndpoints,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+type ExposedEndpointList struct {
+	Items []ExposedEndpoint `json:"items,omitempty"`
```
amisevsk: This seems a strange change to me, as the serialized map would look something like
```yaml
  status:
    exposedEndpoints:
      che-machine-exec:
        items:
        - name: che-machine-exec
          url: ws://serverb3e761b6331b4209-che-machine-exec-4444-devworkspace-controller.192.168.39.8.nip.io
          attributes: #...
      mysql: {}
      theia-ide:
        items:
        - name: theia
          url: http://serverb3e761b6331b4209-theia-ide-3100-devworkspace-controller.192.168.39.8.nip.io
          attributes: #...
```
with the extra nesting for `items`

sleshchenko: Agree. It was done because code-generator reports `map values must be named type`.
It's kind of quick build fix that I need to take a closer look at how we can get rid of it.
BTW it's non-official yet and we can review this part later. Maybe it would be better to rework it in the following way:
```
  exposedContainers:
    - name: che-machine-exec
      endpoints:
        - name: che-machine-exec
          url: ....
          attributes: #....
```

sleshchenko: Agree. It was done because code-generator reports `map values must be named type`.
It's kind of quick build fix that I need to take a closer look at how we can get rid of it.
BTW it's non-official yet and we can review this part later. Maybe it would be better to rework it in the following way:
```
  exposedContainers:
    - name: che-machine-exec
      endpoints:
        - name: che-machine-exec
          url: ....
          attributes: #....
```

go.mod
sleshchenko: > Note: the current PR is incompatible with Go 1.13,

It's kind of strange =) Since I built it on `go1.13.6`
Thanks for the good analyze. Will take a look closer

sleshchenko: Thanks for figuring it out. I wonder why it works for me on go1.13.6 without proposed changes, but it totally makes sense to apply proposed changes and be buildable on more go versions.

pkg/apis/workspace/v1alpha1/workspaceexposure_types.go
sleshchenko: Now it's replaced with `type ExposedEndpointList []ExposedEndpoint`

PR #6, opened 2020-01-29
One more code clean up

sleshchenko:
1. Remove outdated cachingIOUtils. It was needed when we reached requests limit to vscode market place.
2. Restructure packages to void camel-case in file names. It does not bring a lot of profit but still, it's a nice thing to do.
3. Rename WorkspaceExposure to WorkspaceRouting. David and Mario agreed that WorkspaceRouting is a better alternative. Since we don't have a better proposal - I'm applying it.
4. During deploying DevWorkspaceController we get a warning cause OperatorSource (which is OLM specific) is located in the same folder as k8s resources. To avoid it, it's moved to the OLM folder.

It's made in the scope of https://github.com/eclipse/che/issues/15783


pkg/controller/workspace/component/plugins.go
```diff
@@ -83,10 +82,10 @@ func setupPluginInitContainers(names WorkspaceProperties, podSpec *corev1.PodSpe
 
 		if len(def.pluginFQNs) > 0 {
 
-			// TODO: Voir comment le unified broker est défini dans le yaml
-			// et le définir de la même manière ici.
-			// Voir aussi comment ça se fait qu'on ne met pas de volume =>
-			//    => Log du côté operator pour voir ce qu'il y a dans les PluginFQNs
```
amisevsk: :smile:

PR #9, opened 2020-01-31
Rename package che_rest, fill out devfile spec

amisevsk:
### What does this PR do?
Follow up fixes from https://github.com/che-incubator/che-workspace-crd-operator/pull/8:

- Update package name `che_rest` to `server`, since using underscores in go package names is generally seen as poor form
- Add `apiVersion` and `attributes` to devfile spec.

I merged PR #8 to avoid later merge conflicts since it's a fairly large refactor.

### Is it tested? How?
Tested locally on minikube.


pkg/apis/workspace/v1alpha1/devfile.go
```diff
@@ -14,8 +14,15 @@ package v1alpha1
 
 // This schema describes the structure of the devfile object
 type DevfileSpec struct {
+	// Devfile API version
+	APIVersion string `json:"apiVersion,omitempty"`
```
sleshchenko: DevfileSpec is a part of Workspace objects that already contains APIVersion in its model.
I'm not sure how they should coexist, maybe APIVersion on Workspace level should be our way, but I'm OK to add APIVersion to Devfile since we already have it in Devfile model on Che Server.

amisevsk: I suppose I'm not sure of the connection between the Workspace APIVersion (a CRD specific thing) vs devfile APIVersion. If e.g. we were to update the devfile spec, it may be hard to manage the change in a backwards-compatible way since devfile APIVersion is not saved anywhere. If we decide for some reason to change the devfile spec to move commands under components, like
```yaml
components:
  - type: dockerimage
    commands:
      # ...
```
how would we reflect this change in the CRD? I can imagine the operator automatically converting old `v1.0.0` devfiles to the new `v1.1.0` spec, for example.

If this can be managed with the workspace's APIVersion, then I agree we don't need to include it.

sleshchenko: > how would we reflect this change in the CRD? 

I'm not sure at this stage.
Since we still in beta - let's merge it and investigate this topic more later.

PR #11, opened 2020-02-04
Refactor config

sleshchenko:
### What does this PR do?
This PR has a bit refactored config for WorkspaceController and some values which were previously hard-coded - now are configurable.
I guess it's not final look for Config but it's one more step forward.

https://github.com/eclipse/che/issues/15783

### Is it tested? How?
It's tested on minikube with Java Mysql sample


pkg/controller/workspace/config/config.go
```diff
@@ -47,6 +46,23 @@ const (
...
+	IngressGlobalDomain = "ingress.global.domain"
+
+	//WorkspacePVCName config property handles the PVC name that should be created and used for all workspaces within one kubernetes namespace
+	WorkspacePVCName             = "pvc.name"
+	WorkspacePVCStorageClassName = "pvc.storageclass.name"
+
+	UnifiedPluginBrokerImage = "che.workspace.plugin_broker.unified.image"
+	InitPluginBrokerImage    = "che.workspace.plugin_broker.init.image"
+)
+
```
amisevsk: Since these seem to be accessed via `Get` methods, it may make sense to make these not external.

PR #12, opened 2020-02-04
Move go compiling to dockerimage

sleshchenko:
### What does this PR do?
Move go compiling to dockerimage. Why it's a good thing to do:
1. It makes the compiling process independent from developer environment.
2. It's a first that to create CI for DevWorkspace controller and build nightly images and as PR check.

N/A

### Is it tested? How?
I've tested it on minikube with Java MySql sample
`sleshchenko/devworkspace-controller:new-build`


build/Dockerfile
```diff
@@ -1,12 +1,19 @@
+FROM golang:1.13.7-alpine3.11 as builder
+
+COPY . /che-workspace-crd-operator/
+WORKDIR /che-workspace-crd-operator/
+
+RUN CGO_ENABLED=0 GOOS=linux go build -o _output/bin/che-workspace-crd-operator cmd/manager/main.go
```
amisevsk: One detail here I'm not entirely sure about is that this would contradict the `operator-sdk build` command, since that command will
- Build the binary outside of the container using
    ```
    go build -o \
        /home/amisevsk/git/che-workspace-crd-operator/build/_output/bin/che-workspace-crd-operator \
        -gcflags all=-trimpath=/home/amisevsk/git \
        -asmflags all=-trimpath=/home/amisevsk/git \
        github.com/che-incubator/che-workspace-crd-operator/cmd/manager
    ```
- Run the docker build using the dockerfile here. 

Moving the build inside the docker container would make the first build by operator-sdk needless. This may not be a huge deal but we should ensure the tags used in the build make sense.

sleshchenko: > The new go build flag -trimpath removes all file system paths from the compiled executable, to improve build reproducibility.

from https://golang.org/doc/go1.13

So, I'm not sure how that influences the build but seems it makes sense to add it into our build as well

sleshchenko: > Moving the build inside the docker container would make the first build by operator-sdk needless. This may not be a huge deal but we should ensure the tags used in the build make sense.

Do you mean when you execute operator-sdk build?
If yes, then you're right, and in README.md we probably should specify that `docker build -t REPO:TAG -f ./build/Dockerfile .` command should be executed to rebuild dockerimage, there is not need to use operator-sdk anymore...
Actually, it's needed to learn more about operator-sdk build command, maybe it does some additional validation stuff...

PR #21, opened 2020-02-24
Fix operator-sdk generate openapi

amisevsk:
### What does this PR do?

Commit 7a3d45c somehow (?) causes the `operator-sdk generate openapi` command to fail with a nondescriptive error message. On @davidfestal's suggestion, this PR removes file `deploy/crds/workspace.che.eclipse.org_devfilespecs_crd.yaml`, which seems to be in the repo erroneously, and re-runs the operator-sdk `generate` commands.

### Is it tested? How?
Tested on minishift v1.34.2+83ebaab

Fixes https://github.com/eclipse/che/issues/16120


deploy/crds/workspace.che.eclipse.org_workspaceroutings_crd.yaml
```diff
@@ -115,6 +115,7 @@ spec:
               description: Workspace Routing status
               type: string
           type: object
+      type: object
```
sleshchenko: Please double-check if we should add this now
https://github.com/che-incubator/che-workspace-operator/commit/4992687247fd983df1a3253c8a3fe2105897f20a
https://github.com/eclipse/che-operator/pull/140

sleshchenko:
I use `operator-sdk v0.14.0` and it works for me, we need:
- describe in README.md that operator-sdk v0.12.0 must be use
**OR**
- prepare dockerimage with needed operator-sdk version and use it for code generating (probably as a separate issue, later)

amisevsk:
@sleshchenko Part of the problem is that I'm unclear on how the operator-sdk generate commands work:
- It works for me with v0.14 and v0.15, but not v0.13 (or v0.12).
- `generate openapi` is deprecated in favor of `generate crds` and using `openapi-gen` standalone starting in v0.13
- The upgrade instructions for operator-sdk include a step to regenerate crds in their update guide, which suggests that different versions of the sdk are not necessarily compatible (otherwise regenerating CRDs wouldn't be strictly necessary).

deploy/crds/workspace.che.eclipse.org_workspaceroutings_crd.yaml
davidfestal: I agree that's it's a painful one. As long as we plan to target OpenShift 3.11, and at the same time want to get the newest `operator-sdk` releases, we would have to fix this OS 3.11 incompatibility ourselves.
Maybe a bit like what has been done in the `che-operator` recently: https://github.com/eclipse/che-operator/pull/180/files#diff-8a85834afbbc2e0c3afa67168071e391R34

davidfestal:
@amisevsk @sleshchenko I would suggest that for now we would merge this PR to have generation working from the `v0.12.0` SDK release and upper, but **after manually fixing** the `type: object` line pointed by @sleshchenko in comment https://github.com/che-incubator/che-workspace-operator/pull/21#discussion_r383702900, to avoid introducing an incompatibility with OpenShift 3.11. 

And later on you would be able to do something similar to what I did in the [`devworkspace-api`](https://github.com/che-incubator/devworkspace-api#how-to-build) with the [`docker-run.sh`](https://github.com/che-incubator/devworkspace-api/blob/master/docker-run.sh#L21) to ensure a consistent and given `operator-sdk` version is used.

As for upgrading to the last operator sdk release, and using the new `generate crds` and `openapi-gen` tools, I think it should be done, but probably:
- in a distinct issue
- synchronously with the same changs in the `devworkspace-api` repo
- while checking whether it somehow breaks compatibility with Openshift 3.11 (ie Minishift), in which case the fixing any breaking change in generated files should be done in the script proposed [here](https://github.com/che-incubator/che-workspace-operator/pull/21#discussion_r384431977)

PR #22, opened 2020-02-28
Move openshift-oauth-proxy into main workspace deployment

amisevsk:
### What does this PR do?
At a high level, this PR modifies the controller to create artifacts related to the openshift-oauth-proxy inside the main workspace deployment instead of as free-standing deployment. This requires significant changes, however:
- Previously, the entirety of the oauth-proxy was provisioned by the workspacerouting controller. This meant that no coordination between routing objects and deployment objects was required. Additionally, the workspacerouting could avoid creating routes for services that will be exposed through the proxy. With these changes, some coordination is necessary between the two.
- We now have to create a serviceaccount for each workspace, since this service account requires annotations to support the proxy service. The relevant rolebindings also have to be updated to be more generic (matching any serviceaccount in their namespace). 
- There are more changes that I've forgotten since I wrote them :)

This is a draft PR as I'd like to clean up the code and smooth out some rough edges, but I wanted some feedback before continuing to plug away at it. The commit history is a bit of a winding road, so may not be very useful to review (as some commits introduce issues that are later resolved). I've tried to give descriptive commit messages to ease reasoning through the history. 

Commits 966bf67..3d72f37 refactor the oauth-proxy solver to separate out the provisioning of routing objects from deployment objects in order to make it possible to move generating the proxy into the main deployment. Commit 3d72f37 *should* represent a refactor with no real functional changes (i.e the workspace should be similar). Commit f2ca475 moves proxy creation into the workspace controller, and the following commits are cleanup and fixups. Commit 5848259 adds basic authentication to the oauth proxy, limiting workspace access to users who can exec into pods.

### Background
The requirements for setting up the oauth-proxy correctly, and having it work as expected, are:
- The ServiceAccount used for the pod containing the proxy container must be annotated with `serviceaccounts.openshift.io/oauth-redirectreference.<value>`. This means that the common `che-workspace` SA cannot be used, and a new SA must be created for each workspace. This reference must refer to the proxy route by name (linking the generation of the two)
- The proxy container has to be started with parameters for correct listen and proxy ports. This links container creation to service creation.
- The proxy route must point at the proxy service on the correct port. 
- The secret used for the `service.alpha.openshift.io/serving-cert-secret-name` annotation on the proxy service must match a secret mounted into the proxy container.
- Any misconfiguration in any of these objects means connecting to the workspace will fail with absolutely no description of what is wrong.
- Additionally, on the workspace controller side, the internal ComponentInstanceStatus must be updated to accomodate the proxy.

### Known issues/TODOs
- Coalesce name generation and common metadata into one place; there are a few places where names have to be hard-coded to ensure everything stays in sync (e.g. that the ServiceAccount and Route and Service all have the correct names)
- On minishift, I am unable to open a terminal due to errors connecting from the Theia side (self-signed cert):
    ```
    Extension-Host:Unable to create ws connection for task plugin. Cause: Error: self signed certificate in certificate chain
    ```
- ~~The route generated for the proxy can be too long for the URL field in the workspace status, resulting in part being cut off (e.g. `URL=s://server95b4730759e311ea-oauth-proxy-theia-4400-che-workspace-con.192.168.42.207.nip.io`)~~ Fixed
- The openshift-oauth-proxy has some limitations that have ugly workarounds:
  - Each proxy container can only listen on one port; this means we have to add another container for *each* secured port in the workspace
  - The single port limitation means that each secured endpoint will require a separate login; this is a significant issue we need to figure out
  - This limitation also means that adding the proxy to the che-machine-exec route breaks terminal access
- ~~General code cleanup; it's written quickly and messily.~~ Good enough for now.

https://github.com/eclipse/che/issues/15786

### Is it tested? How?
Tested workspaces `java-mysql.yaml` (no oauth), `java-mysql-oauth.yaml` (with oauth), and `cloud-shell.yaml` on minishift. 

TODO: Test on minikube


amisevsk:
@sleshchenko I'm marking this as "ready to review", mostly because I've stared at it for too long at this point so I have trouble seeing what's a real issue and what's not. There are a few TODOs in the additions that I don't have a good answer for right now.

amisevsk:
I've rebased this PR onto the current master to avoid conflicts. 

I've also added a commit https://github.com/che-incubator/che-workspace-operator/pull/22/commits/238bb09b0e51a38365ec5ec65d4982268ffaa617 which adds an option to disable webhooks added in #19 to enable the operator to be run locally, let me know what you think @sleshchenko

sleshchenko:
Is closed in the favor of https://github.com/che-incubator/che-workspace-operator/pull/36

PR #23, opened 2020-03-03
Add validating webhook that denies to exec for everybody except creator

sleshchenko:
### What does this PR do?
It adds a couple of webhooks handlers to make sure that only creator has exec access to workspace-related pod. Now we have:
1. MutatingWebhook for Workspaces which on:
  CREATE: creator annotation is provisioned, the current user id is used;
  UPDATE: make sure that creator annotation is not modified;
2. MutatingWebhook for Deployments&Pod with `che.workspace_id` label which on:
  CREATE: make sure that the creator annotation is provisioned.
  UPDATE: make sure that creator is not changed. It's recovered in case the new object does not contain it.
3. ValidatingWebhook for all execs which do nothing if referenced pod does not have workspace id label. Otherwise checks if the current user id is the same as creator specified in the referenced pod.

Once it's merged - it won't be possible to open CloudShell anymore since it's not possible to exec with SA token. It's why webhooks are disabled by default and can be enabled via configuration.
CloudShell will work once OpenShiftOAuthRoutingController will be able to provision `USE_BEARER_TOKEN` env var for cloud-shell editor.

https://github.com/eclipse/che/issues/16252

### Is it tested? How?
It's tested on CRC:
1. Log in as developer.
2. Create a workspace, try to exec into workspace-related pod.
3. Log in as kube:admin.
4. Try to exec into workspace-related pod. Expected: error.
5. Try to modify creator in deployment/pod. Expected: error
6. Try to remove creator in deployment/pod. Expected: the previous value if recovered.

### What is not addressed in this PR?
1. User/Admin is still able to modify/delete workspace_id label.
User/Admin is still able to modify `Deployment.spec.template.metadata`.
It's needed to deny changing workspace_id label and creator annotation as well.

2. Controller make sure that there is not owned by itself webhooks configurations are webhooks are disabled in confg map.


pkg/webhook/creator/mutate.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if updated != nil {
+		return returnAnnotatedUnstructured(req.Object.Raw, newObject, updated)
+	}
+
+	return admission.Allowed("new deployment has the same creator as old one")
+}
+
+func (a *CreatorChecker) checkAnnotations(old, new map[string]string, kind string) (map[string]string, string) {
```
amisevsk: Consider returning `(map[string]string, error)` instead, using `fmt.Errorf()`

pkg/webhook/creator/validate_exec.go
```diff
@@ -0,0 +1,95 @@
...
+	if req.Resource.Resource == "pods" && req.Operation == v1beta1.Connect {
+		return v.handleCreateExec(ctx, req)
+	}
+	return admission.Denied(fmt.Sprintf("This admission is not supposed to handle %s operation for %s. Let administrator to know about this issue", req.Operation, req.Kind))
+}
+
+func (v *WorkspaceExecValidator) handleCreateExec(ctx context.Context, req admission.Request) admission.Response {
+	c, err := controller.CreateClient()
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
```
amisevsk: This would be better explained by a 500 error, wouldn't it?

pkg/webhook/creator/validate_exec.go
```diff
@@ -0,0 +1,95 @@
...
+type WorkspaceExecValidator struct {
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+func (v *WorkspaceExecValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if req.Resource.Resource == "pods" && req.Operation == v1beta1.Connect {
+		return v.handleCreateExec(ctx, req)
+	}
+	return admission.Denied(fmt.Sprintf("This admission is not supposed to handle %s operation for %s. Let administrator to know about this issue", req.Operation, req.Kind))
```
amisevsk: When is this path triggered? Or is this just a sanity check?

sleshchenko: > Or is this just a sanity check?

I believe so ) Just to make sure that configuration and code work well together.

pkg/webhook/creator/mutate.go
sleshchenko: Actually I'm using `fmt.Sprintf` but not `fmt.Errorf()`. Putting error here - mean that we need to create an error object which then won't be used since `admission.Denied` requires string...
Leaving as-is and we'll change it later if needed.

PR #24, opened 2020-03-03
Make Workspace Controller generate TLS certs for webhook server

sleshchenko:
### What does this PR do?
This PR makes Workspace Controller generate TLS certs for webhook server

It's not working yet:
```
Error from server (InternalError): error when creating "./samples/cloud-shell.yaml": Internal error occurred: failed calling admission webhook "mutate-workspaces.che-workspace-controller.svc": Post https://workspace-controller.che-workspace-controller.svc:443/mutate-workspaces?timeout=30s: x509: certificate is valid for ..svc.cluster.local, not workspace-controller.che-workspace-controller.svc
```

TODO read secret/configmap on start and write content into file to avoid pod restarting when cert is generated.

### Is it tested? How?
It's not working yet


sleshchenko:
is not going to be finalized since it uses non-supported functionality from operator-sdk. See https://github.com/operator-framework/operator-sdk/issues/419#issuecomment-552682642

PR #27, opened 2020-03-20
Propagate access token by OpenShift OAuth proxy

sleshchenko:
### What does this PR do?
I've extracted it from my draft to make others able to try USE_BEARER_TOKEN feature implemented in https://github.com/eclipse/che-machine-exec/pull/85 before I merge webhooks which restrict exec access for owner only.

It's needed to try https://github.com/eclipse/che-machine-exec/pull/85

### Is it tested? How?
I've tested it on crc.
1. oc apply -f ./samples/cloud-shell-oauth.yaml
2. oc edit deployment WORKPSACE_DEPLOYMENT and provision the following env var:
```
- name: LOG_LEVEL
   value: debug
```
3. Wait until pod is affected and opened terminal, make sure that SA token is used
Screenshot_20200320_172613
4. oc edit deployment WORKPSACE_DEPLOYMENT and provision the following env var:
```
- name: USE_BEARER_TOKEN
   value: "true"
```
5. Wait until pod is affected and opened terminal, make sure that bearer token is used
Screenshot_20200320_172802

### DIFF
--- a/samples/cloud-shell-oauth.yaml
+++ b/samples/cloud-shell-oauth.yaml
@@ -4,6 +4,7 @@ metadata:
   name: cloud-shell
 spec:
   started: true
+  routingClass: openshift-oauth
   devfile:
     apiVersion: 0.0.1
     metadata:
--- a/deploy/exec-cluster-role.yaml.yaml
+++ b/deploy/exec-cluster-role.yaml.yaml
@@ -1,0 +1,18 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pods-exec
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - get
+      - list
+  - apiGroups:
+      - ""
+    resources:
+      - pods/exec
+    verbs:
+      - create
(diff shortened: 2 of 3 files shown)


samples/cloud-shell-oauth.yaml
```diff
@@ -0,0 +1,23 @@
...
+    components:
+      - alias: cloud-shell
+        type: cheEditor
+        id: eclipse/cloud-shell/nightly
+      - type: dockerimage
+        memoryLimit: 256Mi
+        alias: dev
+        image: 'quay.io/eclipse/che-sidecar-openshift-connector:0.1.2-2601509'
+        args: ["tail", "-f", "/dev/null"]
+        env:
```
amisevsk: Does it make sense to add `USE_BEARER_TOKEN` here? Even if `false` by default it's a clue that it's configurable.

sleshchenko: Well, it's environment variable of cheEditor but not target container...
So, it makes sense to add it but - If I'm not mistaken - workspace-controller does not provision env vars for cheEditor...
The value of this env var should be provisioned by WorkspaceRouter Controller according to routing class... So, not sure we should promote it on devfile level.

sleshchenko: Merging this PR, for `USE_BEARER_TOKEN` will create a separate PR if agree that it makes sense.

PR #28, opened 2020-03-24
Deny changing workspace id label

sleshchenko:
### What does this PR do?
This PR denies changing:
- workspace id label in pods/deployment metadata
- workspace id label/creator annotation in deployment.spec.template.metadata

In addition, this PR refactors webhook files and packages.

There is separated [commit](https://github.com/che-incubator/che-workspace-operator/pull/28/commits/1d6d73a8f2c41ab26b69b1045ced8da9a851874f) that classes hierarchy
            WebhookHandler:
ResourceValidator ResourceMutator 

Personally I like the look with it, let me know if you don't and we can discuss and revert it.

eclipse/che#16252

### Is it tested? How?
1. `kc apply -f ./samples/cloud-shell.yaml `
2. `kc get workspace -o=yaml | grep creator`
3. `kc edit deployment workspace ....` #tried to modify creator, workspace id in metadata and template, made sure that they are restored if omitted.
4. `kc edit pod workspace ....` #tried to modify creator, workspace id in metadata, made sure that they are restored if omitted.
5. `kc exec -ti workspace ... -c dev sh` # as admin and creator, admin is not able to attach


pkg/webhook/workspace/handler/exec.go
```diff
@@ -0,0 +1,61 @@
...
+
+	_, ok := p.Labels[model.WorkspaceIDLabel]
+	if !ok {
+		return admission.Allowed("It's not workspace related pod")
+	}
+
+	creator, ok := p.Annotations[model.WorkspaceCreatorAnnotation]
+	if !ok {
+		return admission.Denied("The workspace info is missing in the workspace-related pod")
+	}
```
amisevsk: The logic here is a bit strange: if `p.Annotations == nil`, we'll deny admission a few lines later. Instead, I would do 
```go
	_, ok := p.Labels[model.WorkspaceIDLabel]
	if !ok {
		return admission.Allowed("It's not workspace related pod")
	}
	// Note: a nil map is like an empty map for reading.
	if creator, ok := p.Annotations[model.WorkspaceCreatorAnnotation]; !ok {
		return admission.Denied("The workspace info is missing in the workspace-related pod")
	}
```

pkg/webhook/workspace/handler/handler.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"encoding/json"
+	"k8s.io/apimachinery/pkg/runtime"
+	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type WebhookHandler struct {
+	Client  client.Client
```
amisevsk: Is this `client.Client` used currently?

pkg/webhook/workspace/handler/metadata.go
```diff
@@ -0,0 +1,101 @@
...
+			return fmt.Errorf("'%s' label is missing", model.WorkspaceIDLabel)
+		}
+	}
+
+	return nil
+}
+
+func (h *WebhookHandler) mutateMetadataOnUpdate(oldMeta, newMeta *metav1.ObjectMeta) (bool, error) {
+	if oldMeta.Annotations == nil {
+		oldMeta.Annotations = map[string]string{}
```
amisevsk: Isn't `oldMeta.Annotations == nil` or `oldMeta.Labels == nil` an automatic failure? It's strange to create maps here only to check for a specific value in them later.

Alternatively, we could fallback to nil map ~= empty map when reading here.

pkg/webhook/workspace/handler/metadata.go
```diff
@@ -0,0 +1,101 @@
...
+
+func mutateAnnotationsOnUpdate(old, new map[string]string) (bool, error) {
+	oldCreator, found := old[model.WorkspaceCreatorAnnotation]
+	if !found {
+		return false, fmt.Errorf("'%s' annotation is required. Update Controller and restart your workspace", model.WorkspaceCreatorAnnotation)
+	}
+
+	newCreator, found := new[model.WorkspaceCreatorAnnotation]
+	if !found {
+		new[model.WorkspaceCreatorAnnotation] = oldCreator
```
amisevsk: Is this the correct logic? It's more permissive but harder to reason about -- if my update to the workspace removes the owner annotation, I'm not sure we should just add it and pretend nothing is wrong. I'm not 100% sure here though.

pkg/webhook/workspace/handler/workspace.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	oldCreator, found := oldWksp.Annotations[model.WorkspaceCreatorAnnotation]
+	if !found {
+		log.Info(fmt.Sprintf("WARN: Worskpace %s does not have workspace annotation. It happens only for old "+
+			"workspaces or when mutating webhook is not configured properly", newWksp.ObjectMeta.UID))
+		newWksp.Annotations[model.WorkspaceCreatorAnnotation] = req.UserInfo.UID
+		return h.returnPatched(req, newWksp)
```
amisevsk: Does this mean that an old workspace without the annotation can be taken over (e.g. if I take your old workspace and add my creator annotation?)

pkg/webhook/workspace/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"github.com/che-incubator/che-workspace-operator/pkg/webhook/workspace/handler"
+	"k8s.io/api/admission/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// ResourcesValidator validates execs process all exec requests and:
+// if related pod DOES NOT have workspace_id label - just skip it
+// if related pod DOES have workspace_id label - make sure that exec is requested by workspace workspace
```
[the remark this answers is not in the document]
sleshchenko: Sometimes refactoring feature trying to help too much :)

pkg/webhook/workspace/handler/metadata.go
sleshchenko: I implemented this logic by keeping in mind that user is able to do `kubectl apply -f ...`  where original file does not contain creator annotations.
But since we don't really want user to modify that deployments manually, I'm OK with making webhook for pod and deployment validating and just throwing error when creator or workspace id is missing, WDYT?

pkg/webhook/workspace/handler/workspace.go
sleshchenko: Good point!
But actually restart here mean recreate. Will throw error here!

pkg/webhook/workspace/handler/metadata.go
amisevsk: Ah we definitely need to support `kubectl apply -f`. In that scenario though, does the `old` workspace not have the creator annotation? I figure it should be set upon creation, so even if I `apply` a new file over it, the old annotation will always be there. 

If `oldWksp` does not have an annotation, I can claim your workspace by grabbing your workspace yaml from the cluster, making a no-op modification, and applying it.

pkg/webhook/workspace/validate.go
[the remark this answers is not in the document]
amisevsk: Ah yeah I've had that happen too; it's *very* strange when it does.

pkg/webhook/workspace/handler/metadata.go
amisevsk: Ah you're right, `kubectl apply -f` is probably a case we need to support. Ignore this comment :)

amisevsk: Ah you're right, `kubectl apply -f` is probably a case we need to support. Ignore this comment :)

pkg/webhook/workspace/config.go
```diff
@@ -94,12 +93,12 @@ func SetUp(webhookServer *webhook.Server, ctx context.Context) error {
...
 		}
-		log.Info("Updated creator validating webhook configuration")
+		log.Info("Updated workspace validating webhook configuration")
 	} else {
-		log.Info("Created creator validating webhook configuration")
+		log.Info("Created workspace validating webhook configuration")
 	}
 
-	webhookServer.Register(validateWebhookPath, &webhook.Admission{Handler: &WorkspaceExecValidator{}})
+	webhookServer.Register(validateWebhookPath, &webhook.Admission{Handler: NewResourcesValidator()})
```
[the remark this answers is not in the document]
sleshchenko: Changed `NewWorkspaceResourcesMutator` -> `NewResourcesMutator` to be consist with type name.

pkg/webhook/workspace/mutate.go
```diff
@@ -0,0 +1,80 @@
...
+			case handler.V1PodKind:
+				return m.MutatePodOnUpdate(ctx, req)
+			case handler.AppsV1DeploymentKind:
+				return m.MutateDeploymentOnUpdate(ctx, req)
+			}
+		}
+	}
+	// Do not allow operation if the corresponding handler is not found
+	// It indicates that the webhooks configuration is not a valid or incompatible with this version of controller
+	return admission.Denied(fmt.Sprintf("This admission is not supposed to handle %s operation for %s. Let administrator to know about this issue", req.Operation, req.Kind))
```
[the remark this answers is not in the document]
sleshchenko: applied

pkg/webhook/workspace/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	return &ResourcesValidator{&handler.WebhookHandler{}}
+}
+
+func (v *ResourcesValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if req.Kind == handler.V1PodExecOptionKind && req.Operation == v1beta1.Connect {
+		return v.ValidateExecOnConnect(ctx, req)
+	}
+	// Do not allow operation if the corresponding handler is not found
+	// It indicates that the webhooks configuration is not a valid or incompatible with this version of controller
+	return admission.Denied(fmt.Sprintf("This admission is not supposed to handle %s operation for %s. Let administrator to know about this issue", req.Operation, req.Kind))
```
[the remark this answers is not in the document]
sleshchenko: applied

pkg/webhook/workspace/handler/exec.go
sleshchenko:
```go
	// Note: a nil map is like an empty map for reading.
```
You're right. I missed it. Removed `p.Annotations == nil` check

pkg/webhook/workspace/handler/handler.go
sleshchenko: Good catch! Now it's used, I don't create client in exec.go but reuse this one.

pkg/webhook/workspace/handler/metadata.go
sleshchenko: You're right, I missed that nil map in go behaves as an empty map in case of reading. So, it makes sense to initialize only newLabels, and newAnnotations since they may be modified.

pkg/webhook/workspace/handler/exec.go
```diff
@@ -0,0 +1,61 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+var V1PodExecOptionKind = metav1.GroupVersionKind{Kind: "PodExecOptions", Group: "", Version: "v1"}
+
+func (h *WebhookHandler) ValidateExecOnConnect(ctx context.Context, req admission.Request) admission.Response {
+	c, err := controller.CreateClient()
```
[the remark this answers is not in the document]
sleshchenko: Now I use client provided by Webhook server instead of creating new one.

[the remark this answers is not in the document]
sleshchenko: But It makes sense to revise all places we create client and maybe use cached one. But it's out of scope of this PR.

sleshchenko:
retested after changes are applied

PR #30, opened 2020-03-27
Fix workspaces with OpenShift OAuth and webhooks enabled

sleshchenko:
### What does this PR do?
Not every plugin needs bearer token, so it's not the best solution but I decided to create such a quick fix since it will be replaced with new controller architecture.

https://github.com/eclipse/che/issues/16252

### Is it tested? How?
1. Create cloud-shell-oauth workspace as developer
2. Try to open it with kube:admin
Screenshot_20200327_143249
3. Try to open it with developer
Screenshot_20200327_143330

I've tested the use case without webhooks enabled as well.

### DIFF
--- a/pkg/controller/workspace/component/plugins_utils.go
+++ b/pkg/controller/workspace/component/plugins_utils.go
@@ -155,6 +155,14 @@ func convertContainers(pluginContainers []brokerModel.Container, wkspCtx model.W
 		if err != nil {
 			return nil, err
 		}
+
+		if config.ControllerCfg.GetWebhooksEnabled() == "true" {
+			container.Env = append(container.Env, corev1.EnvVar{
+				Name: "USE_BEARER_TOKEN",
+				Value: "true",
+			})
+		}
+
 		containers = append(containers, container)
 	}
 	return containers, nil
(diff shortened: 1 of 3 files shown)


pkg/controller/workspace/component/plugins_utils.go
```diff
@@ -155,6 +155,18 @@ func convertContainers(pluginContainers []brokerModel.Container, wkspCtx model.W
 		if err != nil {
 			return nil, err
 		}
+
+		if config.ControllerCfg.GetWebhooksEnabled() == "true" {
+			if container.Env == nil {
+				container.Env = []corev1.EnvVar{}
+			}
```
amisevsk: You technically don't need to initialize a nil *slice* (unlike a map ;) ) before appending to it

pkg/controller/workspacerouting/openshift_oauth_proxy_solver.go
```diff
@@ -18,7 +18,7 @@ import (
 	"strconv"
 
 	workspacev1alpha1 "github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
-	. "github.com/che-incubator/che-workspace-operator/pkg/controller/workspace/config"
+	cfg "github.com/che-incubator/che-workspace-operator/pkg/controller/workspace/config"
```
amisevsk: :+1:

pkg/controller/workspace/component/plugins_utils.go
sleshchenko: :+1:

PR #32, opened 2020-03-30
Add an ability to configure default routing class on controller level

sleshchenko:
### What does this PR do?
Adds an ability to configure default routing class on controller level.
It allows using the same samples after enabling webhooks.

N/A

### Is it tested? How?
1. Configured Controller with the following config:
```yaml
apiVersion: v1
kind: ConfigMap
data:
  che.default_routing_class: openshift-oauth
  che.webhooks.enabled: "true"
  che.workspace.plugin_broker.artifacts.image: quay.io/eclipse/che-plugin-artifacts-broker:v3.1.0
  cherestapis.image.name: quay.io/dfestal/che-workspace-crd-rest-apis:newone
  ingress.global.domain: apps-crc.testing
  plugin.registry.url: http://che-plugin-registry-che-workspace-controller.apps-crc.testing
```
2. kubectl apply -f ./samples/cloud-shell.yaml
3. make sure that the only workspace creator has an ability to open CloudShell


amisevsk:
cc: @davidfestal as well.

PR #33, opened 2020-03-31
Make certs generate themselves from container

JPinkney:
### What does this PR do?
This PR makes the certs generate themselves inside of the controller by using a job. It will run the job only if needed (there haven't been any previous completed jobs). If the job creates the certs then it will update the deployment with the correct volume/volumeMounts

- [ ] There is an issue with running cloud shell that I'm currently investigating

### Is it tested? How?

tested on crc and started theia workspace, currently diagnosing why cloud shell isn't starting properly


JPinkney:
Closing in favour of following: https://docs.openshift.com/container-platform/4.1/authentication/certificates/service-serving-certificate.html

PR #35, opened 2020-03-31
Added in cert generation for openshift

JPinkney:
### What does this PR do?
This PR adds in cert generation for openshift using https://docs.openshift.com/container-platform/4.2/authentication/certificates/service-serving-certificate.html.

According to the documentation:
> The certificate and key are automatically replaced when they get close to expiration. The service CA certificate, which signs the service certificates, is only valid for one year after OpenShift Container Platform is installed.

### Is it tested? How?

Currently testing on crc with webhooks enabled and che.default_routing_class: "openshift-oauth".

Test image is: `jpinkney/che-workspace-controller:7.3.0`


deploy/os/controller.yaml
```diff
@@ -0,0 +1,72 @@
...
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: "che-workspace-operator"
+          ports:
+            - name: webhook-server
+              containerPort: 8443
+          volumeMounts:
+            - name: webhook-tls-certs
+              mountPath: /tmp/k8s-webhook-server/serving-certs
```
sleshchenko, line 35: Consider duplicating it into env var and then use the value here https://github.com/che-incubator/che-workspace-operator/blob/2dc7f2e252f7912548fa405dac1d24cfda30e9fb/pkg/webhook/server/server.go#L26
it would make a bit transparent that the controller is configured with this value.
@amisevsk WDYT?

deploy/os/controller.yaml
```diff
@@ -0,0 +1,72 @@
...
+              readOnly: true
+      volumes:
+        - name: webhook-tls-certs
+          projected:
+            sources:
+              - configMap:
+                  name: che-workspace-controller-secure-service
```
sleshchenko, line 42: Will deployment be triggered when OpenShift rotates sources?
if no - maybe it makes sense to watch this configmap and secret from WorkspaceController
Or maybe there is another mechanism on how to restart the controller in such case.

deploy/os/controller.yaml
amisevsk: I'm not sure I understand fully, but I feel like reading the value from a mounted file is the more canonical approach; that's what's done in e.g. openshift oauth proxy:
```
      volumeMounts:
        - mountPath: /etc/tls/private
          name: proxy-tls
...
  volumes:
    - name: proxy-tls
      secret:
        defaultMode: 420
        secretName: proxy-tls
```

deploy/os/controller.yaml
amisevsk: We have a mechanism in place for watching the main controller configmap (e.g. [here](https://github.com/amisevsk/che-workspace-crd-operator/blob/dde776de73562554987d33854d51bdf60f13226a/pkg/config/config.go#L141)) but I don't know if it's suitable if we need to *restart* the controller pod. I don't know that controllers typically roll out their own deployments, especially since this could look like an issue in the controller itself.

JPinkney: > Will deployment be triggered when OpenShift rotates sources?

The documentation as far as I can tell doesn't say what would happen when the sources are rotated. All it says is:
> The certificate and key are automatically replaced when they get close to expiration

I'll probably have to dig a little bit deeper into that code to see what would happen

deploy/os/controller.yaml
sleshchenko: I just meant here moving hard-coded path to an environment variable, and we would be able to have:
```
     env:
       - name: webhooks.certs.path 
         value: /tmp/k8s-webhook-server/serving-certs
...
      volumeMounts:
        - mountPath: /tmp/k8s-webhook-server/serving-certs
          name: webhook-tls
...
  volumes:
    - name: webhook-tls
      secret:
        secretName: webhook-tls
```
but it's not critical

deploy/os/controller.yaml
sleshchenko: Screenshot_20200402_123226

It makes me thinking that instead of mounting files we could:
- watch secret - at the starttime - initialize files on the filesystem with the current values.
  - once secret is modified - update the files content. Probably cert watcher will get updated values and restart webbhook server - it's needed to be checked.
- watch configmap - update WebhooksCfgs when certs are changed in configmap

But since by default certificates has 1-year expiration - we could postpone this task.

PR #36, opened 2020-04-01
Rework reconcile loop for workspace controller

amisevsk:
### What does this PR do?
Reworks the controller reconcile loop pretty thoroughly. For reviewing this PR, the diff will be nearly useless -- apart from a few copied sections (e.g. webhooks, some specs), the state of this PR represents a ground-up rewrite of most of the controller code.

A more detailed commit history is contained in repository https://github.com/amisevsk/che-workspace-operator-rework; see the [design doc](https://github.com/amisevsk/che-workspace-operator-rework/blob/master/docs/design.adoc) for a high-level overview.

The vast majority of the additions are because we store `PodAdditions` (a list of elements to be added to the workspace deployment, e.g. pods) in the status of subcomponents. This results in their full spec being included in the relevant CRDs. To make even looking at this PR slightly less daunting:
```
$ git diff master --stat pkg/
82 files changed, 4405 insertions(+), 4194 deletions(-)
```

https://github.com/eclipse/che/issues/16494
https://github.com/eclipse/che/issues/15786
supercedes PR https://github.com/che-incubator/che-workspace-operator/pull/22

### Is it tested? How?
Tested java-mysql and cloud shell with and without webhooks, oauth, on crc.


JPinkney:
Tried this out and ran into a few things:
- When running `oc get workspace` the url given doesn't have http infront anymore, not a big issue I just like being able to click on it in the terminal 
```
NAME          WORKSPACE ID                PHASE      URL
cloud-shell   workspacee0cc88538f2c4e58   Starting   workspacee0cc88538f2c4e58-cloud-shell-proxy-4400.apps-crc.testing
```

- When testing with webhooks enabled and routing as openshift-oauth I'm running into `The only workspace creator has exec access`. For some reason `org.eclipse.che.workspace/creator` is set to ''. I feel like I've run into that issue on master before though 

Tested with webhooks disabled and everything was working 👍 

In terms of the code, it LGTM but I don't have that much experience in go/kubernetes/the operator world yet. I'm going to take another look through it all again tomorrow though

amisevsk:
> - When testing with webhooks enabled and routing as openshift-oauth I'm running into `The only workspace creator has exec access`. For some reason `org.eclipse.che.workspace/creator` is set to ''. I feel like I've run into that issue on master before though 

I've seen this before, but I think usually it's due to starting with webhooks disabled and then enabling them (then the webhook would add the annotation but not have a user for it)? I'm not sure exactly what causes it, but could not reproduce with a fresh deployment

I've added https(s) back into the workspace URL :) -- I didn't add it initially because clicking the url never works in my terminal for some reason.

While looking into including http(s) in the URL status, I came across a few issues:
- We weren't listing endpoints in the runtime annotation if:
    - The endpoint was for a plugin
    - The plugin had an alias
    - The plugin alias didn't match the container name for that plugin
  
  This means the problem basically only manifested for cloud-shell, where it doesn't matter; in the java-mysql workspace theia is aliased to `theia-ide`.

  The solution for now is to disable aliases for Che plugins and editors; from checking the full Che server, I don't see how aliases are used currently. We might still run into issues with component name <-> container name matching, but I don't have a good solution in mind. **Suggestions here are more than welcome**.

- We currently *still* rely on `ingress.global.domain` for correctly setting hostnames, even when running on OpenShift where it's not necessary. I have no idea how I didn't have this issue during testing, but for now it's required to set `ingress.global.domain` manually (e.g. `app-crc.testing` for `crc`). I'll look into fixing this more generally, but the issue is that even with openshift-oauth routing we create ingresses.

- The makefile I merged had some minor issues, I've opened PR to address those https://github.com/che-incubator/che-workspace-operator/pull/37 (they're largely from rebasing over #35). The changes are duplicated to this PR but will be removed once #37 is merged.

Makefile
```diff
@@ -81,15 +99,16 @@ update: _update_yamls _update_crds _reset_yamls
...
+# It's safer to delete all workspaces before deleting the controller; otherwise we could
+# leave workspaces in a hanging state if we add finalizers.
 ifneq ($(shell command -v kubectl),)
 	kubectl delete workspaces.workspace.che.eclipse.org --all-namespaces --all
 else
 	$(info WARN: kubectl is not installed: unable to delete all workspaces)
 endif
 	$(TOOL) delete namespace $(NAMESPACE)
 	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io workspaceroutings.workspace.che.eclipse.org
+	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io components.workspace.che.eclipse.org
```
sleshchenko, line 111:
```suggestion
	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io components.workspace.che.eclipse.org
```

cmd/manager/main.go
```diff
@@ -123,20 +130,43 @@ func main() {
...
+	// Setup all webhooks
 	log.Info("Setting up webhooks")
 	if err := webhook.SetUpWebhooks(mgr, ctx); err != nil {
 		log.Error(err, "unable to register webhooks to the manager")
 		os.Exit(1)
 	}
 
+	// TODO: Required to filter GVK for metrics, since we add routes and templates to the scheme.
+	// TODO: see: https://github.com/operator-framework/operator-sdk/pull/2606
+	//if err = serveCRMetrics(cfg); err != nil {
```
sleshchenko, line 142: Not sure I undestand why it's commented but does it make sense to create a separate issue to uncomment it and eventually make CR metrics working?

deploy/crds/workspace.che.eclipse.org_workspaces_crd.yaml
```diff
@@ -5,24 +5,24 @@ metadata:
 spec:
   additionalPrinterColumns:
   - JSONPath: .status.workspaceId
-    name: Id
+    description: The workspace's unique id
+    name: Workspace ID
```
sleshchenko, line 9: it's not related to this PR but still interesting topic to raise:
Does Workspace ID makes sense since workspace is CR?
Do you think we can reuse UID?

Sometime ago I tried to remove it, but workspace id now it's workspace word + cut version of UID, which is used for objects/names generation. But I wonder if it makes sense in general and if we can drop it out.

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
```diff
@@ -0,0 +1,189 @@
...
+			routes = append(routes, routeV1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      common.RouteName(workspaceMeta.WorkspaceId, endpointName),
+					Namespace: workspaceMeta.Namespace,
+					Labels: map[string]string{
+						config.WorkspaceIDLabel: workspaceMeta.WorkspaceId,
+					},
+				},
+				Spec: routeV1.RouteSpec{
+					Host: hostname,
```
sleshchenko, line 120: I wonder if we should specify hostname at all? I know it helps to solve issues with route lengh and uniqueness but I don't think it's possible on the secure openshift instances, like OSIO
Screenshot_20200403_155155
Or you think that controller must be able to set host?

deploy/crds/workspace.che.eclipse.org_workspaces_crd.yaml
amisevsk: One reason for leaving it in is that we use workspace ID (which is a just the object's UID, but modified) in various places and as a label for all created objects. We *could* use UID directly, but parsing the UID to a UUID [can error](https://github.com/amisevsk/che-workspace-crd-operator/blob/93fdedbd3983cc501aab36a40f16f412f0b888d6/pkg/controller/workspace/workspace_controller.go#L281-L287), so it would complicate using a unique ID the way we do. We would also still have to store it to pass to subcomponents, since their UID is different from the main workspace UID.

Another option would be using the CR's name as a base, so everything created for `cloud-shell` would be prefixed by `cloud-shell`.

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
amisevsk: The main issue is that hostname must be set on ingresses; it's done here mostly because the route code is a port of the ingress code; we *should* drop it here, but the PR will need to be reworked to create only routes and not ingresses when running on OpenShift.

Also, I don't have a minishift instance handy for testing, but I wonder how it would work there since we currently rely on using `$(minishift ip).nip.io`.

cmd/manager/main.go
amisevsk: It was initially commented to resolve a log spamming issue Josh found while porting webhooks to the forked repo, and I didn't have time to delve deeply enough to figure out the full problem. I agree that it should be removed and an issue created, though.

Makefile
amisevsk: :+1:

amisevsk:
@sleshchenko Commit 22ab54d  *should* fix your issue with `command -v kubectl` in the makefile, please double check.

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
```diff
@@ -0,0 +1,117 @@
...
+			"--https-address=:" + strconv.FormatInt(proxyEndpoint.publicEndpoint.Port, 10),
+			"--http-address=127.0.0.1:" + strconv.FormatInt(proxyEndpoint.publicEndpointHttpPort, 10),
+			"--provider=openshift",
+			"--openshift-service-account=" + common.ServiceAccountName(meta.WorkspaceId),
+			"--upstream=http://localhost:" + strconv.FormatInt(proxyEndpoint.upstreamEndpoint.Port, 10),
+			"--tls-cert=/etc/tls/private/tls.crt",
+			"--tls-key=/etc/tls/private/tls.key",
+			"--cookie-secret=SECRET",
+			// Currently: block anyone who can't exec in the current namespace
+			"--openshift-sar=" + fmt.Sprintf(openShiftProxySARFmt, "", ""),
```
sleshchenko, line 75:
```suggestion
```
`""` seems to be a default namespace but not the current. So, it does not work for me when I test workspace as a developer
Screenshot_20200406_102950

pkg/controller/workspace/env/common_env_vars.go
```diff
@@ -10,50 +10,49 @@
...
 		{
 			Name:  "CHE_AUTH_ENABLED",
-			Value: AuthEnabled,
+			Value: config.AuthEnabled,
 		},
 		{
 			Name:  "CHE_WORKSPACE_NAMESPACE",
-			Value: wkspCtx.Namespace,
+			Value: namespace,
 		},
```
sleshchenko, line 56: As a quick solution for CloudShell with OpenShiftOAuth: 
```suggestion
		},
		{
			Name:  "USE_BEARER_TOKEN",
			Value: config.ControllerCfg.GetWebhooksEnabled(),
		},
```
As a bit better solution I wonder if OpenShift OAuth Routing could contribute env var to all containers via PodAdditions... But then we have another question that SA with exec rights is not needed anymore, then who should contribute it?
I'm OK with trying to solve this issue in scope of a dedicated issue.

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
```diff
@@ -0,0 +1,189 @@
...
+		proxyPorts[proxyEndpoint.machineName] = append(proxyPorts[proxyEndpoint.machineName], proxyEndpoint.publicEndpoint)
+	}
+	for machineName, machineEndpoints := range noProxy {
+		proxyPorts[machineName] = append(proxyPorts[machineName], machineEndpoints...)
+	}
+	// Use common service for all unproxied endpoints
+	proxyServices := getServicesForEndpoints(proxyPorts, workspaceMeta)
+	for idx := range proxyServices {
+		proxyServices[idx].Annotations = map[string]string{
+			"service.alpha.openshift.io/serving-cert-secret-name": "proxy-tls",
```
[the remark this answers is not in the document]
sleshchenko: P.S. Controller does not seem to update this field after reconciling, not sure it's expected
Screenshot_20200406_133853

pkg/adaptor/plugin.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, plugin := range plugins {
+		component, err := adaptChePluginToComponent(workspaceId, plugin)
+		if err != nil {
+			return nil, nil, err
+		}
+		// TODO: Alias for plugins seems to be ignored in regular Che
```
sleshchenko, line 47: Not really. From che.openshift.io:
```json
{
  "namespace": "sleshche",
  "temporary": false,
  "id": "workspace6wt7edpromg4w9w0",
  "status": "RUNNING",
  "runtime": {
    "machines": {
      "nodejs": {
        "attributes": {
          "component": "nodejs",
          "memoryRequestBytes": "209715200",
          "memoryLimitBytes": "1073741824",
          "source": "recipe",
          "cpuLimitCores": "2.0",
          "cpuRequestCores": "0.125"
        }
      }
    },
    "commands": [
      {
        "commandLine": "yarn run lint",
        "name": "lint",
        "attributes": {
          "componentAlias": "nodejs",
          "machineName": "nodejs",
          "workingDir": "${CHE_PROJECTS_ROOT}/angular-realworld-example-app"
        },
        "type": "exec"
      }
    ]
  },
  "devfile": {
    "metadata": {
      "name": "wksp-sek5"
    },
    "components": [
      {
        "mountSources": true,
        "endpoints": [
          {
            "name": "angular",
            "port": 4200
          }
        ],
        "memoryLimit": "1Gi",
        "type": "dockerimage",
        "alias": "nodejs",
        "image": "quay.io/eclipse/che-nodejs10-community:nightly"
      }
    ],
    "apiVersion": "1.0.0",
    "commands": [
      {
        "name": "lint",
        "actions": [
          {
            "workdir": "${CHE_PROJECTS_ROOT}/angular-realworld-example-app",
            "type": "exec",
            "command": "yarn run lint",
            "component": "nodejs"
          }
        ]
      }
    ]
  }
}
```

So, alias is propagated to machine's and commands attributes. Then Theia use them to match command to container.
I'm OK with skipping alias for time being but it would be good to make comment up to date.

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
```diff
@@ -0,0 +1,117 @@
...
+			{
+				//Name:          endpoint.upstreamEndpoint.Name,
+				ContainerPort: int32(proxyEndpoint.publicEndpoint.Port),
+				Protocol:      corev1.ProtocolTCP,
+			},
+		},
+		ImagePullPolicy: corev1.PullPolicy(config.ControllerCfg.GetSidecarPullPolicy()),
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      "proxy-tls",
```
sleshchenko, line 59: to be consist in proposal in https://github.com/che-incubator/che-workspace-operator/pull/36/files#r403919074
```suggestion
				Name:      "proxy-tls-" + meta.WorkspaceId,
```


P.S. It surprised me a bit that volumeMounts is used as secret name. It's OK for time being but it would be better to design more straightforward solution. Something like https://github.com/che-incubator/che-workspace-operator/commit/117c52150aade53fb90e0d0e329908f22333781f

pkg/controller/component/component_controller.go
```diff
@@ -0,0 +1,182 @@
...
+
+	var components []workspacev1alpha1.ComponentDescription
+	dockerimageDevfileComponents, pluginDevfileComponents, err := adaptor.SortComponentsByType(instance.Spec.Components)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	commands := instance.Spec.Commands
+
+	dockerimageComponents, err := adaptor.AdaptDockerimageComponents(instance.Spec.WorkspaceId, dockerimageDevfileComponents, commands)
```
sleshchenko, line 116: In general dockerimage is no a required component, so additional check is needed to avoid 
```
{"level":"error","ts":1586176842.3509479,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"workspace-controller","request":"dev-ws/cloud-shell","error":"Component.workspace.che.eclipse.org \"components-workspaced46b62f0ebe44f3c-docker\" is invalid: spec.components: Invalid value: \"null\": spec.components in body must be of type array: \"null\"","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.3.0/pkg/internal/controller/controller.go:218\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.3.0/pkg/internal/controller/controller.go:192\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.3.0/pkg/internal/controller/controller.go:171\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190817020851-f2f3a405f61d/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190817020851-f2f3a405f61d/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190817020851-f2f3a405f61d/pkg/util/wait/wait.go:88"}
```
https://github.com/che-incubator/che-workspace-operator/commit/64751c6ee3ead5790b1b78288458fe5ac4f866d9
tested with theia-simple https://github.com/che-incubator/che-workspace-operator/commit/a0b4bb05e4adeb78f6178e834f6e5ce794c65203
P.S. I was lucky not being able to start any other workspace with plugins and dockerimage containers because of memory lack and it's how I found this bug.

P.P.S. Not sure if we should be able to run workspace without editor/plugins, like with dockerimage component. In general it could work, so:
- we should add a corresponding check below;
- add some devfile validating before workspace creating;

pkg/adaptor/plugin.go
amisevsk: Yes I'm fine to update this -- it was on my TODO list for Friday that I didn't get to. I recognize that alias is used for dockerimage components, but I could not find it being used for plugin components. I need to look into it more deeply.

pkg/controller/workspace/runtime/runtime_annotation.go
```diff
@@ -0,0 +1,72 @@
...
+		for containerName, container := range component.ComponentMetadata.Containers {
+			servers := map[string]v1alpha1.CheWorkspaceServer{}
+			// TODO: This is likely not a good choice for matching, since it'll fail if container name does not match an endpoint key
+			for _, endpoint := range endpoints[containerName] {
+				protocol := endpoint.Attributes[v1alpha1.PROTOCOL_ENDPOINT_ATTRIBUTE]
+
+				servers[endpoint.Name] = v1alpha1.CheWorkspaceServer{
+					Attributes: endpoint.Attributes,
+					Status:     v1alpha1.RunningServerStatus, // TODO: This is just set so the circles are green
+					URL:        fmt.Sprintf("%s://%s", protocol, endpoint.Url),
```
sleshchenko, line 53: It leads to url with double protocol like `"http://https://workspace383a324431b54d9f-theia-proxy-4400.apps-crc.testing`

Something like https://github.com/amisevsk/che-workspace-crd-operator/commit/84dd6135e7f777033ea1faec8d73a6dc585f9ade could be done to solve this issue.

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
```diff
@@ -0,0 +1,117 @@
...
+
+var openShiftProxySARFmt = `{"namespace": "%s", "resource": "pods", "name": "%s", "verb": "exec"}`
+
+func getProxyPodAdditions(proxyEndpoints map[string]proxyEndpoint, meta WorkspaceMetadata) *v1alpha1.PodAdditions {
+	var proxyContainers []corev1.Container
+	for _, proxyEndpoint := range proxyEndpoints {
+		proxyContainers = append(proxyContainers, getProxyContainerForEndpoint(proxyEndpoint, meta))
+	}
+	proxyVolumes := getProxyVolumes(proxyContainers)
+	serviceAcctAnnotations := getProxyServiceAcctAnnotations(proxyEndpoints, meta)
```
sleshchenko, line 35:
```suggestion
	serviceAccAnnotations := getProxyServiceAccAnnotations(proxyEndpoints, meta)
```

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
amisevsk: If it's a namespace issue, then you should get a forbidden page; 500 internal server error is weird here.

pkg/controller/workspacerouting/solvers/solver.go
```diff
@@ -0,0 +1,32 @@
...
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package solvers
+
+import (
+	"github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
+	v12 "github.com/openshift/api/route/v1"
```
sleshchenko, line 17: openshift api is so far from v12 ;)
```suggestion
	routev1 "github.com/openshift/api/route/v1"
```

amisevsk: Yeah, this is a goland problem, I thought I caught all instances of it doing this :angry: 

If you have an import alias conflict it just throws a `2` on the end silently.

pkg/config/property.go
```diff
@@ -21,10 +33,10 @@ const (
...
 	defaultPluginArtifactsBrokerImage = "quay.io/eclipse/che-plugin-artifacts-broker:v3.1.0"
 
-	//routingClass defines the default routing class that should be used if user does not specify it explicitly
+	// routingClass defines the default routing class that should be used if user does not specify it explicitly
 	routingClass        = "che.default_routing_class"
 	defaultRoutingClass = "basic"
 
 	webhooksEnabled        = "che.webhooks.enabled"
-	defaultWebhooksEnabled = "false"
+	defaultWebhooksEnabled = "true"
```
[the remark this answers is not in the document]
amisevsk: Good point -- I set the default before I was aware of the limitation.

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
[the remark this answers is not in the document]
amisevsk: Ah good catch, we currently ignore `ObjectMeta` when checking if a service needs to be updated. I'll fix this.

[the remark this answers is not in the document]
amisevsk: Looking into it a bit, this is maybe an task for another issue -- we can't naively check that annotations match since some are added by the cluster, and I don't want to hard-code checking `service.alpha.openshift.io/serving-cert-secret-name` in the generic case without further discussion.

For now, this annotation should not change (unless you rollout a new operator that changes the semantics).

pkg/controller/component/component_controller.go
amisevsk: Good catch -- I had completely missed this. I'm adding checks to `pkg/controller/workspace/provision/component.go` to avoid this. For now, I'm allowing workspaces with no editors/plugins (though they're kind of meaningless). We can decide separately if we e.g. want a default as is in upstream or if we want to just set status to failed. 

As for devfile validation, I expect that will come once we move to the devfile 2.0 spec, as validation will be built in.

pkg/controller/workspace/runtime/runtime_annotation.go
amisevsk: This is an artifact of me implementing Josh's request for protocol in status -- I forgot to remove it here. It doesn't work properly anyways, since protocol is *always* `http` even for https routes. We should have protocol included in URL everywhere now, so checking is unnecessary.

amisevsk: Ah except when it doesn't:
```
ws://http://workspace82ba65b9c2b84595-che-machine-exec-4444.apps-crc.testing
```

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
amisevsk: I prefer Acct as it's a more common abbreviation of account

amisevsk:
Tested again briefly on `crc`, and I suspect there may be a regression somewhere. With the java-mysql workspace I'm getting frequent issues opening a terminal.

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
amisevsk: Fixed :)

pkg/controller/workspacerouting/solvers/oauth_proxy_container.go
sleshchenko: OK

pkg/controller/workspace/provision/deployment.go
```diff
@@ -0,0 +1,312 @@
...
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					config.CheOriginalNameLabel: config.CheOriginalName,
+					config.WorkspaceIDLabel:     workspace.Status.WorkspaceId,
+					config.WorkspaceNameLabel:   workspace.Name,
+				},
+			},
+			Strategy: appsv1.DeploymentStrategy{
+				Type: "RollingUpdate",
```
sleshchenko, line 165: Maybe it's better to start with Recreate since workspaces will fails to update if PVC is used (at least on production-ready cluster)

pkg/controller/workspace/runtime/runtime_annotation.go
sleshchenko: So, now we get https URL even when protocol is ws, right?

samples/no-editor.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: workspace.che.eclipse.org/v1alpha1
+kind: Workspace
+metadata:
+  name: cloud-shell
```
sleshchenko, line 4: is it really cloud-shell? Will controller propagate a default editor?

pkg/apis/workspace/v1alpha1/workspace_types.go
```diff
@@ -13,48 +13,92 @@
...
+	// Human-readable message indicating details about last transition.
+	Message string `json:"message,omitempty"`
+}
+
+type WorkspacePhase string
+
+// Valid workspace Statuses
+const (
+	WorkspaceStatusStarting WorkspacePhase = "Starting"
+	WorkspaceStatusReady    WorkspacePhase = "Ready"
```
sleshchenko, line 70: Previously it was Running, right?
I don't have a strong preferences but just reminder that OpenShift Console depends on Running state and we need inform them after PR is merged.

amisevsk:
Tested again, and the issues seem to be related to `crc` being somewhat unresponsive. Cloud shell works with both basic routing and OAuth + webhooks.

I'm merging this PR to unblock things; we can address edge cases in separate issues.

samples/no-editor.yaml
amisevsk: My bad, these got added half-accidentally as I threw them together for testing. I'll update the names since they could still be useful for test cases.

pkg/apis/workspace/v1alpha1/workspace_types.go
amisevsk: You're right, great catch -- I mixed up the workspace phase and the workspaceRouting phase. Fixed in PR https://github.com/che-incubator/che-workspace-operator/pull/46

amisevsk:
@sleshchenko I missed your last comments before merging; they are addressed in PR https://github.com/che-incubator/che-workspace-operator/pull/46

PR #37, opened 2020-04-03
Fix makefile issues due to rebasing over PR #35

amisevsk:
### What does this PR do?
Fix the makefile after rebasing over #35, which moved some yamls around in the deploy folder -- I failed to check for this before merging.

It also adds support for setting default routing class when deploying.

### Is it tested? How?
Tested on crc with options
```
export IMG=amisevsk/che-workspace-controller:dev
export OPERATOR_NAME=che-workspace-operator
export WEBHOOK_ENABLED=true
export DEFAULT_ROUTING=openshift-oauth
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ TOOL ?= oc
 CLUSTER_IP ?= 192.168.99.100
 PULL_POLICY ?= Always
 WEBHOOK_ENABLED ?= false
+DEFAULT_ROUTING ?= basic
 
 all: help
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -81,8 +99,8 @@ update: _update_yamls _update_crds _reset_yamls
 
 ### uninstall: remove namespace and all CRDs from cluster
 uninstall:
-	# It's safer to delete all workspaces before deleting the controller; otherwise we could
-	# leave workspaces in a hanging state if we add finalizers.
+# It's safer to delete all workspaces before deleting the controller; otherwise we could
+# leave workspaces in a hanging state if we add finalizers.
 ifneq ($(shell command -v kubectl),)
```
sleshchenko, line 104: it does not work for me for some reasons I don't understand because of limited knowledge of makefiles
Screenshot_20200403_124352

amisevsk: Figured it out, that was subtle: https://stackoverflow.com/questions/5618615/check-if-a-program-exists-from-a-makefile#comment76110496_34756868

PR #38, opened 2020-04-03
Add instructions on how to deploy with webhooks and openshift oauth

sleshchenko:
### What does this PR do?
Add instructions on how to deploy with webhooks and openshift oauth.
It will help people to get such configuration easily without learning all of configuration options.

N/A

### Is it tested? How?
No need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,17 @@ The controller can be deployed to a cluster provided you are logged in with clus
 ```bash
 export IMG=quay.io/che-incubator/che-workspace-controller:nightly
 export TOOL=oc # Use 'export TOOL=kubectl' for kubernetes
-export WEBHOOK_ENABLED=false
+make deploy
+```
+
+By default, controller will expose workspace servers without any authentications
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,7 +9,17 @@ The controller can be deployed to a cluster provided you are logged in with clus
 ```bash
 export IMG=quay.io/che-incubator/che-workspace-controller:nightly
 export TOOL=oc # Use 'export TOOL=kubectl' for kubernetes
-export WEBHOOK_ENABLED=false
+make deploy
+```
+
+By default, controller will expose workspace servers without any authentications
+which is not accessible in public available clusters.
```
amisevsk, line 16:
```suggestion
By default, controller will expose workspace servers without any authentication; this is not advisable for public clusters, as any user could access the created workspace via URL.
```

sleshchenko: fixup: Add instructions on how to deploy with webhooks and openshift oauth

PR #39, opened 2020-04-03
Add an ability to configure already deployed controller with make

sleshchenko:
### What does this PR do?
Adds a command to update controller configuration according to the set env vars and actual configuration files state (for example while testing some branch).

Trigger reconfigure during rollout to make sure that the right image is used with the right configuration files.

N/A

### Is it tested? How?
Used it but not sure about corner cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,11 @@ _update_yamls: _set_registry_url
 ifeq ($(TOOL),oc)
 	sed -i "s|image: .*|image: $(IMG)|g" ./deploy/os/controller.yaml
 	sed -i "s|imagePullPolicy: Always|imagePullPolicy: $(PULL_POLICY)|g" ./deploy/os/controller.yaml
+	sed -i "s|kubectl.kubernetes.io/restartedAt: .*|kubectl.kubernetes.io/restartedAt: '$$(date --iso-8601=seconds)'|g" ./deploy/os/controller.yaml
 else
 	sed -i "s|image: .*|image: $(IMG)|g" ./deploy/k8s/controller.yaml
 	sed -i "s|imagePullPolicy: Always|imagePullPolicy: $(PULL_POLICY)|g" ./deploy/k8s/controller.yaml
+	sed -i "s|kubectl.kubernetes.io/restartedAt: .*|kubectl.kubernetes.io/restartedAt: '$$(date --iso-8601=seconds)'|g" ./deploy/k8s/controller.yaml
 endif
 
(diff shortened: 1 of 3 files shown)


deploy/k8s/controller.yaml
```diff
@@ -13,6 +13,8 @@ spec:
     metadata:
       labels:
         app: che-workspace-controller
+      annotations:
+        kubectl.kubernetes.io/restartedAt: ""
```
sleshchenko, line 17: it's a tricky thing we can do to make sure that the deployed is retriggered after we apply `./deploy` and there is no change in deployment but is in configmap.

Instead we could just:
```
configure: _update_yamls _apply_controller_cfg _reset_yamls restart
```
but then deployment may be triggered twice if there is changes in deployment as well.

Controller has some code to reload configuration after config map is changed, but I don't think it will work with webhooks_enabled...

@amisevsk WDYT? Maybe it is the option to go and having +1 triggering of deployment is not critical at all:
```
configure: _update_yamls _apply_controller_cfg _reset_yamls restart
```
or updating restartAt is good enough to proceed with?

Makefile
```diff
@@ -94,6 +98,9 @@ endif
 ### rollout: rebuild and push docker image and restart cluster deployment
 rollout: docker restart
 
+### configure: configures already deployed controller according to set env variables
+configure: _update_yamls _apply_controller_cfg _reset_yamls
```
sleshchenko, line 102: probably configure and update could be merged but not sure.
Maybe it makes sense to rename update to update_crds

Makefile
```diff
@@ -94,8 +98,11 @@ endif
...
+
 ### update: update CRDs defined on cluster
-update: _update_yamls _update_crds _reset_yamls
+update: _update_crds
```
sleshchenko, line 105: remove updating yamls here since it does not modify CRDs
and sed output in misleading when `make update`

Makefile
amisevsk: I'll do this in a followup PR.

deploy/k8s/controller.yaml
amisevsk: Makes sense to me :+1:

Makefile
amisevsk: Actually, looking at this more, we do need the `_update_yamls` step, since `_update_crds` reapplies the controller configmap, so this would reset any changes defined by env vars to their defaults.

PR #43, opened 2020-04-06
Switch context to admin when it's needed

sleshchenko:
### What does this PR do?
I use developer admin as test one from terminal, so each time when I do `make rollout` I faced an issue that image is rebuild but deployment is not restarted, so I need to do `kubectx - && make restart && kubectx -`

This PR allows to set env var with admin context and then make will take care of this routine.

I'm not so sure that we should merge this feature in our makefile,
- from one hand - it makes developing easier;
- from another hand - it makes makefile even more difficult but not much ;)

@amisevsk @JPinkney WDYT?

N/A

### Is it tested? How?
Screenshot_20200406_180255

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,4 +165,4 @@ help: Makefile
 	@echo "    CLUSTER_IP         - For Kubernetes only, the ip address of the cluster (minikube ip)"
 	@echo "    PULL_POLICY        - Image pull policy for controller"
 	@echo "    WEBHOOK_ENABLED    - Whether webhooks should be enabled in the deployment"
-	@cho  "    ADMIN_CTX          - Kubectx entry that should be used during work with cluster. The current will be used if omitted"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -147,3 +165,4 @@ help: Makefile
 	@echo "    CLUSTER_IP         - For Kubernetes only, the ip address of the cluster (minikube ip)"
 	@echo "    PULL_POLICY        - Image pull policy for controller"
 	@echo "    WEBHOOK_ENABLED    - Whether webhooks should be enabled in the deployment"
+	@cho  "    ADMIN_CTX          - Kubectx entry that should be used during work with cluster. The current will be used if omitted"
```
JPinkney, line 168: is this supposed to be echo not cho?

amisevsk: Fixed, as it turns out I have push access to the branch :thinking:

PR #45, opened 2020-04-07
Deploy registry only if needed

JPinkney:
### What does this PR do?
This PR creates an internal registry that holds the cloud shell meta. This makes it so that you won't have to deploy the plugin registry when just using cloud shell.

https://github.com/eclipse/che/issues/16313

### Is it tested? How?

Tested on crc with first setting env variable `REGISTRY_ENABLED = false`. As expected, the plugin registry wasn't deployed and when launching a cloud shell workspace it pulled from the internal registry and successfully started. When you try and start a workspace that uses theia, you will get an error saying something like `could not retrieve meta.yaml from registry`

Then I turned on `REGISTRY_ENABLED = true` and ran `make deploy` and the plugin registry was successfully deployed. Tried to start a cloud shell workspace and it worked. Then I tried to start python-sample.yaml and it worked


pkg/internal_registry/registry.go
```diff
@@ -0,0 +1,57 @@
...
+
+func InternalRegistry() map[string]brokerModel.PluginMeta {
+
+	return map[string]brokerModel.PluginMeta{
+		"eclipse/cloud-shell/nightly": {
+			APIVersion:  "v2",
+			Publisher:   "eclipse",
+			Description: "Cloud Shell provides an ability to use terminal widget like an editor.",
+			DisplayName: "Cloud Shell Editor",
+			ID:          "",
```
amisevsk, line 15: I think ID needs to be set as well -- I'm seeing
```
containers:
        che-machine-exec:
          attributes:
            plugin: ""
            source: tool
          ports:
          - 4444
```
in the component description, which looks wrong.

amisevsk: c.f.:
```yaml
      containers:
        che-machine-exec:
          attributes:
            plugin: eclipse/cloud-shell/nightly
            source: tool
          ports:
          - 4444

```
in master (though this has no bearing on cloud-shell)

amisevsk: Found where `getPluginMetadata` sets ID explicitly: https://github.com/eclipse/che-plugin-broker/blob/master/utils/metadata_resolver.go#L65-L73

pkg/internal_registry/registry.go
```diff
@@ -0,0 +1,57 @@
...
+							{
+								ExposedPort: 4444,
+							},
+						},
+						Command: []string{
+							"/go/bin/che-machine-exec",
+							"--static",
+							"/cloud-shell",
+							"--url",
+							"127.0.0.1:4444",
```
amisevsk, line 47: I put a `json.MarshalIndent` on the plugin meta we get from `utils.GetPluginMeta` and I see (formatted as json)
```yaml
apiVersion: v2
spec:
  endpoints:
  - name: cloud-shell
    public: true
    targetPort: 4444
    attributes:
      cookiesAuthEnabled: "true"
      discoverable: "false"
      protocol: http
      secure: "true"
      type: ide
  containers:
  - name: che-machine-exec
    image: quay.io/eclipse/che-machine-exec:nightly
    env: []
    commands: []
    volumes: []
    ports:
    - exposedPort: 4444
    mountSources: false
    command: []
    args: []
  initContainers: []
  workspaceEnv: []
  extensions: []
id: eclipse/cloud-shell/nightly
name: cloud-shell
displayName: Cloud Shell Editor
publisher: eclipse
type: Che Editor
description: Cloud Shell provides an ability to use terminal widget like an editor.
version: nightly
title: Cloud Shell Editor
icon: https://www.eclipse.org/che/images/logo-eclipseche.svg

Comment (amisevsk) [reply pkg/internal_registry/registry.go:47]: The command in your meta.yaml was added in https://github.com/eclipse/che-plugin-registry/commit/74fcb49, which is not included in the 7.8.0 release of the registry. I don't know if it matters significantly, but this does bring up the issue of how we maintain the offline registry (we can discuss in a separate issue)

Comment (amisevsk) [reply pkg/internal_registry/registry.go:47]: The command actually prevents machine-exec from listening outside the local network, which means cloud shell won't work with basic routing. It's not an issue for openshift-oauth as the proxy pod is included in the deployment and forwards requests.

Created issue https://github.com/eclipse/che/issues/16568 to discuss this.

File: pkg/internal_registry/registry.go [hunk 3]
```diff
@@ -0,0 +1,50 @@
+package registry
+
+import (
+	brokerModel "github.com/eclipse/che-plugin-broker/model"
+)
+
+func InternalRegistry() map[string]brokerModel.PluginMeta {
+
+	return map[string]brokerModel.PluginMeta{
+		"eclipse/cloud-shell/nightly": {
```
sleshchenko, line 10: Can we move it into yaml files... `/deploy/internal-registry/eclipse/cloud-shell/nightly`
Then it will be possible to customize it like with mount PVC or rebuild controller image with modified internal registries files?

pkg/internal_registry/registry.go
sleshchenko: Probably we also may need some endpoint diagnostics tools in workspace controller, like now with nightly plugin registry we'll get Running but not working workspace
Screenshot_20200408_124846

pkg/internal_registry/registry.go
```diff
@@ -0,0 +1,50 @@
+package registry
```
sleshchenko, line 1: Please add license header

pkg/internal_registry/registry.go
JPinkney: :+1:

pkg/internal_registry/registry.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/eclipse/che-plugin-broker/model"
+	brokerModel "github.com/eclipse/che-plugin-broker/model"
+	"gopkg.in/yaml.v1"
```
amisevsk, line 22: There is also a `yaml.v2` API version that might be better, though I don't know the difference. Also this isn't in `go.mod` -- we need to run `go mod tidy` and commit `go.mod` and `go.sum`.

deploy/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v2
```
sleshchenko, line 1: I would say that it's not a part of `deploy` configuration but I'm not sure.
I can propose to make `internal-registry` as a root folder or subfolder of `build` folder...

PR #47, opened 2020-04-08
Create OAuthClient for OpenShift OAuth Proxy

sleshchenko:
### What does this PR do?
Use OAuthClient per Workspace to get user:full token with OpenShift OAuth Proxy.

Drawbacks:
- currently, oauthclient secrets are hard-coded but there is a separate issue where this part should be addressed along with cookie-secret;
- Since WorkspaceRouting are namespaced-scoped and OAuthClient are cluster-scoped, WorkspaceRouting can't own OAuthClient. So, I added finalizer and OAuth clients are removed along with workspaces but when OAuthClient are removed or updated - the reconciling loop is not reinvoked. Probably we're able to listen to all OAuthClients as we do with an object owned by us (Workspace,WorkspaceRouting), check if it's workspace-related OAuthClient and reinvoke reconcile loop for WorkspaceRouting. Not sure if we really should implement it. Seems Che Operator does not watch it's oauthclient as well.

https://github.com/eclipse/che/issues/16515

### Is it tested? How?
oauth-client

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,7 @@ _update_controller_configmap:
 
 _apply_controller_cfg:
 	$(TOOL) apply -f ./deploy
+	$(TOOL) apply -f ./deploy/role.yaml
 ifeq ($(TOOL),oc)
 	$(TOOL) apply -f ./deploy/os/
 else
--- a/Makefile
+++ b/Makefile
@@ -121,8 +121,13 @@ _do_uninstall:
 # leave workspaces in a hanging state if we add finalizers.
 ifneq ($(shell command -v kubectl 2> /dev/null),)
 	kubectl delete workspaces.workspace.che.eclipse.org --all-namespaces --all
+# Have to wait for routings to be deleted in case there are finalizers
+	kubectl delete workspaceroutings.workspace.che.eclipse.org --all-namespaces --all --wait
 else
-	$(info WARN: kubectl is not installed: unable to delete all workspaces)
+ifneq ($(TOOL) get workspaces.workspace.che.eclipse.org --all-namespaces,"No resources found.")
+	$(info To automatically remove all workspaces when uninstalling, ensure kubectl is installed)
+	$(error Cannot uninstall operator, workspaces still running. Delete all workspaces and workspaceroutings before proceeding)
+endif
 endif
 	$(TOOL) delete namespace $(NAMESPACE)
 	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io workspaceroutings.workspace.che.eclipse.org
--- a/pkg/controller/workspacerouting/sync_oauthclients.go
+++ b/pkg/controller/workspacerouting/sync_oauthclients.go
@@ -1,0 +1,116 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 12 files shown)


pkg/controller/workspacerouting/sync_oauthclients.go
```diff
@@ -0,0 +1,116 @@
...
+	return found.Items, nil
+}
+
+func (r *ReconcileWorkspaceRouting) deleteOAuthClients(routing *v1alpha1.WorkspaceRouting) error {
+	labelSelector, err := labels.Parse(fmt.Sprintf("%s=%s", config.WorkspaceIDLabel, routing.Spec.WorkspaceId))
+	if err != nil {
+		return err
+	}
+	listOptions := &client.DeleteAllOfOptions{
+		ListOptions: *&client.ListOptions{
```
amisevsk, line 110: Looks like some IDE "magic": `*&client.ListOptions`

pkg/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -195,9 +230,28 @@ func (r *ReconcileWorkspaceRouting) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, statusErr
 	}
 
+	oauthClientInSync, err := r.syncOAuthClient(instance, *oauthClient)
```
amisevsk, line 233: This looks like a potential nil dereference -- `routingObjects.OAuthClient` should be nil for `basic` routing.

pkg/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -245,3 +299,21 @@ func getSecureProtocol(protocol string) string {
...
+}
+
+func remove(list []string, s string) []string {
+	for i, v := range list {
+		if v == s {
+			list = append(list[:i], list[i+1:]...)
+		}
+	}
+	return list
+}
```
amisevsk, line 319: Gotta love golang.

pkg/controller/workspacerouting/sync_oauthclients.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (r *ReconcileWorkspaceRouting) getClusterOAuthClients(routing *v1alpha1.WorkspaceRouting) ([]oauthv1.OAuthClient, error) {
+	found := &oauthv1.OAuthClientList{}
+	labelSelector, err := labels.Parse(fmt.Sprintf("%s=%s", config.WorkspaceIDLabel, routing.Spec.WorkspaceId))
+	if err != nil {
+		return nil, err
+	}
+	listOptions := &client.ListOptions{
+		Namespace:     routing.Namespace,
```
amisevsk, line 93: I had issues with the namespaced list options here, since OAuthClients are cluster wide.

amisevsk:
Fixed the issues I mentioned and rebased onto master to resolve conflict.

Makefile
```diff
@@ -100,6 +100,7 @@ _update_controller_configmap:
 
 _apply_controller_cfg:
 	$(TOOL) apply -f ./deploy
+	$(TOOL) apply -f ./deploy/role.yaml
```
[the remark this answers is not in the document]
sleshchenko: Screenshot_20200422_145654
It's related ) For some reason oauthclients disappear from role when applying the whole folder. And I have a plan to remove applying the whole folder in a separate PR.

PR #50, opened 2020-04-09
Create routes on OpenShift and ingresses on Kubernetes

amisevsk:
### What does this PR do?
Change WorkspaceRoutings controller to always create routes on OpenShift and ingresses on Kubernetes.

One of the goals in this PR was to eliminate the need for `ingress.global.domain` when running on OpenShift by not specifying hostnames when creating routes. However, OpenShift's automatically-generated route hostnames are `<route-name>-<namespace>.<routing-suffix>`, which means they're frequently invalid (if e.g. deployed in namespace `che-workspace-controller`. As a result, I've renamed `ingress.global.domain` to `cluster.routing.suffix` and added setting the value via the makefile.

Note: the controller does contain logic to [automatically set routing suffix](https://github.com/che-incubator/che-workspace-operator/blob/ae2b1ca06f69b3ad6d4ed245e7483f8637e70be9/pkg/config/config.go#L226) on OpenShift; however, I have had it fail/fall out of sync in the past so the separate option is still useful.

### Is it tested? How?
Tested on `crc` with the usual matrix of settings. Have not currently tested on minikube but will before merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 NAMESPACE = che-workspace-controller
 IMG ?= quay.io/che-incubator/che-workspace-controller:nightly
 TOOL ?= oc
-CLUSTER_IP ?= 192.168.99.100
+ROUTING_SUFFIX ?= 192.168.99.100.nip.io
 PULL_POLICY ?= Always
 WEBHOOK_ENABLED ?= false
 DEFAULT_ROUTING ?= basic
@@ -14,6 +14,7 @@ _print_vars:
 	@echo "Current env vars:"
 	@echo "    IMG=$(IMG)"
 	@echo "    PULL_POLICY=$(PULL_POLICY)"
+	@echo "    ROUTING_SUFFIX=$(ROUTING_SUFFIX)"
 	@echo "    WEBHOOK_ENABLED=$(WEBHOOK_ENABLED)"
 	@echo "    DEFAULT_ROUTING=$(DEFAULT_ROUTING)"
 	@echo "    REGISTRY_ENABLED=$(REGISTRY_ENABLED)"
@@ -40,9 +41,9 @@ ifeq ($(REGISTRY_ENABLED),true)
 ifeq ($(TOOL),oc)
 	$(TOOL) apply -f ./deploy/registry/local/os
 else
-	sed -i.bak -e "s|192.168.99.100|$(CLUSTER_IP)|g" ./deploy/registry/local/k8s/ingress.yaml
+	sed -i.bak -e  "s|192.168.99.100.nip.io|$(ROUTING_SUFFIX)|g" ./deploy/registry/local/k8s/ingress.yaml
 	$(TOOL) apply -f ./deploy/registry/local/k8s
-	sed -i.bak -e "s|$(CLUSTER_IP)|192.168.99.100.nip.io|g" ./deploy/registry/local/k8s/ingress.yaml
+	sed -i.bak -e "s|$(ROUTING_SUFFIX)|192.168.99.100.nip.io|g" ./deploy/registry/local/k8s/ingress.yaml
 	rm ./deploy/registry/local/k8s/ingress.yaml.bak
 endif
 endif
@@ -59,6 +60,7 @@ _update_yamls: _set_registry_url
 	sed -i.bak -e "s|plugin.registry.url: .*|plugin.registry.url: http://$(PLUGIN_REGISTRY_HOST)|g" ./deploy/controller_config.yaml
 	sed -i.bak -e 's|che.webhooks.enabled: .*|che.webhooks.enabled: "$(WEBHOOK_ENABLED)"|g' ./deploy/controller_config.yaml
 	sed -i.bak -e 's|che.default_routing_class: .*|che.default_routing_class: "$(DEFAULT_ROUTING)"|g' ./deploy/controller_config.yaml
+	sed -i.bak -e 's|cluster.routing_suffix: .*|cluster.routing_suffix: $(ROUTING_SUFFIX)|g' ./deploy/controller_config.yaml
 	rm ./deploy/controller_config.yaml.bak
 ifeq ($(TOOL),oc)
 	sed -i.bak -e "s|image: .*|image: $(IMG)|g" ./deploy/os/controller.yaml
@@ -76,6 +78,7 @@ _reset_yamls: _set_registry_url
 	sed -i.bak -e "s|http://$(PLUGIN_REGISTRY_HOST)|http://che-plugin-registry.192.168.99.100.nip.io/v3|g" ./deploy/controller_config.yaml
 	sed -i.bak -e 's|che.webhooks.enabled: .*|che.webhooks.enabled: "false"|g' ./deploy/controller_config.yaml
 	sed -i.bak -e 's|che.default_routing_class: .*|che.default_routing_class: "basic"|g' ./deploy/controller_config.yaml
+	sed -i.bak -e 's|cluster.routing_suffix: .*|cluster.routing_suffix: 192.168.99.100.nip.io|g' ./deploy/controller_config.yaml
 	rm ./deploy/controller_config.yaml.bak
 ifeq ($(TOOL),oc)
 	sed -i.bak -e "s|image: $(IMG)|image: quay.io/che-incubator/che-workspace-controller:nightly|g" ./deploy/os/controller.yaml
@@ -176,15 +179,15 @@ fmt:
 .PHONY: help
 ### help: print this message
 help: Makefile
-	@echo "Available rules:"
+	@echo 'Available rules:'
 	@sed -n 's/^### /    /p' $< | awk 'BEGIN { FS=":" } { printf "%-22s -%s\n", $$1, $$2 }'
-	@echo ""
-	@echo "Supported environment variables:"
-	@echo "    IMG                - Image used for controller"
-	@echo "    NAMESPACE          - Namespace to use for deploying controller"
-	@echo "    TOOL               - CLI tool for interfacing with the cluster: kubectl or oc; if oc is used, deployment is tailored to OpenShift, otherwise Kubernetes"
-	@echo "    CLUSTER_IP         - For Kubernetes only, the ip address of the cluster (minikube ip)"
-	@echo "    PULL_POLICY        - Image pull policy for controller"
-	@echo "    WEBHOOK_ENABLED    - Whether webhooks should be enabled in the deployment"
-	@echo "    ADMIN_CTX          - Kubectx entry that should be used during work with cluster. The current will be used if omitted"
-	@echo "    REGISTRY_ENABLED   - Whether the plugin registry should be deployed"
+	@echo ''
+	@echo 'Supported environment variables:'
+	@echo '    IMG                - Image used for controller'
+	@echo '    NAMESPACE          - Namespace to use for deploying controller'
+	@echo '    TOOL               - CLI tool for interfacing with the cluster: kubectl or oc; if oc is used, deployment is tailored to OpenShift, otherwise Kubernetes'
+	@echo '    ROUTING_SUFFIX     - Cluster routing suffix (e.g. $$(minikube ip).nip.io, apps-crc.testing)'
+	@echo '    PULL_POLICY        - Image pull policy for controller'
+	@echo '    WEBHOOK_ENABLED    - Whether webhooks should be enabled in the deployment'
+	@echo '    ADMIN_CTX          - Kubectx entry that should be used during work with cluster. The current will be used if omitted'
+	@echo '    REGISTRY_ENABLED   - Whether the plugin registry should be deployed'
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -72,8 +72,8 @@ func (wc *ControllerConfig) GetPluginRegistry() string {
 	return wc.GetPropertyOrDefault(pluginRegistryURL, "")
 }
 
-func (wc *ControllerConfig) GetIngressGlobalDomain() string {
-	return wc.GetPropertyOrDefault(ingressGlobalDomain, defaultIngressGlobalDomain)
+func (wc *ControllerConfig) GetRoutingSuffix() string {
+	return wc.GetPropertyOrDefault(routingSuffix, defaultRoutingSuffix)
 }
 
 func (wc *ControllerConfig) GetPVCStorageClassName() *string {
@@ -218,7 +218,7 @@ func buildDefaultConfigMap(cm *corev1.ConfigMap) {
 	cm.Namespace = ConfigMapReference.Namespace
 
 	cm.Data = map[string]string{
-		ingressGlobalDomain:        defaultIngressGlobalDomain,
+		routingSuffix:              defaultRoutingSuffix,
 		pluginArtifactsBrokerImage: defaultPluginArtifactsBrokerImage,
 	}
 }
@@ -252,7 +252,7 @@ func fillOpenShiftRouteSuffixIfNecessary(nonCachedClient client.Client, configMa
 	host := testRoute.Spec.Host
 	if host != "" {
 		prefixToRemove := "che-workspace-controller-test-route-" + configMap.Namespace + "."
-		configMap.Data["ingress.global.domain"] = strings.TrimPrefix(host, prefixToRemove)
+		configMap.Data[routingSuffix] = strings.TrimPrefix(host, prefixToRemove)
 	}
 
 	err = nonCachedClient.Update(context.TODO(), configMap)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"errors"
 	"github.com/che-incubator/che-workspace-operator/internal/cluster"
+	"github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
 	"os"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"strings"
@@ -118,6 +119,13 @@ func (wc *ControllerConfig) GetPropertyOrDefault(name string, defaultValue strin
 		return val
 	}
 	return defaultValue
+}
+
+func (wc *ControllerConfig) Validate() error {
+	if !wc.isOpenShift && wc.GetDefaultRoutingClass() == string(v1alpha1.WorkspaceRoutingOpenShiftOauth) {
+		return fmt.Errorf("controller appears to be running in non-OpenShift cluster, but default routing class is '%s'", v1alpha1.WorkspaceRoutingOpenShiftOauth)
+	}
+	return nil
 }
 
 func updateConfigMap(client client.Client, meta metav1.Object, obj runtime.Object) {
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -42,6 +42,9 @@ const (
 	// WorkspaceIDLabel is label key to store workspace identifier
 	WorkspaceIDLabel = "che.workspace_id"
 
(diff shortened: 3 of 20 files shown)


pkg/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -175,33 +176,95 @@ func (r *ReconcileWorkspaceRouting) Reconcile(request reconcile.Request) (reconc
...
+			if len(ingress.Spec.Rules) == 1 {
+				protocol := endpoint.Attributes[workspacev1alpha1.PROTOCOL_ENDPOINT_ATTRIBUTE]
+				url := fmt.Sprintf("%s://%s", protocol, ingress.Spec.Rules[0].Host)
+				return url, nil
+			} else {
+				return "", fmt.Errorf("ingress %s contains multiple rules", ingress.Name)
+			}
+		}
+	}
+	return "", fmt.Errorf("could not get URL for endpoint %s", endpoint.Name)
```
sleshchenko, line 265: Is it really an error? While workspace is starting I got quite a lot of spam
Screenshot_20200414_152116

Are we able to return only that endpoints which are already exposed and schedule requeuing? 
other (without ingress/route) could be skipped or returned with status - Exposing (not sure if we have endpoint status now).

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
```diff
@@ -60,41 +53,27 @@ func (s *OpenShiftOAuthSolver) GetSpecObjects(spec v1alpha1.WorkspaceRoutingSpec
...
-			endpointName := common.EndpointName(endpoint.Name)
-			hostname := common.EndpointHostname(workspaceMeta.WorkspaceId, endpointName, endpoint.Port, workspaceMeta.IngressGlobalDomain)
-			url := fmt.Sprintf("%s://%s", getSecureProtocol(endpoint.Attributes[v1alpha1.PROTOCOL_ENDPOINT_ATTRIBUTE]), hostname)
-
-			// NOTE: openshift oauth-proxy only supports listening on a single port; as a result, we can't proxy more than
-			// one endpoint or we run into cookie issues (each proxied port gets a container and sets a cookie in the browser
-			// once auth is completed).
 			var tls *routeV1.TLSConfig = nil
 			if endpoint.Attributes[v1alpha1.SECURE_ENDPOINT_ATTRIBUTE] == "true" {
 				if endpoint.Attributes[v1alpha1.TYPE_ENDPOINT_ATTRIBUTE] == "terminal" {
```
sleshchenko, line 79: It seems to be workaround. Recently I've seen an issue on Che Server side where someone asks about an ability to configure InsecureEdgeTerminationPolicy on the endpoint level...
Won't it work if we use like TLSTerminationReencrypt for every endpoint?

pkg/config/constants.go
```diff
@@ -42,6 +42,9 @@ const (
 	// WorkspaceIDLabel is label key to store workspace identifier
 	WorkspaceIDLabel = "che.workspace_id"
 
+	// WorkspaceEndpointNameLabel is the label key for storing an endpoint's name from the devfile representation
+	WorkspaceEndpointNameLabel = "che.workspace.endpoint_name"
```
sleshchenko, line 46: It would mean that we're making endpoint name DNS name compatible. Do we really want to do it?
AFAIS we don't fetch objects by this label, so we're able to use annotation which is less strict in the values format.

Storing WorkspaceEndpointName would mean that we create one Ingress/Route per endpoint but after we reached Route number limit on OSIO, we started reuse Routes if possible...
Maybe we should store EndpointNames instead.

pkg/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Makes sense -- where are you running the operator? It may be a route/ingress exposure issue (i.e. it takes longer than expected). I don't see this on `crc`.

I'll add this logic to the PR.

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
amisevsk: I'm not sure, the logic here is copied from long ago and I haven't had time to dig into the why of it all. Perhaps input from @davidfestal would be helpful here.

pkg/config/constants.go
amisevsk: Makes sense, though note that we're already making endpoint names DNS compatible, since we use them as a part of the route/ingress name.

We can assign them as annotations as well, I'll do that.

I'm not clear on what you mean by 
> Storing WorkspaceEndpointName would mean that we create one Ingress/Route per endpoint but after we reached Route number limit on OSIO, we started reuse Routes if possible...

On OpenShift, routes only support a single port, so we'll need to create one per endpoint in any case. I would consider merging endpoints into a single route (which I suppose we could do with e.g. the webviews endpoint) as a separate PR -- WDYT?

pkg/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Looking into this a bit more, I'm not sure what's causing the issue; we only return that error when we cannot find any ingress with the appropriate label. If the url field is empty, we just return empty.

amisevsk:
Added:
- Use annotations for storing endpoint name instead of labels, use original endpoint name in annotation instead of DNS-sanitized version
- Fix makefile not resetting registry ingress hostname when deploying to k8s
- Don't try to sync routes when we're running on k8s, as we cannot list Routes
- Add some validation to controller config: fail startup if we're not in OpenShift but use openshift-oauth as default routing
- Set workspace routing phase to "Preparing" when we can't get route/ingress URLs.

pkg/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: I tried it on local crc

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
sleshchenko: BTW it's out of the PR scope. I'll create a separate issue for that

pkg/config/constants.go
sleshchenko: Well, theia and webview it's another case since we need unique hosts there and it's why the corresponding attribute is there https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-theia/7.11.0/meta.yaml#L33

I briefly checked our plugins and devfile and sharing host by endpoint is more imaginary case then currently used.
But theoretically I wonder if it makes sense for cloud-shell to have two different endpoint:
```
spec:
  endpoints:
   -  name: cloud-shell
      public: true
      targetPort: 4444
      attributes:
        protocol: http
        type: ide
        discoverable: false
        secure: true
        cookiesAuthEnabled: true
   -  name: machine-exec
      public: true
      targetPort: 4444
      attributes:
        protocol: ws
        type: terminal
        discoverable: false
        secure: true
        cookiesAuthEnabled: true
```

>  I would consider merging endpoints into a single route (which I suppose we could do with e.g. the webviews endpoint) as a separate PR -- WDYT?

:+1: 
If now we create Ingress/Route per endpoint - is then https://github.com/eclipse/che/issues/15990 actual?

pkg/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: > Looking into this a bit more, I'm not sure what's causing the issue; we only return that error when we cannot find any ingress with the appropriate label

Exactly. Now I see the same error but only once
route-not-found

So, instead of failing reconciling loop it would be better to make endpoint as problematic without host and retry later. I think it worths the dedicated PR because it needs introducing some phases,conditions for endpoints.

sleshchenko: This error also appears after removing the workspacerouting routeScreenshot_20200415_133739

pkg/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -175,33 +175,109 @@ func (r *ReconcileWorkspaceRouting) Reconcile(request reconcile.Request) (reconc
...
 		return nil
 	}
 	instance.Status.Phase = workspacev1alpha1.RoutingReady
 	instance.Status.PodAdditions = routingObjects.PodAdditions
-	instance.Status.ExposedEndpoints = routingObjects.ExposedEndpoints
+	instance.Status.ExposedEndpoints = exposedEndpoints
 	return r.client.Status().Update(context.TODO(), instance)
 }
 
+func getExposedEndpoints(
```
sleshchenko, line 222: consider moving endpoints related funcs into a separate file, like to be sync with other helper files -> `resolve_endpoints`. Does it make sense?

pkg/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -175,33 +175,109 @@ func (r *ReconcileWorkspaceRouting) Reconcile(request reconcile.Request) (reconc
...
 
-	routesInSync, err := r.syncRoutes(instance, routes)
+	routesInSync, clusterRoutes, err := r.syncRoutes(instance, routes)
 	if err != nil || !routesInSync {
 		reqLogger.Info("Routes not in sync")
 		return reconcile.Result{Requeue: true}, err
 	}
 
-	return reconcile.Result{}, r.reconcileStatus(instance, routingObjects)
+	exposedEndpoints, endpointsReady, err := getExposedEndpoints(instance.Spec.Endpoints, clusterIngresses, clusterRoutes)
```
sleshchenko, line 190: `exposedEndpoints, endpointsReady` makes an impression that two lists are returned - all exposed, only ready ones.

`exposedEndpoints, areReady` might be a bit better, might be not. Up to you

pkg/controller/workspacerouting/solvers/common.go
```diff
@@ -81,87 +81,105 @@ func getServicesForEndpoints(endpoints map[string][]v1alpha1.Endpoint, workspace
...
+			Selector: meta.PodSelector,
 			Type:     corev1.ServiceTypeClusterIP,
 		},
 	})
 
 	return services
 }
 
-func getIngressesForSpec(endpoints map[string][]v1alpha1.Endpoint, workspaceMeta WorkspaceMetadata) ([]v1beta1.Ingress, map[string][]v1alpha1.ExposedEndpoint) {
+func getRoutingForSpec(endpoints map[string][]v1alpha1.Endpoint, meta WorkspaceMetadata) ([]v1beta1.Ingress, []routeV1.Route) {
```
sleshchenko, line 100: AFAIU we only need ingresses or routes but not both at the same time. Then I wonder if we're able to build some general interface to work with Routing and that just assign right implementation according to the current platform, like:
```go
struct Routing {
  Name string
  Host string
  Target ServicePortReference
}
interface RoutingManager {
  generateForEndpoint(Endpoint) runtime.Object
  sync([]runtime.Object) error
  resolveURLForEndpoint([]runtime.Object, Endpoint) string, err
}
```

or simpler in the implementation terms ( no needs to cast objects and play with their types)
```go
struct Routing {
  Route routev1.Route
  Ingress ingressv1.Ingress
}
interface RoutingManager {
  generateForEndpoint(Endpoint) Routing
  sync([]Routing) error
  resolveURLForEndpoint([]Routing, Endpoint) string, err
}
```

It would make code simpler in some places but make more difficult in others...
We should consider such refactoring later to play and implement if we like it.

pkg/controller/workspacerouting/workspacerouting_controller.go
amisevsk: @sleshchenko I think the error you describe initially (`could not get URL...`) *is* a failure; we should never reach that point in the reconcile and not have routes/ingresses available. The previous step, prior to attempting this matching, is to make sure routes/ingresses are in sync between the cluster and spec. Not being able to match the two means that we have an endpoint (which defines the spec), that does not have a route/ingress associated with it. Until we have a concrete case where this can occur, it should mark the workspace as failed.

The latter failure, `already exists`, is a kind of familiar issue -- I haven't gotten around to implementing it, but we shouldn't log that error and just continue. It just means the state of the cluster has changed since we started our reconcile loop. The correct solution there would be to just requeue if `errors.IsConflict(err)` or `errors.IsAlreadyExists(err)`

amisevsk: > Exactly. Now I see the same error but only once

This is a different error message, to be clear.

pkg/config/constants.go
amisevsk: > If now we create Ingress/Route per endpoint - is then eclipse/che#15990 actual?

Technically no, but I can't test since moving past 7.8.0 is blocked by https://github.com/eclipse/che/issues/16568.

Makefile
```diff
@@ -163,14 +167,14 @@ fmt:
...
-	@echo "    CLUSTER_IP         - For Kubernetes only, the ip address of the cluster (minikube ip)"
-	@echo "    PULL_POLICY        - Image pull policy for controller"
-	@echo "    WEBHOOK_ENABLED    - Whether webhooks should be enabled in the deployment"
-	@echo  "    ADMIN_CTX          - Kubectx entry that should be used during work with cluster. The current will be used if omitted"
+	@echo ''
+	@echo 'Supported environment variables:'
+	@echo '    IMG                - Image used for controller'
+	@echo '    NAMESPACE       http://127.0.0.1:44299/api/v1/namespaces/kubernetes-dashboard/services/http:kubernetes-dashboard:/proxy/#/overview?namespace=che-workspace-controller   - Namespace to use for deploying controller'
+	@echo '    TOOL               - CLI tool for interfacing with the cluster: kubectl or oc; if oc is used, deployment is tailored to OpenShift, otherwise Kubernetes'
+	@echo '    ROUTING_SUFFIX     - Cluster routing suffix (e.g. $$(minikube ip).nip.io, apps-crc.testing)'
```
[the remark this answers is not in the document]
amisevsk: The issue I've had is that *not* setting it correctly initially sometimes (but not always) results in the wrong routing suffix being used, even on OpenShift. I'll create an issue if I can find a consistent reproducer.

The functionality for rewriting routingSuffix (formerly ingress.global.domain) has existed since we started working on the project but we've all been bit by it being set incorrectly at some point.

pkg/controller/workspacerouting/solvers/common.go
amisevsk: Yeah; my primary concern is that this kind of repetition is built into the language. I want to try to avoid shoe-horning OOP-style thinking of "make everything an interface" unless it really increases clarity. 

If we want to start thinking about abstracting out interfaces, the `getClusterX` and `syncX` functions are all practically identical:
```
syncX(workspace, ...) {
  spec := getSpecForX()
  cluster := getClusterForX(name, namespace, client)
  return syncSpecAndCluster(spec, cluster, client)
```
but I'm not ready to fight golang to implement it cleanly. We currently have 5-6 sets of functions where the only difference is the type signature, but abstracting that to an interface means dealing with `runtime.Object` and casting.

pkg/config/config.go
```diff
@@ -252,7 +260,7 @@ func fillOpenShiftRouteSuffixIfNecessary(nonCachedClient client.Client, configMa
 	host := testRoute.Spec.Host
 	if host != "" {
 		prefixToRemove := "che-workspace-controller-test-route-" + configMap.Namespace + "."
```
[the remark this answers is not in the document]
amisevsk: It's legacy functionality I didn't even know existed until I worked on this PR; I'm inclined to remove it entirely if there's doubts as to its usefulness. Thus far, the concerns are
- It depends on deployed namespace being `che-workspace-controller`, will fail if we change that
- Will fail if how default hostnames are computed is changed
- Only works on OpenShift

This kind of points towards "we shouldn't support this at the moment". Executing this only when the entry in the cm is blank means that we just move all those failure cases down the line, except it makes it look like we're changing config requirements rather than the cluster is changing how it resolves things.

Regarding 
>Also, che-workspace-controller-test-route-che-workspace-controller is near to drain 64 chars limit ) Consider making test route name shorter.

This is actually not a problem; OpenShift will generously generate an invalid hostname and only tell you its a problem later:
```
$ oc get route che-plugin-registry-abcdefghijklmnopqrstuvwxyz -o yaml | grep host:
  host: che-plugin-registry-abcdefghijklmnopqrstuvwxyz-che-workspace-controller.apps-crc.testing
      message: 'host name validation errors: spec.host: Invalid value: "che-plugin-registry-abcdefghijklmnopqrstuvwxyz-che-workspace-controller.apps-crc.testing":
    host: che-plugin-registry-abcdefghijklmnopqrstuvwxyz-che-workspace-controller.apps-crc.testing
```

However, it might make sense to try and use `.status.ingress[].routerCanonicalHostname`:
```
$ oc get route che-plugin-registry-abcdefghijklmnopqrstuvwxyz -o yaml | yq '.status.ingress[].routerCanonicalHostname'
"apps-crc.testing"
```

pkg/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Moved these functions to `resolve_endpoints.go`

pkg/controller/workspacerouting/workspacerouting_controller.go
amisevsk: renamed `endpointsReady` -> `endpointsAreReady`. I prefer to keep `endpoints` in the name as otherwise we just have a variable `areReady` floating around. WDYT?

pkg/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: > This is a different error message, to be clear.

My fault. But seems there were two tries to create ingress and it can indicate in the issues in the reconciling loop, because I did not create such a route but hand.

Will test more precisely. BTW it's not a blocker

pkg/config/config.go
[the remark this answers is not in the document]
sleshchenko: > This is actually not a problem; OpenShift will generously generate an invalid hostname and only tell you its a problem later :) :

:-D It's true and it may be strange but it helps use in this case

pkg/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: Yeah, it's `already exists` issue and seems routes are not created immediately 
Screenshot_20200416_151950

RetryAfter 200 helps me to solve an issue. 100 is not enough... 
```go
	routesInSync, clusterRoutes, err := r.syncRoutes(instance, routes)
	if err != nil || !routesInSync {
		reqLogger.Info("Routes not in sync")
		return reconcile.Result{ RequeueAfter: 100 * time.Millisecond}, err
	}
```
But not sure if it's a really good solution since this limit depends on the infrastructure, I'm OK with leaving this error propagated constantly for time being.

amisevsk: The main thing stopping us implementing this is that it's going to be a fair bit of boilerplate error checking; it might also be improved by eventually filtering reconciles (by determining if changes are necessary)

pkg/config/config.go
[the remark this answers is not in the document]
amisevsk: Looked into using the status field, but it's not populated immediately. For now I'm leaving the functionality in place, but we should test its reliability in the future and potentially remove/improve it.

pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
sleshchenko: @amisevsk FYI here is an issue for supporting tls-termination passtrough in the devfile and probably in the plugin-meta.yaml...

PR #51, opened 2020-04-14
Provision pvc only if needed

JPinkney:
### What does this PR do?
This PR makes it so that that a PVC is only created when it's needed. It does this by delaying the generation of the PVC and checking the contents of the devfile to see if it should create the PVC or not.

It was built on top of the internal-registry PR so it will be easier once that PR is merged. For the meantime heres a link to the commit: https://github.com/che-incubator/che-workspace-operator/commit/696995bc9ac529e0bd7c34ee96fad0fd7279fbdf

https://github.com/eclipse/che/issues/16530

### Is it tested? How?

Tested on crc with cloud-shell -> made sure pvc is not there
Tested on crc with go-sample -> made sure pvc is created


pkg/controller/workspace/prerequisites/check_prerequisites.go
```diff
@@ -14,50 +14,48 @@ package prerequisites
...
-					reqLogger.Error(err, "")
-				}
-			}
 		}
 	}
 	return nil
 }
+
+// ProvisionPrereqs provisions the runtime objects
+func ProvisionPrereqs(prereq runtime.Object, prereqAsMetaObject metav1.Object, client client.Client, reqLogger logr.Logger) error {
```
sleshchenko, line 46: I wonder if we still need prerequisites entity anymore...

pkg/controller/workspace/prerequisites/generate_prerequisites.go
```diff
@@ -14,37 +14,13 @@ package prerequisites
...
-				},
-				Resources: corev1.ResourceRequirements{
-					Requests: corev1.ResourceList{
-						"storage": pvcStorageQuantity,
-					},
-				},
-				StorageClassName: config.ControllerCfg.GetPVCStorageClassName(),
-			},
-		},
+func generatePrerequisites(namespace string) []runtime.Object {
```
sleshchenko, line 22: It would be more precisely if we name if RBAC, Roles&RolesBinding...

pkg/controller/workspace/provision/pvc.go
```diff
@@ -0,0 +1,76 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsPVCRequired checks to see if we need a PVC for the given devfile. If cloud-shell is present in the devfile,
+// we do not need PVC else we do
+func IsPVCRequired(components []v1alpha1.ComponentSpec) bool {
+	for _, comp := range components {
+		if strings.Contains(comp.Id, config.CloudShellID) && comp.Type == v1alpha1.CheEditor {
```
sleshchenko, line 34: I think it would be a bit smarter if we check volumes in podsAdditional instead of sticking to component id...

sleshchenko: Here is my draft proposal https://github.com/sleshchenko/che-workspace-operator/commit/91206b47e32c761ce1bfa79edc0d5ad1abf27384
Maybe we even should get rid of object.go#SyncObject because each object type should be sync in different way: like PVC is immutable, like you can't assign new size.
While probably we should update Roles and RolesBindings...

JPinkney: I've just been debugging this for a bit and it looks like podAdditions.Volumes always has length 0 for some reason meaning the PVC will never be created.

If we compare the volumeMounts to the PVCVolumeName it works though:
```go
func IsPVCRequired(components []v1alpha1.ComponentDescription) bool {
	volumeName := config.ControllerCfg.GetWorkspacePVCName()
	for _, comp := range components {
		for _, cont := range comp.PodAdditions.Containers {
			for _, vm := range cont.VolumeMounts {
				if vm.Name == volumeName {
					return true
				}
			}
		}
	}
	return false
}
```

pkg/controller/workspace/provision/deployment.go
```diff
@@ -53,11 +54,12 @@ func SyncDeploymentToCluster(
 	workspace *v1alpha1.Workspace,
 	podAdditions []v1alpha1.PodAdditions,
 	saName string,
-	clusterAPI ClusterAPI) DeploymentProvisioningStatus {
+	clusterAPI ClusterAPI,
+	isPVCRequired bool) DeploymentProvisioningStatus {
```
amisevsk, line 58: Considering that `isPVCRequired` is in the `provision` package, and depends only on the workspace spec, we can just repeat the check where it's needed (i.e. in `getSpecDeployment`). This would keep the signature for `SyncDeploymentToCluster` simpler and would allow `workspace_controller.go` to be simplified
```
	if provision.IsPVCRequired(workspace.Spec.Devfile.Components) {
		provision.GeneratePVC(workspace, r.client, reqLogger)
	}
```

pkg/controller/workspace/provision/pvc.go
```diff
@@ -0,0 +1,76 @@
...
+func GeneratePVC(workspace *v1alpha1.Workspace, client client.Client, reqLogger logr.Logger) error {
+
+	pvcStorageQuantity, err := resource.ParseQuantity(config.PVCStorageSize)
+	if err != nil {
+		return err
+	}
+
+	pvc := []runtime.Object{&corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      config.ControllerCfg.GetWorkspacePVCName(),
```
amisevsk, line 51: It's still a bit strange to be optionally creating a PVC but never cleaning it up, but this is a side-effect of basically using the `common` strategy from che-server. With the current set up, your namespace won't get a PVC until you start a non-cloudshell workspace, at which point the PVC will exist until you manually delete it or delete the namespace.

No suggestions here, but it's something we should think about.

pkg/controller/workspace/prerequisites/check_prerequisites.go
amisevsk, line 46: nitpick: should be named `provisionPrereq` since we only handle one at a time.

pkg/controller/workspace/provision/pvc.go
amisevsk: :+1: on checking for the pvc volume mount.

The reason you're not finding it in `podAdditions.Volumes` is because we currently handle it separately -- otherwise we'd have to deduplicate the list of `podAdditions.Volumes`, which would prevent us catching issues like two components trying to add the same volume.

pkg/controller/workspace/provision/pvc.go
```diff
@@ -0,0 +1,72 @@
+package provision
```
sleshchenko, line 1: Please add license headers

pkg/controller/workspace/provision/pvc.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if err := SyncObject(pvc, client, reqLogger); err != nil {
+		return err
+	}
+	return nil
+}
+
+func generatePVC(workspace *v1alpha1.Workspace, components []v1alpha1.ComponentDescription) (*corev1.PersistentVolumeClaim, error) {
+	if !IsPVCRequired(components) {
```
sleshchenko, line 30: it seems a bit redundant since a caller should/and already make sure that PVC is required before generating it.

pkg/controller/workspace/provision/pvc.go
```diff
@@ -0,0 +1,72 @@
...
+					"storage": pvcStorageQuantity,
+				},
+			},
+			StorageClassName: config.ControllerCfg.GetPVCStorageClassName(),
+		},
+	}, nil
+}
+
+// IsPVCRequired checks to see if we need a PVC for the given devfile.
+// If there is any PodAdditions with Volume - we need PVC, otherwise we don't
```
sleshchenko, line 59: The doc seems a bit outdated.

pkg/controller/workspace/provision/rbac.go
```diff
@@ -1,50 +1,27 @@
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func generatePrerequisites(namespace string) ([]runtime.Object, error) {
-	pvcStorageQuantity, err := resource.ParseQuantity(config.PVCStorageSize)
-	if err != nil {
-		return nil, err
+func SyncRBAC(workspace *v1alpha1.Workspace, client client.Client, reqLogger logr.Logger) (err error) {
```
sleshchenko, line 13: Please add a few lines of documentation

pkg/controller/workspace/workspace_controller.go
```diff
@@ -212,6 +204,15 @@ func (r *ReconcileWorkspace) Reconcile(request reconcile.Request) (reconcileResu
 		componentDescriptions = append(componentDescriptions, cheRestApisComponent)
 	}
 
+	if err := provision.SyncPVC(workspace, componentDescriptions, r.client, reqLogger); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	//TODO Previously we check rbac before components creation. Now we are able to do the same but what the purpose of creating RBAC if like there is component with unreached plugin/editor?
```
sleshchenko, line 211: 1. Just remove this TODO or 2. move to old place.
+1 for 1 option but up to you

pkg/controller/workspace/provision/pvc.go
amisevsk: I ended up using [addlicense](https://github.com/google/addlicense) plus a license template to add headers to my rework PR -- perhaps we should include the header in the repo itself to ease this process/automate a check.

JPinkney: +1 to that

amisevsk: :tada: https://github.com/che-incubator/che-workspace-operator/pull/54

PR #52, opened 2020-04-15
Use configmap for che-rest-apis requirements

amisevsk:
### What does this PR do?
- Provision configmap during reconcile that contains the runtime json annotation and devfile yaml
- Mount this configmap into the che-rest-apis sidecar
- Remove runtime annotation from workspace status
- Remove roles/rolebindings required for che-rest-apis to read workspace CR

Depends on https://github.com/che-incubator/che-api-sidecar/pull/2

### Is it tested? How?
Tested on `crc`, will test more deeply before merging.

Current changeset is built in image `amisevsk/che-workspace-controller:rest-apis-cm`. To test the changes it's necessary to also patch the configmap to use the patched che-rest-apis image:

```bash
export IMG=amisevsk/che-workspace-controller:rest-apis-cm
# set other env vars as required
make deploy
oc patch cm/che-workspace-controller --patch='{"data":{"cherestapis.image.name": "amisevsk/che-rest-apis:configmap"}}'
make restart
```


pkg/controller/workspace/prerequisites/generate_prerequisites.go
```diff
@@ -58,19 +58,6 @@ func generatePrerequisites(namespace string) ([]runtime.Object, error) {
 				},
 			},
 		},
-		&rbacv1.Role{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "view-workspaces",
```
sleshchenko, line 63: Reading Workspace CR directly has its own benefit: Rest API Sidecar always gets the latest changes. With this approach Theia won't be restarted if command is updated, and Rest API Sidecar won't get the latest changes.

So, in general I would leave reading Workspace CR but try to avoid using Runtime additional field in the WorkspaceStatus, instead I would investigate the possible model for WorkspaceStatus to cover Rest API needs and make sidecar able to generate Runtime for itself based on the status. Let me imagine:
```yaml
kind: Workspace
spec:
  started: true
  devfile:
    components:
      - type: cheEditor
        alias: theia
        id: eclipse/che-theia/nightly
      - type: dockerimage
        alias: maven
        dockerimage: maven:3.10.0
    commands:
      - name: build
        componenent: maven
        exec: mvn clean install
status:
  editorURL: "https://workspace123-theia.apps-testing.com"
  endpoints:
    - name: ide
      component: theia
      url: https://workspace123-theia.apps-testing.com
      status: Exposing
      type: ide
    - name: webviews
      component: theia
      url: https://workspace123-webviews.apps-testing.com
      status: Unknown
 containers:
    - name: theia
      status: Running
      image: eclipse/che-theia:next
      component: theia
       #? podName: ?
    - component: maven
      image: maven:3.10.0
      resources:
        limit:
          memory: 1Gi
          cpu: 2
#^ commands are not in the status since Sidecar is able to match commands from spec to containers from status using component field.
```

amisevsk: Mounting a configmap in the way we do correctly propagates changes to che-rest-apis:

update-workspace-propagation

pkg/common/naming.go
```diff
@@ -50,3 +50,7 @@ func RouteName(workspaceId, endpointName string) string {
 func OAuthProxySecretName(workspaceId string) string {
 	return "proxy-tls-" + workspaceId
 }
+
+func CheRestAPIsConfigmapName(workspaceId string) string {
+	return "che-rest-apis-" + workspaceId
```
sleshchenko, line 55: Consider alligning naming, I prefer workspaceId to be at the beginning a bit more but it's not a strong preference.

Screenshot_20200416_145007

pkg/controller/workspace/provision/restApisConfigmap.go
```diff
@@ -0,0 +1,171 @@
...
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package provision
```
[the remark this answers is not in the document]
JPinkney: +1 on using snake case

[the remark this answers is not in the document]
amisevsk: Renamed to snake case.

Regarding moving to a separate package, the reason it's in provision currently is so we can reuse `ProvisioningStatus` more directly -- the overall action and how it's treated is not too different. I do agree though, that having the functionality spread across files in separate packages is pretty bad.

I've made the following rearrangements:
- All che-rest-apis files are in `workspace/restapis`
- Moved ideUrl.go into status.go for simplicity.

pkg/common/naming.go
amisevsk: Changed all names in common/naming.go to use workspaceId as prefix.

amisevsk:
Tested again with latest changes and with new default rest-apis image on `crc`.

PR #53, opened 2020-04-15
Update Makefile so that it works on OS X

JPinkney:

### What does this PR do?
This PR makes it so that you can run the Makefile on both OS X and Linux. Previously you would get 

```
sed: 1: "./deploy/controller_con ...": invalid command code .
```

because of incompatibility between both sed and date.

### Is it tested? How?

Tested on both OS X and Linux to see that the yamls are changed correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ else
 	sed -i.bak -e "s|192.168.99.100|$(CLUSTER_IP)|g" ./deploy/registry/local/k8s/ingress.yaml
 	rm ./deploy/registry/local/k8s/ingress.yaml.bak
 	$(TOOL) apply -f ./deploy/registry/local/k8s
+	sed -i "s|$(ROUTING_SUFFIX)|192.168.99.100.nip.io|g" ./deploy/registry/local/k8s/ingress.yaml
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -40,7 +40,8 @@ ifeq ($(REGISTRY_ENABLED),true)
 ifeq ($(TOOL),oc)
 	$(TOOL) apply -f ./deploy/registry/local/os
 else
-	sed -i "s|192.168.99.100|$(CLUSTER_IP)|g" ./deploy/registry/local/k8s/ingress.yaml
+	sed -i.bak -e "s|192.168.99.100|$(CLUSTER_IP)|g" ./deploy/registry/local/k8s/ingress.yaml
+	rm ./deploy/registry/local/k8s/ingress.yaml.bak
```
amisevsk, line 44: Could you also mirror this commit from my PR: https://github.com/che-incubator/che-workspace-operator/pull/50/commits/83cb27d073d6d47cf474107e19aef2b6652bf75f? We're currently not resetting `k8s/ingress.yaml` after creating the route.

benoitf:
FYI in chectl I used https://github.com/che-incubator/chectl/blob/e747a60e2f46a39f38a739e10d9c7a5410fed2e2/make-release.sh#L29-L36

it avoids to create ugly .bak files

PR #59, opened 2020-04-23
Patch deployment when using webhooks for kubernetes

JPinkney:

### What does this PR do?
This PR makes it so that we patch the deployment when we are using webhooks with kubernetes

### Is it tested? How?

Tested the changes on minikube

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ _apply_controller_cfg:
 ifeq ($(TOOL),oc)
 	$(TOOL) apply -f ./deploy/os/
 else
-	$(TOOL) apply -f ./deploy/k8s/
+	$(TOOL) apply -f ./deploy/k8s/controller.yaml
 endif
 
 _do_restart:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -133,12 +133,10 @@ docker: _print_vars
...
 
-### webhook: generate certificates for webhooks and deploy to cluster; no-op if webhooks are disabled or running on OpenShift
+### webhook: generate certificates for webhooks and deploy to cluster; no-op if running on OpenShift
 webhook:
-ifeq ($(WEBHOOK_ENABLED),true)
 ifeq ($(TOOL),kubectl)
 	./deploy/webhook-server-certs/deploy-webhook-server-certs.sh kubectl
-endif
 else
 	@echo "Webhooks disabled, skipping certificate generation"
```
sleshchenko, line 141: This message going to appear when deploying on OpenShift and it seems not to be actual anymore.

Makefile
```diff
@@ -133,12 +133,10 @@ docker: _print_vars
 	docker build -t $(IMG) -f ./build/Dockerfile .
 	docker push $(IMG)
 
-### webhook: generate certificates for webhooks and deploy to cluster; no-op if webhooks are disabled or running on OpenShift
+### webhook: generate certificates for webhooks and deploy to cluster; no-op if running on OpenShift
 webhook:
-ifeq ($(WEBHOOK_ENABLED),true)
 ifeq ($(TOOL),kubectl)
 	./deploy/webhook-server-certs/deploy-webhook-server-certs.sh kubectl
```
sleshchenko, line 139: I wonder if it makes sense optionally generate TLS certs as previously but instead extract volume and volumes bound from main k8s controller yaml, like:
`kubectl patch -f ./deploy/k8s/controller-tls.yaml --merge` ->
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: che-workspace-controller
  namespace: che-workspace-controller
spec:
  template:
    spec:
      containers:
        - name: che-workspace-controller
          volumeMounts:
            - name: webhook-tls-certs
              mountPath: /tmp/k8s-webhook-server/serving-certs
              readOnly: true
      volumes:
        - name: webhook-tls-certs
          secret:
            secretName: webhook-server-tls
```
Not sure if it's going to work

JPinkney: I was able to get this to work and have updated the PR!

PR #60, opened 2020-04-23
Initialize KubeConfig in Containers

JPinkney:
### What does this PR do?
~~This PR does two things:~~
~~1. Mounts a empty dir volume where the kubeconfig is going to live (config lives at `/tmp/.kube/config` by default), this basically makes it so that we know /tmp/.kube will exist when we are creating the kubeconfig in che-machine-exec~~
~~2. Sets the kubeconfig Env variable that will be propagated to the containers in the workspace pod.~~

This PR has been changed so that both the meta.yaml and the cloud-shell devfile set the KubeConfig Env variable. The Env variable is needed in the meta.yaml so that che-machine-exec will know where to create the kubeconfig. The kubeconfig in the cloud-shell devfile is set so that oc/kubectl knows where the kubeconfig is. Since che-machine-exec is doing the creating of the kubeconfig, the volume allows the configs to be shared between che-machine-exec container and the dev container. The volume and volumeMounts are now only added when KUBECONFIG env variable is set.

The creation of the kubeconfig itself had to be done on che-machine-exec side: https://github.com/eclipse/che-machine-exec/pull/95

Before:
cloud-shell-before

After:
cloud-shell-after

https://github.com/eclipse/che/issues/16576

### Is it tested? How?

Tested on CRC with webhooks enabled + openshift-oauth and replaced the che-machine-exec image in the internal registry to be `jpinkney/che-machine-exec:latest`

### DIFF
--- a/pkg/controller/workspace/provision/deployment.go
+++ b/pkg/controller/workspace/provision/deployment.go
@@ -134,9 +134,18 @@ func getSpecDeployment(
 		return nil, err
 	}
 
+	kubeConfigPath := findKubeConfigPath(workspace)
+	// Kubeconfig env var was found
+	if kubeConfigPath != "" {
+		podAdditions.Volumes = append(podAdditions.Volumes, getKubeConfigVolume())
+	}
+
 	commonEnv := env.CommonEnvironmentVariables(workspace.Name, workspace.Status.WorkspaceId, workspace.Namespace)
 	for idx := range podAdditions.Containers {
 		podAdditions.Containers[idx].Env = append(podAdditions.Containers[idx].Env, commonEnv...)
+		if kubeConfigPath != "" {
+			podAdditions.Containers[idx].VolumeMounts = append(podAdditions.Containers[idx].VolumeMounts, kubeConfigVolumeMount(kubeConfigPath))
+		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/config/property.go
```diff
@@ -39,4 +39,7 @@ const (
 
 	webhooksEnabled        = "che.webhooks.enabled"
 	defaultWebhooksEnabled = "false"
+
+	kubeConfigPath        = "che.kubectx.path"
```
amisevsk, line 43: Maybe `che.kubeconfig.path`?

pkg/controller/workspace/provision/deployment.go
```diff
@@ -181,6 +182,7 @@ func getSpecDeployment(
 				Spec: corev1.PodSpec{
 					InitContainers:                podAdditions.InitContainers,
 					Containers:                    podAdditions.Containers,
+					Volumes:                       append(podAdditions.Volumes, getPersistentVolumeClaim(), getKubeConfigVolume()),
```
amisevsk, line 185: At this point I would consider extracting this to a variable for clarity.

JPinkney:
Closing as this PR is no longer needed

PR #61, opened 2020-04-27
Add an ability to configure path for endpoint

sleshchenko:
### What does this PR do?
It's needed for proxied cloud shell via OpenShift Console proxy to avoid absolute redirection from `/` to `/static`. Since CloudShell Server Side is under reverse proxy, and the path is rewritten - it's not possible to send right redirection, so the only way - go to right endpoint path from the beginning 

Note that such a feature is already supported by Che Server.

It's done during an investigation for https://github.com/eclipse/che/issues/16633

### Is it tested? How?
I did not manage to test it with OpenShift OAuth Proxy, my crc is got crazy and OpenShift OAuth Proxy can't access OS API. I hope to retest it later.

But I tested the following with basic routing on crc
1. Run CloudShell.
2. Open IDE URL of running workspace.
3. Make sure the CloudShell is opened without absolute redirection.
Screenshot_20200427_182646

### DIFF
--- a/pkg/controller/workspacerouting/resolve_endpoints.go
+++ b/pkg/controller/workspacerouting/resolve_endpoints.go
@@ -14,6 +14,7 @@ package workspacerouting
 
 import (
 	"fmt"
+	"strings"
 
 	workspacev1alpha1 "github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
 	"github.com/che-incubator/che-workspace-operator/pkg/config"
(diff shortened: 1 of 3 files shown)


pkg/controller/workspacerouting/resolve_endpoints.go
```diff
@@ -51,31 +52,37 @@ func getExposedEndpoints(
...
+	if secure && endpoint.Attributes[workspacev1alpha1.SECURE_ENDPOINT_ATTRIBUTE] == "true" {
+		protocol = getSecureProtocol(protocol)
+	}
+	path := endpoint.Attributes[workspacev1alpha1.PATH_ENDPOINT_ATTRIBUTE]
+	if path != "" {
+		if !strings.HasPrefix(path, "/") {
+			path = "/" + path
+		}
+	}
+	return fmt.Sprintf("%s://%s%s", protocol, host, path)
```
amisevsk, line 87: As soon as we have to worry about leading/trailing slashes, it's time to move to net/url IMO:

```suggestion
	url := url.URL{
		Scheme: protocol,
		Host: host,
		Path: path,
	}
	return url.String()
```

PR #63, opened 2020-04-30
[proposal] Add annotation to controller to support immutable workspaces

amisevsk:
### What does this PR do?
Add annotation org.eclipse.che.workspace/immutable to signify that workspace should not allow further modification, to make secure configuration more persistent.

Once added, further modifications to workspaces are rejected by existing webhooks. If added to a workspace when webhooks are disabled, startup is failed.

We'll likely run into situations where modifications to an existing workspace can be used to extract a user's OpenShift token. This is a first step in preventing that without impacting other flows.

### Is it tested? How?
Tested on crc

### DIFF
--- a/pkg/webhook/workspace/handler/workspace.go
+++ b/pkg/webhook/workspace/handler/workspace.go
@@ -13,6 +13,8 @@ package handler
 import (
 	"context"
 	"fmt"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"net/http"
 
 	"github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/webhook/workspace/handler/workspace.go
```diff
@@ -45,6 +45,10 @@ func (h *WebhookHandler) MutateWorkspaceOnUpdate(_ context.Context, req admissio
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
+	immutable := oldWksp.Annotations[config.WorkspaceImmutableAnnotation]
+	if immutable == "true" {
+		return admission.Denied(fmt.Sprintf("workspace '%s' is immutable. To make modifications it must be deleted and recreated", oldWksp.Name))
```
sleshchenko, line 50: Probably we need to allow changing `spec.started` attributed even for immutable workspaces.
It's not really actual for cloud-shell workspace but do you think in future it makes sense to deny changing commands as well?

amisevsk: Yeah I imagine once the started field is implemented this can be revisited. This PR is more of a question: is it worth our while adding this and supporting it? I can add an 'limited-mutation' branch to the webhook but don't want to spend the time on it if we don't want the overall feature.

amisevsk:
Rebased on master and added exception for `started` field, to allow starting/stopping immutable workspaces.

amisevsk:
Marked as ready for review so we can assess whether this feature is useful.

PR #67, opened 2020-05-07
Update apiVersion in samples to 1.0.0

sleshchenko:
### What does this PR do?
Probably we need to add some validation on controller or CRD side, but it only updates apiVersion in samples to 1.0.0 to avoid confusion.

N/A

### Is it tested? How?
I believe it should not influence anything.


amisevsk:
Most of our additional files are currently out of date (e.g. olm, etc.)

PR #68, opened 2020-05-07
Generate secrets for workspace

JPinkney:
### What does this PR do?
This is a draft PR that solves https://github.com/eclipse/che/issues/16647. The general idea is that only new keys will be added to the SecretStore on the cluster

https://github.com/devfile/devworkspace-operator/issues/434

### Is it tested? How?

tested on crc


sleshchenko:
Closing this PR since we don't have a plan to finalize it.
Changes should be available in any case here if we later will prioritize https://github.com/devfile/devworkspace-operator/issues/434

PR #69, opened 2020-05-07
Update CRDs to support k8s 1.18

amisevsk:
### What does this PR do?
Updates CRDs to workaround issue https://github.com/eclipse/che/issues/16791

Adds a script that does the required patch, depending on `yq`. This script is executed during `make generate`.

This PR is broken into three commits
- 76059d3 is a no-op formatting change -- I ran the existing CRDs through `yq` to make the formatting consistent and avoid spurious changes (in short, `yq` formats yaml differently than operator-sdk
- 121ef2b adds script `patch/patch_crds.sh` which will add `protocol` to the list of required fields in containerPorts
- b618f6b regenerates CRDs and runs the script (`make generate`)

Workaround for https://github.com/eclipse/che/issues/16791

### Is it tested? How?
Tested applying CRDs to minikube with kubernetes 1.18. Could not test full deployment/workspace start as my minikube network is borked for some reason.

### DIFF
--- a/patch/patch_crds.sh
+++ b/patch/patch_crds.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+#
+# Copyright (c) 2012-2018 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# This script is required to make automatically generated CRDs compatible with
+# Kubernetes >1.18
+#
+# Patching is done in-place. Note that the formatting output by yq is slightly
+# different than the one output by operator-sdk generate commands: list items
+# are indented one level further:
+#     list:            list:
+#     - item1    -->     - item1
+#     - item2            - item2
+#
+
+set -e
+
+COMPONENTS_CRD_PATH="deploy/crds/workspace.che.eclipse.org_components_crd.yaml"
+ROUTINGS_CRD_PATH="deploy/crds/workspace.che.eclipse.org_workspaceroutings_crd.yaml"
+
+# CRD path from root to status field, in jq filter format
+STATUS_PATH='.spec.validation.openAPIV3Schema.properties["status"]'
+# CRD path podAdditions to containerPorts required field, in jq filter format
+# Note this path takes one jq arg: "CONTAINERS_FIELD" can be used to specify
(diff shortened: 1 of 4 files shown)


amisevsk:
Tested workspace starts on minikube (k8s 1.18), found no issues.

patch/patch_crds.sh
```diff
@@ -0,0 +1,78 @@
...
+# Update components CRD using yq; no-op if already patched.
+# Args:
+#  $1 - podAdditions field to update ("containers" or "initContainers")
+function update_components_crd() {
+  field="$1"
+  local yq_check_script="${COMPONENTS_SPEC_PATH}"' | index("protocol")'
+  local yq_patch_script="${COMPONENTS_SPEC_PATH}"' |= . + ["protocol"]'
+  already_patched=$(yq -r "$yq_check_script" "$COMPONENTS_CRD_PATH" --arg "CONTAINERS_FIELD" "$field")
+  if [[ "$already_patched" == "null" ]]; then
+    yq -Y -i "$yq_patch_script" "$COMPONENTS_CRD_PATH" --arg "CONTAINERS_FIELD" "$field"
```
sleshchenko, line 47: Seems we have different yq... it does not work for me =(
Screenshot_20200508_080946

patch/patch_crds.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo "Updated CRD $COMPONENTS_CRD_PATH"
+  else
+    echo "Updating CRD $COMPONENTS_CRD_PATH not necessary"
+  fi
+}
+
+# Update workspaceroutings CRD using yq; no-op if already patched.
+# Args:
+#  $1 - podAdditions field to update ("containers" or "initContainers")
+function update_routings_crd() {
```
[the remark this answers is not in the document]
amisevsk: We could, I didn't do it because the scripting and bash function params were getting complicated.

patch/patch_crds.sh
amisevsk: Ah you need to update `yq`; in-place editing was added more recently.

amisevsk: The precise requirements are

- `jq`: version 1.5.0 or higher (https://github.com/stedolan/jq/releases/tag/jq-1.5)
- `yq`: version 2.9.0 or higher (https://github.com/kislyuk/yq/releases/tag/v2.9.0)

sleshchenko: OK, thanks for the clarification.
Previously, I installed it with apt-get, but seems new version is not available there. I'll download the latest versions manually.

PR #70, opened 2020-05-08
Add Command Line Terminal cheEditor into internal Plugin Registry

sleshchenko:
### What does this PR do?
This PR adds Command Line Terminal cheEditor into internal Plugin Registry, it's needed since currently, OpenShift Console depends on the nightly CloudShell https://github.com/openshift/console/blob/cd751e4badfd55d79df916ca8bca0b6126c446b3/frontend/packages/console-app/src/components/cloud-shell/cloud-shell-utils.ts#L70
and since code-freeze OpenShift 4.5 is coming it's critical to put some long-term supported version.

N/A

### Is it tested? How?
not yet

### DIFF
--- a/internal-registry/che-incubator/command-line-terminal/4.5.0/meta.yaml
+++ b/internal-registry/che-incubator/command-line-terminal/4.5.0/meta.yaml
@@ -1,0 +1,31 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: 4.5.0
+type: Che Editor
+displayName: Command Line Terminal
+title: Command Line Terminal
+description: Command Line Terminal provides the ability to start a terminal inside the OpenShift Console.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/eclipse/che-machine-exec/
+firstPublicationDate: "2020-05-13"
+category: Other
+spec:
+  endpoints:
+   -  name: command-line-terminal
+      public: true
+      targetPort: 4444
+      attributes:
+        protocol: http
+        type: ide
+        path: /static/
+        discoverable: false
+        secure: true
+        cookiesAuthEnabled: true
+  containers:
+   - name: command-line-terminal
+     image: "quay.io/eclipse/che-machine-exec:nightly"
+#    ^ it'll be replaced with new image below after https://github.com/eclipse/che/issues/16942 is fixed
+#    image: "quay.io/che-incubator/command-line-terminal:4.5.0"
+     ports:
+       - exposedPort: 4444
--- a/internal-registry/che-incubator/command-line-terminal/nightly/meta.yaml
+++ b/internal-registry/che-incubator/command-line-terminal/nightly/meta.yaml
@@ -1,0 +1,31 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: nightly
+type: Che Editor
+displayName: Command Line Terminal
+title: Command Line Terminal
+description: Command Line Terminal provides the ability to start a terminal inside the OpenShift Console.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/eclipse/che-machine-exec/
+firstPublicationDate: "2020-05-13"
+category: Other
+spec:
+  endpoints:
+   -  name: command-line-terminal
+      public: true
+      targetPort: 4444
+      attributes:
+        protocol: http
+        type: ide
+        path: /static/
+        discoverable: false
+        secure: true
+        cookiesAuthEnabled: true
+  containers:
+   - name: command-line-terminal
+     image: "quay.io/eclipse/che-machine-exec:nightly"
+#    ^ it'll be replaced with new image below after https://github.com/eclipse/che/issues/16942 is fixed
+#    image: "quay.io/che-incubator/command-line-terminal:nightly"
+     ports:
+       - exposedPort: 4444
--- a/samples/command-line-terminal.yaml
+++ b/samples/command-line-terminal.yaml
@@ -1,0 +1,31 @@
+# It's just an example of workspace configuration OpenShift Console creates
+# when user requests terminal
+apiVersion: workspace.che.eclipse.org/v1alpha1
+kind: Workspace
+metadata:
+  name: command-line-terminal
+  annotations:
+    # it's important to make workspace immutable to make sure nobody with right access
+    # won't set custom editor to steal creator's token
+    org.eclipse.che.workspace/immutable: "true"
+  labels:
+    # it's a label OpenShift console uses a flag to mark terminal's workspaces
+    console.openshift.io/cloudshell: "true"
+spec:
+  started: true
+  devfile:
+    apiVersion: 1.0.0
+    metadata:
+      name: command-line-terminal
+    components:
+      - alias: command-line-terminal
+        type: cheEditor
+        id: che-incubator/command-line-terminal/4.5.0
+      - type: dockerimage
+        memoryLimit: 256Mi
+        alias: dev
+        image: 'registry.redhat.io/codeready-workspaces/plugin-openshift-rhel8:2.1'
+        args: ["tail", "-f", "/dev/null"]
+        env:
+          - value: '\[\e[34m\]>\[\e[m\]\[\e[33m\]>\[\e[m\]'
+            name: PS1


sleshchenko:
I mixed latest and openshift4.5 versions... Will finalize it on Tuesday or feel free anybody to create your PR or finalize this one

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+      attributes:
+        protocol: http
+        type: ide
+        path: /static/
+        discoverable: false
+        secure: true
+        cookiesAuthEnabled: true
+  containers:
+   - name: che-machine-exec
+     image: "quay.io/eclipse/che-machine-exec:nightly"
```
sleshchenko, line 27: later it will be replaced with a released che-machine-exec version that is compatible with openshift4.5 or even maybe image built for openshift4.5

samples/openshift-cloud-terminal.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: workspace.che.eclipse.org/v1alpha1
+kind: Workspace
+metadata:
+  name: openshift-terminal
+  annotations:
+    org.eclipse.che.workspace/immutable: "true"
```
sleshchenko, line 6: It's needed to make sure that nobody (who has edit rights to a workspace) is not able to modify workspace and put custom editor that will be like send personal user's token somewhere or just print it into logs.

samples/openshift-cloud-terminal.yaml
```diff
@@ -0,0 +1,26 @@
...
+  labels:
+    console.openshift.io/cloudshell: "true"
```
sleshchenko, line 8: just a marker for OpenShift Console to pick up the workspace as a terminal

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v2
+publisher: eclipse
+name: openshift-cloud-terminal
```
sleshchenko, line 3: it's proposed by Mario as well as alternative variant: `eclipse/cloud-shell/openshift4.5`

pdaverh: Cloud-terminal is better than cloud-shell. I'm even fine with "openshift-terminal". Thanks!

samples/openshift-cloud-terminal.yaml
amisevsk: Isn't this label used by the console to store username? Does putting "true" here work?

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+version: 4.5.0
+type: Che Editor
+displayName: OpenShift Cloud Terminal
+title: OpenShift Cloud Terminal
+description: OpenShift Cloud Terminal provides an ability to use terminal inside OpenShift Console.
```
amisevsk, line 8:
```suggestion
description: OpenShift Cloud Terminal provides the ability to start a terminal inside the OpenShift Console.
```

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
nickboldt: If you're going to rebrand it I have two thoughts here:
a) do you have permission to call it OpenShift ___ ? You need Brand/Marketing/PM approval to use the name in anything producty/projecty like this.

b) if you DO have approval to use OpenShift, then why not use the name that's currently in plan for the productized version?  https://issues.redhat.com/browse/CRW-600 has "OpenShift Command Line Terminal"

cc: @l0rd

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
nickboldt: I really hope you don't plan to have two different che-machine-exec branches... one for 4.4 and earlier, and another for 4.5+. 

Once 4.5 is out, CRW support plan is basically 4.5+, so moving upstream to 4.5+ too is likely OK as well.

samples/openshift-cloud-terminal.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  started: true
+  devfile:
+    apiVersion: 1.0.0
+    metadata:
+      name: cloud-shell
```
nickboldt, line 14: I see you have cloud-shell and openshift-cloud-terminal mixed together here. 

Also is in intentional that you're using codeready-workspaces containers in *gasp* the upstream project? :) I'm OK with that but... does that also introduce a circular dependency?

samples/openshift-cloud-terminal.yaml
sleshchenko: OpenShift Console uses `console.openshift.io/cloudshell-user` annotation to store username and `console.openshift.io/cloudshell` is supposed to have true/false value. 
It eventually makes sense to rename it to `console.openshift.io/openshift-terminal` I think, but it's example that shows the current state.

P.S. I had a conversation with Abhi about username annotation OpenShift Console uses and we agreed that it makes sense to just reuse `org.eclipse.che.workspace/creator` annotation with UID controller provisions.

samples/openshift-cloud-terminal.yaml
sleshchenko: it's better to change this particular cloud-shell word. Will do.

Apart from that there is still cloud-shell cheEditor plugin in internal registry which I did not touch for backward compatibility. I think it eventually will be removed https://github.com/che-incubator/che-workspace-operator/tree/master/internal-registry/eclipse/cloud-shell/nightly

> Also is in intentional that you're using codeready-workspaces containers in gasp the upstream project? :) I'm OK with that but... does that also introduce a circular dependency?

It's an example that shows which workspace configuration OpenShift Console is supposed to create. The actual configuration is stored in OpenShift Console Repo https://github.com/openshift/console/blob/cd751e4badfd55d79df916ca8bca0b6126c446b3/frontend/packages/console-app/src/components/cloud-shell/cloud-shell-utils.ts#L76

So, I don't think it should lead to circular dependency... The thing we must release here - openshift-cloud-terminal configuration via controller image + OLM...
But I may be wrong and more questions/concerns are welcome if words above do not answer a question.

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
sleshchenko: @l0rd any preferences here?

samples/openshift-cloud-terminal.yaml
nickboldt: Is anyone using the "cloud-shell cheEditor plugin" today? Does it work? If this "cloud-terminal" is the replacement for that, and the old one is a version <1.0, then it's incubating, subject to change/deletion. 

I only see a "nightly" version here: 

https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/cloud-shell/nightly/meta.yaml

So I think you're safe to just delete it. Unversioned "nightly" releases are throwaway builds, and are never IMHO supported. 

However I see the machine-exec nightly is ALSO affected by this old cloud-shell:

https://github.com/eclipse/che-plugin-registry/commit/74fcb49997f86a9fe4a8419292c42f8b107268c2#diff-6bd840228d47bceca757b6a33c8ad2a2

Will the machine exec continue to reference the "cloud-shell" or will it need to be updated to the new "cloud-terminal" ?

https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-machine-exec-plugin/nightly/meta.yaml#L30

samples/openshift-cloud-terminal.yaml
nickboldt: If you're renaming "cloud-shell" to "cloud-terminal" then I suspect you need to rename here too.

samples/openshift-cloud-terminal.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: workspace.che.eclipse.org/v1alpha1
+kind: Workspace
+metadata:
+  name: openshift-terminal
```
nickboldt, line 4: is it "openshift-terminal" or "openshift-cloud-terminal" ??

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
nickboldt: I see that @pdaverh *DID* suggest cloud-terminal is better... but I'm not sure Branding will be cool with "openshift-terminal" or "openshift-cloud-terminal". Has anyone confirmed this naming scheme with the OpenShift PM yet?

samples/openshift-cloud-terminal.yaml
sleshchenko: it's just a sample, the actual value will be generated by OpenShift Console. 
But at this point to makes sense to align it with plugin name. I'm going to rename it to openshift-cloud-terminal until another name won't be chosen under thread https://github.com/che-incubator/che-workspace-operator/pull/70#discussion_r424388778

internal-registry/eclipse/openshift-cloud-terminal/4.5.0/meta.yaml
l0rd: @sleshchenko that was an error on my side. @nickboldt is right. 

Upstream we should not prefix it with openshift and it would be better if we don't use the eclipse org because that's not an eclipse project yet. `quay.io/che-incubator/command-line-terminal:4.5.0`.

Downstream the product name should be "OpenShift Command Line Terminal" hence we should have `publisher: redhat`, `name: openshift-command-line-terminal`, `version: 4.5.0`. For the image name I don't know what's the best org: `quay.io/redhat/openshift-command-line-terminal:4.5.0`, `quay.io/openshift/command-line-terminal:4.5.0` or `quay.io/crw/openshift-command-line-terminal:4.5.0`?

sleshchenko: @l0rd and me had a meeting where we agreed on 
`che-incubator/command-line-terminal/4.5.0` for cheEditor used by OpenShift Console for workspaces.
The image that will be in that editor will be tagged as `quay.io/che-incubator/command-line-terminal:4.5.0` but technically it'll be https://github.com/eclipse/che-machine-exec

the PR is updated in accordance, please review and give more feedback if any

samples/openshift-cloud-terminal.yaml
sleshchenko: > Is anyone using the "cloud-shell cheEditor plugin" today? Does it work? If this "cloud-terminal" is the replacement for that, and the old one is a version <1.0, then it's incubating, subject to change/deletion.

Now it's used by OpenShift console https://github.com/openshift/console/blob/cd751e4badfd55d79df916ca8bca0b6126c446b3/frontend/packages/console-app/src/components/cloud-shell/cloud-shell-utils.ts#L70
Once we merge this PR and they update source code to use command-line-terminal - I think we're safe to remove cloud-shell at all.

> Will the machine exec continue to reference the "cloud-shell" or will it need to be updated to the new "cloud-terminal" ?

Actually that our front-end is not used anywhere.
Theia based workspaces - uses Theia extension.
OpenShift Console uses their own frontend.

And CloudShell Frontend in machine exec needs to be revised and eventually removed or used somewhere as separately.
And I'll create a PR into plugin-registry to stop serving CloudShell UI by che-machine-exec.

internal-registry/che-incubator/command-line-terminal/4.5.0/meta.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: 4.5.0
+type: Che Editor
+displayName: Command Line Terminal
+title: Command Line Terminal
+description: Command Line Terminal provides the ability to start a terminal inside the OpenShift Console.
```
nickboldt, line 8: does this also work with k8s? Or is it exclusively targetting OpenShift?

internal-registry/che-incubator/command-line-terminal/nightly/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+        protocol: http
+        type: ide
+        path: /static/
+        discoverable: false
+        secure: true
+        cookiesAuthEnabled: true
+  containers:
+   - name: command-line-terminal
+     image: "quay.io/eclipse/che-machine-exec:nightly"
+#    ^ eventually it'll be replaced with new image below, but currently it's not in place yet
```
nickboldt, line 28: personal preference here would be to mark this with a `TODO` and a link to the issue tracking the work so it's easier to find and fix in future...

samples/command-line-terminal.yaml
```diff
@@ -0,0 +1,31 @@
+# It's just an example of workspace configuration OpenShift Console creates
+# when user requests terminal
+apiVersion: workspace.che.eclipse.org/v1alpha1
```
nickboldt, line 3: just a question about the version here... v1alpha1? why not something simpler like v0.0.1? on line 17 you use apiVersion 1.0.0 so I'm just a bit confused how the apis are versioned.

samples/command-line-terminal.yaml
```diff
@@ -0,0 +1,31 @@
...
+spec:
+  started: true
+  devfile:
+    apiVersion: 1.0.0
+    metadata:
+      name: command-line-terminal
+    components:
+      - alias: command-line-terminal
+        type: cheEditor
+        id: che-incubator/command-line-terminal/4.5.0
```
nickboldt, line 23: should this be a moving tag `:4.5` or do you plan to release a new release for every OCP 4.5.x release too?

internal-registry/che-incubator/command-line-terminal/nightly/meta.yaml
sleshchenko: Makes sense. Will do

samples/command-line-terminal.yaml
sleshchenko: I did not have enough time to learn much enough how CRD is supposed to be designed, so my simple answer is: 
- apiVersion of k8s object consist of apiGroup - workspace.che.eclipse.org + version v1alpha1;
- Our Workspace CRD is not stable enough to move it from alpha version I think.

I think K8s has conversion about versions for apps, and versions are not supposed to be MAJOR.MINOR.PATCH. It should be `vXalphaX`, `vXbetaX`, `vX`.

@amisevsk Do you have something to add?

samples/command-line-terminal.yaml
sleshchenko: Good question :+1:
I think it's a safer option to release command-line-terminal version specially for the concrete OpenShift version. I can imagine a situation where OpenShift 4.5.0 and OpenShift 4.5.1 possible need different APIs from command-line-terminal.

Or instead, we're able not stick to OpenShift version and do the right Major.minor.patch versioning where we make sure that we do not provide breaking changes.
Then OpenShift Console will get bug fixes without any changes in their code base, and version in OpenShift Console will be changed only after OpenShift Console Frontend adaptation.

internal-registry/che-incubator/command-line-terminal/4.5.0/meta.yaml
sleshchenko: It exclusively targeting OpenShift.
If be more precisely, it provides an API for OpenShift Console to initialize container before opening terminal (create kubeconfig).
There is already API to cache terminal sessions, but OpenShift Console does not use it yet.

Possible Kubernetes Dashboard could be improved to use it as well but nobody's taken a look at that.
So, at this point, we declare only OpenShift Terminal Providing.

samples/command-line-terminal.yaml
amisevsk: Yeah, kubernetes API version conventions dictate this sort of naming: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning

internal-registry/che-incubator/command-line-terminal/4.5.0/meta.yaml
nickboldt: works for me, just making sure your description was accurate since most of Che features are for BOTH k8s & OS.

pdaverh: Can we please update with name "OpenShift Terminal" - short and sweet - instead of "Command Line Terminal"?

l0rd: @pdaverh are we changing the name again? We aligned to [CRW-600](https://issues.redhat.com/browse/CRW-600). We are avoiding to prefix it with OpenShift since it's not an OpenShift product (from @nickboldt comment above "You need Brand/Marketing/PM approval to use the name in anything producty/projecty like this"). Moreover this is the name of a `cheEditor`, not the name of the Operator hence I don't think it really matters because it won't be exposed to users.

@sleshchenko I would merge this PR like this. @pdaverh we can later update this cheEditor meta.yaml if we all agree on a better naming. We may need a naming dedicated call with @nickboldt.

sleshchenko:
I've merged this PR to move forward but we'll still continue the discussion on better naming and are able to change in separate PRs

internal-registry/che-incubator/command-line-terminal/4.5.0/meta.yaml
pdaverh: Thanks for the information that the name here will not be exposed to users. For the operator, we are working with brand and marketing to approve "OpenShift Terminal". OpenShift PM is fine with it.

PR #73, opened 2020-05-13
Only propagate creator annotation if present

amisevsk:
### What does this PR do?
Only propagate creator annotation to workspace deployment if it is nonempty, to avoid applying an empty annotation to deployments when running with webhooks disabled.

Fix for first part of https://github.com/eclipse/che/issues/16627

### Is it tested? How?
Tested on crc with and without webhooks enabled. With webhooks, behavior is unmodified; without no creator annotation is created on deployments.

### DIFF
--- a/pkg/controller/workspace/provision/deployment.go
+++ b/pkg/controller/workspace/provision/deployment.go
@@ -200,8 +200,8 @@ func getSpecDeployment(
 		deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, getPersistentVolumeClaim())
 	}
 
-	workspaceCreator := workspace.Annotations[config.WorkspaceCreatorLabel]
-	if workspaceCreator != "" {
+	workspaceCreator, present := workspace.Labels[config.WorkspaceCreatorLabel]
+	if present {
 		deployment.Labels[config.WorkspaceCreatorLabel] = workspaceCreator
 		deployment.Spec.Template.Labels[config.WorkspaceCreatorLabel] = workspaceCreator
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/workspace/provision/deployment.go
```diff
@@ -203,6 +200,13 @@ func getSpecDeployment(
 		deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, getPersistentVolumeClaim())
 	}
 
+	workspaceCreator := workspace.Annotations[config.WorkspaceCreatorAnnotation]
+	if workspaceCreator != "" {
```
sleshchenko, line 204: 1. It means that kube:admin is not able to create workspaces anymore if webhooks are enabled since validation webhooks won't allow deployment creation without creator annotation.
Screenshot_20200514_105951
2. There is one more place - podTemplate which probably should be changed as well.

Probably it should be changed in the following way:

```go
workspaceCreator, present := workspace.Annotations[config.WorkspaceCreatorAnnotation]
	if present {
		deployment.Annotations = map[string]string{
			config.WorkspaceCreatorAnnotation: workspaceCreator,
		}
		deployment.Spec.Template.Annotations = map[string]string{
			config.WorkspaceCreatorAnnotation: workspaceCreator,
		}
	}
```
but it rewrites whole annotations... It may be a bit safer if we initialize annotations above with empty map and then here put needed value

amisevsk: Updated, thanks for the suggestion, my initial PR was a bit sloppy. I've rebased on master and included the Pod label as well.

amisevsk: Ah I had to rewrite this portion and that check got lost in the rebase. I'll fix.

amisevsk: Ah I had to rewrite this portion and that check got lost in the rebase. Should be fixed now.

PR #75, opened 2020-05-15
Add rhel.Dockefile that uses rhel based go builder image

sleshchenko:
### What does this PR do?
Add rhel.Dockefile that uses rhel based go builder image

There is no created issue.

### Is it tested? How?
Build manually
1. Create SA on the http://access.redhat.com/terms-based-registry
2. Login docker with the created SA
3. Build dockerimage with new dockerfile
`docker build --file ./build/rhel.Dockerfile . -t WANTED_TAG`
4. Deploy built dockerimage controller.

Or test built image
1. `export IMG=sleshchenko/che-workspace-operator:rhel8-go-toolset-1.13.4-15`
2. Deploy and test workspaces as usual.

### DIFF
--- a/build/rhel.Dockerfile
+++ b/build/rhel.Dockerfile
@@ -1,0 +1,37 @@
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhel8/go-toolset
+FROM registry.redhat.io/rhel8/go-toolset:1.13.4-15 as builder
+
+ENV GOPATH=/go/
+
+USER root
+
+WORKDIR /che-workspace-operator
+
+# Populate the module cache based on the go.{mod,sum} files.
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+# copy the rest of the sources code
+COPY . .
+# compile workspace controller binaries
+RUN CGO_ENABLED=0 GOOS=linux go build \
+  -o _output/bin/che-workspace-controller \
+  -gcflags all=-trimpath=/ \
+  -asmflags all=-trimpath=/ \
+  cmd/manager/main.go
+
+FROM registry.access.redhat.com/ubi8-minimal:8.1-279
+COPY --from=builder /che-workspace-operator/_output/bin/che-workspace-controller /usr/local/bin/che-workspace-controller
+COPY --from=builder /che-workspace-operator/internal-registry  internal-registry
+
+ENV USER_UID=1001 \
+    USER_NAME=che-workspace-controller
+
+COPY build/bin /usr/local/bin
+RUN  /usr/local/bin/user_setup
+
+USER ${USER_UID}
+
+ENTRYPOINT ["/usr/local/bin/entrypoint"]
+CMD /usr/local/bin/che-workspace-controller


build/Dockerfile
```diff
@@ -1,4 +1,14 @@
-FROM golang:1.13.7-alpine3.11 as builder
+FROM registry.access.redhat.com/devtools/go-toolset-rhel7:1.12.12-4 as builder
+
+ENV PATH=/opt/rh/go-toolset-1.12/root/usr/bin:$PATH \
+    GOPATH=/go/
+
+USER root
+
+RUN case $(uname -m) in \
```
sleshchenko, line 8: I'm not sure what it does, just copied/pasted from che-operator Dockerfile. Sorry about that =)

nickboldt: right now it does nothing but set a variable `ARCH` to one of two values based on the architecture of the platform.

You probably want to then use this when running the go compilation on line 75...

```
- RUN CGO_ENABLED=0 GOOS=linux ...
+ RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} ...
```

Also if you want to support ppc64le as well as x86_64/amd64 and s390x, you could replace the case statement with this:

```
ARCH="$(uname -m)" && if [[ ${ARCH} == "x86_64" ]]; then export ARCH="amd64"; fi
```

build/Dockerfile
```diff
@@ -1,4 +1,14 @@
-FROM golang:1.13.7-alpine3.11 as builder
+FROM registry.access.redhat.com/devtools/go-toolset-rhel7:1.12.12-4 as builder
```
nickboldt, line 1: If you use go 1.13, you can do this:

```
# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhel8/go-toolset
FROM rhel8/go-toolset:1.13.4-15 as builder
ENV GOPATH=/go/
```

amisevsk:
```
Error response from daemon: pull access denied for rhel8/go-toolset, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Do we need to start implementing/maintaining `Dockerfile` and `rhel.Dockerfile`?

build/Dockerfile
```diff
@@ -1,4 +1,14 @@
...
+
+ENV PATH=/opt/rh/go-toolset-1.12/root/usr/bin:$PATH \
+    GOPATH=/go/
+
+USER root
```
amisevsk, line 6: Is using root required?

nickboldt: > Do we need to start implementing/maintaining Dockerfile and rhel.Dockerfile?

If the plan is to build this in Brew, then yes. I still don't have a clear answer from @parag about what we're doing here.

nickboldt: seems we are going the brew route and will be setting up repos via comet to host the code.

nickboldt: for reference here's the latest CRW operator dockerfile. http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-operator/tree/Dockerfile?h=crw-2.2-rhel-8 Note it's using go-toolset 1.12.

sleshchenko: Without user root I'm not even able to build to locally:
Screenshot_20200601_133940

build/Dockerfile
sleshchenko: @nickboldt now `registry.redhat.io/rhel8/go-toolset:1.13.4-15` is used in rhel.Dockerfile.

sleshchenko:
I've merged this PR and new comments/suggestions are going to be fixed in separate PRs

PR #76, opened 2020-05-16
In cluster cert generation for webhook server

JPinkney:
### What does this PR do?
This PR brings cert generation into the controller. If webhooks are enabled the controller will create the needed service and configmap then update the deployment with the new volume and volumemount.

https://github.com/eclipse/che/issues/16954

### Is it tested? How?

Tested on CRC with webhooks enabled. Controller initially starts without needed volume and volumemount and then the deployment is updated with them.


internal/controller/ownerref.go
```diff
@@ -73,3 +74,65 @@ func findFinalOwnerRef(ctx context.Context, client crclient.Client, ns string, o
 	Log.V(1).Info("Pods owner found", "Kind", ownerRef.Kind, "Name", ownerRef.Name, "Namespace", ns)
 	return ownerRef, nil
 }
+
+//FindControllerOwner returns OwnerReferent that owns controller process
+//it starts searching from the current pod and then resolves owners recursively
+//until object without owner is not found
```
amisevsk, line 80: In addition to @sleshchenko's comment, I prefer comments formatted with a space:
```suggestion
// FindControllerOwner returns OwnerReferent that owns controller process
// it starts searching from the current pod and then resolves owners recursively
// until object without owner is not found
```

internal/controller/ownerref.go
```diff
@@ -73,3 +74,65 @@ func findFinalOwnerRef(ctx context.Context, client crclient.Client, ns string, o
...
+	// Get current Pod the operator is running in
+	pod, err := k8sutil.GetPod(ctx, client, ns)
+	if err != nil {
+		return nil, err
+	}
+	// Get Owner that the Pod belongs to
+	//TODO Take a look, maybe it's a better thing to use
+	//podOwnerRefs := metav1.NewControllerRef(pod, pod.GroupVersionKind())
+	ownerRef := metav1.GetControllerOf(pod)
+	deployment, err := findDeployment(ctx, client, ns, ownerRef)
```
amisevsk, line 96: I would move this first ownerref into `findDeployment` -- it doesn't make as much sense to start one step up the ladder instead of at the base:

```go
deployment, err := findDeployment(crx, client, ns, pod)
// ...
func findDeployment(...) {
  ownerRef := metav1.GetControllerOf(pod)
  // continue as normal
}
```
I believe you can use a type such as `metav1.Object` for this purpose and then reflect the final result to a Deployment.

internal/controller/ownerref.go
```diff
@@ -73,3 +74,65 @@ func findFinalOwnerRef(ctx context.Context, client crclient.Client, ns string, o
...
+	if err != nil {
+		return nil, err
+	}
+	newOwnerRef := metav1.GetControllerOf(obj)
+	if newOwnerRef != nil {
+		return findDeployment(ctx, client, ns, newOwnerRef)
+	}
+
+	Log.V(1).Info("Deployment is not found =(", "Kind", ownerRef.Kind, "Name", ownerRef.Name, "Namespace", ns)
+	return nil, nil
```
amisevsk, line 137: I'd invert the order of these checks:
```suggestion
	newOwnerRef := metav1.GetControllerOf(obj)
	if newOwnerRef == nil {
		Log.V(1).Info("Deployment is not found =(", "Kind", ownerRef.Kind, "Name", ownerRef.Name, "Namespace", ns)
		return nil, nil
	}
	return findDeployment(ctx, client, ns, newOwnerRef)
```

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,177 @@
...
+	log.Info("Updating the deployment")
+	err = updateDeployment(ctx, crclient)
+	if err != nil {
+		return nil
+	}
+
+	return nil
+}
+
+func createSecureService() *v1.Service {
```
amisevsk, line 53: To avoid confusion, I would avoid naming this with `create` (since it could imply that the service is created on cluster). Maybe something like `getSecureServiceSpec()`? 

Similar suggestion for `createSecureConfigMap`

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,177 @@
...
+			log.Info("There was an error")
+			log.Info(err.Error())
+			return err
+		}
+	}
+
+	return nil
+}
+
+// append the volume mount if it is missing. Indicates if the volume mount is missing
```
amisevsk, line 157: godoc generally suggests starting function docs with the function name:
```suggestion
// appendVolumeMountIfMissing appends the volume mount if it is missing. Indicates if the volume mount is missing
```

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,177 @@
...
+	crclient, err := controller.CreateClient()
+	if err != nil {
+		return err
+	}
+
+	log.Info("Creating the secure service")
+	secureService := createSecureService()
+	err = crclient.Create(ctx, secureService)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
JPinkney: Yeah, I think it wrote it once wrong by accident the copied it throughout the file haha

internal/controller/ownerref.go
JPinkney: (This wasn't my original code, all the changes I took in ownerrefs was from an old closed PR on this repo) but I think the reason why it was done this way is `newOwnerRef` is passed back into findDeployment later so we'd need to pass in the ownerref and not the pod

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,228 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secureServiceName,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				"service.beta.openshift.io/inject-cabundle": "true",
+			},
+		},
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ConfigMap",
```
sleshchenko, line 148: Is it really needed even when you create v1.ConfigMap object?

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,228 @@
...
+	} else {
+		log.Info("Created secure configmap")
+	}
+	return nil
+}
+
+func getSecureConfigMapSpec(namespace string) *v1.ConfigMap {
+	return &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secureServiceName,
```
sleshchenko, line 140: Seems here should be used const like `certCAConfigMapName`

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,228 @@
...
+							},
+						},
+					},
+				},
+			},
+		})
+
+	isVMMissing := appendVolumeMountIfMissing(&deployment.Spec.Template.Spec.Containers[0].VolumeMounts,
+		*&v1.VolumeMount{
+			Name:      "webhook-tls-certs",
```
sleshchenko, line 193: Such duplicated hard-coded values are error-prone. I propose to move configuration objects to separate file and move everything (or at least that is used more than once) to const
See https://github.com/sleshchenko/che-workspace-operator/commit/5ca6b6d56c14c2548fa03cee61cef73ae6a68c8c

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+		log.Info("Updated secure service")
+	} else {
+		log.Info("Created secure service")
+	}
+	return nil
+}
+
+func getSecureServiceSpec(namespace string) *v1.Service {
+	label := map[string]string{"app": "che-workspace-controller"}
```
sleshchenko, line 87: It would be much safer if we make sure that deployment really contains such label and probably provision if missing.

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -0,0 +1,228 @@
...
+			Annotations: map[string]string{
+				"service.beta.openshift.io/serving-cert-secret-name": "workspace-controller",
+			},
+		},
+		Spec: v1.ServiceSpec{
+			Ports: []v1.ServicePort{
+				{
+					Port:       port,
+					Protocol:   "TCP",
+					TargetPort: intstr.FromString("webhook-server"),
```
sleshchenko, line 104: Consider adding a check that deployment has such target port.

internal/controller/ownerref.go
amisevsk: Sure but wouldn't it be cleaner to do

```go
// findDeployment tries to locate the final controller/owner based on the owner reference provided.
func findDeployment(ctx context.Context, client crclient.Client, ns string, metaObj metav1.Object) (*v1.Deployment, error) {
	ownerRef := metav1.GetControllerOf(metaObj)

	if ownerRef == nil {
		return nil, nil
	}

	if ownerRef.Kind == "Deployment" {
		d := &v1.Deployment{}
		err := client.Get(ctx, types.NamespacedName{Namespace: ns, Name: ownerRef.Name}, d)
		d.APIVersion = ownerRef.APIVersion
		d.Kind = ownerRef.Kind
		if err != nil {
			return nil, err
		}
		return d, nil
	}
	obj := &unstructured.Unstructured{}
	obj.SetAPIVersion(ownerRef.APIVersion)
	obj.SetKind(ownerRef.Kind)
	err := client.Get(ctx, types.NamespacedName{Namespace: ns, Name: ownerRef.Name}, obj)
	if err != nil {
		return nil, err
	}

	return findDeployment(ctx, client, ns, obj)
}
```

internal/controller/ownerref.go
```diff
@@ -14,6 +14,8 @@ import (
 	"context"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/apps/v1"
```
sleshchenko, line 18: Seems one of `k8s.io/api/apps/v1` is redundant

internal/controller/ownerref.go
```diff
@@ -73,3 +75,57 @@ func findFinalOwnerRef(ctx context.Context, client crclient.Client, ns string, o
...
+
+	deployment, err := findDeployment(ctx, client, ns, pod)
+	if err != nil {
+		return nil, err
+	}
+	if deployment != nil {
+		return deployment, nil
+	}
+
+	// Default to returning Pod as the Owner
```
sleshchenko, line 100: the comment does not seem to be correct

pkg/webhook/server/cert_cfg.go
```diff
@@ -0,0 +1,119 @@
...
+func getSecureConfigMapSpec(namespace string) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      certConfigMapName,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				"service.beta.openshift.io/inject-cabundle": "true",
+			},
+		},
+		TypeMeta: metav1.TypeMeta{
```
sleshchenko, line 66: So, is it needed? Why we don't have it or Service for example?

pkg/webhook/server/cert_cfg.go
```diff
@@ -0,0 +1,119 @@
...
+			Annotations: map[string]string{
+				"service.beta.openshift.io/serving-cert-secret-name": certSecretName,
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Port:       port,
+					Protocol:   "TCP",
+					TargetPort: intstr.FromString("webhook-server"),
```
sleshchenko, line 47: Seems container port now gets its name from webhookServerName const

PR #79, opened 2020-05-22
Rename ide url

amisevsk:
### What does this PR do?
Rename `workspace.status.ideUrl` to `workspace.status.mainIdeUrl` to support compatibility with [devfile 2.0 spec](https://github.com/devfile/kubernetes-api/blob/06293a5b9f8aef0a82e20c3085739d04766c5295/pkg/apis/workspaces/v1alpha1/devworkspace_types.go#L21). 

OpenShift Terminal should be compatible with both current CRD and devfile 2.0 if possible. Console backend proxy depends on ideUrl for proxying requests

### Is it tested? How?
Tested deploying updated CRDs on `crc`, URL field in `oc get workspace` is set correctly.

### DIFF
--- a/deploy/crds/workspace.che.eclipse.org_workspaces_crd.yaml
+++ b/deploy/crds/workspace.che.eclipse.org_workspaces_crd.yaml
@@ -12,7 +12,7 @@ spec:
     description: The current workspace startup phase
     name: Phase
     type: string
-  - JSONPath: .status.ideUrl
+  - JSONPath: .status.mainIdeUrl
     description: Url endpoint for accessing workspace
     name: URL
     type: string
@@ -253,14 +253,14 @@ spec:
                 - type
                 type: object
               type: array
-            ideUrl:
+            mainIdeUrl:
               type: string
             phase:
               type: string
             workspaceId:
               type: string
           required:
-          - ideUrl
+          - mainIdeUrl
           - workspaceId
           type: object
       type: object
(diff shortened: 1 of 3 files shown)


amisevsk:
@sleshchenko That's a question to bring up for @davidfestal and @l0rd :)

I'll hold on merging this PR until next week.

davidfestal:
> Changes look good but I wonder if we are not going to rename `mainIdeUrl` to like
> 
> 1. `mainEndpointUrl` since on component level we have endpoint entity
> 2. or `editorUrl`, since we have editor component

In Devfile 2.0, there's not `editor` component anymore. And plugins are mainly devfile fragments.

> 
> Main ide url makes impression that there are multiple IDEs and one is main....

Isn't it a bit the case ? Isn't it possible, in full generality, to have bith a theia ide plugin ad a jupyter ide plugin in a single workspace ?

sleshchenko:
> Isn't it a bit the case ? Isn't it possible, in full generality, to have bith a theia ide plugin ad a jupyter ide plugin in a single workspace ?

For sure it's possible but how it's going to work?
1. Who defines from which plugin IDE is the main? If user, what should be defined in devfile to achieve it?
^ now, plugins meta.yaml contains `attributes.type: ide` and the first or last will be set as IDEUrl of the workspace.
2. How the user is supposed to use multiple IDEs in workspace? What is the scenario of changing IDE, how user gets all IDEs URLs? If there are really many IDEs, then maybe workspace status should have arrays of named URLs that easily allow access wanted one.

Is there any case when devfile is used not for running IDE?
Let's imagine developer solves any issue with Postgres, to do it they need PgAdmin and database with predefined tests set... Should we call PgAdmin (`Open Source administration and development platform for PostgreSQ` (from https://www.pgadmin.org/)) IDE? Or since devfile does not have IDE component - we should use the corresponding abstract naming for the main URL as well. 
As far as I understand such abstract entity may be `endpoint`.

sleshchenko:
We agree that the corresponding field in DevWorkspace will be renamed to `ideUrl`.

An alternative solution we discussed - use some common wording like `mainEndpointUrl`, but whether we implement it or not depends on the time we have before OpenShift 4.5 code freeze

PR #80, opened 2020-05-25
Resolve minor issues with webhooks deployment

amisevsk:
### What does this PR do?
Minor fixes for webhooks service deployment.

- Don't fail when running locally with webhooks enabled
- Avoid incompatible changes to service spec.

Fixes https://github.com/eclipse/che/issues/17003

### Is it tested? How?
Tested briefly on `crc` but should be tested more thoroughly.

### DIFF
--- a/pkg/webhook/server/webhook_cert_init.go
+++ b/pkg/webhook/server/webhook_cert_init.go
@@ -14,11 +14,16 @@ package server
 import (
 	"context"
 	"errors"
+	"fmt"
+	"os"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/che-incubator/che-workspace-operator/pkg/config"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 
 	"github.com/che-incubator/che-workspace-operator/internal/controller"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -32,7 +37,10 @@ func InitWebhookServer(ctx context.Context) error {
 	}
 
 	ns, err := k8sutil.GetOperatorNamespace()
-	if err != nil {
+	if err == k8sutil.ErrRunLocal {
+		ns = os.Getenv("WATCH_NAMESPACE")
+		log.Info(fmt.Sprintf("Running operator in local mode; watching namespace %s", config.ConfigMapReference.Namespace))
+	} else if err != nil {
 		return err
 	}
 
@@ -60,13 +68,20 @@ func syncService(ctx context.Context, crclient client.Client, namespace string) 
 		if !apierrors.IsAlreadyExists(err) {
 			return err
 		}
-		existingCfg := &v1.Service{}
+		existingCfg := &corev1.Service{}
 		err := crclient.Get(ctx, types.NamespacedName{
 			Name:      secureService.Name,
 			Namespace: secureService.Namespace,
 		}, existingCfg)
+		if err != nil {
+			return err
+		}
+		clusterIP := existingCfg.Spec.ClusterIP
+		existingCfg.Spec = secureService.Spec
+		existingCfg.Spec.ClusterIP = clusterIP
+		copyLabelsAndAnnotations(secureService, existingCfg)
 
-		err = crclient.Update(ctx, secureService)
+		err = crclient.Update(ctx, existingCfg)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/webhook/server/webhook_cert_init.go
```diff
@@ -32,7 +34,10 @@ func InitWebhookServer(ctx context.Context) error {
 	}
 
 	ns, err := k8sutil.GetOperatorNamespace()
-	if err != nil {
+	if err == k8sutil.ErrRunLocal{
+		ns = os.Getenv("WATCH_NAMESPACE")
+		log.Info(fmt.Sprintf("Running operator in local mode; watching namespace %s", config.ConfigMapReference.Namespace))
+	}	else if err != nil {
```
sleshchenko, line 40: This and lines above are changed after `make fmt`

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -60,13 +65,19 @@ func syncService(ctx context.Context, crclient client.Client, namespace string)
...
 		}, existingCfg)
+		if err != nil {
+			return err
+		}
+		clusterIP := existingCfg.Spec.ClusterIP
+		existingCfg.Spec = secureService.Spec
+		existingCfg.Spec.ClusterIP = clusterIP
 
-		err = crclient.Update(ctx, secureService)
+		err = crclient.Update(ctx, existingCfg)
```
sleshchenko, line 80: It means that we may lose critical annotation `service.beta.openshift.io/serving-cert-secret-name`.
I suppose it might be a safer option to restore ClusterIP in new object
```go
		secureService.ResourceVersion = existingCfg.ResourceVersion
		// ClusterIP is immutable and assigned by cluster
		secureService.Spec.ClusterIP = existingCfg.Spec.ClusterIP
		err = crclient.Update(ctx, secureService)
```

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -83,13 +94,17 @@ func syncConfigMap(ctx context.Context, crclient client.Client, namespace string
...
 		err := crclient.Get(ctx, types.NamespacedName{
 			Name:      secureConfigMap.Name,
 			Namespace: secureConfigMap.Namespace,
 		}, existingCfg)
-
-		err = crclient.Update(ctx, secureConfigMap)
+		if err != nil {
+			return err
+		}
+		existingCfg.Data = secureConfigMap.Data
```
sleshchenko, line 105: secureConfigMap data are always empty. 
The main configuration secureConfigMap should have - `"service.beta.openshift.io/inject-cabundle": "true"` annotation.
So, I think here it makes sense port resourceVersion to secureConfigMap, and update with new object.

pkg/webhook/server/webhook_cert_init.go
```diff
@@ -32,7 +34,10 @@ func InitWebhookServer(ctx context.Context) error {
 	}
 
 	ns, err := k8sutil.GetOperatorNamespace()
-	if err != nil {
+	if err == k8sutil.ErrRunLocal{
+		ns = os.Getenv("WATCH_NAMESPACE")
+		log.Info(fmt.Sprintf("Running operator in local mode; watching namespace %s", config.ConfigMapReference.Namespace))
```
sleshchenko, line 39: Should not we early fail here with a message like - `Operator is run in local mode; webhooks are not supported. Please disable webhooks and rerun operator.`
Screenshot_20200525_114053
?
What is the purpose to create/update service and configmap?

pkg/webhook/server/webhook_cert_init.go
amisevsk: I can copy in annotations and labels, the problem is that most of the metadata fields are populated by the cluster and we don't necessarily want to overwrite those (e.g. resourceVersion, etc.).

pkg/webhook/server/webhook_cert_init.go
amisevsk: Mostly that switching between local and in-cluster development requires updating the configmap, making it tedious to switch between testing something on the cluster and starting up a local instance for debugging. If you're running locally, you don't care about webhooks because they'll never work, and if you're running in-cluster the operator has to make the services consistent anyways.

Probably the best solution is to just assume webhooks are disabled when running locally.

pkg/webhook/server/webhook_cert_init.go
amisevsk: Fixed to copy only labels/annotations to existing cfg.

pkg/webhook/server/webhook_cert_init.go
amisevsk: Add step of copying labels/annotations over to existingCfg.

pkg/webhook/server/webhook_cert_init.go
amisevsk: Then again, I used local debugging to step through the service creation process, so it's still useful even if webhooks end up not initialized.

pkg/webhook/server/webhook_cert_init.go
sleshchenko: I'm not sure about the use-case of keeping existing labels and annotation for che-workspace-controller service, like who is supposed to fill in them? Admin? Won't it break installation?
But this PR solves quite annoying bug and I'm OK to proceed with the proposed solution. We're able to change it later if we really need.

PR #81, opened 2020-05-25
Fixes for summit at sites talk

davidfestal:
### What does this PR do?

This PR proposes somes fixes I worked on in order to use the controller for the talk we did last week at Summit@sites

It provides the following fixes:
- Always respect the pull policy config map setting, for all sidecars containers, which was not the case until now
- Implements the removal of the `WorkspaceRouting` custom resource as a first step of routing reconciliation if the routing class on the `Workspace` custom resource is not the same as the the class on the existing `WorkspaceRouting` custom resource. This allows switching to openshift-oauth in a given workspace.
- Replace the `CHE_PROJECTS_ROOT` in case of plugins, as already done for `dockerImage` components
- Fix conditions in the workspace status, to have them working with the `kubectl wait ... --for=condition=Ready`:
    - use the `conditions` field name instead of `condition`
    - Change the `DeploymentReady` condition to `Ready` since it's expected to be the last one.
- Allow the ability to point to a plugin registry service by passing an environment variable with the registry service name to the controller deployment. This will allow installing the registry service with OLM along with the rest of the operator bits and have it automatically taken in account and configured inside the controller.  

No PR

### Is it tested? How?

Tested and used to prepare the Workspace CRD talk demos at Summit@sites


pkg/adaptor/plugin.go
```diff
@@ -146,7 +146,7 @@ func convertPluginContainer(workspaceId, pluginID string, brokerContainer broker
 	for _, brokerEnv := range brokerContainer.Env {
 		env = append(env, corev1.EnvVar{
 			Name:  brokerEnv.Name,
-			Value: brokerEnv.Value,
+			Value: strings.ReplaceAll(brokerEnv.Value, "$(CHE_PROJECTS_ROOT)", config.DefaultProjectsSourcesRoot),
```
sleshchenko, line 149: That should be extracted somewhere in common file, but I'm OK with merging as is and we'll update it separately.

pkg/apis/workspace/v1alpha1/workspace_types.go
```diff
@@ -72,7 +72,7 @@ const (
 	WorkspaceComponentsReady     WorkspaceConditionType = "ComponentsReady"
 	WorkspaceRoutingReady        WorkspaceConditionType = "RoutingReady"
 	WorkspaceServiceAccountReady WorkspaceConditionType = "ServiceAccountReady"
-	WorkspaceDeploymentReady     WorkspaceConditionType = "DeploymentReady"
+	WorkspaceReady               WorkspaceConditionType = "Ready"
```
sleshchenko, line 75: After the latest changes, it makes sense to have DeploymentReady and WorkspaceReady separately, since we have endpoints health check after deployment is marked as ready. WDYT?

amisevsk: :+1: Agree here, though we could use `WorkspaceReady` now and add `DeploymentReady` later, if need be. I've been generally operating under the idea that conditions represent different elements of the workspace, but not an overall summary of workspace state. E.g. `WorkspaceRoutingReady` might be false even though `WorkspaceDeploymentReady` is true. I'd prefer to limit overlap between components for this reason -- `WorkspaceReady` ends up being 'all other conditions are set'.

pkg/config/config.go
```diff
@@ -195,6 +195,11 @@ func WatchControllerConfig(ctr controller.Controller, mgr manager.Manager) error
 		return err
 	}
 
+	err = fillRegistryIfNecessary(nonCachedClient, configMap)
```
amisevsk, line 198: I'm wondering how this works now that plugin registry is optional.

pkg/apis/workspace/v1alpha1/workspace_types.go
davidfestal: +1 to have workspaceReady distinguished from DeploymentReady. I wondered exactly the same when changing it for the talk. Could this be done in a distinct PR ? It would also impact issue https://github.com/devfile/kubernetes-api/issues/44#issuecomment-633530139 which I'm currently working on.

pkg/config/config.go
davidfestal: it only works if a given env variable is passed to the controller deployment with the name of te registry K8S service.
This allows setting this env variable in the deployment section of the CSV for OLM.

amisevsk: But if we rely on the environment variable and do the check every time, why not just store the registry URL internally rather than updating a configmap. It's strange to have a deployment managing its own configmap in this way, since data should generally flow from a configmap to the deployment, not vice-versa.

davidfestal: It's just building the default registry URL from the service name when the config map doesn't exist (=> default value), which will be the case with OLM deployments.

pkg/apis/workspace/v1alpha1/workspace_types.go
sleshchenko: > Could this be done in a distinct PR ?

Sure!

PR #85, opened 2020-05-27
Add dev version of command-line-terminal that doesn't use TLS

amisevsk:
### What does this PR do?
Add plugin `che-incubator/command-line-terminal-dev/[nightly|4.5.0]` to the internal registry. This plugin is identical to the command-line-terminal plugin, except it does not enable TLS in the editor.

The normal command-line-terminal plugin only works if all components are running in the cluster
- `cluster-tls` routing is inaccessible if the controller is running locally
- `basic` routing causes a crashloop since the plugin expects certificates to be mounted.

This also blocks testing openshift-console changes locally, for the same reasons.

### Is it tested? How?
Tested on OpenShift 4.5

### DIFF
--- a/internal-registry/che-incubator/command-line-terminal-dev/4.5.0/meta.yaml
+++ b/internal-registry/che-incubator/command-line-terminal-dev/4.5.0/meta.yaml
@@ -1,0 +1,35 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: 4.5.0
+type: Che Editor
+displayName: Command Line Terminal
+title: Command Line Terminal
+description: Command Line Terminal provides the ability to start a terminal inside
+  the OpenShift Console.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/eclipse/che-machine-exec/
+firstPublicationDate: "2020-05-13"
+category: Other
+spec:
+  endpoints:
+    - name: command-line-terminal
+      public: true
+      targetPort: 4444
+      attributes:
+        protocol: http
+        type: ide
+        discoverable: false
+        secure: true
+        cookiesAuthEnabled: true
+  containers:
+    - name: command-line-terminal
+      image: "quay.io/eclipse/che-machine-exec:nightly"
+      command: ["/go/bin/che-machine-exec",
+                "--authenticated-user-id", "$(CHE_WORKSPACE_CREATOR)",
+                "--idle-timeout", "$(CHE_WORKSPACE_IDLE_TIMEOUT)"]
+      ports:
+        - exposedPort: 4444
+      env:
+        - name: USE_BEARER_TOKEN
+          value: true
--- a/internal-registry/che-incubator/command-line-terminal-dev/4.5.0/meta.yaml
+++ b/internal-registry/che-incubator/command-line-terminal-dev/4.5.0/meta.yaml
@@ -1,15 +1,16 @@
 apiVersion: v2
 publisher: che-incubator
-name: command-line-terminal
+name: command-line-terminal-dev
 version: 4.5.0
 type: Che Editor
-displayName: Command Line Terminal
-title: Command Line Terminal
+displayName: Dev Command Line Terminal
+title: Dev Command Line Terminal
 description: Command Line Terminal provides the ability to start a terminal inside
-  the OpenShift Console.
+  the OpenShift Console. The development version does not run with TLS enabled and
+  is intended for development purposes only.
 icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
 repository: https://github.com/eclipse/che-machine-exec/
-firstPublicationDate: "2020-05-13"
+firstPublicationDate: "2020-06-01"
 category: Other
 spec:
   endpoints:
--- a/internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
+++ b/internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
@@ -1,0 +1,35 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: nightly
+type: Che Editor
+displayName: Command Line Terminal
+title: Command Line Terminal
+description: Command Line Terminal provides the ability to start a terminal inside
+  the OpenShift Console.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/eclipse/che-machine-exec/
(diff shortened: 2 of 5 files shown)


internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+        type: ide
+        discoverable: false
+        secure: true
+        cookiesAuthEnabled: true
+  containers:
+    - name: command-line-terminal
+      image: "quay.io/eclipse/che-machine-exec:nightly"
+      command: ["/go/bin/che-machine-exec",
+                "--authenticated-user-id", "$(CHE_WORKSPACE_CREATOR)",
+                "--idle-timeout", "$(CHE_WORKSPACE_IDLE_TIMEOUT)"]
```
sleshchenko, line 30: Do you think we should exclude that from the internal registry for released dockerimage?
Or it's safe enough to provide it as well?

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: nightly
+type: Che Editor
+displayName: Command Line Terminal
+title: Command Line Terminal
+description: Command Line Terminal provides the ability to start a terminal inside
```
sleshchenko, line 8: I think it worths mentioning here why dev version exist and how it's supposed to be used - for openshift local bridge...

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  the OpenShift Console.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+repository: https://github.com/eclipse/che-machine-exec/
+firstPublicationDate: "2020-05-13"
```
sleshchenko, line 12: I know it changes nothing but it's copied/pasted date )

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
+version: nightly
+type: Che Editor
+displayName: Command Line Terminal
```
sleshchenko, line 6:
```suggestion
displayName: Dev Version of Command Line Terminal
```

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
```
sleshchenko, line 3:
```suggestion
name: dev-command-line-terminal
```

internal-registry/che-incubator/command-line-terminal-dev/4.5.0/meta.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v2
+publisher: che-incubator
+name: command-line-terminal
```
sleshchenko, line 3:
```suggestion
name: dev-command-line-terminal
```

internal-registry/che-incubator/command-line-terminal/nightly/meta.yaml
```diff
@@ -12,26 +12,26 @@ firstPublicationDate: "2020-05-13"
...
-        protocol: http
-        type: ide
-        discoverable: false
-        secure: true
-        cookiesAuthEnabled: true
+  - name: command-line-terminal
+    public: true
+    targetPort: 4444
+    attributes:
+      protocol: http
```
sleshchenko, line 19: Since we explicitly run the process with `tls`, does it make sense to specify `https` here?

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
amisevsk: I think it's okay to provide as well -- someone could always deploy a custom registry and use that to run whatever plugins they want. If it's not included in the release, then testing anything that depends on this change would require building and deploying the operator (which could be tedious for e.g. console changes). 

If there's a security concern, though, I'm okay with leaving it out.

internal-registry/che-incubator/command-line-terminal/nightly/meta.yaml
amisevsk: It likely does, but our story on using configuring TLS is kind of unclear -- we also have the `secure` attribute, and as far as I can tell, protocol is converted `http -> https / ws -> wss` based on that property.

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
amisevsk: Updated description to mention that it doesn't run with TLS enabled. I think there are uses beyond running bridge locally, as this version is required for all `routingClass`es other than `cluster-tls`.

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
amisevsk: Good catch :)

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
amisevsk: Updated

internal-registry/che-incubator/command-line-terminal-dev/4.5.0/meta.yaml
amisevsk: Updated to `command-line-terminal-dev` to match path.

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
amisevsk: Updated to `command-line-terminal-dev` to match path.

internal-registry/che-incubator/command-line-terminal-dev/nightly/meta.yaml
sleshchenko: Let's figure out later.

PR #86, opened 2020-05-28
Update default config map properties when controller-config isn't cre…

JPinkney:
…ated on cluster

### What does this PR do?
This PR updates the default config map properties for when we are deploying the operator with OLM and we cannot deploy the che-workspace-controller configmap.

https://github.com/eclipse/che/issues/17042

### Is it tested? How?

Tested by removing all references to controller-config.yaml in the Makefile and deleting the controller-config yaml file from the repo.

Then deploying the changes with `make deploy` and then creating a cloud shell and see that everything starts correctly

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -235,10 +235,7 @@ func buildDefaultConfigMap(cm *corev1.ConfigMap) {
 	cm.Name = ConfigMapReference.Name
 	cm.Namespace = ConfigMapReference.Namespace
 
-	cm.Data = map[string]string{
-		routingSuffix:              defaultRoutingSuffix,
-		pluginArtifactsBrokerImage: defaultPluginArtifactsBrokerImage,
-	}
+	cm.Data = map[string]string{}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -238,6 +238,7 @@ func buildDefaultConfigMap(cm *corev1.ConfigMap) {
 	cm.Data = map[string]string{
 		routingSuffix:              defaultRoutingSuffix,
 		pluginArtifactsBrokerImage: defaultPluginArtifactsBrokerImage,
+		webhooksEnabled:            defaultNoConfigMapWebhooksEnabled,
```
sleshchenko, line 241: Do we really need defaults on configmap level? Maybe embedded in controller defaults are enough?
Then it would be easier to change defaults if we need it.
Otherwise - we can't check if it's a default in configmap we should be free to change or it's admin set value?

If we talk about webhooks enabled - then it's going to be unmodified I believe.
But if we talk about plugin broker - I believe the default value is going to be changed pretty soon.

So, I would create empty may just for admin easier find to place where put custom configuration but without values.
Maybe only ones which are evaluated depending on the cluster, like routingSuffix on OpenShift Infra.
@amisevsk @JPinkney WDYT?

PR #89, opened 2020-06-03
Make workspace-related objects immutable except by controller

amisevsk:
### What does this PR do?
- Make objects owned by workspace (e.g. routes, services) immutable except by workspace controller 
- Make immutable workspaces editable by creator and controller.

Resolves https://github.com/eclipse/che/issues/16943

### Is it tested? How?
Tested on `crc`:
1. Create a workspace
2. Try to edit service, route, ingress, pod, deployment -> denied 
3. Set workspace to immutable
4. Edit workspace as owner -> changes are still propagated (controller can update resources)
5. Login as another user, edit workspace -> denied (not owner)

### DIFF
--- a/pkg/webhook/workspace/config.go
+++ b/pkg/webhook/workspace/config.go
@@ -15,7 +15,12 @@ import (
 	"context"
 	"errors"
 
+	"fmt"
+	"os"
+
+	"github.com/che-incubator/che-workspace-operator/pkg/config"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/che-incubator/che-workspace-operator/internal/controller"
 	"github.com/che-incubator/che-workspace-operator/pkg/webhook/server"
@@ -53,6 +58,11 @@ func Configure(ctx context.Context) error {
 		return nil
 	}
 
+	saUID, err := controllerSAUID(ctx, c)
+	if err != nil {
+		return err
+	}
+
 	ownRef, err := controller.FindControllerOwner(ctx, c)
 	if err != nil {
 		return err
@@ -85,7 +95,7 @@ func Configure(ctx context.Context) error {
 		log.Info("Created workspace mutating webhook configuration")
 	}
 
-	server.GetWebhookServer().Register(mutateWebhookPath, &webhook.Admission{Handler: NewResourcesMutator()})
+	server.GetWebhookServer().Register(mutateWebhookPath, &webhook.Admission{Handler: NewResourcesMutator(saUID)})
 
 	validateWebhookCfg.SetOwnerReferences([]metav1.OwnerReference{*ownRef})
 
@@ -110,7 +120,7 @@ func Configure(ctx context.Context) error {
 		log.Info("Created workspace validating webhook configuration")
 	}
 
-	server.GetWebhookServer().Register(validateWebhookPath, &webhook.Admission{Handler: NewResourcesValidator()})
+	server.GetWebhookServer().Register(validateWebhookPath, &webhook.Admission{Handler: NewResourcesValidator(saUID)})
 
 	return nil
 }
@@ -131,4 +141,25 @@ func getValidateWebhook(ctx context.Context, c client.Client, validateWebhookCfg
 		Namespace: validateWebhookCfg.Namespace,
 	}, existingCfg)
 	return existingCfg, err
+}
+
+func controllerSAUID(ctx context.Context, c client.Client) (string, error) {
+	saName := os.Getenv(config.ControllerServiceAccountNameEnvVar)
+	if saName == "" {
+		return "", fmt.Errorf("could not get service account name")
+	}
+	namespace, err := k8sutil.GetOperatorNamespace()
+	if err != nil {
+		return "", err
+	}
+	namespacedName := types.NamespacedName{
+		Namespace: namespace,
+		Name:      saName,
+	}
+	sa := &corev1.ServiceAccount{}
+	err = c.Get(ctx, namespacedName, sa)
+	if err != nil {
+		return "", err
+	}
+	return string(sa.UID), nil
 }
--- a/pkg/webhook/workspace/config.go
+++ b/pkg/webhook/workspace/config.go
@@ -58,7 +58,7 @@ func Configure(ctx context.Context) error {
 		return nil
 	}
 
-	saUID, err := controllerSAUID(ctx, c)
+	saUID, saName, err := controllerSAUID(ctx, c)
 	if err != nil {
 		return err
 	}
@@ -95,7 +95,7 @@ func Configure(ctx context.Context) error {
 		log.Info("Created workspace mutating webhook configuration")
 	}
 
-	server.GetWebhookServer().Register(mutateWebhookPath, &webhook.Admission{Handler: NewResourcesMutator(saUID)})
+	server.GetWebhookServer().Register(mutateWebhookPath, &webhook.Admission{Handler: NewResourcesMutator(saUID, saName)})
 
 	validateWebhookCfg.SetOwnerReferences([]metav1.OwnerReference{*ownRef})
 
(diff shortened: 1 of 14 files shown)


pkg/webhook/workspace/handler/immutable.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (h *WebhookHandler) handleImmutableWorkspace(oldWksp, newWksp *v1alpha1.Workspace, uid string) admission.Response {
+	creatorUID := oldWksp.Labels[config.WorkspaceCreatorLabel]
+	if uid == creatorUID || uid == h.ControllerUID {
+		return admission.Allowed("immutable workspace is updated by owner or controller")
+	}
+	if cmp.Equal(oldWksp, newWksp, StopStartDiffOption[:]...) {
+		return admission.Allowed("immutable workspace is started/stopped")
+	}
+	return admission.Denied(fmt.Sprintf("workspace '%s' is immutable. To make modifications it must be deleted and recreated", oldWksp.Name))
```
sleshchenko, line 38: Seems the message should be rewritten, like `create your own workspace and then modify it` :stuck_out_tongue_closed_eyes:

pkg/webhook/workspace/config.go
```diff
@@ -31,6 +38,11 @@ func Configure(ctx context.Context) error {
 		return err
 	}
 
+	saUID, err := controllerUID(ctx, c)
```
sleshchenko, line 41: Consider moving in below `if !server.IsSetUp()` condition to make that env var optional when it's not used - local operator run, or when webhooks are disabled.

pkg/webhook/workspace/config.go
```diff
@@ -117,7 +129,28 @@ func Configure(ctx context.Context) error {
 		log.Info("Created workspace validating webhook configuration")
 	}
 
-	server.GetWebhookServer().Register(validateWebhookPath, &webhook.Admission{Handler: NewResourcesValidator()})
+	server.GetWebhookServer().Register(validateWebhookPath, &webhook.Admission{Handler: NewResourcesValidator(saUID)})
 
 	return nil
 }
+
+func controllerUID(ctx context.Context, c client.Client) (string, error) {
```
sleshchenko, line 137:
```suggestion
func controllerSAUID(ctx context.Context, c client.Client) (string, error) {
```

sleshchenko: Does it make sense to move it somewhere like `/internal/controller/service_account.go` ?

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -0,0 +1,49 @@
+package handler
```
sleshchenko, line 1: missing license header

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
+	"github.com/che-incubator/che-workspace-operator/pkg/config"
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+func (h *WebhookHandler) HandleImmutable(ctx context.Context, req admission.Request) admission.Response {
```
sleshchenko, line 16: I think `HandleImmutable` and `handleImmutableWorkspace` should be both private or public.

pkg/webhook/workspace/mutating_cfg.go
```diff
@@ -29,72 +29,111 @@ func buildMutateWebhookCfg() *v1beta1.MutatingWebhookConfiguration {
...
 	labelExistsOp := metav1.LabelSelectorOpExists
 	equivalentMatchPolicy := v1beta1.Equivalent
-	return &v1beta1.MutatingWebhookConfiguration{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: mutateWebhookCfgName,
+
+	workspaceMutateWebhook := v1beta1.MutatingWebhook{
+		Name:          "mutate.che-workspace-controller.svc",
+		FailurePolicy: &mutateWebhookFailurePolicy,
+		ClientConfig: v1beta1.WebhookClientConfig{
```
sleshchenko, line 36: Consider extracting client config to a separate variable.

pkg/webhook/workspace/mutating_cfg.go
```diff
@@ -29,72 +29,111 @@ func buildMutateWebhookCfg() *v1beta1.MutatingWebhookConfiguration {
...
+			{
+				Operations: []v1beta1.OperationType{v1beta1.Create, v1beta1.Update},
+				Rule: v1beta1.Rule{
+					APIGroups:   []string{"apps"},
+					APIVersions: []string{"v1"},
+					Resources:   []string{"deployments"},
 				},
 			},
+			{
+				Operations: []v1beta1.OperationType{v1beta1.Update},
```
sleshchenko, line 102: Do you think it makes sense to deny creating other workspace-related objects for everyone except creator and controller?
For sure except pods - since they are managed by replicasSets...

pkg/webhook/workspace/handler/immutable.go
amisevsk: Good catch -- done :)

pkg/webhook/workspace/config.go
amisevsk: Done.

pkg/webhook/workspace/config.go
amisevsk: For now, it's only used for webhooks, but we could always move it to an internal utility if we find we need the UID elsewhere.

pkg/webhook/workspace/handler/immutable.go
amisevsk: Fixed

pkg/webhook/workspace/handler/immutable.go
amisevsk: The convention I was going for was:
- Function takes `(context, request)` and returns `admission.Response` -> public
- Function takes `old, new <object>` and returns (continue, message)` -> private

Perhaps the naming of the functions is not clear, but the goal is to have standalone methods that can be called from `mutate.go` be exported, and internal methods to be helpers for e.g. `handleImmutableWorkspace`.

pkg/webhook/workspace/mutating_cfg.go
amisevsk: Done, good point.

pkg/webhook/workspace/mutating_cfg.go
amisevsk: I think it makes sense; added in https://github.com/che-incubator/che-workspace-operator/pull/89/commits/aa9415d6364d8abb21cfe92dd252b5f351eb1b03

sleshchenko:
@amisevsk :warning: It seems not to work after the latest changes 
```
,"error":"admission webhook \"mutate-ws-resources.che-workspace-controller.svc\" denied the request: Only the workspace controller can create workspace objects
```

sleshchenko:
really strange
```go
return admission.Denied("Only the workspace controller can create workspace objects. " + req.UserInfo.Username + ":" + req.UserInfo.UID + " - " + h.ControllerUID)
```
says:
```
Only the workspace controller can create workspace objects. 
system:serviceaccount:che-workspace-controller:che-workspace-controller: 
- ac8d6cba-f8cb-4cea-9e18-f6f26d1f57c0
```

amisevsk:
Turns out *only* routes have this UID issue, opened issue https://github.com/eclipse/che/issues/17114 to track (though it's probably an OpenShift issue?)

Added a workaround where we compare on username for routes specifically, which is less secure.

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (h *WebhookHandler) handleImmutableRoute(oldObj, newObj runtime.Object, username string) (allowed bool, msg string) {
+	if username == h.ControllerSAName {
+		return true, ""
+	}
+	if reflect.DeepEqual(oldObj, newObj) {
+		return true, ""
+	}
+	return false, "object is owned by workspace and cannot be updated."
+}
```
sleshchenko, line 88:
```suggestion
}

```

amisevsk: Forgot the `make fmt` :)

PR #94, opened 2020-06-04
Revert changes to label used for workspace ID

amisevsk:
### What does this PR do?
The machine-exec plugin expects the workspace pod to be labelled with
```
    che.workspace_id
```
Changing this label name causes pod resolution to break in the machine-exec plugin.

We should reapply this change after changing machine-exec to accomodate the alternate label.

Fixes https://github.com/eclipse/che/issues/17099


JPinkney:
Just confirmed, yeah its a regression

PR #95, opened 2020-06-05
Make sure webhooks reference the correct service

JPinkney:

### What does this PR do?
This PR makes sure that the webhooks reference the correct service when the operator is deployed in a different namespace than che-workspace-controller

https://github.com/eclipse/che/issues/17098

### Is it tested? How?

Tested on CRC with webhooks enabled

### DIFF
--- a/pkg/webhook/workspace/config.go
+++ b/pkg/webhook/workspace/config.go
@@ -14,6 +14,8 @@ package workspace
 import (
 	"context"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+
 	"github.com/che-incubator/che-workspace-operator/internal/controller"
 	"github.com/che-incubator/che-workspace-operator/pkg/webhook/server"
 	"k8s.io/api/admissionregistration/v1beta1"
@@ -27,6 +29,11 @@ import (
 func Configure(ctx context.Context) error {
 	log.Info("Configuring workspace webhooks")
 	c, err := controller.CreateClient()
+	if err != nil {
+		return err
+	}
+
+	namespace, err := k8sutil.GetOperatorNamespace()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/webhook/workspace/config.go
```diff
@@ -31,6 +32,11 @@ func Configure(ctx context.Context) error {
 		return err
 	}
 
+	namespace, err := k8sutil.GetOperatorNamespace()
```
sleshchenko, line 35: Just wondering if it makes sense to evaluate it once and return from Config object.

pkg/config/config.go
```diff
@@ -106,6 +106,10 @@ func (wc *ControllerConfig) GetWebhooksEnabled() string {
 	return wc.GetPropertyOrDefault(webhooksEnabled, defaultWebhooksEnabled)
 }
 
+func (wc *ControllerConfig) GetNamespace() string {
```
amisevsk, line 109: There's also [`k8sutil.GetOperatorNamespace`](https://godoc.org/github.com/operator-framework/operator-sdk/pkg/k8sutil#GetOperatorNamespace) that we use elsewhere -- may be more reliable.

JPinkney: Yeah, thats what I originally was using but I didn't know if it made sense to constantly use that or do
>Just wondering if it makes sense to evaluate it once and return from Config object.

sleshchenko: I tried to play with it and it's what I've done https://github.com/JPinkney/che-workspace-operator/compare/fix-webhook-reference...sleshchenko:fix-webhook-reference-namespace-as-config?expand=1

I'm not fully happy with that since it seems all usage of `k8sUtils.GetOperatorNamespace` can't be replaced with `Config.GetOperatorNamespace` because init of config it not done as soon as possible...
But using ConfigMap's namespace it's not really the same as OperatorNamespace.
So, I'm OK with reverting to use K8sUtil and maybe later we'll refactor it in a nice and clear manner :crossed_fingers:

PR #96, opened 2020-06-05
Rework cert generation so that it uses two deployments

JPinkney:

### What does this PR do?
This PR reworks the way we do cert generation completely and has a few massive benefits:
1. We can mount the configmap in the controller deployment itself, meaning that we don't need to restart the controller once the configmaps have been created
2. We save SO much time because the controller starts right away
3. We save ourself from things like https://github.com/eclipse/che/issues/16980#issuecomment-638257576
4. Fixes a deployment issue when installing via OLM. Basically what was happening was that If you modify the deployment to add volumes after (like we previously did with cert generation) then OLM will automatically use the deployment template from the CSV, instead of our changed deployment for creating the new pod.

The way it works is that when webhooks are enabled we deploy an additional deployment called `che-workspace-controller-cert-gen` that has the sole responsibility of creating the configmap and service needed for the certs. Then, instead of mounting the configmap into the controller deployment later we mount the configmap in the controller deployment yaml right away. That way when the controller deployment is ready and started it already has all the certs available.

Since the `che-workspace-controller-cert-gen` binary is so small it actually finishes running before the controller deployment even finishes getting setup. In the worst case, the controller deployment will block for 1-2 seconds (this is only theoretical as I've never actually seen it happen). When the controller starts setting up the webhooks (and the certs are available) it will delete the `che-workspace-controller-cert-gen` deployment.

Ideally, we'd use a Job for creating `che-workspace-controller-cert-gen` but since OLM doesn't allow you to specify jobs in the bundle, we will have to use deployment.

Part of https://github.com/eclipse/che/issues/16980

### Is it tested? How?

`make docker_cert`
`make docker`
`make deploy`

with both webhooks on and then webhooks off


cert-generation/main.go
```diff
@@ -50,66 +87,12 @@ func getSecureServiceSpec(namespace string) *corev1.Service {
...
-		ReadOnly:  true,
-	}
-}
 
-func getPort() corev1.ContainerPort {
-	port := int32(8443)
-	return corev1.ContainerPort{
-		ContainerPort: port,
-		Name:          webhookServerName,
+	// Wait until its cleaned up by the workspace-controller
```
sleshchenko, line 95:
```suggestion
	// Wait until it's cleaned up by the workspace-controller
```

deploy/os/che-workspace-controller-cert-gen-deployment.yaml
```diff
@@ -0,0 +1,23 @@
...
+        app: che-workspace-controller-cert-gen
+      annotations:
+        kubectl.kubernetes.io/restartedAt: ""
+    spec:
+      serviceAccountName: che-workspace-controller
+      containers:
+        - name: che-workspace-controller-cert-gen
+          image: quay.io/che-incubator/che-workspace-controller-cert-gen:latest
+          imagePullPolicy: Always
+
```
sleshchenko, line 23:
```suggestion
```

cert-generation/main.go
```diff
@@ -50,66 +87,12 @@ func getSecureServiceSpec(namespace string) *corev1.Service {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      certConfigMapName,
-			Namespace: namespace,
-			Annotations: map[string]string{
-				"service.beta.openshift.io/inject-cabundle": "true",
-			},
-		},
+	_, err = clientset.CoreV1().Services(namespace).Create(service)
+	if err != nil {
+		os.Exit(1)
```
sleshchenko, line 92: figuring out why cert generation pod failed without any error logs could be a nightmare =)

cert-generation/main.go
```diff
@@ -50,66 +87,12 @@ func getSecureServiceSpec(namespace string) *corev1.Service {
...
-	}
-}
 
-func getPort() corev1.ContainerPort {
-	port := int32(8443)
-	return corev1.ContainerPort{
-		ContainerPort: port,
-		Name:          webhookServerName,
+	// Wait until its cleaned up by the workspace-controller
+	for {
```
sleshchenko, line 96: Again adding logs could be useful like `I've done my work and waiting to be utilized. Don't feel sorry for me I was initially designed to die if everything is fine` :smile:

deploy/os/che-workspace-controller-cert-gen-deployment.yaml
```diff
@@ -0,0 +1,23 @@
...
+    matchLabels:
+      app: che-workspace-controller-cert-gen
+  template:
+    metadata:
+      labels:
+        app: che-workspace-controller-cert-gen
+      annotations:
+        kubectl.kubernetes.io/restartedAt: ""
+    spec:
+      serviceAccountName: che-workspace-controller
```
sleshchenko, line 18: does it need SA or it's just for a namespace to be mount?
If so - isn't it better just to provision namespace value with env var and valueFrom: metadata.namespace?

deploy/os/che-workspace-controller-cert-gen-deployment.yaml
```diff
@@ -0,0 +1,23 @@
...
+      containers:
+        - name: che-workspace-controller-cert-gen
+          image: quay.io/che-incubator/che-workspace-controller-cert-gen:latest
```
sleshchenko, line 21: I think ci for this image is needed before merging

cert-generation/main.go
sleshchenko: Screenshot_20200609_155615

deploy/os/che-workspace-controller-cert-gen-deployment.yaml
JPinkney: It needs the service account for the configmap/service to be created. Otherwise it just ends up erroring. Now that I've fixed the logging issue it will show the error in console if you don't have the service account

Makefile
```diff
@@ -258,3 +278,4 @@ help: Makefile
 	@echo '    ADMIN_CTX                  - Kubectx entry that should be used during work with cluster. The current will be used if omitted'
 	@echo '    REGISTRY_ENABLED           - Whether the plugin registry should be deployed'
 	@echo '    DEVWORKSPACE_API_VERSION   - Branch or tag of the github.com/devfile/kubernetes-api to depend on. Defaults to master'
+	@echo '    CERT_IMG           		  - The name of the cert generator image'
```
sleshchenko, line 281: It does not seem to be aligned to other messages

cert-generation/main.go
```diff
@@ -0,0 +1,127 @@
...
+			log.Fatal("Failed when attempting to create configmap: ", err)
+		}
+	} else {
+		_, err = clientset.CoreV1().ConfigMaps(namespace).Update(configMap)
+		if err != nil {
+			log.Fatal("Failed when attempting to update configmap: ", err)
+		}
+	}
+
+	label := map[string]string{"app": "che-workspace-controller"}
```
sleshchenko, line 80: Should we change that label to something like `devworkspace-controller`?

cert-generation/main.go
```diff
@@ -0,0 +1,127 @@
...
+		service.ResourceVersion = clusterService.ResourceVersion
+
+		_, err = clientset.CoreV1().Services(namespace).Update(service)
+		if err != nil {
+			log.Fatal("Failed when attempting to update service: ", err)
+		}
+	}
+
+	// Wait until it's cleaned up by the workspace-controller
+	log.Println("I've done my work and waiting to be utilized. Don't feel sorry for me I was initially designed to die if everything is fine")
```
sleshchenko, line 124: If was a joke but up to you to leave it or reformulate )

JPinkney: I thought it was a funny secret haha

PR #98, opened 2020-06-08
Stop webhooks from being removed once webhook configuration is turned…

JPinkney:
… off

### What does this PR do?
This PR makes it so that if you go from `webhooks enabled` -> `webhooks disabled` the controller will error out with a message that `Webhooks have previously been set up and cannot be removed automatically ...` 

Part of https://github.com/eclipse/che/issues/16980

### Is it tested? How?

First enable webhooks and wait for the controller to start, then edit the configmap to disable webhooks and then run `make _do_restart`. You'll notice that the controller fails with an error message telling the admin what to do to fix it.

To test that it works after you remove the validating/admission webhooks first remove them:
`kubectl delete validatingwebhookconfigurations/workspace.che.eclipse.org`
`kubectl delete mutatingwebhookconfigurations/workspace.che.eclipse.org`

then run `make _do_restart`

and it should be started succesfully

### DIFF
--- a/pkg/webhook/workspace/config.go
+++ b/pkg/webhook/workspace/config.go
@@ -13,6 +13,7 @@ package workspace
 
 import (
 	"context"
+	"errors"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 
@@ -22,6 +23,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/webhook/workspace/config.go
```diff
@@ -31,32 +33,19 @@ func Configure(ctx context.Context) error {
...
-		}
-		if err = c.Delete(ctx, &v1beta1.ValidatingWebhookConfiguration{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: validateWebhookCfgName,
-			}}); err != nil {
-			if !apierrors.IsNotFound(err) {
-				return err
-			}
-		}
+		_, mutatingWebhookErr := checkMutatingWebhookExistance(ctx, c, mutateWebhookCfg)
```
amisevsk, line 39: Typo :)
```suggestion
		_, mutatingWebhookErr := checkMutatingWebhookExistence(ctx, c, mutateWebhookCfg)
```

pkg/webhook/workspace/config.go
```diff
@@ -121,3 +107,21 @@ func Configure(ctx context.Context) error {
...
+func checkMutatingWebhookExistance(ctx context.Context, c client.Client, mutatingWebhookCfg *v1beta1.MutatingWebhookConfiguration) (*v1beta1.MutatingWebhookConfiguration, error) {
+	existingCfg := &v1beta1.MutatingWebhookConfiguration{}
+	err := c.Get(ctx, types.NamespacedName{
+		Name:      mutatingWebhookCfg.Name,
+		Namespace: mutatingWebhookCfg.Namespace,
+	}, existingCfg)
+	return existingCfg, err
+}
+
+func checkValidatingWebhookExistance(ctx context.Context, c client.Client, validateWebhookCfg *v1beta1.ValidatingWebhookConfiguration) (*v1beta1.ValidatingWebhookConfiguration, error) {
```
amisevsk, line 120: I think naming-wise, `get[Mutating|Validating]Webhook` is more accurate here

pkg/webhook/workspace/config.go
```diff
@@ -31,32 +33,19 @@ func Configure(ctx context.Context) error {
...
-				return err
-			}
-		}
+		_, mutatingWebhookErr := getMutatingWebhook(ctx, c, mutateWebhookCfg)
+		_, validatingWebhookErr := getValidatingWebhook(ctx, c, validateWebhookCfg)
 
-		log.Info("Existing workspace related webhook configurations are removed")
+		// No errors from either means that they are on the cluster
+		if mutatingWebhookErr == nil || validatingWebhookErr == nil {
+			return errors.New(`Webhooks have previously been set up and cannot be removed automatically. The admin must make sure that all workspaces are stopped, webhooks are removed, webhooks are reconfigured then restart the controller`)
```
sleshchenko, line 44: 1. It makes sense to reformulate `webhooks are removed` something like `workspace-related webhooks configurations are removed`
2. `webhooks are reconfigured` could you clarify what do you mean here? Register `AdmissionWebhooksConfiguration` on the cluster? Or set `controller.webhooks.enabled` for controller to false? If so - isn't user expected to fall here only when webhooks are disabled?
```suggestion
			return errors.New(`Webhooks have previously been set up and cannot be removed automatically. Configure the controller to use webhooks again or make sure that all workspaces are stopped, webhooks configuration are removed and then restart the controller`)
```

pkg/webhook/workspace/config.go
```diff
@@ -121,3 +107,21 @@ func Configure(ctx context.Context) error {
...
+func getMutatingWebhook(ctx context.Context, c client.Client, mutatingWebhookCfg *v1beta1.MutatingWebhookConfiguration) (*v1beta1.MutatingWebhookConfiguration, error) {
+	existingCfg := &v1beta1.MutatingWebhookConfiguration{}
+	err := c.Get(ctx, types.NamespacedName{
+		Name:      mutatingWebhookCfg.Name,
+		Namespace: mutatingWebhookCfg.Namespace,
+	}, existingCfg)
+	return existingCfg, err
+}
+
+func getValidatingWebhook(ctx context.Context, c client.Client, validateWebhookCfg *v1beta1.ValidatingWebhookConfiguration) (*v1beta1.ValidatingWebhookConfiguration, error) {
```
amisevsk, line 120:
```suggestion
func getValidatingWebhook(ctx context.Context, c client.Client, validatingWebhookCfg *v1beta1.ValidatingWebhookConfiguration) (*v1beta1.ValidatingWebhookConfiguration, error) {
```

PR #99, opened 2020-06-08
Use che-machine-exec target port selector

JPinkney:

### What does this PR do?
This PR uses these changes from che-machine-exec: https://github.com/eclipse/che-machine-exec/pull/110 to set the target pod selector. This PR will then set TARGET_POD_SELECTOR to be `io.devfile.workspace/workspace_id=$(WORKSPACE_ID)` for the workspace operator, allowing us to change the workspaceID label so that the workspace operator and che operator can live on the same cluster.

https://github.com/eclipse/che/issues/16924

### Is it tested? How?

Use the changes from https://github.com/eclipse/che-machine-exec/pull/110 for che-machine-exec. Deploy the workspace operator (webhooks enabled) and che operator.  Try to start workspace in che itself, then try to start cloud shell workspace. Both should start successfully

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -30,7 +30,7 @@ const (
 	PVCStorageSize            = "1Gi"
 
 	// WorkspaceIDLabel is label key to store workspace identifier
-	WorkspaceIDLabel = "che.workspace_id"
+	WorkspaceIDLabel = "org.eclipse.che.workspace/id"
 
 	// WorkspaceEndpointNameAnnotation is the annotation key for storing an endpoint's name from the devfile representation
 	WorkspaceEndpointNameAnnotation = "controller.devfile.io/endpoint_name"
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -30,7 +30,7 @@ const (
 	PVCStorageSize            = "1Gi"
 
 	// WorkspaceIDLabel is label key to store workspace identifier
-	WorkspaceIDLabel = "org.eclipse.che.workspace/id"
+	WorkspaceIDLabel = "controller.devfile.io/workspace_id"
 
 	// WorkspaceEndpointNameAnnotation is the annotation key for storing an endpoint's name from the devfile representation
 	WorkspaceEndpointNameAnnotation = "controller.devfile.io/endpoint_name"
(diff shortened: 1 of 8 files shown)


pkg/controller/workspace/env/common_env_vars.go
```diff
@@ -66,5 +66,9 @@ func CommonEnvironmentVariables(workspaceName, workspaceId, namespace, creator s
 			Name:  "CHE_WORKSPACE_IDLE_TIMEOUT",
 			Value: config.ControllerCfg.GetWorkspaceIdleTimeout(),
 		},
+		{
+			Name:  "TARGET_POD_SELECTOR",
```
sleshchenko, line 70: I wonder if it should be 1. common env var or 2. setting specific for che-machine-exec https://github.com/che-incubator/che-workspace-operator/blob/master/internal-registry/redhat-developer/web-terminal/4.5.0/meta.yaml#L29

If 1. - then it would be better to rename that env var to something workspace related, like `WORKSPACE_POD_SELECTOR` 
if 2 - the current name is good

amisevsk: To add -- if [we make the machine-exec side accept a commandline arg](https://github.com/eclipse/che-machine-exec/pull/110#pullrequestreview-426702439) in addition to an env var, this could be configured by updating the meta.yaml for machine-exec plugins: https://github.com/che-incubator/che-workspace-operator/compare/master...amisevsk:fix-label

pkg/config/constants.go
```diff
@@ -30,7 +30,7 @@ const (
 	PVCStorageSize            = "1Gi"
 
 	// WorkspaceIDLabel is label key to store workspace identifier
-	WorkspaceIDLabel = "che.workspace_id"
+	WorkspaceIDLabel = "controller.devfile.io/id"
```
sleshchenko, line 33: I wonder if `controller.devfile.io/workspace_id` is a bit cleaner

sleshchenko:
Tested cloud-shell and web-terminal-dev workspaces. Work as expected

pkg/controller/workspace/provision/deployment.go
```diff
@@ -140,11 +140,11 @@ func getSpecDeployment(
 	creator := workspace.Labels[config.WorkspaceCreatorLabel]
 	commonEnv := env.CommonEnvironmentVariables(workspace.Name, workspace.Status.WorkspaceId, workspace.Namespace, creator)
 	for idx := range podAdditions.Containers {
-		podAdditions.Containers[idx].Env = append(podAdditions.Containers[idx].Env, commonEnv...)
+		podAdditions.Containers[idx].Env = append(commonEnv, podAdditions.Containers[idx].Env...)
```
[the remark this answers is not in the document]
sleshchenko: I've reverted that changes and configured che-machine-exec with an argument instead of env var.
There is a dedicated issue to deal with env var interpolation https://github.com/eclipse/che/issues/17136

PR #101, opened 2020-06-09
Move annotations/labels to controller.devfile.io, rename webhook

amisevsk:
### What does this PR do?
- Move annotations/labels to use controller.devfile.io as a base
- Rename webhook to controller.devfile.io

Part of https://github.com/eclipse/che/issues/17122

### Is it tested? How?
Tested on `crc`.


pkg/config/constants.go
```diff
@@ -33,21 +33,21 @@ const (
 	WorkspaceIDLabel = "che.workspace_id"
 
 	// WorkspaceEndpointNameAnnotation is the annotation key for storing an endpoint's name from the devfile representation
-	WorkspaceEndpointNameAnnotation = "io.devfile.workspace/endpoint_name"
+	WorkspaceEndpointNameAnnotation = "controller.devfile.io/endpoint_name"
 
 	// WorkspaceNameLabel is label key to store workspace name
-	WorkspaceNameLabel = "io.devfile.workspace/name"
+	WorkspaceNameLabel = "controller.devfile.io/name"
```
sleshchenko, line 39: Are we going to use `controller.devfile.io/workspace_id` for workspace id?
Since there is no workspace word in prefix, I think it makes sense consider using `controller.devfile.io/workspace_name`. But I'm not sure... WDYT?
```suggestion
	WorkspaceNameLabel = "controller.devfile.io/workspace_name"
```

PR #103, opened 2020-06-10
Move subresources to controller.devfile.io APIGroup

amisevsk:
### What does this PR do?
Moves Component and WorkspaceRouting subresources from `workspace.che.eclipse.org` API group to `controller.devfile.io`.

Commits should be broken down into a step-by-step process, with separate commits for autogenerated files.

This PR also removes CSV/OLM related files, since they were already out of date and will be updated separately.

Part of https://github.com/eclipse/che/issues/17122

### Is it tested? How?
Tested on `crc`.


pkg/controller/workspacerouting/solvers/openshift_oauth_solver.go
```diff
@@ -15,7 +15,7 @@ package solvers
 import (
 	"fmt"
 
-	"github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
+	"github.com/che-incubator/che-workspace-operator/pkg/apis/controller/v1alpha1"
```
sleshchenko, line 18: Wow. Seems we need to rename module and replace all imports with new operator location )

pkg/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -17,7 +17,7 @@ import (
 	"fmt"
 
 	"github.com/che-incubator/che-workspace-operator/internal/cluster"
-	workspacev1alpha1 "github.com/che-incubator/che-workspace-operator/pkg/apis/workspace/v1alpha1"
+	workspacev1alpha1 "github.com/che-incubator/che-workspace-operator/pkg/apis/controller/v1alpha1"
```
sleshchenko, line 20: I think the alias now should be renamed

sleshchenko: renamed to controllerv1alpha1

pkg/apis/controller/group.go
```diff
@@ -0,0 +1,6 @@
+// Package controller contains controller API versions.
```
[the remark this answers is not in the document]
amisevsk: It should -- forgot to run `make fmt_license` but @sleshchenko caught it.

pkg/controller/workspacerouting/workspacerouting_controller.go
amisevsk: :+1: Thank you

PR #107, opened 2020-06-11
Fix issues with foreground deletion

amisevsk:
### What does this PR do?
If a deletion timestamp is set, stop the reconcile to avoid creating resources.

It's possible for the controller to race the API server, by e.g. recreating a Component after it is deleted but before the devworkspace is cleaned up. For foreground deletes, this can result in a component being re-created (without the deletion timestamp) and blocking removal of the devworkspace.

Fixes https://github.com/eclipse/che/issues/17146

### Is it tested? How?
Tested on `crc` via the process in https://github.com/eclipse/che/issues/17146. Components are correctly removed in a foreground delete. Devworkspace is removed, but takes slightly longer, since it waits for deployment and workspaceRouting to be cleaned up before it is removed.

### DIFF
--- a/pkg/controller/workspace/workspace_controller.go
+++ b/pkg/controller/workspace/workspace_controller.go
@@ -176,6 +176,11 @@ func (r *ReconcileWorkspace) Reconcile(request reconcile.Request) (reconcileResu
 		return reconcile.Result{}, err
 	}
 
+	if workspace.DeletionTimestamp != nil {
+		reqLogger.V(5).Info("Skipping reconcile of deleted resource")
+		return reconcile.Result{}, nil
+	}
+
 	// Ensure workspaceID is set.
 	if workspace.Status.WorkspaceId == "" {
 		workspaceId, err := getWorkspaceId(workspace)
--- a/pkg/webhook/workspace/handler/immutable.go
+++ b/pkg/webhook/workspace/handler/immutable.go
@@ -17,11 +17,10 @@ import (
 	"reflect"
 
 	"github.com/devfile/devworkspace-operator/pkg/config"
-	"github.com/google/go-cmp/cmp/cmpopts"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	devworkspace "github.com/devfile/kubernetes-api/pkg/apis/workspaces/v1alpha1"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
@@ -104,12 +103,12 @@ func changePermitted(oldObj, newObj runtime.Object) bool {
 	newCopy := newObj.DeepCopyObject()
 	oldMeta, ok := oldCopy.(metav1.Object)
 	if !ok {
-		log.Error(fmt.Errorf("Object is not a valid k8s object: does not have metadata"), "Failed to compare objects")
+		log.Error(fmt.Errorf("Object %s is not a valid k8s object: does not have metadata", oldObj.GetObjectKind()), "Failed to compare objects")
 		return false
 	}
 	newMeta, ok := newCopy.(metav1.Object)
 	if !ok {
-		log.Error(fmt.Errorf("Object is not a valid k8s object: does not have metadata"), "Failed to compare objects")
+		log.Error(fmt.Errorf("Object %s is not a valid k8s object: does not have metadata", newObj.GetObjectKind()), "Failed to compare objects")
 		return false
 	}
 	newMeta.SetFinalizers(oldMeta.GetFinalizers())
--- a/pkg/webhook/workspace/handler/immutable.go
+++ b/pkg/webhook/workspace/handler/immutable.go
@@ -17,12 +17,22 @@ import (
 	"reflect"
 
 	"github.com/devfile/devworkspace-operator/pkg/config"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	devworkspace "github.com/devfile/kubernetes-api/pkg/apis/workspaces/v1alpha1"
 	"github.com/google/go-cmp/cmp"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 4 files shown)


amisevsk:
Fixes https://github.com/eclipse/che/issues/17146

Turns out our webhooks are blocking the removal of the `foregroundDeletion` finalizer. Commit 08fffad757cfdf5eadd610cfc17dcaa560d3722c fixes this.

pkg/webhook/workspace/handler/workspace.go
```diff
@@ -26,7 +26,7 @@ import (
 // StopStartDiffOption is comparing options that should be used to check if there is no other changes except changing started
 var StopStartDiffOption = []cmp.Option{
```
sleshchenko, line 27: Does it makes sense to rename that DiffOptions? It does not seem to be startStop anymore

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -81,8 +84,26 @@ func (h *WebhookHandler) handleImmutableRoute(oldObj, newObj runtime.Object, use
 	if username == h.ControllerSAName {
 		return true, ""
 	}
-	if reflect.DeepEqual(oldObj, newObj) {
+	if changePermitted(oldObj, newObj) {
```
sleshchenko, line 87: Does not go have the same func convention as Java like isChangePermitted?

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -81,8 +84,26 @@ func (h *WebhookHandler) handleImmutableRoute(oldObj, newObj runtime.Object, use
...
+		return false
+	}
+	newMeta, ok := newCopy.(metav1.Object)
+	if !ok {
+		log.Error(fmt.Errorf("Object is not a valid k8s object: does not have metadata"), "Failed to compare objects")
+		return false
+	}
+	newMeta.SetFinalizers(oldMeta.GetFinalizers())
+	newMeta.SetDeletionTimestamp(oldMeta.GetDeletionTimestamp())
+	return reflect.DeepEqual(oldMeta, newMeta)
```
sleshchenko, line 108: Can't we compare it in the same way as we do for workspace? I assume there is trouble with types but just reading)

pkg/controller/workspace/workspace_controller.go
```diff
@@ -176,6 +176,11 @@ func (r *ReconcileWorkspace) Reconcile(request reconcile.Request) (reconcileResu
 		return reconcile.Result{}, err
 	}
 
+	if workspace.DeletionTimestamp != nil {
+		reqLogger.V(5).Info("Skipping reconcile of deleted resource")
```
sleshchenko, line 180: Why 5 but not 6 or 7?)
For we that int log levels are still not clear )

I tried to introduce some constants for typical levels - info, debug, trace but I did not manage to finalize it.
It probably should be handled in the scope of https://github.com/eclipse/che/issues/16659

pkg/webhook/workspace/handler/immutable.go
amisevsk: I don't think there's any such naming convention, but Go generally tries to avoid the verbosity of Java :)

pkg/webhook/workspace/handler/immutable.go
amisevsk: The main issue is that [`cmp` is meant for tests only](https://godoc.org/github.com/google/go-cmp/cmp):

> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal. 

so we probably shouldn't be using it at all :)

pkg/controller/workspace/workspace_controller.go
amisevsk: It's pretty arbitrary, I agree.

pkg/webhook/workspace/handler/workspace.go
amisevsk: Fixed

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -17,13 +17,23 @@ import (
 	"reflect"
 
 	"github.com/devfile/devworkspace-operator/pkg/config"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
sleshchenko, line 21: That changes pretty nothing but out `make fmt` does not seem to sort imports nicely... Like that apimachinery import is separated from other apimachinery imports. If I sort them manually I'd put them like:
```go
	devworkspace "github.com/devfile/kubernetes-api/pkg/apis/workspaces/v1alpha1"

	"github.com/devfile/devworkspace-operator/pkg/config"

	"github.com/google/go-cmp/cmp"
	"github.com/google/go-cmp/cmp/cmpopts"

	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
```
^ good thing here that make fmt does not touch such manually sorted list as well )
Just for us an item to take a look later

pkg/webhook/workspace/handler/immutable.go
```diff
@@ -81,8 +93,26 @@ func (h *WebhookHandler) handleImmutableRoute(oldObj, newObj runtime.Object, use
...
 	}
 	return false, "object is owned by workspace and cannot be updated."
 }
+
+func changePermitted(oldObj, newObj runtime.Object) bool {
+	oldCopy := oldObj.DeepCopyObject()
+	newCopy := newObj.DeepCopyObject()
+	oldMeta, ok := oldCopy.(metav1.Object)
+	if !ok {
+		log.Error(fmt.Errorf("Object is not a valid k8s object: does not have metadata"), "Failed to compare objects")
```
sleshchenko, line 107: Does it make sense to add at least object kind to logs?

pkg/webhook/workspace/handler/immutable.go
amisevsk: Yeah the issue is that goimports will preserve whitespace dividing sections. The only actual formatting it does is to put libraries at the top, separated from other imports, and sorts imports within each section as usual.

amisevsk: Yeah the issue is that goimports will preserve whitespace dividing sections. The only actual formatting it does is to put libraries at the top, separated from other imports, and sorts imports within each section as usual.

pkg/webhook/workspace/handler/immutable.go
amisevsk: Done

pkg/webhook/workspace/handler/immutable.go
amisevsk: Fixed

PR #111, opened 2020-06-17
Make controller deployable to any namespace via Makefile

amisevsk:
### What does this PR do?
Make it possible to manually deploy the controller to any namespace by setting the `$NAMESPACE` env var.

The only ugly workaround is that the clusterrolebinding's subject requires a namespace; this is updated via `sed` and means that manually applying files will not work.

Fixes https://github.com/eclipse/che/issues/17098

### Is it tested? How?
```
git checkout master
make uninstall
git checkout amisevsk/customize-namespace
export NAMESPACE=web-terminal-operator
make deploy
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-NAMESPACE = che-workspace-controller
 OPERATOR_SDK_VERSION = v0.17.0
+NAMESPACE ?= che-workspace-controller
 IMG ?= quay.io/che-incubator/che-workspace-controller:nightly
 TOOL ?= oc
 ROUTING_SUFFIX ?= 192.168.99.100.nip.io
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -114,23 +115,25 @@ _update_crds: update_devworkspace_crds
 	$(TOOL) apply -f ./devworkspace-crds/deploy/crds
 
 _update_controller_configmap:
-	$(TOOL) apply -f ./deploy/controller_config.yaml
+	$(TOOL) apply -f ./deploy/controller_config.yaml -n $(NAMESPACE)
 
 _apply_controller_cfg:
-	$(TOOL) apply -f ./deploy
+	sed -i.bak -e "s|namespace: che-workspace-controller|namespace: $(NAMESPACE)|" ./deploy/role_binding.yaml
```
sleshchenko, line 121: I think it makes sense to change hard-coded namespace to placeholder, like:
```
# The ${NAMESPACE} must be replaced with the actual namespace where controller lives
namespace: ${NAMESPACE}
```
It would help users detect the issue with deployment earlier if namespace is not `che-workspace-controller` since they  will get error message after `oc apply` while without it - deployment is deployed and run but workspace won't be started because of permissions leak (<- I haven't checked that assumption). WDYT?

amisevsk: Good idea -- done.

PR #112, opened 2020-06-17
Rename config properties

amisevsk:
### What does this PR do?
Reworks config property names to be more consistent:
- config properties that reference cluster properties are prefixed by `cluster`
- config properties that impact the controller are prefixed with `controller`
- config properties that impact workspaces are prefiex with `devworkspace`

Fixes https://github.com/eclipse/che/issues/17092

### Is it tested? How?
Deployed on `crc` and found no issues.

### DIFF
--- a/pkg/config/property.go
+++ b/pkg/config/property.go
@@ -13,34 +13,40 @@
 package config
 
 const (
-	pluginRegistryURL = "plugin.registry.url"
+	// URL of external plugin registry; will be used when devworkspace uses plugin
+	// not included in internal registry or when devfile does not include explicit
+	// registry URL.
+	pluginRegistryURL = "controller.plugin_registry.url"
 
-	routingSuffix        = "cluster.routing_suffix"
-	defaultRoutingSuffix = ""
-
-	webhooksEnabled        = "che.webhooks.enabled"
+	webhooksEnabled        = "controller.webhooks.enabled"
 	defaultWebhooksEnabled = "true"
 
-	cheAPISidecarImage = "che.workspace.che_api_sidecar.image"
+	cheAPISidecarImage = "devworkspace.api_sidecar.image"
 	// by default that functionality is not available since it's not fully supported
 	defaultCheAPISidecarImage = ""
 
-	sidecarPullPolicy        = "che.workspace.sidecar.image_pull_policy"
+	// image pull policy that is applied to every container within workspace
+	sidecarPullPolicy        = "devworkspace.sidecar.image_pull_policy"
 	defaultSidecarPullPolicy = "Always"
 
 	// workspacePVCName config property handles the PVC name that should be created and used for all workspaces within one kubernetes namespace
-	workspacePVCName        = "che.workspace.pvc.name"
+	workspacePVCName        = "devworkspace.pvc.name"
 	defaultWorkspacePVCName = "claim-che-workspace"
 
-	workspacePVCStorageClassName = "che.workspace_pvc.storage_class.name"
+	workspacePVCStorageClassName = "devworkspace.pvc.storage_class.name"
 
-	pluginArtifactsBrokerImage        = "che.workspace.plugin_broker.artifacts.image"
+	pluginArtifactsBrokerImage        = "controller.plugin_artifacts_broker.image"
 	defaultPluginArtifactsBrokerImage = "quay.io/eclipse/che-plugin-artifacts-broker:v3.1.0"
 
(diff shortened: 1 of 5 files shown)


pkg/config/property.go
```diff
@@ -13,19 +13,19 @@
 package config
 
 const (
-	pluginRegistryURL = "plugin.registry.url"
+	pluginRegistryURL = "controller.plugin_registry.url"
 
 	routingSuffix        = "cluster.routing_suffix"
```
sleshchenko, line 18: that is used only for workspaceRoutings, right?
What do you think about
```
devworkspace.routing.default_class
devworkspace.routing.cluster_host_suffix
```
?

sleshchenko: I prefer a bit more `devworkspace.routing.cluster_host_suffix` than `cluster.routing_suffix` since on K8s infrastructure, we can assume that admin may use different hosts for different services(Che itself is deployed on one, workspaces on other), and such naming a bit clearly says that it's used for workspaces ingresses hosts suffix.
I'm OK with both options just choose whatever you find more logical.

P.S. On Che Server side it's called `che.infra.kubernetes.ingress.domain`.

amisevsk: > devworkspace.routing.default_class

routingClass is in the [devfile spec](https://github.com/devfile/kubernetes-api/blob/28e3e6acdc368b503672738697b149dd23c86770/pkg/apis/workspaces/v1alpha1/devworkspace_types.go#L12) at this point, so it doesn't make sense to me to group it with controller-specific subresources.

Applied suggestion re: cluster_host_suffix, though -- you make a good point (and it was called `ingress.domain` here as well but we changed it since ingresses are more k8s-specific).

pkg/config/property.go
```diff
@@ -13,19 +13,19 @@
 package config
 
 const (
-	pluginRegistryURL = "plugin.registry.url"
+	pluginRegistryURL = "controller.plugin_registry.url"
```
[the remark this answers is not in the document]
amisevsk: Added doc.

pkg/config/property.go
```diff
@@ -13,19 +13,19 @@
...
+	webhooksEnabled        = "controller.webhooks.enabled"
 	defaultWebhooksEnabled = "true"
 
-	cheAPISidecarImage = "che.workspace.che_api_sidecar.image"
+	cheAPISidecarImage = "devworkspace.api_sidecar.image"
 	// by default that functionality is not available since it's not fully supported
 	defaultCheAPISidecarImage = ""
 
-	sidecarPullPolicy        = "che.workspace.sidecar.image_pull_policy"
+	sidecarPullPolicy        = "devworkspace.sidecar.image_pull_policy"
```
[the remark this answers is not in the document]
amisevsk: Applied -- thanks!

pkg/config/property.go
```diff
@@ -34,13 +35,18 @@ const (
...
-	routingClass        = "che.workspace.default_routing_class"
+	routingClass        = "devworkspace.default_routing_class"
 	defaultRoutingClass = "basic"
 
+	// routingSuffix is the default domain for routes/ingresses created on the cluster. All
+	// routes/ingresses will be created with URL http(s)://<unique-to-workspace-part>.<routingSuffix>
+	routingSuffix        = "devworkspace.routing.cluster_host_suffix"
+	defaultRoutingSuffix = ""
+
 	workspaceIdleTimeout        = "che.workspace.idle_timeout"
```
sleshchenko, line 50:
```suggestion
	workspaceIdleTimeout        = "devworkspace.idle_timeout"
```
?

sleshchenko: There are also "che.workspace.pvc.name" and "che.workspace_pvc.storage_class.name" which are not critical in WebTerminal case but still

amisevsk: Fixed -- renamed to `devworkspace.pvc.*` and `devworkspace.idle_timeout`

There is also `	defaultTerminalDockerimageProperty = "che.workspace.default_dockerimage.redhat-developer.web-terminal"` in cmd_terminal.go. I can't come up with a good name for this one :D

sleshchenko: I have a good proposal neither. 
We can try to proceed with just `devworkspace.default_dockerimage.redhat-developer.web-terminal` and later we'll be able to migrate that if needed.

amisevsk: Done, and squashed the PR commits into one.

PR #115, opened 2020-06-23
Don't delete cert generation deployment after its finished

JPinkney:
### What does this PR do?
This PR makes it so the cert generation deployment isn't deleted anymore and reasonable memory limits are set for the cert gen pod

### Is it tested? How?

Tested locally with webhooks enabled


pkg/webhook/workspace/config.go
```diff
@@ -50,12 +45,6 @@ func Configure(ctx context.Context) error {
 		return err
 	}
 
-	// At this point we no longer need the cert generation deployment
-	err = deleteCertGenerationDeployment(ctx, c, namespace)
```
sleshchenko, line 54: If we don't remove cert generation deployment - we need to update the message here https://github.com/devfile/devworkspace-operator/blob/master/cert-generation/main.go#L124

cert-generation/main.go
```diff
@@ -120,8 +120,7 @@ func main() {
 		}
 	}
 
-	// Wait until it's cleaned up by the workspace-controller
-	log.Println("Certs have been successfully created. Waiting to be cleaned up by devworkspace controller")
+	log.Println("Certs have been successfully created.")
```
sleshchenko, line 123: Just idea for future: if possible - that deployment could listen to created service removing and recreate it after it's removed. Otherwise - it needs to be restarted to restore broker installation. But it's not critical for that PR and maybe even does not worth to be implemented.

PR #116, opened 2020-06-23
Introduce blocker webhook to delay workspace creation until real webhooks are created

JPinkney:
### What does this PR do?
This PR introduces a blocker webhook that delays workspace creation until real webhooks are created. Basically what will happen is if you try and do `oc apply -f samples/cloud-shell.yaml` you'll get a response of:

```Error from server (InternalError): error when creating "samples/cloud-shell.yaml": Internal error occurred: failed calling webhook "mutate.che-workspace-controller.svc": Post https://workspace-controller.che-workspace-controller.svc:443/mutate?timeout=30s: dial tcp 10.128.0.55:8443: connect: connection refused```.

until the actual webhook server is connected. Once the webhook server is started then you'll be able to create a dev workspace successfully.

https://github.com/eclipse/che/issues/17200

### Is it tested? How?

Webhooks enabled, then create the cert gen image `make docker_cert` then the regular controller image `make docker` then `make deploy`. Right after you run `make deploy` start doing `oc apply -f samples/cloud-shell.yaml`

### DIFF
--- a/cert-generation/config/constants.go
+++ b/cert-generation/config/constants.go
@@ -1,0 +1,21 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package config
+
+// Constants for cert generation
+const (
+	SecureServiceName = "devworkspace-controller"
+	CertConfigMapName = "devworkspace-controller-secure-service"
+	CertSecretName    = "devworkspace-controller"
+	WebhookServerName = "webhook-server"
+)
--- a/cert-generation/main.go
+++ b/cert-generation/main.go
@@ -16,6 +16,8 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/devfile/devworkspace-operator/cert-generation/config"
 	"github.com/devfile/devworkspace-operator/cert-generation/webhooks"
@@ -119,7 +121,8 @@ func main() {
 		log.Fatal("Failed when attempting to setup the webhook: ", err)
 	}
 
-	log.Println("Certs have been successfully created.")
-	for {
-	}
+	log.Println("Certificates and webhooks are up to date. Idling until interrupted")
+	var shutdownChan = make(chan os.Signal, 1)
+	signal.Notify(shutdownChan, syscall.SIGTERM)
+	<-shutdownChan
 }
--- a/cert-generation/main.go
+++ b/cert-generation/main.go
@@ -17,31 +17,25 @@ import (
 	"log"
 	"os"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/intstr"
-
+	"github.com/devfile/devworkspace-operator/cert-generation/config"
+	"github.com/devfile/devworkspace-operator/cert-generation/webhooks"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-)
-
-const (
-	secureServiceName = "workspace-controller"
-	certConfigMapName = "che-workspace-controller-secure-service"
-	certSecretName    = "workspace-controller"
-	webhookServerName = "webhook-server"
 )
 
 func main() {
 	log.SetOutput(os.Stdout)
 
-	config, err := rest.InClusterConfig()
+	clusterConfig, err := rest.InClusterConfig()
 	if err != nil {
 		log.Fatal("Failed when attempting to retrieve in cluster config: ", err)
 	}
 
-	clientset, err := kubernetes.NewForConfig(config)
+	clientset, err := kubernetes.NewForConfig(clusterConfig)
 	if err != nil {
 		log.Fatal("Failed when attempting to retrieve in cluster config: ", err)
 	}
@@ -55,7 +49,7 @@ func main() {
 	configMapData := make(map[string]string, 0)
 	configMap := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      certConfigMapName,
+			Name:      config.CertConfigMapName,
 			Namespace: namespace,
 			Annotations: map[string]string{
 				"service.beta.openshift.io/inject-cabundle": "true",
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: cert-generation/config/constants.go [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package config
+
+import "k8s.io/api/admissionregistration/v1beta1"
+
+// Constants for webhook
+const (
+	MutateWebhookName          = "mutate.che-workspace-controller.svc"
```
amisevsk, line 19: I wonder if now is the time to change this to something more Che-independent.

cert-generation/webhooks/webhook.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"github.com/devfile/devworkspace-operator/cert-generation/config"
+	v1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// WebhookInit Initialize the webhook that denies all everything until controller is started succesfully
```
amisevsk, line 23:
```suggestion
// WebhookInit Initialize the webhook that denies everything until controller is started succesfully
```

cert-generation/webhooks/webhook.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// Create dummy mutating webhook or update it if it already exists
+	if _, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(config.MutateWebhookName, metav1.GetOptions{}); apierrors.IsNotFound(err) {
+		_, err = clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Create(configuration)
+		if err != nil {
+			return err
+		}
+	} else {
+		_, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Update(configuration)
```
amisevsk, line 66: I don't know if it's required to update the webhook if it already exists. By default the full webhook's failure policy is Fail, so once the regular webhook is created, a controller update will cause the webhooks to fail until the controller is running.

sleshchenko: Initially, my idea was to create stub mutating webhook configuration (if it's missing) that will be updated by controller. It would simplify workflow a bit I think.
Currently, we can imagine that cert-generation deployment is restarted for some reason (e.g. due to node restart) and that blocker webhook will be recreated and the controller will clean it but only when its next restart.

cert-generation/webhooks/webhook.go
```diff
@@ -0,0 +1,72 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: config.MutateWebhookCfgName,
+		},
+		Webhooks: []v1beta1.MutatingWebhook{
+			workspaceObjMutateWebhook,
+		},
+	}
+
+	// Create dummy mutating webhook or update it if it already exists
+	if _, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(config.MutateWebhookName, metav1.GetOptions{}); apierrors.IsNotFound(err) {
```
sleshchenko, line 60:
```suggestion
	if _, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(config.MutateWebhookCfgName, metav1.GetOptions{}); apierrors.IsNotFound(err) {
```
?

pkg/webhook/workspace/config.go
```diff
@@ -36,6 +36,7 @@ import (
 )
 
 const certGenDeploymentName = "che-workspace-controller-cert-gen"
+const dummyWebhookConfigurationName = "controller.devfile.io-blocker"
```
sleshchenko, line 39: It's better to avoid duplication. Can we just use the value from cert-generation/config?

cert-generation/config/constants.go
```diff
@@ -0,0 +1,31 @@
...
+	MutateWebhookCfgName       = "controller.devfile.io-blocker"
```
sleshchenko, line 20: When I see the code I confuse that with constant from controller. It's better to name Blocker and usual WebhookCfg differently

cert-generation/config/constants.go
```diff
@@ -0,0 +1,21 @@
...
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package config
+
+// Constants for cert generation
+const (
+	SecureServiceName = "workspace-controller"
+	CertConfigMapName = "che-workspace-controller-secure-service"
+	CertSecretName    = "workspace-controller"
```
sleshchenko, line 19:
```suggestion
	CertSecretName    = "devworkspace-controller"
```
?

cert-generation/config/constants.go
```diff
@@ -0,0 +1,21 @@
...
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package config
+
+// Constants for cert generation
+const (
+	SecureServiceName = "workspace-controller"
+	CertConfigMapName = "che-workspace-controller-secure-service"
```
sleshchenko, line 18:
```suggestion
	CertConfigMapName = "devworkspace-controller-secure-service"
```

cert-generation/webhooks/webhook.go
```diff
@@ -0,0 +1,34 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// WebhookInit Initialize the webhook that denies everything until controller is started succesfully
+func WebhookInit(clientset *kubernetes.Clientset, namespace string) error {
+	configuration := workspace.BuildMutateWebhookCfg(namespace)
+
+	// Create mutating webhook
+	if _, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(configuration.Name, metav1.GetOptions{}); apierrors.IsNotFound(err) {
```
sleshchenko, line 27: It could be done with one operation - `create` and just check error if it's not already exist

cert-generation/webhooks/webhook.go
```diff
@@ -0,0 +1,31 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+// WebhookInit Initialize the webhook that denies everything until controller is started succesfully
+func WebhookInit(clientset *kubernetes.Clientset, namespace string) error {
+	configuration := workspace.BuildMutateWebhookCfg(namespace)
+
+	// Create mutating webhook
+	_, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Create(configuration)
+	if !apierrors.IsNotFound(err) {
```
sleshchenko, line 27:
```suggestion
	if !apierrors.IsAlreadyExists(err) {
```

cert-generation/main.go
```diff
@@ -120,6 +114,11 @@ func main() {
 		}
 	}
 
+	err = webhooks.WebhookInit(clientset, namespace)
+	if err != nil {
+		log.Fatal("Failed when attempting to setup the webhook: ", err)
+	}
+
 	log.Println("Certs have been successfully created.")
 	for {
```
sleshchenko, line 123: I suppose it's better to check sigterm here

cert-generation/webhooks/webhook.go
sleshchenko: From debug point of view it makes sense to log message is webhook configuration is created

sleshchenko:
Screenshot_20200630_110650

locally tests are finished successfully after specifying image built from that PR.
I'm merging it and e2e tests are going to be improved to avoid such failure in the scope of https://github.com/eclipse/che/issues/17264

PR #117, opened 2020-06-23
Variable images

amisevsk:
### What does this PR do?
Extracts some hard-coded images to be determined by environment variables, to support restricted networks and OLM.

Fixes https://github.com/eclipse/che/issues/17224

### Is it tested? How?
Tested on `crc`

### Open questions
Do we need to do similar for e.g. the plugin artifacts broker (which is currently defined in the configmap?)

### DIFF
--- a/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
+++ b/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
@@ -24,7 +24,7 @@ spec:
       cookiesAuthEnabled: true
   containers:
   - name: che-machine-exec
-    image: "quay.io/eclipse/che-machine-exec:nightly"
+    image: "${IMAGE_plugin_eclipse_cloud_shell_nightly}"
     ports:
     - exposedPort: 4444
     command: ["/go/bin/che-machine-exec",
--- a/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
+++ b/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
@@ -24,7 +24,7 @@ spec:
       cookiesAuthEnabled: true
   containers:
   - name: che-machine-exec
-    image: "${IMAGE_plugin_eclipse_cloud_shell_nightly}"
+    image: "${RELATED_IMAGE_plugin_eclipse_cloud_shell_nightly}"
     ports:
     - exposedPort: 4444
     command: ["/go/bin/che-machine-exec",
--- a/internal-registry/redhat-developer/web-terminal/nightly/meta.yaml
+++ b/internal-registry/redhat-developer/web-terminal/nightly/meta.yaml
@@ -23,9 +23,7 @@ spec:
       cookiesAuthEnabled: true
   containers:
   - name: web-terminal
-    image: "quay.io/eclipse/che-machine-exec:nightly"
-    #    ^ it'll be replaced with new image below after https://github.com/eclipse/che/issues/16942 is fixed
-    #    image: "quay.io/redhat-developer/web-terminal:nightly"
+    image: "${IMAGE_plugin_redhat_developer_web_terminal_nightly}"
     command: ["/go/bin/che-machine-exec",
               "--authenticated-user-id", "$(CHE_WORKSPACE_CREATOR)",
               "--idle-timeout", "$(CHE_WORKSPACE_IDLE_TIMEOUT)",
--- a/internal-registry/redhat-developer/web-terminal/nightly/meta.yaml
+++ b/internal-registry/redhat-developer/web-terminal/nightly/meta.yaml
@@ -23,7 +23,7 @@ spec:
       cookiesAuthEnabled: true
   containers:
   - name: web-terminal
-    image: "${IMAGE_plugin_redhat_developer_web_terminal_nightly}"
+    image: "${RELATED_IMAGE_plugin_redhat_developer_web_terminal_nightly}"
     command: ["/go/bin/che-machine-exec",
               "--authenticated-user-id", "$(CHE_WORKSPACE_CREATOR)",
               "--idle-timeout", "$(CHE_WORKSPACE_IDLE_TIMEOUT)",
--- a/internal/images/image.go
+++ b/internal/images/image.go
@@ -14,15 +14,39 @@ package images
 
 import (
 	"fmt"
-	"github.com/eclipse/che-plugin-broker/model"
 	"os"
 	"regexp"
+
+	"github.com/eclipse/che-plugin-broker/model"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 var log = logf.Log.WithName("container-images")
 
 var envRegexp = regexp.MustCompile(`\${(.*)}`)
+
+const (
+	webTerminalToolingImageEnvVar  = "IMAGE_web_terminal_tooling"
+	openshiftOAuthProxyImageEnvVar = "IMAGE_openshift_oauth_proxy"
+)
+
+func GetWebTerminalToolingImage() string {
+	val, ok := os.LookupEnv(webTerminalToolingImageEnvVar)
+	if !ok {
+		log.Error(fmt.Errorf("environment variable %s is not set", webTerminalToolingImageEnvVar), "Could not get web terminal tooling image")
+		return ""
+	}
+	return val
+}
+
+func GetOpenShiftOAuthProxyImage() string {
+	val, ok := os.LookupEnv(openshiftOAuthProxyImageEnvVar)
+	if !ok {
+		log.Error(fmt.Errorf("environment variable %s is not set", openshiftOAuthProxyImageEnvVar), "Could not get OpenShift OAuth proxy image")
+		return ""
+	}
+	return val
+}
 
 func FillPluginMetaEnvVars(pluginMeta model.PluginMeta) (model.PluginMeta, error) {
 	for idx, container := range pluginMeta.Spec.Containers {
--- a/internal/images/image.go
+++ b/internal/images/image.go
@@ -23,7 +23,7 @@ import (
 
 var log = logf.Log.WithName("container-images")
 
-var envRegexp = regexp.MustCompile(`\${(.*)}`)
+var envRegexp = regexp.MustCompile(`\${(IMAGE_.*)}`)
 
 const (
 	webTerminalToolingImageEnvVar  = "IMAGE_web_terminal_tooling"
--- a/internal/images/image.go
+++ b/internal/images/image.go
@@ -10,6 +10,14 @@
 //   Red Hat, Inc. - initial API and implementation
 //
 
+// Package images is intended to support deploying the operator on restricted networks. It contains
+// utilities for translating images referenced by environment variables to regular image references,
+// allowing images that are defined by a tag to be replaced by digests automatically. This allows all
+// images used by the controller to be defined as environment variables on the controller deployment.
+//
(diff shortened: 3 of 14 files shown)


internal/images/image.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"regexp"
+
+	"github.com/eclipse/che-plugin-broker/model"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var log = logf.Log.WithName("container-images")
+
+var envRegexp = regexp.MustCompile(`\${(.*)}`)
```
sleshchenko, line 26: I think it makes sense to introduce convention that the only `IMAGE_*` env var could be filled in in plugin configuration, otherwise, every workspace creator could get any env var from controller with custom plugin configuration. WDYT?

internal/images/image.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	webTerminalToolingImageEnvVar  = "IMAGE_web_terminal_tooling"
```
sleshchenko, line 29: Why we have mixed upper and lower case here? Would not be it a bit consistent to use upper case only, like `IMAGE_WEB_TERMINAL_TOOLING`, or there is some requirement from OLM side?

internal-registry/eclipse/cloud-shell/nightly/meta.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       cookiesAuthEnabled: true
   containers:
   - name: che-machine-exec
-    image: "quay.io/eclipse/che-machine-exec:nightly"
+    image: "${IMAGE_plugin_eclipse_cloud_shell_nightly}"
```
sleshchenko, line 27: It makes sense to extract cloud-shell to env vars as well but that it's mainly useful for local testing.
I suppose for WebTerminal operator we should not deliver it as well as web-terminal-dev. WDYT?

pkg/config/cmd_terminal.go
```diff
@@ -32,12 +34,12 @@ var (
 		MemoryLimit: "256Mi",
 		Container: devworkspace.Container{
 			Name:  "dev",
-			Image: "registry.redhat.io/codeready-workspaces/plugin-openshift-rhel8:2.1",
+			Image: images.GetWebTerminalToolingImage(),
```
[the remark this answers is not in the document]
amisevsk: WDYM?

internal/images/image.go
amisevsk: How secret are env vars supposed to be here? Either way I'll add it to the regex.

internal/images/image.go
amisevsk: It's the style chosen by the CRW operator, and I feel like it reads a little better. Reading all-caps is tiring.

internal-registry/eclipse/cloud-shell/nightly/meta.yaml
amisevsk: We can strip it out but I don't know what we gain except complexity -- we'll need this *eventually* anyways. If we intend this operator to ultimately be used for full workspaces as well as web terminals, we'll end up implementing this within a few months anyways.

internal/images/image.go
sleshchenko: > How secret are env vars supposed to be here?

I know this is not really suitable example but still https://github.com/eclipse/che/blob/master/deploy/openshift/templates/multi/postgres-template.yaml#L44
Who knows all env vars K8s infrastructure injects? And how we make sure that like SA token is not injected with env var some day?

pkg/config/cmd_terminal.go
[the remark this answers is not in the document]
sleshchenko: I mean that images are also part of the configuration, right? And if it's evaluated from config map or environment variable - it's implementation details... So, I suppose it makes sense to move func like `GetWebTerminalToolingImage` to https://github.com/devfile/devworkspace-operator/tree/master/pkg/config

I remember you mentioned that it makes be handy to be able to use config map for containers... So, that method may check both places - config map entry and env var and return whichever has more priority.

internal/images/image.go
sleshchenko: > Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names. ***Uppercase and lowercase letters shall retain their unique identities and shall not be folded together***. The name space of environment variable names containing lowercase letters is reserved for applications. Applications can define any environment variables with names from this name space without modifying the behavior of the standard utilities.

from https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html

Google's shell convention declares upper case... https://google.github.io/styleguide/shellguide.html#s7.3-constants-and-environment-variable-names

I would follow a common convention that introducing our own but related images seem to be untypical )
There is a first place where I found `IMAGE_my_image` format https://docs.google.com/document/d/1n4MJgVk8g254s_X6oOgzC1z44RcJur3g2WaMuriv1fg/edit#
Design proposal uses lower case for related images names but it's not env vars: https://github.com/operator-framework/operator-lifecycle-manager/blob/release-4.3/doc/contributors/design-proposals/related-images.md
There is knative operator that uses: mixed upper and lower case, `-`, `_`, `__` for related image name and the same for the corresponding env vars
https://github.com/openshift-knative/serverless-operator/blob/master/olm-catalog/serverless-operator/1.8.0/serverless-operator.v1.8.0.clusterserviceversion.yaml

Seems CRW is going to rename that env vars from `IMAGE__*` to `RELATED_IMAGE_*` https://github.com/eclipse/che-operator/pull/274/files#diff-e7334c795c728921b138e6dde55e412dR48

The quick research I've done makes me even more confused. So, I'm OK with any option :-D

pkg/config/cmd_terminal.go
sleshchenko, line 37: Now it can be "" which will lead deployment to fail. Maybe it's better to fail workspace start in such case.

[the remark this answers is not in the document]
sleshchenko: If be more precise - it's what I meant https://github.com/amisevsk/che-workspace-crd-operator/compare/variable-images...sleshchenko:variable-images-proposal?expand=1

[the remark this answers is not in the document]
amisevsk: My concern is that this functionality is specific to restricted networks (and thus not a "standard" configuration property). Users should never have to set these env vars themselves, it's purely for integration with OpenShift and the build system, hence placing it in `/internal`. 

Your changeset spreads this functionality across packages -- the internal registry is responsible for filling its images, `pkg/config` is responsible for the web terminal tooling image, etc. I personally prefer to keep all code related to supporting restricted networks in one place, with utility functions to do what we need to do. The `config` parts of the code should specifically be for configuration properties a user might be expected to set.

internal/images/image.go
amisevsk: Fair enough :)

internal/images/image.go
amisevsk: I would add that most conventions (e.g.  IEEE Std 1003.1-2001) are specific to use in shells and shell utilities. The overall spec for env vars leaves it open to any character in the printable character set, and there's technically nothing in the spec preventing mixing case. I think it's a lot easier to parse the `RELATED_IMAGES_` prefix when case is mixed. 

```
RELATED_IMAGE_plugin_redhat_developer_web_terminal_4_5_0
RELATED_IMAGE_plugin_redhat_developer_web_terminal_nightly
RELATED_IMAGE_plugin_redhat_developer_web_terminal_dev_4_5_0
RELATED_IMAGE_plugin_redhat_developer_web_terminal_dev_nightly
RELATED_IMAGE_plugin_eclipse_cloud_shell_nightly
RELATED_IMAGE_web_terminal_tooling
RELATED_IMAGE_openshift_oauth_proxy
```
vs
```
RELATED_IMAGE_PLUGIN_REDHAT_DEVELOPER_WEB_TERMINAL_4_5_0
RELATED_IMAGE_PLUGIN_REDHAT_DEVELOPER_WEB_TERMINAL_NIGHTLY
RELATED_IMAGE_PLUGIN_REDHAT_DEVELOPER_WEB_TERMINAL_DEV_4_5_0
RELATED_IMAGE_PLUGIN_REDHAT_DEVELOPER_WEB_TERMINAL_DEV_NIGHTLY
RELATED_IMAGE_PLUGIN_ECLIPSE_CLOUD_SHELL_NIGHTLY
RELATED_IMAGE_WEB_TERMINAL_TOOLING
RELATED_IMAGE_OPENSHIFT_OAUTH_PROXY
```

Using this format also separates them from actual environment variables users might have to think about.

pkg/config/cmd_terminal.go
amisevsk: Either way the deployment is broken; if that env var resolves to "" the only thing that will realistically fix it is a respin + re-release. These aren't configuration properties so them not being set should never reach production.

I've set it to fail workspace startup if this env var is unset. Going forward though -- if *any* of the related_images envrionment variables is unset it will break the controller. Do we need to validate each of them?

internal/images/image.go
sleshchenko: > I think it's a lot easier to parse the RELATED_IMAGES_ prefix when case is mixed.

Agree

internal-registry/redhat-developer/web-terminal/nightly/meta.yaml
```diff
@@ -23,9 +23,7 @@ spec:
       cookiesAuthEnabled: true
   containers:
   - name: web-terminal
-    image: "quay.io/eclipse/che-machine-exec:nightly"
-    #    ^ it'll be replaced with new image below after https://github.com/eclipse/che/issues/16942 is fixed
-    #    image: "quay.io/redhat-developer/web-terminal:nightly"
+    image: "${RELATED_IMAGE_plugin_redhat_developer_web_terminal_nightly}"
```
sleshchenko, line 26: Is there a strong reason to set RELATED IMAGE for nightly version?
If we use related images for airgap mode - then we replace that value with digest, and nightly is not nightly anymore.

internal/images/image.go
```diff
@@ -0,0 +1,86 @@
...
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package images
```
sleshchenko, line 13: Consider adding a few lines description to that package

internal-registry/redhat-developer/web-terminal/nightly/meta.yaml
amisevsk: In general, once we start using digests stuff like "nightly" and "latest" are near-impossible to support. I don't know the preferred practice for this -- @nickboldt any ideas here?

I don't see a good way of avoiding ErrImagePull if we leave these images as tags. Also, once this image is set to a reg.redhat.io build, nightly will cease to exist anyways (unless we do releases every day). I'm not sure the best way to support this on the community front (except that leaving the env-var as is doesn't change functionality).

internal/images/image.go
amisevsk: Done -- documented the exported functions as well.

PR #118, opened 2020-06-24
Don't set ownerrefs for webhooks

JPinkney:
### What does this PR do?
This PR makes it so that ownerrefs aren't set for the mutating and validating webhooks -- meaning they won't be cleaned up when the operator is uninstalled.

Fixes https://github.com/eclipse/che/issues/16980

### Is it tested? How?

In order to build it all together, I cherry-picked the two commits ontop of the `olm-os` branch and then build the olm stuff from there to test. You can also test using:
```
cat <<EOF | oc apply -f -
kind: CatalogSource
metadata:
  name: che-workspace-crd-registry
spec:
  sourceType: grpc
  image: quay.io/jpinkney/che-workspace-operator-index:latest
  publisher: Red Hat
  displayName: Che Workspace Operator Catalog
EOF
```

Just install the web terminal operator, check that validating and mutating webhooks are available then uninstall the web terminal operator and see that they are still on the cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,8 @@ endif
 	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io components.controller.devfile.io
 	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io devworkspaces.workspace.devfile.io
 	$(TOOL) delete customresourcedefinitions.apiextensions.k8s.io devworkspacetemplates.workspace.devfile.io
+	$(TOOL) delete mutatingwebhookconfigurations controller.devfile.io
+	$(TOOL) delete validatingwebhookconfigurations controller.devfile.io
 
 ### docker: build and push docker image
 docker: _print_vars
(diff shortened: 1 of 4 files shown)


JPinkney:
> Are webhooks still automatically removed when the operator is uninstalled?

After this change webhooks will stay around until they're cleaned.

> dedicated-admin installs the operator

is there a list of permissions that dedicated-admin gets so we can check if they can remove the webhooks? I'm not familiar with dedicated-admin

sleshchenko:
@amisevsk good example with dedicated admin. I'll check dedicated admins rights.
But maybe it's better to lock the cluster than leave users' and dedicated admins' sensetive info unsecure. Dedicated admin is still able to unlock the cluster by installing Terminal Operator again, right?
The best approach for sure is removing webhooks but we need OLM to clean up all workspaces CRs.

BTW +1 not be in rush in merging and investigate it more.

sleshchenko:
dedicated-admin

Seems dedicated admins are not able to read nor remove webhooks configurations.

Dropping approval until we agree on the uninstallation procedure for WebTerminal Operator.

PR #119, opened 2020-07-01
[WIP] Simplify webhook certificates setup for controller

amisevsk:
### What does this PR do?
- Remove cert-generation deployment
- Add functionality for discovering tls cert and key at startup
- Use service service CA to mount secret in path matching OLM
- Split deploy folder again, to include devel subdir

Marked as WIP as I still need to do some testing on the OLM side.

It removes the need for a second deployment dedicated to just generating certs.

### Is it tested? How?
Tested locally without issue (though that's to be expected. Need to integrate with changes in https://github.com/redhat-developer/web-terminal-operator/pull/2 to test if it'll work with OLM.


amisevsk:
Closing in favor of https://github.com/devfile/devworkspace-operator/pull/121

PR #120, opened 2020-07-02
Sync labels with CSV to simplify generation

amisevsk:
### What does this PR do?
Sync labels applied to CSV to deployment.yamls to ease generation

https://github.com/redhat-developer/web-terminal-operator/pull/2/files#r448209113

### DIFF
--- a/deploy/os/controller.yaml
+++ b/deploy/os/controller.yaml
@@ -2,17 +2,20 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: che-workspace-controller
+  labels:
+    app.kubernetes.io/name: devworkspace-controller
+    app.kubernetes.io/part-of: devworkspace-operator
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app: che-workspace-controller
+      app.kubernetes.io/name: devworkspace-controller
+      app.kubernetes.io/part-of: devworkspace-operator
   template:
     metadata:
       labels:
-        app: che-workspace-controller
         app.kubernetes.io/name: devworkspace-controller
-        app.kubernetes.io/part-of: devworkspace-controller
+        app.kubernetes.io/part-of: devworkspace-operator
       annotations:
         kubectl.kubernetes.io/restartedAt: ""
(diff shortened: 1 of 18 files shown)


deploy/os/controller.yaml
```diff
@@ -11,6 +11,8 @@ spec:
     metadata:
       labels:
         app: che-workspace-controller
+        app.kubernetes.io/name: devworkspace-controller
```
sleshchenko, line 14: 1. It's pod template labels, does it makes sense to include them into deployment's labels as well?

2. Should not match labels be the same as template's labels? If I'm not mistaken it's a requirement from some K8s/OpenShift version.

sleshchenko: I moved labels to k8s recommended one.
Please review https://github.com/sleshchenko/che-workspace-operator/commit/489f89a1df3c854cdb6aa48aabfac29dbf53da44 and pull it if it's OK

And again reference to Recommended labels for k8s objects https://kubernetes.cn/docs/concepts/overview/working-with-objects/common-labels/

amisevsk: Makes sense -- I only copied what was present on the web-terminal-operator side. I'll add labels as you suggest.

amisevsk: Added PR with a minor fixup: app label in the aggregate edit/view roles changed to `devworkspace-controller` instead of `che-plugin-registry`.

sleshchenko:
tests failed because of used `cert-gen:latest`. Rebased against master to fix them

sleshchenko:
Seems, tests should be adapted before merging this PR:

> Pod not created yet with selector app=che-workspace-controller in namespace che-workspace-controller

I'll handle that

sleshchenko:
@JPinkney You're right. I already have a fix locally and verifying that it's work. I would not merge PR if it's possible not to break e2e tests.

sleshchenko:
Screenshot_20200706_175740
Now tests pass but not on PR check since cert-gen image must be rebuilt

PR #121, opened 2020-07-03
Simplify webhook certificates setup for controller

amisevsk:
### What does this PR do?
- Remove cert-generation deployment
- Add functionality for discovering tls cert and key at startup
- Use service service CA to mount secret in path matching OLM

Also adds a sample to `deploy/crds/samples` because it makes CSV generation simpler.

Remove need for second cert-generation deployment

### Is it tested? How?
Tested on `crc` both deployed manually and via OLM. For testing via OLM, see PR https://github.com/redhat-developer/web-terminal-operator/pull/2


deploy/os/controller.yaml
```diff
@@ -51,16 +51,9 @@ spec:
...
-            sources:
-              - configMap:
-                  name: devworkspace-controller-secure-service
-                  items:
-                    - key: service-ca.crt
-                      path: ./ca.crt
-              - secret:
-                  name: devworkspace-controller
+          secret:
+            secretName: workspace-controller
```
sleshchenko, line 59:
```suggestion
            secretName: devworkspace-controller
```
?

deploy/os/service.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: workspace-controller
+  labels:
+    app: che-workspace-controller
```
sleshchenko, line 7: let's not introduce one more mentioning of che here

pkg/webhook/server/server.go
```diff
@@ -83,3 +92,25 @@ func GetWebhookServer() *webhook.Server {
...
+	if _, err := os.Stat(filepath.Join(webhookServerCertDir, "tls.key")); err == nil {
+		key = "tls.key"
+		if _, err := os.Stat(filepath.Join(webhookServerCertDir, "tls.crt")); err != nil {
+			return "", "", fmt.Errorf("could not find certificates for webhook: could not find tls.crt")
+		}
+		crt = "tls.crt"
+		return key, crt, nil
+	}
+	// Mounted by OLM: apiserver.key and apiserver.crt
+	if _, err := os.Stat(filepath.Join(webhookServerCertDir, "apiserver.key")); err == nil {
```
sleshchenko, line 107: When I install WebTerminalOperator with `make olm_install_local` I get 
Screenshot_20200703_135935

Does OLM here mean something different?

sleshchenko: I got from standup that apiserver files are mounted when no annotated service is used but when webhooks configurations are described in OLM manifests.

Makefile
```diff
@@ -122,10 +114,7 @@ _apply_controller_cfg:
 	$(TOOL) apply -f ./deploy -n $(NAMESPACE)
 	mv ./deploy/role_binding.yaml.bak ./deploy/role_binding.yaml
 ifeq ($(TOOL),oc)
-ifeq ($(WEBHOOK_ENABLED),true)
-	$(TOOL) apply -f ./deploy/os/che-workspace-controller-cert-gen-deployment.yaml -n $(NAMESPACE)
```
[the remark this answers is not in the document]
amisevsk: The *other* cert gen is for running on k8s, since it doesn't support the service annotation. I thought I got everything related to this cert gen but I'll double-check. We'll eventually need to decide how we want to support k8s, but for now I think leaving the existing tools in place is a good idea.

deploy/os/service.yaml
amisevsk: It should match the other labels, that's why it's there now. I can update all of our labels in this PR I suppose.

pkg/webhook/server/server.go
amisevsk: The webhooks are not added to the CSV yet, they're still mounted via secret. I think a follow-up task is to check how adding the webhooks to the CSV will work (if e.g. it'll resolve our webhook delay to have them created by OLM) but it's not done yet to keep the scope of this PR smaller.

Potential issues we'll run into are
- We check/modify webhooks in the controller itself, which could potentially lead to a loop if the specs go out of sync
- We need to manage two separate flows (webhooks created by controller vs webhooks created by OLM)
- We'll need to figure out how CSV generation works to ensure adding webhooks to the CSV isn't a manual step.

Makefile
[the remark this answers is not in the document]
amisevsk: Looking at your diff -- it looks like I lost some changes in a rebase. I was *sure* I removed those files :)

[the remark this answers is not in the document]
amisevsk: Cherry-picked your changes to the branch. Thanks for prepping a working commit :)

deploy/os/controller.yaml
amisevsk: :+1: Changed here and in `service.yaml`.

deploy/os/service.yaml
amisevsk: Done :)

sleshchenko:
Tests do not have a chance to pass since `nightly` image is not compatible with made deployment changes. We can merge without tests after making sure that tests passed locally.

sleshchenko:
This PR breaks uninstallation with OLM since service is not automatically removed.
Screenshot_20200706_123052
And again dedicated admin - seems they can't remove services in `openshift-` namespaces:
```
> oc auth can-i delete service -n openshift-operators
no
> oc whoami
sleshchenko
> oc get group
NAME                     USERS
dedicated-admins   sleshchenko
```

amisevsk:
Superceded by https://github.com/devfile/devworkspace-operator/pull/126

PR #122, opened 2020-07-03
Fail workspace startup when creator is missing but webhooks enabled

sleshchenko:
### What does this PR do?
This PR improve behavior when a workspace is created before webhooks are set.

Previously, we got Starting workspace forever with endless error in controller logs:
Screenshot_20200703_111058

With that PR we get FAILED workspace with one error log in a controller:
Screenshot_20200703_114243

What could be done better:
1. DevWorkspace status must have a failure message or the corresponding event must be propagated, otherwise user does not have a clue why the workspace is failed (which is better comparing to previous state when workspace is staring forever). There is a dedicated issue for it https://github.com/eclipse/che/issues/17326
2. Probably there might be some errors thrown by deployment#getSpecDeployment after which it makes sense to reconcile devworkspace again but to archive it it's needed to do so refactoring where thrown error contains info if it's critical in terms of workspace start-up or not... I think the proposed solution is good enough to proceed with, users is able to start/stop workspace is any case. And later we'll do such refactoring.

N/A

### Is it tested? How?
1. `make deploy` when image is not pulled.
2. create devworkspace just after `make deploy`
3. Check devworkspace status and controller logs when controller is up.
Screenshot_20200703_114715
Screenshot_20200703_114243

### DIFF
--- a/pkg/controller/workspace/provision/deployment.go
+++ b/pkg/controller/workspace/provision/deployment.go
@@ -14,6 +14,7 @@ package provision
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


sleshchenko:
> what happens if you try to add a creator ID after the workspace is in the failed state?

I believe it should be:

> "label 'creator...' is missing. Please recreate workspace to get it initialized

https://github.com/devfile/devworkspace-operator/blob/master/pkg/webhook/workspace/handler/workspace.go#L52

But let me retest before merging.

sleshchenko:
Tested. Any update attempt is denied with error:
> error: devworkspaces.workspace.devfile.io "web-terminal" could not be patched: admission webhook "mutate.che-workspace-controller.svc" denied the request: label 'controller.devfile.io/creator' is missing. Please recreate workspace to get it initialized

PR #126, opened 2020-07-07
Separate webhook server into its own deployment

JPinkney:
### What does this PR do?
This PR reworks the webhook server into a seperate deployment. This will make it so that when uninstalling the operator existing workspaces will be secure until admin deletes the seperate webhook server deployment. Also it avoids the issue with https://github.com/redhat-developer/web-terminal-operator/pull/2#pullrequestreview-443109054.

The first commit removes the old cert generation stuff
The second commit introduces the new webhook server in its own deployment

Update: I've tested it in OLM and when uninstall the operator the webhook server stays around

:warning: Note if we go this way - we need to update OpenShift Console not to translate webhooks availability as WebTerminal installed, instead it should depend on Operator Subscription... At the same time, it makes sense to check webhooks availability as well.

Fixes https://github.com/eclipse/che/issues/16967
Fixes https://github.com/eclipse/che/issues/16980

The main purpose - implement proper uninstallation described in https://issues.redhat.com/browse/WTO-13

### Is it tested? How?
```
export IMG=my_img
make docker

make test_e2e
# make sure test passes

export DEFAULT_ROUTING=openshift-oauth
make deploy
# wait until webhook server is started
oc project devworkspace-controller
oc apply -f samples/cloud-shell.yaml
oc get devworkspace -w
# make sure workspace is started and you're able to open terminal by its URL
oc delete all --selector app.kubernetes.io/name=devworkspace-controller
# make sure that you're still able to exec into pods, you can refresh Cloud Shell Workspace URL
# Note that after that you won't be able to do make uninstall until do: remove mutating webhook and set workspacerouting finalizers to []
```

It's not tested after the latest changed via OLM yet


build/Dockerfile
```diff
@@ -16,13 +16,23 @@ RUN CGO_ENABLED=0 GOOS=linux go build \
...
+
 FROM registry.access.redhat.com/ubi8-minimal:8.1-279
 COPY --from=builder /che-workspace-operator/_output/bin/che-workspace-controller /usr/local/bin/che-workspace-controller
+COPY --from=builder /che-workspace-operator/_output/bin/webhook-server /usr/local/bin/webhook-server
 COPY --from=builder /che-workspace-operator/internal-registry  internal-registry
 
 ENV USER_UID=1001 \
     USER_NAME=che-workspace-controller
 
+RUN chmod 775 /usr/local/bin/webhook-server
```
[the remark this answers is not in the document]
sleshchenko: Seems it still works after removing this line

pkg/webhook/webhook.go
```diff
@@ -1,43 +1,52 @@
...
-// This program and the accompanying materials are made
-// available under the terms of the Eclipse Public License 2.0
-// which is available at https://www.eclipse.org/legal/epl-2.0/
-//
-// SPDX-License-Identifier: EPL-2.0
-//
-// Contributors:
-//   Red Hat, Inc. - initial API and implementation
-//
 package webhook
```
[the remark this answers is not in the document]
sleshchenko: Reverted

pkg/webhook/webhook_certs.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		log.Info("Failed creating the secure configmap")
+		return err
+	}
+	return nil
+}
+
+func createSecureService(client crclient.Client, ctx context.Context, namespace string) error {
+	label := map[string]string{
+		"app.kubernetes.io/name": "devworkspace-controller",
```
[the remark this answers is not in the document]
sleshchenko: Changed to `"app.kubernetes.io/name": "devworkspace-controller",`

pkg/webhook/webhook_deployment.go
```diff
@@ -0,0 +1,174 @@
...
+
+	var user *int64
+	if !config.ControllerCfg.IsOpenShift() {
+		uID := int64(1234)
+		user = &uID
+	}
+
+	saName := os.Getenv(config.ControllerServiceAccountNameEnvVar)
+	if saName == "" {
+		return nil
```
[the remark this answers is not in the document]
sleshchenko: Now error it returned here

pkg/webhook/webhook_deployment.go
```diff
@@ -0,0 +1,174 @@
...
+														Key: "service-ca.crt",
+														Path: "./ca.crt",
+													},
+												},
+											},
+										},
+										{
+											Secret: &corev1.SecretProjection{
+												LocalObjectReference: corev1.LocalObjectReference{
+													Name: SecureServiceName,
```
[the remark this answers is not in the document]
sleshchenko: Fixed

pkg/webhook/webhook_certs.go
```diff
@@ -0,0 +1,168 @@
...
+			return err
+		}
+		log.Info("Updating webhook server secure cert service")
+	} else {
+		log.Info("Updating webhook server secure cert service")
+	}
+	return nil
+}
+
+func createConfigMap(client crclient.Client, ctx context.Context, namespace string) error {
```
amisevsk, line 99: Why is creating a configmap necessary? If you properly annotate the service, the secret that's created contains everything necessary to secure the webhooks server. In fact, the value from the configmap just results in the same cert being mounted twice:

```
sh-4.4$ cat ca.crt
-----BEGIN CERTIFICATE-----
MIIDUTCCAjmgAwIBAgIIM9zMSeK<snip>LGdCTeNeKyWJGCZlqUpjSUw==
-----END CERTIFICATE-----


sh-4.4$ cat tls.crt
-----BEGIN CERTIFICATE-----
MIIEJjCCAw6gAwIBAgIIdbYMKGO<snip>oidBU1KQMlDjEreW/cFaFmXEjf8Vxgu4g0hyoh
VW58v/oaewmr4g==
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
MIIDUTCCAjmgAwIBAgIIM9zMSeK<snip>LGdCTeNeKyWJGCZlqUpjSUw==
-----END CERTIFICATE-----
```

The second cert in `tls.crt` is identical to `ca.crt`.

See https://github.com/devfile/devworkspace-operator/pull/121/commits/449520373dfb820e0768ea04e03e4d4eb54ee052 -- I got it to work with just a service.

pkg/webhook/webhook_roles.go
```diff
@@ -0,0 +1,99 @@
...
+		log.Info("Created webhook server cluster role")
+	}
+
+	return nil
+}
+
+func getSpecRoleBinding(saName string) (*v1.ClusterRole, error) {
+
+	labels := map[string]string{
+		"app.kubernetes.io/name":    saName,
```
sleshchenko, line 57: I would not use here service account name

sleshchenko: Is addressed in https://github.com/devfile/devworkspace-operator/pull/126/commits/681ffb7526cd805834b9baf2543caede45322266

sleshchenko:
> Error from server: admission webhook "validate-exec.devworkspace-controller.svc" denied the request: the cache is not started, can not read objects

Thanks for catching it. I just verified that workspace is started but no more at yesterday point.
So, we manually started Webhook server while it's supposed to be done via manager https://github.com/devfile/devworkspace-operator/pull/126/commits/e3a053141a563a2b9d854ef0549cd0110c2da11e

Now I'm able to start cloud-shell with OpenShift OAuth proxy with one logged error, Josh faced as well:

> "Denied request to UPDATE resource with kind /v1, Kind=Service by user system:serviceaccount:openshift-service-ca:service-ca: object is owned by workspace and cannot be updated."

It's the bug in unmodifiable objects logic we have. I'll create a separate issue to deal with that. I believe it should be reproducible on master as well.

pkg/webhook/webhook_certs.go
sleshchenko: Good catch. As far as I see ca.crt is not even used by webhook server. So, removed configmap.

sleshchenko:
There is an issue dedicated for 
> "Denied request to UPDATE resource with kind /v1, Kind=Service by user system:serviceaccount:openshift-service-ca:service-ca: object is owned by workspace and cannot be updated."

https://github.com/eclipse/che/issues/17366

pkg/webhook/webhook_cluster_role_binding.go
```diff
@@ -0,0 +1,95 @@
...
+		Subjects: []v1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      saName,
+				Namespace: namespace,
+			},
+		},
+		RoleRef: v1.RoleRef{
+			Kind: "ClusterRole",
+			//Isn't that role removed along with Operator? Should not we create a dedicated role for webhook server?
```
JPinkney, line 73: Looks like I created the role but forgot to switch from devworkspace-controller role to the new role

pkg/webhook/webhook_cluster_role_binding.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := client.Create(ctx, clusterRoleBinding); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		existingCfg, err := getClusterRoleBinding(ctx, client)
+		if err != nil {
+			return err
+		}
+		clusterRoleBinding.ResourceVersion = existingCfg.ResourceVersion
+		err = client.Update(ctx, clusterRoleBinding)
```
amisevsk, line 44: I had an old clusterrolebinding on the cluster from an earlier test of this PR, and this step fails when the binding needs to be updated. Maybe it makes more sense to 
1. Get existing cfg
2. Compare to spec
   a. If different, delete and recreate
   b. If same, do nothing.

pkg/webhook/webhook_roles.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/devfile/devworkspace-operator/webhook/server"
+	v1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func CreateWebhookRole(client crclient.Client,
```
amisevsk, line 26:
```suggestion
func CreateWebhookClusterRole(client crclient.Client,
```
And a similar rename for the filename

pkg/webhook/webhook_roles.go
```diff
@@ -0,0 +1,104 @@
...
+			{
+				APIGroups: []string{
+					"admissionregistration.k8s.io",
+				},
+				Resources: []string{
+					"mutatingwebhookconfigurations",
+					"validatingwebhookconfigurations",
+				},
+				Verbs: []string{
+					"*",
```
amisevsk, line 71: Do we need `delete/deleteCollection`? I worry that taking unnecessary permissions could make people hesitate to install the operator. We'd also need to update `role.yaml` with this change.

webhook/server/server.go
```diff
@@ -11,34 +11,48 @@
...
 
 var log = logf.Log.WithName("webhook.server")
 var webhookServer *webhook.Server
 var CABundle []byte
 
-func ConfigureWebhookServer(mgr manager.Manager, ctx context.Context) error {
-	if config.ControllerCfg.GetWebhooksEnabled() == "false" {
-		log.Info("Webhooks are disabled. Skipping setting up webhook server")
-		return nil
+var WebhookServerAppLabels = func() map[string]string {
```
amisevsk, line 48: Same comment as above -- is it necessary for this to be a constant function?

pkg/webhook/webhook_cluster_role_binding.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return nil
+}
+
+func getSpecClusterRoleBinding(namespace string) (*v1.ClusterRoleBinding, error) {
+	clusterRoleBinding := &v1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      server.WebhookServerSAName,
+			Namespace: namespace,
```
amisevsk, line 60: ClusterRoles and bindings are cluster-scoped, so `namespace` shouldn't be here.

Makefile
```diff
@@ -143,6 +135,15 @@ else
 	kubectl rollout restart -n $(NAMESPACE) deployment/devworkspace-controller
 endif
 
+_do_restart_webhook_server:
+ifeq ($(TOOL),oc)
+	oc patch deployment/devworkspace-operator-webhook-server \
```
amisevsk, line 140:
```suggestion
	oc patch deployment/devworkspace-webhook-server \
```

pkg/config/constants.go
```diff
@@ -12,6 +12,14 @@
 
 package config
 
+// Labels which should be used for controller related objects
+var ControllerAppLabels = func() map[string]string {
```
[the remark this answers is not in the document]
sleshchenko: To avoid occasional change by clients since we can't put map as const. So I came with function that will return always new instance which clients will be able to fill in with extra items if needed.
@amisevsk WDYT? Can you suggest another way to archive it?

pkg/webhook/webhook_cluster_role_binding.go
sleshchenko: I faced the same
+1 to fix before merging

sleshchenko:
> I currently see Denied request to UPDATE 'workspace12fd04c1205a4aff-service' Service by user system:serviceaccount:openshift-service-ca:service-ca: object is owned by workspace and cannot be updated." when deploying a workspace with cluster-tls routing, so we need to special-case the mutating webhook for services.

It doesn't block workspace start, right?
It's not something that is introduced in that PR and the same issue can be reproduced in master. I believe I've create a dedicated issue for that but I might forget )

pkg/webhook/webhook_roles.go
sleshchenko: Better request what we only need

amisevsk:
Added a commit to resolve the service webhook issue.

JPinkney:
Yeah this is the issue for the service webhook: https://github.com/eclipse/che/issues/17366

amisevsk:
Ah I must've missed it. We can separate out the commit if need be. I also added the minor changes from my last review, I hope that's okay.

Fixes https://github.com/eclipse/che/issues/17366

pkg/webhook/webhook_cluster_role_binding.go
amisevsk: This can be done as a separate PR, IMO

pkg/config/constants.go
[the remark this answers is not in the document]
amisevsk: That makes sense, it just looked like a strange construct at first glance.

pkg/webhook/webhook_roles.go
amisevsk: Limited to `get, list, create, patch, update`.

pkg/webhook/cluster_role_bindings.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if err := client.Create(ctx, roleBinding); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		existingCfg, err := getClusterRoleBinding(ctx, client)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("\n%s\n\n", cmp.Diff(roleBinding, existingCfg))
```
sleshchenko, line 46: Seems it should be removed

Makefile
```diff
@@ -75,16 +72,14 @@ _update_yamls: _set_registry_url
 	rm ./deploy/controller_config.yaml.bak
 ifeq ($(TOOL),oc)
 	sed -i.bak -e "s|image: .*|image: $(IMG)|g" ./deploy/os/controller.yaml
+	sed -i.bak -e "s|value: \"quay.io/che-incubator/che-workspace-controller:nightly\"|value: $(IMG)|g" ./deploy/os/controller.yaml
```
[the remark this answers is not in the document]
sleshchenko: I don't think so )
it's `value:` that might be used for other env vars... I'll take a look if we can use multilines here and add env var name for matching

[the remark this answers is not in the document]
sleshchenko: Seems matching multi lines with sed is not an easy thing. Leaving as is for time being.

webhook/main.go
```diff
@@ -34,28 +35,29 @@ var log = logf.Log.WithName("cmd")
 func main() {
 	logf.SetLogger(zap.Logger())
 
-	clusterConfig, err := rest.InClusterConfig()
+	// Get a config to talk to the apiserver
+	cfg, err := config.GetConfig()
 	if err != nil {
-		log.Error(err, "Failed when attempting to retrieve in cluster config")
+		log.Error(err, "")
```
JPinkney, line 41:
```suggestion
		log.Error(err, "Failed when retrieving config")
```

JPinkney:
I can't really approve my own PR so I'll just give a 👍 for approval

webhook/main.go
sleshchenko: I think `""` here is to be inlined with controller/main.go and the motivation behind it, I think - error should already have clear message when it happened...

But I assume there might be some errors thrown like `value must not be nil`, which won't show the place where it happened on controller level. So, I'm not +1 to duplicate message in the way Joshua suggests but it can be done as separate PR for all places where we log `""`

sleshchenko:
Even e2e test passed :-D

sleshchenko:
Merging it and since I don't expect any critical bugs here. Minor fixes could be provided separately

webhook/main.go
amisevsk: Yeah woops, this was copied from the main controller code. I can't recall if it's autogenerated or not :smile:

PR #131, opened 2020-07-09
Add images documenting webhook and workspace creation flows

amisevsk:
### What does this PR do?
Adds a few images to `./doc` to document the workspace startup and webhook flows. There's no other documentation at the moment, but these could ostensibly be reused once we start more detailed documentation.

I'm adding web-terminal specific images here (instead of in [web-terminal-operator](https://github.com/amisevsk/che-workspace-crd-operator/pull/new/add-doc-images) since the core logic lives here.


amisevsk:
>Do you think we should have a WTO subdirectory under doc and then have another subdirectory specific for devworkspace?

We probably should :) I just threw this PR together since I created the images for the OLM Jira I opened.

amisevsk:
Probably it makes sense to have these be editable.

PR #132, opened 2020-07-13
Remove CICO scripts and references to the corresponding Jobs

sleshchenko:
### What does this PR do?
This PR removes CICO scripts and references to the corresponding Jobs.

🚧 since it depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/6401

Fixes https://github.com/eclipse/che/issues/17179

### Is it tested? How?
Nothing to test


sleshchenko:
> I wonder if it makes sense to move cico_openshift_e2e.sh into ./test.

I thought about it and it makes sense for sure. I did not do it because it would need changing configuration somewhere I'm not sure about )

@flacatus Could you point me where `cico_openshift_e2e.sh` is triggered, and then I'll rename e2e test script since it's not CentOS CI related.

sleshchenko:
ci/prow/v4-openshift-web-terminal-e2e fails because of
> {"level":"error","ts":1594723317.025547,"logger":"cmd","msg":"","error":"clusterroles.rbac.authorization.k8s.io is forbidden: User \"system:serviceaccount:devworkspace-controller:devworkspace-controller\" cannot create resource \"clusterroles\" in API group \"rbac.authorization.k8s.io\" at the cluster scope","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nmain.main\n\t/devworkspace-operator/cmd/manager/main.go:136\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}

It must not be related to the PR's changes and I'm going to rerun tests after new next tag is built where webhook server is separated.

PR #136, opened 2020-07-16
Fail workspace startup if workspace is older than webhooks

amisevsk:
### What does this PR do?
- Fail startup at start of reconcile loop if creator id isn't present, instead of during deployment provisioning
- Fail workspace startup if creationTimestamp is older than mutating webhook's creationTimestamp
- Fix a minor issue in clusterrole (watch permissions seem to be required by controller-runtime)

Fixes https://github.com/eclipse/che/issues/16914

Web Terminal Operator manifests are updated in https://github.com/redhat-developer/web-terminal-operator/pull/11

### Is it tested? How?
Tested on `crc`:

```bash
# set env vars as usual
make uninstall
kubectl apply -f devworkspace-crds/deploy/crds/ \
  && kubectl apply -f deploy/crds/ \
  && oc new-project ${NAMESPACE} \
  && kubectl apply -f samples/cloud-shell.yaml
kubectl patch devworkspace cloud-shell --type='merge' \
  -p '{"metadata": {"labels": {"controller.devfile.io/creator": "test"}}}'
make docker deploy
# Check that workspace start fails and message is logged
```

### DIFF
--- a/pkg/controller/workspace/validation.go
+++ b/pkg/controller/workspace/validation.go
@@ -1,0 +1,40 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package workspace
+
+import (
+	"fmt"
+
+	"github.com/devfile/devworkspace-operator/pkg/config"
+	"github.com/devfile/devworkspace-operator/pkg/webhook"
+	devworkspace "github.com/devfile/kubernetes-api/pkg/apis/workspaces/v1alpha1"
+)
+
+func (r *ReconcileWorkspace) validateCreatorTimestamp(workspace *devworkspace.DevWorkspace) error {
+	if config.ControllerCfg.GetWebhooksEnabled() != "true" {
+		return nil
+	}
+	if _, present := workspace.Labels[config.WorkspaceCreatorLabel]; !present {
+		return fmt.Errorf("devworkspace does not have creator label -- it must be recreated to resolve the issue")
+	}
+
+	webhooksTimestamp, err := webhook.GetWebhooksCreationTimestamp(r.client)
+	if err != nil {
+		return fmt.Errorf("webhooks not set up yet: %w", err)
+	}
+	if workspace.CreationTimestamp.Before(&webhooksTimestamp) {
+		return fmt.Errorf("devworkspace created before webhooks -- it must be recreated to resolve the issue")
+	}
+
+	return nil
+}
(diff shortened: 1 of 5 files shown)


pkg/controller/workspace/validation.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+
+	"github.com/devfile/devworkspace-operator/pkg/config"
+	"github.com/devfile/devworkspace-operator/pkg/webhook"
+	devworkspace "github.com/devfile/kubernetes-api/pkg/apis/workspaces/v1alpha1"
+)
+
+func (r *ReconcileWorkspace) validateCreatorTimestamp(workspace *devworkspace.DevWorkspace) error {
+	if config.ControllerCfg.GetWebhooksEnabled() == "true" {
```
sleshchenko, line 24: Personally I prefer to reduce the nesting of the code, but I'm OK with 
```go
	if config.ControllerCfg.GetWebhooksEnabled() == "true" {
    // skip validation creation timestamp if webhook are disabled
    return nil
  }

	if _, present := workspace.Labels[config.WorkspaceCreatorLabel]; !present {
		return fmt.Errorf("devworkspace does not have creator label -- it must be recreated to resolve the issue")
	}

	webhooksTimestamp, err := webhook.GetWebhooksCreationTimestamp(r.client)
	if err != nil {
		return fmt.Errorf("webhooks not set up yet: %w", err)
	}
	if workspace.CreationTimestamp.Before(&webhooksTimestamp) {
		return fmt.Errorf("devworkspace created before webhooks -- it must be recreated to resolve the issue")
	}
```

amisevsk: Yeah, my bad -- I copy pasted from the main function where the nesting was necessary

pkg/webhook/info.go
```diff
@@ -0,0 +1,38 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var webhooksCreationTimestamp = metav1.Time{}
+
+func GetWebhooksCreationTimestamp(client client.Client) (metav1.Time, error) {
+	if webhooksCreationTimestamp.IsZero() {
+		webhook := &v1beta1.MutatingWebhookConfiguration{}
+		err := client.Get(context.TODO(), types.NamespacedName{Name: webhookCfg.MutateWebhookCfgName}, webhook)
```
sleshchenko, line 31: Can we watch webhook config here and reset the cached value in case webhooks removed or recreated?

amisevsk: We can, but it would increase the complexity significantly. Ideally we could have some sort of expiring cache or similar, but this is a reasonable stop-gap measure I think.

amisevsk:
> Also, the use case that came to my mind that may need a separate issue if it's relevant: can we watch for webhooks configurations by controller and FAIL all workspace (remove their deployments) when webhook configurations are removed? Probably it's out of our power since it's operator framework who invoke reconcile loop... But if somebody ask that question - we can tell that we thought about it )

Yeah, this is something that's kind of outside of our scope. It's tricky since the devworkspace controller doesn't really know how many devworkspaces are on the cluster and doesn't track any state between reconciles (at least, per workspace), so the only way to really do it would be yet another deployment that watches webhooks and then iterates through all namespaces to delete workspaces.

amisevsk:
Created https://github.com/eclipse/che/issues/17433 as a follow-up

PR #140, opened 2020-07-21
Update README to remove references to Che

l0rd:
### What does this PR do?

Update the README now that the operator is supposed to be a dependency not only of Che but of the Web Terminal too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,7 @@ make local
 operator-sdk up local --namespace ${NAMESPACE}
 ```
 
-When running locally, only a single namespace is wat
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,7 +79,8 @@ make local
 operator-sdk up local --namespace ${NAMESPACE}
 ```
 
-When running locally, only a single namespace is watched; as a result, all workspaces have to be deployed to `${NAMESPACE}`
+When running locally, only a single namespace is wat
```
benoitf, line 82: looks a typo error

l0rd: Fixed. Thank you @benoitf

PR #141, opened 2020-07-21
Remove Che references from code

sleshchenko:
### What does this PR do?
One does not simply get rid of Che references =) It's why this PR appeared.
It contains three commits with self-describing messages.
The only thing that makes sense to mention, we still have che references in:
- che REST API sidecar;
- che plugin broker;
- che plugin registry;
- Che Related env vars, like CHE_WORKSPACE_ID, CHE_WORKSPACE_NAMESPACE, ... I assume they are used by Theia and it's better to migrate them later.

https://github.com/eclipse/che/issues/17179

### Is it tested? How?
I've tested that CloudShell workspace with OpenShift OAuth still work.

### DIFF
--- a/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
+++ b/internal-registry/eclipse/cloud-shell/nightly/meta.yaml
@@ -28,7 +28,7 @@ spec:
     ports:
     - exposedPort: 4444
     command: ["/go/bin/che-machine-exec",
-              "--authenticated-user-id", "$(CHE_WORKSPACE_CREATOR)",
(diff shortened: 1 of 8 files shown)


amisevsk:
> Che Related env vars, like CHE_WORKSPACE_ID, CHE_WORKSPACE_NAMESPACE, ... I assume they are used by Theia and it's better to migrate them later.

I agree, I believe Che-Theia and Che REST APIs depend on these being named as they are so more care needs to be taken. Also, I don't think it's that bad to have Che references in env vars used for configuring Che components

PR #144, opened 2020-07-30
Add PR validation with GitHub action

sleshchenko:
### What does this PR do?
This PR adds PR validation with GitHub action.
It's porting with adaptation from https://github.com/devfile/parser/blob/master/.github/workflows/go.yml

What can be done better? - We could cache docker layers to speed up docker container build phase like here https://github.com/che-incubator/configbump/blob/master/.github/workflows/ci.yaml#L29-L36. But we can improve it a bit later.

N/A

### Is it tested? How?
it's tested in https://github.com/devfile/devworkspace-operator/actions/runs/188642947

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,68 @@
+name: Validate PRs
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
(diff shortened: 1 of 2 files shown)


amisevsk:
Great addition!

PR #146, opened 2020-08-28
Fix devworkspace api to be v1alpha1

sleshchenko:
### What does this PR do?
The version in devfile/api is changed to v1alpha2 but devworkspace-operator does not seem to be adapted to that and I faced the following issues:
Screenshot_20200828_145750
Screenshot_20200828_145811

So, this issue fixes devfile/api version to the latest where we had v1alpha1. 
Later when devworkspace API is adapted to v1alpha2 we can return back to master.

:warning: Houston, we have a problem!
<details>

<summary>Details inside</summary>

We have inconsistency in our dependencies:
from go sources we depend on the  Aug 11, 2020 https://github.com/devfile/api/commit/b6daa2c997a523f0859a83d49656cf1336ba7059
but the latest Component CRD is generated somewhere from  Jun 16, 2020 https://github.com/devfile/api/commit/28e3e6acdc368b503672738697b149dd23c86770 or even earlier

So, Web Terminal Operator has different objects (like endpoints) in DevWorkspace and Component:
https://github.com/redhat-developer/web-terminal-operator/blob/master/manifests/controller.devfile.io_components_crd.yaml#L347
https://github.com/redhat-developer/web-terminal-operator/blob/master/manifests/workspace.devfile.io_devworkspaces_crd.yaml#L431
Technically, it does not fail Web Terminal but if user will try to use more from DevWorkspace format - I think devworkspace will fail to reconcile because of that inconsistency.
The possible solutions:
- leave as is and don't care about advanced usage of devworkspaces. But but that it might be an issue to create a valid transforming webhook;
- update web-terminal-operator CRDs to the latest we want to have as v1alpha1 and which is compatible with CR created by OpenShift Console.

</details>
We agree to try to propagate consistent CRD to WTO if it does not influence the release date.

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ WEBHOOK_ENABLED ?= true
 DEFAULT_ROUTING ?= basic
 ADMIN_CTX ?= ""
 REGISTRY_ENABLED ?= true
-DEVWORKSPACE_API_VERSION ?= master
+DEVWORKSPACE_API_VERSION ?= v1alpha1
 
 #internal params
 INTERNAL_TMP_DIR=/tmp/devworkspace-controller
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -265,9 +265,12 @@ update_devworkspace_crds:
 	if git show-ref --verify refs/tags/$(DEVWORKSPACE_API_VERSION) --quiet; then
 		echo 'DevWorkspace API is specified from tag'
 		git checkout tags/$(DEVWORKSPACE_API_VERSION)
-	else
+	elif git rev-parse --verify $(DEVWORKSPACE_API_VERSION); then
```
sleshchenko, line 268: Now I'm not sure it works fine...Maybe it's better to remove it for time being.

amisevsk: It at least doesn't interfere with release, so we can always fix the makefile later (as long as we have the right CRDs used for generation)

PR #148, opened 2020-09-15
Fix makefile update_devworkspace_crds rule 

l0rd:
### What does this PR do?

I have followed the instructions in the README and the command `make deploy` failed (see error below).

I have looked at the error and it was due to the fact that the rule `update_devworkspace_crds` had some multi line bash commands that do not work in a makefile.

The command `make deploy` is currently not working for me (macOS, go1.14.4, GNU Make 3.81).

```bash
$ make deploy
Current env vars:
    IMG=quay.io/devfile/devworkspace-controller:next
    PULL_POLICY=Always
    ROUTING_SUFFIX=192.168.99.100.nip.io
    WEBHOOK_ENABLED=true
    DEFAULT_ROUTING=basic
    REGISTRY_ENABLED=true
    DEVWORKSPACE_API_VERSION=v1alpha1
oc create namespace devworkspace-controller || true
(...)
rm ./deploy/os/controller.yaml.bak
cd devworkspace-crds
if [ ! -d ./.git ]; then
bash: -c: line 1: syntax error: unexpected end of file
make: *** [update_devworkspace_crds] Error 1
```

### Is it tested? How?

On both linux and macOS the command `make deploy` completes successfully.

```bash
$ make deploy
(...)
rm ./deploy/os/controller.yaml.bak
$ echo $?
0
...

### DIFF
--- a/update_devworkspace_crds.sh
+++ b/update_devworkspace_crds.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Copyright (c) 2019-2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+set -e
+
+DEVWORKSPACE_API_VERSION=${1:-v1alpha1}
+
+mkdir -p devworkspace-crds
+cd devworkspace-crds
+if [ ! -d ./.git ]; then
+	git init
+	git remote add origin -f https://github.com/devfile/api.git
+	git config core.sparsecheckout true
+	echo "deploy/crds/*" > .git/info/sparse-checkout
+else
+	git remote set-url origin https://github.com/devfile/api.git
+fi
+git fetch --tags -p origin
+if git show-ref --verify refs/tags/"${DEVWORKSPACE_API_VERSION}" --quiet; then
+	echo 'DevWorkspace API is specified from tag'
+	git checkout tags/"${DEVWORKSPACE_API_VERSION}"
+elif git rev-parse --verify "${DEVWORKSPACE_API_VERSION}"; then
+	echo 'DevWorkspace API is specified from branch'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by sleshchenko — COMMENTED** [review]
Seems it fixes https://github.com/eclipse/che/issues/17645

We just need to make sure that it works on Linux as well.
Later I would like to see it moved to `.sh` script moved as we already did for web-terminal-operator.

**Comment by l0rd** [comment]
@sleshchenko ok I will check on linux as well.

**Review by amisevsk — COMMENTED** [review]
I don't this this works on Linux, as the `.ONESHELL` directive means we try to 
```bash
cd devworkspace-crds
cd devworkspace-crds
```
which fails.

Digging into it a bit, it looks like the issue is that the default version of Make on MacOS is just old enough (at ~14 years) to not include support for `.ONESHELL`. If we want to go the path of not using `.ONESHELL` we'll need to remove the directive from the top, but IIRC we included it because it was necessary for some tricky commands.

Another option is to extract the update rules to scripts and call them in the makefile -- that should avoid the issue.

l0rd:
@sleshchenko @amisevsk I have extracted the shell commands in an external shell script and now `make deploy` works on both linux and macOS.
 
The script is called `update_devworkspace_crds.sh` and is at the root of the repo. Do you want me to move it in a specific folder?

update_devworkspace_crds.sh
```diff
@@ -0,0 +1,37 @@
...
+if git show-ref --verify refs/tags/"${DEVWORKSPACE_API_VERSION}" --quiet; then
+	echo 'DevWorkspace API is specified from tag'
+	git checkout tags/"${DEVWORKSPACE_API_VERSION}"
+elif git rev-parse --verify "${DEVWORKSPACE_API_VERSION}"; then
+	echo 'DevWorkspace API is specified from branch'
+	git checkout "${DEVWORKSPACE_API_VERSION}" && git reset --hard origin/"${DEVWORKSPACE_API_VERSION}"
+else
+	echo 'DevWorkspace API is specified from revision'
+	git checkout "${DEVWORKSPACE_API_VERSION}"
+fi
```
JPinkney, line 37: Looks like we need a newline at the end of the file

l0rd: done

JPinkney (approved):
Just tested on Fedora and its working for me

amisevsk (approved):
LGTM, thanks @l0rd 

I think the script *could* go into `build/scripts` or something similar, but the root of the repo is fine for now.

PR #152, opened 2020-09-18
feat(ubi8): Upgrade ubi8 version to a grade A image

benoitf:
### What does this PR do?
very old image is used
wonder why alpine based image is then using ubi8 and not scratch or alpine ?

https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?tag=8.1-279

### Is it tested? How?
Launch make deploy

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -23,7 +23,7 @@ RUN CGO_ENABLED=0 GOOS=linux go build \
   -asmflags all=-trimpath=/ \
   webhook/main.go
 
-FROM registry.access.redhat.com/ubi8-minimal:8.1-279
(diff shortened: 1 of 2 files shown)


JPinkney (approved):

sleshchenko (approved):

sleshchenko:
> wonder why alpine based image is then using ubi8 and not scratch or alpine ?

probably we should do the same as che operator does: uses rhel based go container https://github.com/eclipse/che-operator/blob/master/Dockerfile
But If I'm not mistaken at time I tried to use it - we had some issues.

e2e tests passed locally on crc 1.15.0 with OpenShift 4.5.7

PR #157, opened 2020-09-23
feat(minikube) : better support out of the box with minikube

benoitf:
### What does this PR do?
When minikube is used as current kube context, grab the minikube ip for ingress domain and set tool to `kubectl` automatically

Also add `make info` to display these variables.

N/A
When using minikube, the default IP is not correct.

### Is it tested? How?
make deploy with minikube


sleshchenko (approved):

JPinkney (approved):

benoitf:
hello, seems I need the ok to test to have tests running ?

amisevsk (approved):
`make info` is a great idea, I don't know why we haven't already added it.

PR #158, opened 2020-09-25
Check license headers in GitHub PR actions

amisevsk:
### What does this PR do?
Adds a step in the PR check github action to check that all files have license headers.

### Is it tested? How?
Tested locally, :crossed_fingers: I got the action right.


amisevsk:
Woops I missed that this is already in the `check_format` step.

PR #159, opened 2020-09-25
Use a job for generating certs on kubernetes

JPinkney:
### What does this PR do?
This PR does a few things:
1. It rejoins kubernetes and openshift deployments so we no longer need them to be seperate
2. It removes the kubernetes cert generation scripts
3. Updates the Makefile to reflect these changes

4. Reformulates a bit of the structure around generating certs
```
> pkg/kubernetes
      # Package that will create the jobs etc on kubernetes
> pkg/webhook/service
      # creates a secure service used by openshift and kubernetes. service.go is in its own package in order to resolve a circular dependency issue. if anyone has a better idea to fix that let me know.
> pkg/webhook/kubernetes
      # Actually generate the cert job using pkg/kubernetes and then create the secure service
> pkg/webhook/openshift
      # Create the secure service that openshift uses
```

https://github.com/eclipse/che/issues/17855

### Is it tested? How?

`make docker`
Update controller_config.yaml to include `devworkspace.api_sidecar.image: amisevsk/che-rest-apis:v0.0.2`
`make deploy`
then use:

```
kind: DevWorkspace
metadata:
  name: theia-nodejs
spec:
  started: true
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - plugin:
          id: eclipse/che-theia/latest
      - plugin:
    commands:
      - exec:
          id: download dependencies
          component: nodejs
          commandLine: npm install
      - exec:
          id: run the app
          component: nodejs
          commandLine: nodemon app.js
      - exec:
          id: run the app (debugging enabled)
          component: nodejs
          commandLine: nodemon --inspect app.js
      - exec:
          id: stop the app
          component: nodejs
          commandLine: >-
              node_server_pids=$(pgrep -fx '.*nodemon (--inspect )?app.js' | tr "\\n" " ") &&
              echo "Stopping node server with PIDs: ${node_server_pids}" &&
              kill -15 ${node_server_pids} &>/dev/null && echo 'Done.'
      - vscodeLaunch:
          id: Attach remote debugger
          inlined: |
            {
              "version": "0.2.0",
              "configurations": [
                {
                  "type": "node",
                  "request": "attach",
                  "name": "Attach to Remote",
                  "address": "localhost",
                  "port": 9229,
                  "localRoot": "${workspaceFolder}",
                  "remoteRoot": "${workspaceFolder}"
                }
              ]
            }

```


pkg/webhook/kubernetes/tls.go
```diff
@@ -0,0 +1,87 @@
...
+	TLSDomain                          = "devworkspace-webhookserver.devworkspace-controller.svc"
+)
+
+// SetupKubernetesWebhookCerts handles TLS secrets required for deployment on Kubernetes.
+func SetupKubernetesWebhookCerts(client crclient.Client, ctx context.Context, namespace string) error {
+	devworkspaceSecret := &corev1.Secret{}
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: server.WebhookServerTLSSecretName}, devworkspaceSecret)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			log.Error(err, "Error getting Che TLS secret " + server.WebhookServerTLSSecretName)
```
sleshchenko, line 43: Seems to be copy/paste error

pkg/webhook/kubernetes/sync_jobs.go
```diff
@@ -0,0 +1,114 @@
...
+	"context"
+	"github.com/devfile/devworkspace-operator/internal/images"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func SyncJobToCluster(
```
sleshchenko, line 25: It seems to be free from TLS or webhook server logic. Maybe then it makes sense to move to some common place.

pkg/webhook/kubernetes/pods.go
```diff
@@ -0,0 +1,61 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// cleanupPods takes a job and cleans up all the pods associated with that job
+func cleanupPods(client crclient.Client, job *batchv1.Job, selectorName string) error {
+	selector := fmt.Sprintf("job-name=%s", selectorName)
```
sleshchenko, line 27: can't we just use job.GetName instead of selector name?

pkg/webhook/kubernetes/job.go
```diff
@@ -0,0 +1,77 @@
...
+func waitForJobCompletion(client crclient.Client, name string, namespace string, timeout time.Duration) error {
+	const interval = 1 * time.Second
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		job, err := getJobInNamespace(client, name, namespace)
+		if err != nil {
+			return false, err
+		}
+
+		if job.Status.Succeeded > 0 {
+			log.Info("Please import public part of DevWorkspace self-signed CA certificate from " + TLSSelfSignedCertificateSecretName + " secret into your browser.", )
```
sleshchenko, line 71: webhook server related CA is not supposted to be imported into browser truststore.

pkg/webhook/kubernetes/tls.go
```diff
@@ -0,0 +1,87 @@
...
+
+// TLS related constants
+const (
+	TLSJobName                         = "devworkspace-tls-job"
+	TLSSelfSignedCertificateSecretName = "devworkspace-self-signed-certificate"
+	TLSDomain                          = "devworkspace-webhookserver.devworkspace-controller.svc"
+)
+
+// SetupKubernetesWebhookCerts handles TLS secrets required for deployment on Kubernetes.
+func SetupKubernetesWebhookCerts(client crclient.Client, ctx context.Context, namespace string) error {
```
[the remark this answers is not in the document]
sleshchenko: `SetupKubernetesWebhookCerts` apart from certs creation does create service. Maybe then it's better to name it `setupSecureService`?

pkg/webhook/create.go
```diff
@@ -48,9 +50,44 @@ func SetupWebhooks(ctx context.Context, cfg *rest.Config) error {
 	if err != nil {
 		return fmt.Errorf("failed to create new client: %w", err)
 	}
+
+	err = setupCluster(ctx, client, namespace)
```
sleshchenko, line 54: I don't get why SA, clusterRole and clusterRole binding is hidden in that setupCluster method while they all are steps of setting up webhook server. I would move everything on the same level or regroup it in the following way: initCfg + RBAC + SecureService + Deployment

sleshchenko (commented):
CI reports that some files need to be formatted.

I quickly tried to refactor it in a bit more generic way to be able reuse it for workspace TLS certificates later but for sure there are some mistakes and places which needs to be fixed
So, please take a look and apply changes which you think makes sense
https://github.com/devfile/devworkspace-operator/compare/kubernetes-cert-job...sleshchenko:kubernetes-cert-job_proposal?expand=1

pkg/webhook/create.go
JPinkney: It was done this way so we don't have 200+ line functions like what Che operator has. It's been changed to initCfg + RBAC + SecureService + Deployment.

internal/cluster/job.go
```diff
@@ -0,0 +1,103 @@
...
+	err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, job)
+	if err != nil {
+		return job, err
+	}
+	return job, nil
+}
+
+// deleteJob deletes a given job and cleans up any pods associated with it
+func DeleteJob(client crclient.Client, job *batchv1.Job) error {
+	err := CleanupPods(client, job)
```
amisevsk, line 50: Is it necessary to clean up pods on a job? I assumed the job would be the owner of its pods.

internal/cluster/log.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package cluster
+
+import logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+var log = logf.Log.WithName("internal-cluster")
```
amisevsk, line 17: I realize it's not a *this-PR* issue but we should avoid using the name `log` directly to avoid confusion with the default `go` log.

internal/cluster/job.go
```diff
@@ -0,0 +1,103 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+)
+
+// CleanJob cleans up a job in a given namespace
+func CleanJob(client crclient.Client, name string, namespace string) error {
```
amisevsk, line 26: I feel like most of these functions do not need to be exported.

pkg/kubernetes/tls/certs.go
```diff
@@ -0,0 +1,100 @@
...
+		return err
+	}
+
+	err = cluster.SyncJobToCluster(client, ctx, job)
+	if err != nil {
+		return err
+	}
+
+	// Wait a maximum of 60 seconds for the job to be completed
+	err = cluster.WaitForJobCompletion(client, jobName, params.Namespace, 60*time.Second)
```
amisevsk, line 66: I wonder if 60 seconds is long enough, given how slow we've seen some image pulls to be

pkg/kubernetes/tls/certs.go
```diff
@@ -0,0 +1,100 @@
...
+func removeCACertificate(client crclient.Client, name, namespace string) error {
+	caSelfSignedCertificateSecret := &corev1.Secret{}
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: name}, caSelfSignedCertificateSecret)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Error getting self-signed certificate secret "+name)
+		return err
+	} else if err != nil && errors.IsNotFound(err) {
+		// We don't have anything to remove in this case since its already not found
+		return nil
+	}
```
amisevsk, line 91:
```suggestion
	if err != nil {
		if !errors.IsNotFound(err) {
			log.Error(err, "Error getting self-signed certificate secret "+name)
			return err
		} else {
			// We don't have anything to remove in this case since its already not found
			return nil
		}
	}
```

pkg/kubernetes/tls/certs.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Remove CA secret because TLS secret is missing (they should be generated together).
+	if err = client.Delete(context.TODO(), caSelfSignedCertificateSecret); err != nil {
+		log.Error(err, "Error deleting self-signed certificate secret "+ TLSSelfSignedCertificateSecretName)
+		return err
+	}
+
+	return nil
+}
```
amisevsk, line 100: Repo needs a format

pkg/kubernetes/tls/job.go
```diff
@@ -0,0 +1,76 @@
...
+	var jobEnvVars []corev1.EnvVar
+	for envVarName, envVarValue := range env {
+		jobEnvVars = append(jobEnvVars, corev1.EnvVar{Name: envVarName, Value: envVarValue})
+	}
+
+	job := &batchv1.Job{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Job",
+			APIVersion: batchv1.SchemeGroupVersion.String(),
+		},
```
amisevsk, line 45: I don't think we need an explicit TypeMeta

pkg/webhook/cluster_roles.go
```diff
@@ -100,6 +100,32 @@ func getSpecClusterRole() (*v1.ClusterRole, error) {
...
+			},
+			{
+				APIGroups: []string{
+					"",
+				},
+				Resources: []string{
+					"secrets",
+				},
+				Verbs: []string{
+					"create",
```
amisevsk, line 126: Don't we also need "list", "get", and "delete"?

amisevsk (commented):
A few comments for now, I'll test more thoroughly tomorrow once I fix my docker install :/

internal/cluster/job.go
JPinkney: It seems to linger around until its cleaned up even though the job doesn't exist anymore:
```
NAME                                          READY   STATUS      RESTARTS   AGE
che-plugin-registry-7c995f8845-n76d7          1/1     Running     0          2m41s
devworkspace-controller-d4f9dc86-qsg7l        1/1     Running     0          2m39s
devworkspace-webhook-server-9b7549977-9dlbh   1/1     Running     0          2m27s
webhook-server-gen-cert-nvzfs                 0/1     Completed   0          2m31s
```
```
[joshpinkney@localhost devworkspace-operator]$ kubectl get job
No resources found in devworkspace-controller namespace.
```

internal/cluster/log.go
JPinkney: I think at one point a `log` got merged and then since then it was just used as a convention for the project but yeah it could create some confusion. There's a lot of code quality things like that in this repo

pkg/kubernetes/tls/certs.go
JPinkney: I've bumped it up to 300 seconds. If theres a way in the kubernetes go code to check if an image is being pulled I could technically do
```
waitInfinite until image is pulled
waitForJobCompletion
```

pkg/webhook/create.go
```diff
@@ -48,13 +51,48 @@ func SetupWebhooks(ctx context.Context, cfg *rest.Config) error {
...
 	}
 
+	err = setUpWebhookServerRBAC(ctx, err, client, namespace)
+	if err != nil {
+		return err
+	}
+
+	if config.ControllerCfg.IsOpenShift() {
+		// Set up the certs for OpenShift
+		log.Info("Setting up the OpenShift webhook server secure certs")
```
sleshchenko, line 69:
```suggestion
		log.Info("Setting up the OpenShift webhook server secure service")
```
?

pkg/webhook/create.go
```diff
@@ -48,13 +51,48 @@ func SetupWebhooks(ctx context.Context, cfg *rest.Config) error {
...
+		err := webhook_openshift.SetupSecureService(client, ctx, namespace)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Set up the certs for kubernetes
+		log.Info("Setting up the Kubernetes webhook server secure certs")
```
sleshchenko, line 76:
```suggestion
		log.Info("Setting up the Kubernetes webhook server secure service")
```
?

pkg/webhook/kubernetes/tls.go
```diff
@@ -0,0 +1,61 @@
...
+func SetupSecureService(client crclient.Client, ctx context.Context, namespace string) error {
+	devworkspaceSecret := &corev1.Secret{}
+
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: tls.TLSSelfSignedCertificateSecretName}, devworkspaceSecret)
+	if err == nil {
+		return nil
+	}
+
+	if !errors.IsNotFound(err) {
+		log.Error(err, "Error getting TLS secret "+server.WebhookServerTLSSecretName)
```
sleshchenko, line 42: What is difference between `WebhookServerTLSSecretName` and `TLSSelfSignedCertificateSecretName` ?
Seems you tried to get `TLSSelfSignedCertificateSecretName` but reported issue about `WebhookServerTLSSecretName`

pkg/webhook/kubernetes/tls.go
```diff
@@ -0,0 +1,61 @@
...
+
+var log = logf.Log.WithName("webhook-k8s")
+
+// SetupSecureService handles TLS secrets required for deployment on Kubernetes.
+func SetupSecureService(client crclient.Client, ctx context.Context, namespace string) error {
+	devworkspaceSecret := &corev1.Secret{}
+
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: tls.TLSSelfSignedCertificateSecretName}, devworkspaceSecret)
+	if err == nil {
+		return nil
```
sleshchenko, line 38: Should we also check that service is in place to recover after unproper uninstallation?

pkg/webhook/service/service.go
```diff
@@ -10,13 +10,12 @@
 //   Red Hat, Inc. - initial API and implementation
 //
 
-package webhook
+package service
```
sleshchenko, line 13: With that dedicated package you solving cyclic import issue, right?

pkg/webhook/service/service.go
```diff
@@ -25,26 +24,14 @@ import (
...
-	err := createSecureService(client, ctx, namespace)
-	if err != nil {
-		log.Info("Failed creating the secure service")
-		return err
-	}
-	return nil
-}
-
-func createSecureService(client crclient.Client, ctx context.Context, namespace string) error {
+func CreateSecureService(client crclient.Client, ctx context.Context, namespace string, annotations map[string]string) error {
```
sleshchenko, line 27: That's not really create but make sure it exists in actual state... Create or update...

pkg/webhook/create.go
```diff
@@ -48,13 +51,48 @@ func SetupWebhooks(ctx context.Context, cfg *rest.Config) error {
...
+	log.Info("Creating the webhook server deployment")
+	err = CreateWebhookServerDeployment(client, ctx, namespace)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// setUpWebhookServerRBAC sets webhook blocking and required service account, cluster role, and cluster role binding
```
sleshchenko, line 93:
```suggestion
// setUpWebhookServerRBAC sets required service account, cluster role, and cluster role binding
```

pkg/kubernetes/tls/certs.go
```diff
@@ -0,0 +1,102 @@
...
+	RequesterName string
+	Domain        string
+	Namespace     string
+	TLSSecretName string
+	CASecretName  string
+}
+
+// TLS related constants
+const (
+	TLSSelfSignedCertificateSecretName = "devworkspace-self-signed-certificate"
```
sleshchenko, line 37: I think it makes sense to make it up to client (webhooks server, workspace routing) to decide where CA should be stored.

sleshchenko (commented):
I'll test it tomorrow

pkg/kubernetes/tls/certs.go
amisevsk: i think at 5 minutes we're in nightmare edge-case territory :)

pkg/webhook/service/service.go
JPinkney: Yeah

internal/cluster/job.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := client.Create(ctx, specJob); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		existingCfg, err := getJobInNamespace(client, specJob.GetName(), specJob.Namespace)
+		if err != nil {
+			return err
+		}
+		specJob.ResourceVersion = existingCfg.ResourceVersion
+		err = client.Update(ctx, specJob)
```
amisevsk, line 94: I'm still not sure this is the correct logic for an update, but I don't have a suggestion right now.

amisevsk (approved):
LGTM; tested on minikube.

pkg/webhook/kubernetes/tls.go
JPinkney: Not sure I understand what you mean, if the service exists then it's just going to be updated. Or did you mean something else?

sleshchenko: >  if the service exists then it's just going to be updated

Isn't it supposed to be updated below here https://github.com/devfile/devworkspace-operator/pull/159/files/d4e58c25e4d83adc706cfaaac3f8ec4587e015bb#diff-3dee34c4acc1e59056a87b123e78432aR46? 

If yes, then how it's going to be updated if we return here?

JPinkney: Yes, I understand that. I just don't understand what you mean by
> Should we also check that service is in place to recover after unproper uninstallation

As in check that the service exists and if it does delete before erroring? 

If it errors here there then the webhook server deployment just won't get setup

sleshchenko: I mean that there can be cases when TLSSelfSignedCertificateSecretName exists but service itself is removed.
and now it seems to be handled properly apart from fact that certs are going to be generated each time
So, code proposal instead of thousand words
https://github.com/devfile/devworkspace-operator/compare/kubernetes-cert-job...sleshchenko:kubernetes-cert-job-proposal2?expand=1

JPinkney: Ah okay that makes sense. Yeah I'll add that in

pkg/webhook/kubernetes/tls.go
```diff
@@ -0,0 +1,65 @@
...
+const (
+	TLSSelfSignedCertificateSecretName = "devworkspace-self-signed-certificate"
+)
+
+// SetupSecureService handles TLS secrets required for deployment on Kubernetes.
+func SetupSecureService(client crclient.Client, ctx context.Context, namespace string) error {
+	devworkspaceSecret := &corev1.Secret{}
+
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: server.WebhookServerTLSSecretName}, devworkspaceSecret)
+	if !errors.IsNotFound(err) {
```
sleshchenko, line 41: I constantly get `Error getting TLS secret ` and interesting part that the whole process does not fail because current err is nil :-D
I'm proving proposal as usual )
https://github.com/devfile/devworkspace-operator/compare/kubernetes-cert-job...sleshchenko:kubernetes-cert-job-proposal-3?expand=1

sleshchenko (commented):
I'm not able to finalize my testing today, will finish tomorrow.
In the proposal I've left TODO with investigating difference between ca.crt and tls.crt https://github.com/devfile/devworkspace-operator/compare/kubernetes-cert-job...sleshchenko:kubernetes-cert-job-proposal-3?expand=1
I've asked Deploy team to help figure out where cert gen source live, probably it could help us

sleshchenko:
Also, I somehow got state where cer-gen job neither its pod is not removed
 Screenshot_20200930_175810
Maybe we need to check if we need to clean up them even if secret exists

pkg/webhook/kubernetes/tls.go
```diff
@@ -0,0 +1,67 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var log = logf.Log.WithName("webhook-k8s")
+
+const (
+	WebhookServerCACertSecretName = "devworkspace-self-signed-certificate"
```
sleshchenko, line 33:
```suggestion
	WebhookServerCACertSecretName = "devworkspace-webhookserver-ca"
```

internal/cluster/job.go
sleshchenko: As more I test I become more confusing how k8s works :-D
After I removed pods removing, I got job related pob left without ownerRef:
```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2020-10-01T09:09:32Z"
  generateName: webhook-server-gen-cert-
  labels:
    app.kubernetes.io/component: cert-generator
    app.kubernetes.io/name: webhook-server-gen-cert
    controller-uid: 2cc52032-5a5c-4ede-a9b7-ec8cc7099125
    job-name: webhook-server-gen-cert
  name: webhook-server-gen-cert-7gq47
  resourceVersion: "10595"
  selfLink: /api/v1/namespaces/devworkspace-controller/pods/webhook-server-gen-cert-7gq47
  uid: 4c6bba20-c069-4533-8026-0db2e2a1e4e4
```
Seems like k8s or k8s-go-client bug

sleshchenko: Alright! https://thenewstack.io/deletion-garbage-collection-kubernetes-objects/
Orphan removal is the default for crclient? Really?!!! We should be careful with that.
So, we should do background removal, in my proposal I still left warning and cleanning up if pods are still there.

sleshchenko (commented):
Almost every my comment should be addressed in my proposal https://github.com/devfile/devworkspace-operator/compare/kubernetes-cert-job...sleshchenko:kubernetes-cert-job-final-review?expand=1
Except job updating when it already exists for some reason.

internal/cluster/pod.go
```diff
@@ -0,0 +1,62 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CleanupPods takes a job and cleans up all the pods associated with that job
+func CleanupPods(client crclient.Client, job *batchv1.Job) error {
+	selector := fmt.Sprintf("job-name=%s", job.GetName())
```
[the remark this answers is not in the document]
amisevsk: I'm not sure we need to have a generic cleanupPods function -- it's adding generality without having a real reason for it. We currently should never be deleting pods directly (at least, that I can think of), so we only gain complexity here.

internal/cluster/job.go
```diff
@@ -0,0 +1,109 @@
...
+	err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, job)
+	if err != nil {
+		return job, err
+	}
+	return job, nil
+}
+
+// deleteJob deletes a given job and cleans up any pods associated with it
+func deleteJob(c crclient.Client, job *batchv1.Job) error {
+	deleteForeground := metav1.DeletePropagationBackground
```
amisevsk, line 55: `deleteBackground := ...` :)

internal/cluster/job.go
```diff
@@ -0,0 +1,109 @@
...
+// Wait for the job to complete. Times out if the job isn't complete after $(timeout) seconds
+func WaitForJobCompletion(client crclient.Client, name string, namespace string, timeout time.Duration) error {
+	const interval = 1 * time.Second
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		job, err := getJobInNamespace(client, name, namespace)
+		if err != nil {
+			return false, err
+		}
+
+		if job.Status.Succeeded > 0 {
```
amisevsk, line 86: Probably a future improvement, but I wonder if it's worth it to check `job.Status.Failed > 0` as well and log a line mentioning it.

internal/cluster/job.go
```diff
@@ -0,0 +1,109 @@
...
+		return false, nil
+	})
+}
+
+func SyncJobToCluster(
+	client crclient.Client,
+	ctx context.Context,
+	specJob *batchv1.Job,
+) error {
+	// Attempt to clean up the job before re-creating it. If a job didn't exist then it will throw isNotFound error
```
amisevsk, line 98: We should include this doc line on `CleanJob` as well (`Returns isNotFound error if job doesn't exist`)

internal/cluster/job.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	err = deleteJob(client, job)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// getJobInNamespace finds a job with a given name in a namespace
+func getJobInNamespace(client crclient.Client, name string, namespace string) (*batchv1.Job, error) {
```
amisevsk, line 44: I prefer exported functions to be listed first in the file, for readability.

amisevsk (approved):
A few minor suggestions but LGTM.

internal/cluster/pod.go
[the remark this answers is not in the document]
sleshchenko: Agree, there is no real reason to keep pods cleaning up. Let's remove it

internal/cluster/job.go
sleshchenko: Ah, it's not easy to deceive you!

internal/cluster/job.go
```diff
@@ -0,0 +1,111 @@
...
+func deleteJob(c crclient.Client, job *batchv1.Job) error {
+	deleteBackground := metav1.DeletePropagationBackground
+	err := c.Delete(context.TODO(), job, &crclient.DeleteOptions{PropagationPolicy: &deleteBackground})
+	if err != nil {
+		log.Error(err, "Error deleting job: "+job.Name)
+		return err
+	}
+	//we don't use job.Selector here since it's autogenerated and contains only generated controller-uid unique for each job instance
+	//if we use job-name which is also generated label but not used in selector - we clean up pods from old jobs
+	pods, err := GetPodsBySelector(c, job.Namespace, "job-name="+job.Name)
```
sleshchenko, line 99: That's up to you if we need to keep that logic or just remove cleaning up job-related pods.

sleshchenko (approved):
I did not retest it but code LGTM

pkg/webhook/create.go
```diff
@@ -78,6 +78,7 @@ func SetupWebhooks(ctx context.Context, cfg *rest.Config) error {
 		if err != nil {
 			return err
 		}
+		log.Info("The provided webhook server cert is not production ready. If you want to use this in production please set up certs with a certification manager")
```
amisevsk, line 81: Might me misleading to suggest using a cert manager when we have no way of consuming those certs (AFAIK), but oh well.
```suggestion
		log.Info("Warning: the webhook server cert in use is not suitable for production. If you want to use this in production please set up certs with a certificate manager")
```

amisevsk (approved):

pkg/webhook/create.go
JPinkney: Yeah, I'm not sure what to do here either

Issue #168, opened 2020-10-02
Make basic routing working again


sleshchenko:
To move forward faster with an ability to use devworkspace controller as workspace engine in chectl, it's needed to provide some quick solution to make basic routing working again, it's needed:
- patch THEIA_HOST from plugin registry meta.yaml to set 0.0.0.0: https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-theia/7.19.2/meta.yaml;
- patch che-machine exec to listen to 0.0.0.0 https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-machine-exec-plugin/7.19.2/meta.yaml;
- switch on TLS for basic routing (if it's still not enabled);
-* unique host may be also needed for webviews to work but that's as best afford

Later these hacks should be replaced with proper routing with authentication enabled.

sleshchenko:
basic routing works again after https://github.com/devfile/devworkspace-operator/pull/169 except Theia Webview feature.

PR #169, opened 2020-10-05
Make basic routing working again

sleshchenko:
### What does this PR do?
This PR hack the code to make basic routing working somehow again.

https://github.com/devfile/devworkspace-operator/issues/168

Limitations:
- Theia webview does not work because of TLS leak + unique host requirement;
- ~~Containers view is empty, probably because of incompatibility with Che Rest API; Containers view populates but components aren't marked as running~~
- ~~Terminal is not working, `connect` websocket connection is closed because of 401 unauthorized;~~

### Is it tested? How?
```
export IMG=sleshchenko/che-workspace-controller:basic-routing
export WEBHOOK_ENABLED=false
export REGISTRY_ENABLED=true
make deploy
kubectl apply -f ./samples/theia.yaml
kubectl get devworkspace -w # & and wait until workspace is running
# Open a workspace URL
```
Screenshot_20201007_143949

### DIFF
--- a/deploy/controller_config.yaml
+++ b/deploy/controller_config.yaml
@@ -15,3 +15,5 @@ data:
   # image pull policy that is applied to all workspace's containers
   devworkspace.sidecar.image_pull_policy: Always
   devworkspace.api_sidecar.image: quay.io/che-incubator/che-api-sidecar:next
+  # Do not enable it in production!!!
+  devworkspace.experimental_features_enabled: "true"
--- a/deploy/controller_config.yaml
+++ b/deploy/controller_config.yaml
@@ -14,3 +14,4 @@ data:
   devworkspace.default_routing_class: "basic"
   # image pull policy that is applied to all workspace's containers
   devworkspace.sidecar.image_pull_policy: Always
+  devworkspace.api_sidecar.image: quay.io/che-incubator/che-api-sidecar:next
--- a/pkg/adaptor/plugin.go
+++ b/pkg/adaptor/plugin.go
@@ -17,6 +17,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/devfile/devworkspace-operator/pkg/adaptor/plugin_patch"
+
 	devworkspace "github.com/devfile/api/pkg/apis/workspaces/v1alpha1"
 	"github.com/devfile/devworkspace-operator/pkg/apis/controller/v1alpha1"
 
@@ -47,24 +49,8 @@ func AdaptPluginComponents(workspaceId, namespace string, devfileComponents []de
 	}
 
 	for _, plugin := range plugins {
-		if plugin.Name == "che-machine-exec-plugin" {
-			for _, value := range plugin.Containers {
-				for i, command := range value.Command {
-					if strings.Contains(command, "127.0.0.1:4444") {
-						value.Command[i] = "0.0.0.0:4444"
-					}
-				}
-			}
-		}
-
-		if plugin.Name == "che-theia" {
-			for _, container := range plugin.Containers {
-				for i, env := range container.Env {
-					if env.Name == "THEIA_HOST" {
-						container.Env[i].Value = "0.0.0.0"
-					}
-				}
-			}
+		if config.ControllerCfg.GetExperimentalFeaturesEnabled() {
+			plugin_patch.PublicAccessPatch(&plugin)
 		}
 
 		component, err := adaptChePluginToComponent(workspaceId, plugin)
--- a/pkg/adaptor/plugin.go
+++ b/pkg/adaptor/plugin.go
@@ -52,6 +52,7 @@ func AdaptPluginComponents(workspaceId, namespace string, devfileComponents []de
 		if config.ControllerCfg.GetExperimentalFeaturesEnabled() {
 			plugin_patch.PublicAccessPatch(&plugin)
 			plugin_patch.PatchMachineSelector(&plugin, workspaceId)
+			plugin_patch.AddMachineNameEnv(&plugin)
 		}
 
 		component, err := adaptChePluginToComponent(workspaceId, plugin)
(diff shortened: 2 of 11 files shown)


benoitf:
it's not related directly to this PR but can we use reference: in the meta.yaml (plugin/reference instead of plugin/id)

deploy/controller_config.yaml
```diff
@@ -14,3 +14,4 @@ data:
   devworkspace.default_routing_class: "basic"
   # image pull policy that is applied to all workspace's containers
   devworkspace.sidecar.image_pull_policy: Always
+  devworkspace.api_sidecar.image: quay.io/che-incubator/che-api-sidecar:next
```
amisevsk, line 17: ++++1, I always forget to add it when deploying :)

pkg/adaptor/plugin.go
```diff
@@ -47,6 +47,26 @@ func AdaptPluginComponents(workspaceId, namespace string, devfileComponents []de
...
+		if plugin.Name == "che-theia" {
+			for _, container := range plugin.Containers {
+				for i, env := range container.Env {
+					if env.Name == "THEIA_HOST" {
+						container.Env[i].Value = "0.0.0.0"
+					}
+				}
+			}
+		}
+
```
amisevsk, line 69: Should we delimit this section with comments to note that it's a temporary solution? We'll eventually need to remove this code.

amisevsk (approved):
At first glance looks good to me; I'll build + test code later.

pkg/adaptor/plugin.go
sleshchenko: I'll move it to a separate file and add comments that it's temporary solution

deploy/controller_config.yaml
JPinkney: +1 as well, I kept having to add this when I was testing the kubernetes cert gen pr

JPinkney (approved):
LGTM, I like the experimental feature toggle as well

amisevsk:
From testing, it seems the second limitation,

> Containers view is empty, probably because of incompatibility with Che Rest API

Is actually because the added devfile doesn't have any container components; adding a container component copied from `theia-nodejs.yaml` works mostly as expected, though the icons for each component are grey instead of green (to signify running). This is an issue I ran into in the past.

amisevsk (approved):
I've tested again on minikube with webhooks disabled and was able to get everything to work. I've opened a PR into this branch that address two of the three listed limitations here: https://github.com/sleshchenko/devworkspace-operator/pull/2

Regarding TLS support, I think it's best to separate that out into another issue/PR. Since it'll potentially be easier to do once https://github.com/devfile/devworkspace-operator/pull/159 is merged, and it'll require a bit of work making sure every component picks up the injected certs on OpenShift (we had to patch machine-exec to get it to work there, IIRC).

pkg/config/property.go
```diff
@@ -47,6 +47,9 @@ const (
 	routingSuffix        = "devworkspace.routing.cluster_host_suffix"
 	defaultRoutingSuffix = ""
 
+	experementalFeaturesEnabled        = "devworkspace.experimental_features_enabled"
```
benoitf, line 50:
```suggestion
	experimentalFeaturesEnabled        = "devworkspace.experimental_features_enabled"
```

pkg/config/config.go
```diff
@@ -74,6 +74,13 @@ func (wc *ControllerConfig) GetPluginRegistry() string {
 	return wc.GetPropertyOrDefault(pluginRegistryURL, "")
 }
 
+//GetExperimentalFeaturesEnabled returns true if experimental features should be enabled.
+//DO NOT TURN ON IT IN THE PRODUCTION.
+//Experimental features are not well tested and may be totally removed without announcement.
+func (wc *ControllerConfig) GetExperimentalFeaturesEnabled() bool {
+	return wc.GetPropertyOrDefault(experementalFeaturesEnabled, defaultExperimentalFeaturesEnabled) == "true"
```
benoitf, line 81:
```suggestion
	return wc.GetPropertyOrDefault(experimentalFeaturesEnabled, defaultExperimentalFeaturesEnabled) == "true"
```

sleshchenko:
Tested on minikube and works as expected.
Going to test on crc and will merge when PR checks are done

sleshchenko:
> it's not related directly to this PR but can we use reference: in the meta.yaml (plugin/reference instead of plugin/id)

@benoitf AFAIS it's not implemented yet. I'll give a quick look if it's easy to implement and will provide a separate PR

sleshchenko:
Tested on crc with the described scenario from PR description
<details>
<summary>Screenshot</summary>

Screenshot_20201007_145458

</details>

pkg/adaptor/plugin_patch/public_access.go
```diff
@@ -0,0 +1,45 @@
...
+
+package plugin_patch
+
+import (
+	"strings"
+
+	"github.com/eclipse/che-plugin-broker/model"
+)
+
+// PublicAccessPatch pathes plugin's configuration to make endpoints publicly available
```
JPinkney, line 21: nitpick: `patches` instead of `pathes`

sleshchenko: eagle eye :eagle:

sleshchenko:
@benoitf I'm happy to inform that using plugin from reference works fine and I just missed that previously.
<details>
<summary>Tested devfile</summary>

```yaml
kind: DevWorkspace
metadata:
  name: theia
spec:
  started: true
  routingClass: 'basic'
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - plugin:
          uri: http://che-plugin-registry-devworkspace-controller.apps-crc.testing/v3/plugins/eclipse/che-theia/latest/meta.yaml
      - plugin:
    commands:
      - exec:
          id: say hello
          component: plugin

```

</details>

PR #172, opened 2020-10-09
Add the initial proposal for devworkspace authentication

sleshchenko:
### What does this PR do?
It's draft since it's not going to be merged here.
This draft is here for more convenient discussion and collaboration.
So, it adds the initial proposal for devworkspace authentication;

For good picture see [README.md](https://github.com/devfile/devworkspace-operator/blob/devworkspace-auth/auth-proposal/README.md)
for commenting use [file changes](https://github.com/devfile/devworkspace-operator/pull/172/files)

https://github.com/eclipse/che/issues/17856

### Is it tested? How?
No way ) We'll test after we agree and implement everything


auth-proposal/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+So, we propose not to satisfy that requirement on any Kubernetes cluster. Then we can propose the following abilities for authentication:
+
+## Basic HTTPs Routing
+
+It means that we don't provide any authentication on the workspace endpoints. Everything that we need here to support it: implement/provision passthrough proxy container into workspace pod, to make workspace server available since they are supposed to listen to 127.0.0.0 only.
+
+* for simplification we can rely on default certificates configured on the Ingress controller side if we have access to CA;
+
+### Basic Authentication HTTPs Routing
```
sleshchenko, line 24: It's a simple solution but UX is not really great. Alternatively we are going to investigate a bit more OpenID Connect https://github.com/devfile/devworkspace-operator/issues/174 which could be used k8s cluster where it's enabled.

musienko-maxim:
test

JPinkney:
I was looking up easier ways we could do this with kubernetes and it looks like https://kubernetes.github.io/ingress-nginx/examples/auth/oauth-external-auth/ provides an interesting solution. Since we are currently using ingresses with nginx what we might be able to do is add an annotation that automatically redirects all ingresses we create to something like oauth2_proxy or dex when a user tries to access them. That way when someone tries to access a workspace related ingress they're automatically redirected to a login page. Then the user would login in then be redirected to the workspace. Technically this means we would have to limit the ingresses to only being nginx though unless other ingress controllers have that feature.

sleshchenko:
> https://kubernetes.github.io/ingress-nginx/examples/auth/oauth-external-auth/

briefly took a look at that. Benefits it could provide:
- authentication server can be set independently from the controller or even devworkspace routing controller. The controller is just configured where the authentication server is located.

Cons:
- devworkspace in-cluster services are not secure, everybody can request any devworkspace endpoint without authentication;
- authorization is missing. So, devworkspace endpoints(che-machine-exec, che-theia) need to do it on its own, evaluate if logged in user should be able to access this devworkspace endpoint or not;

So, https://github.com/sleshchenko/devworkspace-operator/tree/authentication-schemas/authentication-schemas#devworkspaces-servers-authentication-with-openid seems to be a bit better if we build our a bit custom oauth proxy with embedded authorization.

l0rd:
@sleshchenko @JPinkney what about using oauth-external-auth with https://github.com/brancz/kube-rbac-proxy?

sleshchenko:
I think the main reason we did not investigate a lot about kube-rbac proxy:
1. It supports only proxying one port (the same limitations are in oauth2-proxy and openshift oauth proxy).
2. It does not provide a complete workflow for browser clients as oauth2-proxy or openshift oauth proxy provides, where user gets redirect to auth provider when user is not authenticated.
3. kuberbac proxy requires oidc be set on the k8s cluster, which might be OK but still seems to be a heavy prerequisite.

But it's still in our attention area since potentially we can use it if we fork it or contribute there features we miss.

sleshchenko:
It's not tested for a while. Work can be reused if it makes sense from the PR diff

PR #173, opened 2020-10-09
Using cert manager for webhook certs

JPinkney:
### What does this PR do?
This PR introduces cert manager use for webhook certs on kubernetes.

The way this works is a cert manager certificate needs to be created in the same namespace with the name devworkspace-webhook-certificate. This certificate then needs to configured to create a secret named devworkspace-webhookserver-tls.
Then inside of the controller if there is already a devworkspace-webhookserver-tls secret by the time the controller is looking for certs then the cert generation Job WON'T be activated and devworkspace-webhookserver-tls will only contain data
from cert-manager. Then, on the webhook server deployment side the validating and mutating webhooks will be annotated
with `cert-manager.io/inject-ca-from` and inject devworkspace-webhook-certificate CABundles. These CABundles will
automatically be rotated when they are supposed to be renewed.

On kubernetes if cert-manager is not available it will use the cert generation job

On openshift it still uses the secure service.

### Is it tested? How?

Tested both on kubernetes and openshift. For minikube:
```bash
# First install cert-manager
kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v1.0.2/cert-manager.yaml
# Then build and test PR changes
export IMG={my_ctrl_img}
# Make sure that RELATED_IMAGE_devworkspace_webhook_server is set to $IMG so the webhook changes are picked up
make uninstall; make docker; make deploy;
```

For openshift:
```bash
make uninstall; make docker; make deploy
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ SHELL := bash
 OPERATOR_SDK_VERSION = v0.17.0
 NAMESPACE ?= devworkspace-controller
 IMG ?= quay.io/devfile/devworkspace-controller:next
-TOOL ?= oc
+PLATFORM ?= openshift
 ROUTING_SUFFIX ?= 192.168.99.100.nip.io
 PULL_POLICY ?= Always
 WEBHOOK_ENABLED ?= true
@@ -30,9 +30,11 @@ BUMPED_KUBECONFIG=$(INTERNAL_TMP_DIR)/kubeconfig
 RELATED_IMAGES_FILE=$(INTERNAL_TMP_DIR)/environment
 
 # minikube handling
-ifeq ($(shell $(TOOL) config current-context),minikube)
+ifeq ($(PLATFORM),kubernetes)
 	ROUTING_SUFFIX := $(shell minikube ip).nip.io
 	TOOL := kubectl
+else ifeq ($(PLATFORM),openshift)
+	TOOL := oc
 endif
 
 all: help
@@ -47,7 +49,7 @@ _print_vars:
 	@echo "    DEFAULT_ROUTING=$(DEFAULT_ROUTING)"
 	@echo "    REGISTRY_ENABLED=$(REGISTRY_ENABLED)"
 	@echo "    DEVWORKSPACE_API_VERSION=$(DEVWORKSPACE_API_VERSION)"
-	@echo "    TOOL=$(TOOL)"
+	@echo "    PLATFORM=$(PLATFORM)"
 
 _set_ctx:
 ifneq ($(ADMIN_CTX),"")
@@ -79,7 +81,7 @@ endif
 endif
 
 _set_registry_url:
-ifeq ($(TOOL),oc)
+ifeq ($(PLATFORM),openshift)
 	$(eval PLUGIN_REGISTRY_HOST := $(shell $(TOOL) get route che-plugin-registry -n $(NAMESPACE) -o jsonpath='{.spec.host}' || echo ""))
 else
 	$(eval PLUGIN_REGISTRY_HOST := $(shell $(TOOL) get ingress che-plugin-registry -n $(NAMESPACE) -o jsonpath='{.spec.rules[0].host}' || echo ""))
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -205,11 +205,22 @@ docker: _print_vars
 	docker build -t $(IMG) -f ./build/Dockerfile .
 	docker push $(IMG)
 
+### gen_webhook: generate cert manager certificates for kubernetes webhooks and deploy to cluster; no-op if running on OpenShift
+gen_webhook:
+ifeq ($(WEBHOOK_ENABLED),true)
+ifeq ($(TOOL),kubectl)
+	$(eval IS_NOT_FOUND := $(shell kubectl get namespace cert-manager 2>&1 | grep -q "not found"; echo $$?))
+	@if [ "$(IS_NOT_FOUND)" = "0" ]; then echo "Cert Manager has not been found."; else kubectl apply -f deploy/k8s_webhook/cert_manager.yaml; fi
```
amisevsk, line 213: This makes our makefile single-namespace again; I think we need something like 
```bash
sed -i.bak -e "s|namespace: .*|namespace: $(NAMESPACE)|" \
           -e "s|devworkspace-webhookserver.devworkspace-controller.svc|devworkspace-webhookserver.$(NAMESPACE).svc|' \
            deploy/k8s_webhook/cert_manager.yaml
kubectl apply -f deploy/k8s_webhook/cert_manager.yaml -n $(NAMESPACE)
mv deploy/k8s_webhook/cert_manager.yaml.bak deploy/k8s_webhook/cert_manager.yaml
```
which is a pain (as usual).

I think we should also split this onto multiple lines; I'm not fully sure the MacOS `make` version won't cause issues but if it does we should just do line continuations (`\`)

amisevsk (commented):
Generally looks good, I'll test directly later today or tomorrow.

pkg/webhook/kubernetes/tls.go
```diff
@@ -54,14 +56,32 @@ func SetupSecureService(client crclient.Client, ctx context.Context, namespace s
...
+			return nil
 		}
 	}
 
-	err = service.CreateOrUpdateSecureService(client, ctx, namespace, map[string]string{})
+	// Check if the secret found at server.WebhookServerTLSSecretName is a cert manager secret. If it is then we can
+	// just inject the ca automatically using cert manager annotations
+	secureServiceAnnotations := make(map[string]string)
+	if isCertManagerSecret(devworkspaceSecret.GetAnnotations()) {
+		secureServiceAnnotations["cert-manager.io/inject-ca-from-secret"] = server.WebhookServerTLSSecretName
```
sleshchenko, line 68: Could you elaborate more on why we need it? I assume that this annotation should be put info webhook configurations if we don't want to propagate it from WebhookServerTLSSecretName manually, as we currently do on webhook server side.

sleshchenko: I think it's redundant here. Everything that we may be interested to do with `WebhookServerTLSSecretName` secret - check that needed keys are in place, cert-manager provides the following:
Screenshot_20201013_131835
but we require only tls.key and tls.crt

scripts/webhook.sh
```diff
@@ -0,0 +1,45 @@
...
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    '-n'|'--namespace') NAMESPACE="$2"; shift 1;;
+    '-h'|'--help') printHelp="true"; shift 0;;
+  esac
+  shift 1
+done
+
+if [[ "$printHelp" == "true" ]]; then
+  echo "Usage:"
```
sleshchenko, line 24: I think it makes sense to make it clearer that this script is for k8s-webhooks-certs via cert-manager.

webhook/workspace/config.go
```diff
@@ -43,8 +42,13 @@ func Configure(ctx context.Context) error {
 		return err
 	}
 
-	mutateWebhookCfg := BuildMutateWebhookCfg(namespace)
-	validateWebhookCfg := buildValidatingWebhookCfg(namespace)
+	webhookAnnotations, err := GetWebhookAnnotations(c, namespace)
```
sleshchenko, line 45:
```suggestion
	webhookAnnotations, err := GetWebhookCfgAnnotations(c, namespace)
```

sleshchenko (commented):
I'll test it later

sleshchenko (commented):
With this PR cert-manager related logic is spread over different places.
I think it's better to try to refactor it in one place. It's not finalized but I hope it will make my idea more clearer https://github.com/devfile/devworkspace-operator/compare/cert-manager-webhook-k8s...sleshchenko:cert-manager-webhook-k8s-proposal?expand=1

The alternative simpler approach we could go - just refuse the idea to use `cert-manager.io/inject-ca-from`. If we don't use it on openshift, why we need it on kubernetes?... 
We can just declare that we rely on the actual state of the secret with certain name, and then we propagate it to places where it's needed by our own... It would guarantee the consistency between cert used by webhook server and value in webhook cfg.

sleshchenko:
I believe that's already implemented but with kustomize

PR #177, opened 2020-10-20
Derive Web Terminal Tooling image from OpenShift version

JPinkney:

### What does this PR do?
This PR makes it so that you can have different tooling versions on different versions of OpenShift. With this change we can have RELATED_IMAGE_web_terminal_tooling_4_5 that will only be picked up on a 4.5 cluster and RELATED_IMAGE_web_terminal_tooling_4_6 that will only be picked up on a 4.6 cluster

### DIFF
--- a/internal/cluster/openshift.go
+++ b/internal/cluster/openshift.go
@@ -1,0 +1,58 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package cluster
+
+import (
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+)
+
+// Create a new openshift client using the in cluster config
+func openShiftClient() (*configv1client.ConfigV1Client, error) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	if config != nil {
+		client, err := configv1client.NewForConfig(config)
+		if err != nil {
+			return client, err
+		}
+		return client, nil
+	}
+	return nil, nil
+}
+
+// Get the version of the current running OpenShift cluster in semver format E.g. 4.5.9
+func OpenshiftVersion() (string, error) {
+	client, err := openShiftClient()
+	if err != nil {
+		return "", err
+	}
+
+	openshiftAPIServer, err := client.ClusterOperators().Get("openshift-apiserver", metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	if openshiftAPIServer != nil {
+		for _, ver := range openshiftAPIServer.Status.Versions {
+			if ver.Name == "operator" {
+				// Apparently only openshift-apiserver clusteroperator reports the version number
+				return ver.Version, nil
+			}
+		}
+	}
+	return "", nil
+}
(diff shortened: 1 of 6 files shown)


internal/cluster/openshift.go
```diff
@@ -0,0 +1,58 @@
...
+package cluster
+
+import (
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+)
+
+// Create a new openshift client using the in cluster config
+func openshiftClient() (*configv1client.ConfigV1Client, error) {
```
sleshchenko, line 22: let's use OpenShift word

sleshchenko: would not it be more specific if we use like configClient, clusterConfigClient?

sleshchenko (commented):
I'm curious if we need to use master for WTO 1.1. We (I) already contributed some hacks to support basic routing which I'm not sure should appear in WTO even if that code should not be activated by default. Let's discuss it.

internal/cluster/openshift.go
JPinkney: I just left it as lowercase because I don't think it really needs to be used by anything else. The only reason it's used is because the kubernetes client-go doesn't have access to clusteroperators

sleshchenko: I was not clear and meant openShift

internal/images/image.go
```diff
@@ -52,9 +55,30 @@ func GetWebhookServerImage() string {
 }
 
 // GetWebTerminalToolingImage returns the image reference for the default web tooling image. Returns
-// the empty string if environment variable RELATED_IMAGE_web_terminal_tooling is not defined
+// the empty string if environment variable RELATED_IMAGE_web_terminal_tooling is not defined. On OpenShift
```
sleshchenko, line 58:
```suggestion
// the empty string if environment variable RELATED_IMAGE_web_terminal_tooling is not defined or failed to evaluate cluster version. On OpenShift
```

sleshchenko (approved):

internal/cluster/openshift.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	if config != nil {
+		client, err := configv1client.NewForConfig(config)
+		if err != nil {
+			return client, err
+		}
+		return client, nil
+	}
+	return nil, nil
```
amisevsk, line 35: I don't think `rest.InClusterConfig()` returns `nil` if `err` is `nil` -- we shouldn't need the check `config != nil` (or if we do we should return an error here.

This leaves `OpenshiftVersion()` open to a nil error below, since we don't check that `client` is nil.

internal/cluster/openshift.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Get the version of the current running OpenShift cluster in semver format E.g. 4.5.9
+func OpenshiftVersion() (string, error) {
+	client, err := openShiftClient()
+	if err != nil {
+		return "", err
+	}
+
+	openshiftAPIServer, err := client.ClusterOperators().Get("openshift-apiserver", metav1.GetOptions{})
```
amisevsk, line 45: I feel like there should be a better way to get the cluster version -- e.g. `oc version` should need to read the cluster operator (and limited users wouldn't be able to do this anyways), but maybe I'm wrong.

amisevsk (approved):

internal/cluster/openshift.go
JPinkney: Yeah it's quite a pain. I actually found this approach by reading through how `oc version` computes their client/server versions

amisevsk (changes requested):
Looking forward, we should handle versioning more carefully -- what happens if I deploy WTO on a 4.7 cluster, before the version that specifies the 4.7 RELATED_IMAGE env var?

We should have some defaulting to a latest or oldest version of the tooling image to make sure we can always start a web terminal (even if the version doesn't match perfectly).

amisevsk (approved):

tinakurian (approved):

sleshchenko:
@JPinkney Do you think this PR has a future?

Issue #180, opened 2020-10-21
Update the DevWorkspace Controller to Operator SDK 1.0


sleshchenko:
Update the DevWorkspace Controller to Operator SDK 1.0 (compatible with the new kubebuilder)

PR #186, opened 2020-10-27
Update sdk 1.x

amisevsk:
### What does this PR do?
Update to Operator SDK 1.1.

This process requires bootstrapping a new project using kubebuilder and porting files over to the new structure. See [this repo](https://github.com/amisevsk/devworkspace-controller-migration) for a more detailed commit history (though it's still a bit messy and hard to read). I generally tried to avoid making unnecessary changes, to make the transition easier, so the configmap is still used, webhooks are done as they are currently, etc.

Main directory structure changes:
- `deploy/ => config/`
    - The deployment yaml files are now defined using kustomize. 
    - The "main" deployment is built from `config/default`
    - development deployments are generated from `config/devel/`
    - CRDs are stored in `config/crds/`
    - Pieces of the deployment are defined in `config/components/` and combined
    - Plugin registry is defined in `config/registry`
    - RBAC files for the controller are generated from annotations on `*_controller.go` files

- `pkg/apis/ => apis/`
    - subdirectores in `apis/` are named after the resource's group (`controller` in our case)
- `pkg/controller/ => controllers`
    - subdirectories in `controllers` are named for the group of the CR they control (`controller` for `Components`, `WorkspaceRoutings`; `workspace` for `DevWorkspaces`)
    - the `controllers/controller` subdirectory is further divided to distinguish between the controller for Components and WorkspaceRoutings.
- `pkg` now contains mostly helper code from `pkg` in the current repo

### Other notable differences
- No longer have a hard dependency on operator-sdk code.
- Makefile had to be reworked significantly; there might be gaps in functionality.
- Setting up controllers is done differently now which makes some things harder (watching a configmap)
- To simplify importing CRDs for `devfile/api`, I reworked the script to grab the files. It now downloads to a temp dir and copies definitions to the expected directory (`config/crds`). The copied files are gitignored.

### Further TODOs
- Remove the configmap-related code from the controller and use environment variables instead, or figure out proper way to configure an operator.
- **Depends on changes to `devfile/api`**, reflected by changes [here](https://github.com/devfile/api/compare/v1alpha1...amisevsk:v1alpha1-controller-runtime-compat). This is set in `go.mod` as a replace.
- I didn't look into adapting `e2e` tests.

Resolves https://github.com/devfile/devworkspace-operator/issues/180

**We'll likely need to do additional fixes to finalize this work**

### Is it tested? How?
Tested on minikube; **still need to test WTO and `crc` use cases**


amisevsk:
Closing in favor or https://github.com/devfile/devworkspace-operator/pull/187, since that branch is pushed to this repo instead of my fork.

PR #187, opened 2020-10-28
Update to Operator SDK 1.1

amisevsk:
Note this PR is a re-do of https://github.com/devfile/devworkspace-operator/pull/186 based off a branch in this repo.

### What does this PR do?
Update to Operator SDK 1.1.

This process requires bootstrapping a new project using kubebuilder and porting files over to the new structure. See [this repo](https://github.com/amisevsk/devworkspace-controller-migration) for a more detailed commit history (though it's still a bit messy and hard to read). I generally tried to avoid making unnecessary changes, to make the transition easier, so the configmap is still used, webhooks are done as they are currently, etc.

Main directory structure changes:
- `deploy/ => config/`
    - The deployment yaml files are now defined using kustomize. 
    - The "main" deployment is built from `config/default`
    - development deployments are generated from `config/devel/`
    - CRDs are stored in `config/crds/`
    - Pieces of the deployment are defined in `config/components/` and combined
    - Plugin registry is defined in `config/registry`
    - RBAC files for the controller are generated from annotations on `*_controller.go` files

- `pkg/apis/ => apis/`
    - subdirectores in `apis/` are named after the resource's group (`controller` in our case)
- `pkg/controller/ => controllers`
    - subdirectories in `controllers` are named for the group of the CR they control (`controller` for `Components`, `WorkspaceRoutings`; `workspace` for `DevWorkspaces`)
    - the `controllers/controller` subdirectory is further divided to distinguish between the controller for Components and WorkspaceRoutings.
- `pkg` now contains mostly helper code from `pkg` in the current repo

### Other notable differences
- No longer have a hard dependency on operator-sdk code.
- Makefile had to be reworked significantly; there might be gaps in functionality.
- Setting up controllers is done differently now which makes some things harder (watching a configmap)
- To simplify importing CRDs for `devfile/api`, I reworked the script to grab the files. It now downloads to a temp dir and copies definitions to the expected directory (`config/crds`). The copied files are gitignored.

### Further TODOs
- Remove the configmap-related code from the controller and use environment variables instead, or figure out proper way to configure an operator.
- **Depends on changes to `devfile/api`**, reflected by changes [here](https://github.com/devfile/api/compare/v1alpha1...amisevsk:v1alpha1-controller-runtime-compat). This is set in `go.mod` as a replace.
- I didn't look into adapting `e2e` tests.

Resolves https://github.com/devfile/devworkspace-operator/issues/180

**We'll likely need to do additional fixes to finalize this work**

### Is it tested? How?
Tested on minikube; **still need to test WTO and `crc` use cases**


build/rhel.Dockerfile
```diff
@@ -1,38 +1,43 @@
...
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
-# copy the rest of the sources code
+# Copy the go source
 COPY . .
+
 # compile workspace controller binaries
 RUN CGO_ENABLED=0 GOOS=linux go build \
```
JPinkney, line 19: Should we also include GOARCH=amd64 GO111MODULE=on here as well? and same with line 26?

config/components/manager/manager.yaml
```diff
@@ -0,0 +1,69 @@
...
+          - name: RELATED_IMAGE_plugin_redhat_developer_web_terminal_dev_nightly
+            value: "quay.io/eclipse/che-machine-exec:nightly"
+          - name: RELATED_IMAGE_plugin_eclipse_cloud_shell_nightly
+            value: "quay.io/eclipse/che-machine-exec:nightly"
+          - name: RELATED_IMAGE_web_terminal_tooling
+            value: "quay.io/wto/web-terminal-tooling:latest"
+          - name: RELATED_IMAGE_openshift_oauth_proxy
+            value: "openshift/oauth-proxy:latest"
+          - name: RELATED_IMAGE_devworkspace_webhook_server
+            value: quay.io/devfile/devworkspace-controller:next
```
JPinkney, line 67: It's not required but if you quote this then we wouldn't need https://github.com/devfile/devworkspace-operator/pull/178 (and they would conflict anyways)

config/components/prometheus/monitor.yaml
```diff
@@ -0,0 +1,16 @@
+
```
JPinkney, line 1: nitpick: spacing, not sure if this is auto-gen or not. Feel free to ignore

config/components/webhook/service.yaml
```diff
@@ -0,0 +1,12 @@
+
```
JPinkney, line 1: nitpick: spacing

config/components/rbac/role.yaml
```diff
@@ -1,163 +1,180 @@
+
```
JPinkney, line 1: nitpick: spacing

config/devel/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+bases:
+- ../default
+
+commonLabels:
```
JPinkney, line 4: I'm not exactly sure how kustomize works but if it uses the base of `../default` do we still need to define commonLabels in here? Because it looks like they're also defined there as well

config/devel/manager_image_patch.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Deployment
+metadata:
+  name: manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: devworkspace-controller
+        image: docker.io/amisevsk/che-workspace-controller:dev
```
JPinkney, line 11: Should this image be quay.io/devworkspace/devworkspace-controller:next?

controllers/workspace/devworkspace_controller.go
```diff
@@ -394,3 +307,35 @@ func getWorkspaceId(instance *devworkspace.DevWorkspace) (string, error) {
 	}
 	return "workspace" + strings.Join(strings.Split(uid.String(), "-")[0:3], ""), nil
 }
+
+func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// TODO: Set up indexing https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup
```
JPinkney, line 312: Should we create issues for these improvements

update_devworkspace_crds.sh
```diff
@@ -10,29 +10,33 @@
...
+	git checkout --quiet "${DEVWORKSPACE_API_VERSION}" && git reset --hard origin/"${DEVWORKSPACE_API_VERSION}"
 else
 	echo 'DevWorkspace API is specified from revision'
-	git checkout "${DEVWORKSPACE_API_VERSION}"
+	git checkout --quiet "${DEVWORKSPACE_API_VERSION}"
 fi
+cp deploy/crds/workspace.devfile.io_devworkspaces_crd.yaml \
+   $SCRIPT_DIR/config/crd/bases/workspace.devfile.io_devworkspaces.yaml
 
+cd $SCRIPT_DIR
```
JPinkney, line 42: newline

JPinkney (commented):
I just have some initial comments/questions, I'm going to test it on minikube/crc later today

build/Dockerfile
```diff
@@ -1,29 +1,36 @@
...
 # Compile webhook binaries
-RUN CGO_ENABLED=0 GOOS=linux go build \
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
   -o _output/bin/webhook-server \
   -gcflags all=-trimpath=/ \
   -asmflags all=-trimpath=/ \
   webhook/main.go
 
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
```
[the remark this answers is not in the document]
amisevsk: It means I forgot to remove the comment when I gave up on its suggestion and stuck with what we currently do :)

build/rhel.Dockerfile
amisevsk: Good catch; the vars were set by kubebuilder so I'm not 100% clear on if they're actually needed though. I imagine at some point we might have to undo the change to support multi-platform builds.

config/devel/kustomization.yaml
amisevsk: You're right, I probably included it while testing something and forgot to remove. Thanks!

config/components/prometheus/monitor.yaml
amisevsk: The files *were* autogenerated but I don't think anything will touch them going forward. I'll fix the weird spacing.

config/devel/manager_image_patch.yaml
amisevsk: Ah these files accidentally got mangled while I was figuring out the Makefile, they should be full of env vars to be replaced. That must've made testing kind of tedious.

The purpose of the additional files in `config/devel` is that they're templates that we fill with `envsubst` before running kustomize, to get the same functionality with env vars we have in the current makefile (see the `deploy` rule).

controllers/workspace/devworkspace_controller.go
amisevsk: I'll create some issues on Monday, that's a good point.

Makefile
```diff
@@ -29,13 +31,43 @@ INTERNAL_TMP_DIR=/tmp/devworkspace-controller
...
+
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
 endif
 
-all: help
+all: manager
```
sleshchenko, line 70: What do you think if we leave help as default rule?

Makefile
```diff
@@ -238,54 +101,71 @@ update_devworkspace_api:
...
-### start_local_debug: start local instance of controller with debugging enabled
-start_local_debug: _bump_kubeconfig _generate_related_images_env _login_with_devworkspace_sa
-	@source $(RELATED_IMAGES_FILE)
-ifeq ($(WEBHOOK_ENABLED),true)
-	#in cluster mode it comes from Deployment env var
-	export RELATED_IMAGE_devworkspace_webhook_server=$(IMG)
-	#in cluster mode it comes from configured SA propogated via env var
-	export CONTROLLER_SERVICE_ACCOUNT_NAME=devworkspace-controller
-	export KUBECONFIG=$(BUMPED_KUBECONFIG)
+	kubectl apply -f config/registry/local/os -n $(NAMESPACE)
```
sleshchenko, line 158: that's strange but it fails for me on crc
```kubectl apply -f config/registry/local/os -n devworkspace-controller
error: unable to recognize "config/registry/local/os/route.yaml": no matches for kind "Route" in version "v1"
Makefile:154: recipe for target 'deploy_registry' failed
make: *** [deploy_registry] Error 1
```
> ~/projects/devworkspace-operator update-sdk-1.x *3 !1 > crc version                                                                                              > % 14:35:22
> CodeReady Containers version: 1.15.0+e317bed
> OpenShift version: 4.5.7 (embedded in binary)

After changing Route api version to `apiVersion: route.openshift.io/v1 to fail in another place

config/default/kustomization.yaml
```diff
@@ -0,0 +1,71 @@
+# Adds namespace to all resources.
+namespace: devworkspace-operator
```
sleshchenko, line 2: What I did wrong :
```bash
~/projects/devworkspace-operator update-sdk-1.x *3 !1 > make deploy                                                                                           9s % 14:44:25
kubectl create namespace devworkspace-controller || true
Error from server (AlreadyExists): namespaces "devworkspace-controller" already exists
kubectl apply -f config/registry/local -n devworkspace-controller
deployment.apps/che-plugin-registry unchanged
service/che-plugin-registry unchanged
kubectl apply -f config/registry/local/os -n devworkspace-controller
route.route.openshift.io/che-plugin-registry unchanged
mv config/devel/config.properties config/devel/config.properties.bak
mv config/devel/manager_image_patch.yaml config/devel/manager_image_patch.yaml.bak
envsubst < config/devel/config.properties.bak > config/devel/config.properties
envsubst < config/devel/manager_image_patch.yaml.bak > config/devel/manager_image_patch.yaml
/home/sleshche/gocode/bin/kustomize build config/devel | kubectl apply -f - || true
namespace/devworkspace-controller-system created
customresourcedefinition.apiextensions.k8s.io/components.controller.devfile.io configured
customresourcedefinition.apiextensions.k8s.io/devworkspaces.workspace.devfile.io configured
customresourcedefinition.apiextensions.k8s.io/workspaceroutings.controller.devfile.io configured
clusterrole.rbac.authorization.k8s.io/devworkspace-controller-edit-workspaces created
clusterrole.rbac.authorization.k8s.io/devworkspace-controller-manager-role created
clusterrole.rbac.authorization.k8s.io/devworkspace-controller-proxy-role created
clusterrole.rbac.authorization.k8s.io/devworkspace-controller-view-workspaces created
clusterrole.rbac.authorization.k8s.io/devworkspace-controller-metrics-reader created
clusterrolebinding.rbac.authorization.k8s.io/devworkspace-controller-manager-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/devworkspace-controller-proxy-rolebinding created
configmap/devworkspace-controller created
Error from server (NotFound): error when creating "STDIN": namespaces "devworkspace-operator" not found
Error from server (NotFound): error when creating "STDIN": namespaces "devworkspace-operator" not found
Error from server (NotFound): error when creating "STDIN": namespaces "devworkspace-operator" not found
Error from server (NotFound): error when creating "STDIN": namespaces "system" not found
mv config/devel/config.properties.bak config/devel/config.properties

```

Makefile
```diff
@@ -238,54 +101,71 @@ update_devworkspace_api:
...
+debug: _generate_related_images_env
+	source $(RELATED_IMAGES_FILE)
+	dlv debug --listen=:2345 --headless=true --api-version=2 ./main.go --
+
+### install: Install CRDs into a cluster
+install: manifests _kustomize _create_namespace
+	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+
+#### uninstall: Uninstall CRDs from a cluster
+uninstall: manifests _kustomize
```
sleshchenko, line 131: previously it used to remove controller as well. What is rule to remove controller ?

Makefile
```diff
@@ -238,54 +101,71 @@ update_devworkspace_api:
...
+run: _generate_related_images_env
+	source $(RELATED_IMAGES_FILE)
+	go run ./main.go
+
+debug: _generate_related_images_env
+	source $(RELATED_IMAGES_FILE)
+	dlv debug --listen=:2345 --headless=true --api-version=2 ./main.go --
+
+### install: Install CRDs into a cluster
+install: manifests _kustomize _create_namespace
```
sleshchenko, line 127: Why CRD needs created namespace?

Makefile
```diff
@@ -238,54 +101,71 @@ update_devworkspace_api:
...
+	$(KUSTOMIZE) build config/devel | kubectl apply -f - || true
+	mv config/devel/config.properties.bak config/devel/config.properties
+	mv config/devel/manager_image_patch.yaml.bak config/devel/manager_image_patch.yaml
+
+### restart: Restart devworkspace-controller deployment
+restart:
+	kubectl rollout restart -n $(NAMESPACE) deployment/devworkspace-controller-manager
+
+### uninstall_controller: Remove controller resources from the cluster
+uninstall_controller: _kustomize
```
sleshchenko, line 149:
```
~/projects/devworkspace-operator update-sdk-1.x *3 !8 > make uninstall_controller                                                                            INT % 15:17:44
kustomize build config/devel | kubectl delete --ignore-not-found -f -
2020/11/02 15:17:45 got file 'controller.devfile.io_components.yaml', but '/home/sleshche/projects/devworkspace-operator/config/crd/bases/controller.devfile.io_components.yaml' must be a directory to be a root
Error: accumulating resources: recursed accumulation of path '/home/sleshche/projects/devworkspace-operator/config/default': accumulating resources: recursed accumulation of path '/home/sleshche/projects/devworkspace-operator/config/crd': accumulating resources: accumulating resources from 'bases/controller.devfile.io_components.yaml': empty item at spec/validation/openAPIV3Schema/properties/spec/properties/commands/items/properties/commandType/enum/ in object {map[apiVersion:apiextensions.k8s.io/v1beta1 kind:CustomResourceDefinition metadata:map[annotations:map[controller-gen.kubebuilder.io/version:v0.3.0] creationTimestamp:<nil> name:components.controller.devfile.io] spec:map[group:controller.devfile.io names:map[kind:Component listKind:ComponentList plural:components singular:component] scope:Namespaced subresources:map[status:map[]] validation:map[openAPIV3Schema:map[description:Component is the Schema for the components API properties:map[apiVersion:map[description:APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources type:string] kind:map[description:Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds type:string] metadata:map[type:object] spec:map[description:ComponentSpec defines the desired state of Component properties:map[commands:map[description:Commands from devfile, to be matched to components items:map[properties:map[apply:map[description:Command that consists in applying a given component definition, typically bound to a workspace event. 
 For example, when an `apply` command is bound to a `preStart` event, and references a `container` component, it will start the container as a K8S initContainer in the workspace POD, unless the component has its `dedicatedPod` field set to `true`. 
 When no `apply` command exist for a given component, it is assumed the component will be applied at workspace start by default. properties:map[attributes:map[additionalProperties:map[type:string] description:Optional map of free-form additional command attributes type:object] component:map[description:Describes component that will be applied type:string] group:map[description:Defines the group this command is part of properties:map[isDefault:map[description:Identifies the default command for a given group kind type:boolean] kind:map[description:Kind of group the command is part of enum:[build run test debug] type:string]] required:[kind] type:object] id:map[description:Mandatory identifier that allows referencing this command in composite commands, from a parent, or in events. type:string] label:map[description:Optional label that provides a label for this command to be used in Editor UI menus for example type:string]] required:[id] type:object] commandType:map[description:Type of workspace command enum:[Exec <nil>]]]]]]]]]]]]}
No resources found
```

sleshchenko (commented):
I was not able to test it on crc yet

config/components/rbac/role.yaml
```diff
@@ -2,162 +2,178 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: devworkspace-controller
-  labels:
-    app.kubernetes.io/name: devworkspace-controller
-    app.kubernetes.io/part-of: devworkspace-operator
+  creationTimestamp: null
+  name: manager-role
```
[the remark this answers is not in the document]
amisevsk: It's an auto-generated file from annotations (e.g. [here](https://github.com/devfile/devworkspace-operator/pull/187/files#diff-37e268a06048df028153ba3ae56a059ed72ded62c1b945631e411b1f5242b767R47-R49)) but I could look into changing the name.

config/components/rbac/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
...
+- role_binding.yaml
+- leader_election_role.yaml
+- leader_election_role_binding.yaml
+# Cluster-edit and view roles for devworkspaces and related resources
+- edit-workspaces-cluster-role.yaml
+- view-workspaces-cluster-role.yaml
+# Comment the following 4 lines if you want to disable
+# the auth proxy (https://github.com/brancz/kube-rbac-proxy)
+# which protects your /metrics endpoint.
+- auth_proxy_service.yaml
```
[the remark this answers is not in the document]
amisevsk: By default, kubebuilder bootstraps the "manager" (formerly known as "controller") pod to contain the RBAC proxy for metrics. I haven't tested it much yet but it seems like a generally good idea.

Makefile
amisevsk: :+1: This has already tripped me up me a few times during development :)

Makefile
amisevsk: Hm I'll re-test this today. The files are copied over from the current repo IIRC so this shouldn't be happening.

config/default/kustomization.yaml
amisevsk: If this is on `crc` I suspect this might be me not making the project correctly. I'll double check.

Makefile
amisevsk: It doesn't; good catch :)

config/components/manager/manager.yaml
```diff
@@ -0,0 +1,69 @@
...
+    control-plane: controller-manager
+  name: system
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: manager
+  namespace: system
+  labels:
+    control-plane: controller-manager
```
sleshchenko, line 14: That's seems to be too general labels which may overlap with other generated controllers. What if we use here something devworkspace specific?

sleshchenko (commented):
e2e tests need context.TODO() to be added in a few places. Check it with `make test`

controllers/workspace/devworkspace_controller.go
amisevsk: Created #192 and #191

config/components/manager/manager.yaml
amisevsk: I mostly left it because all kubebuilder projects are generated with this label and I didn't see anything about it in the documentation. We can probably remove it.

Makefile
amisevsk: Interesting; `oc` recognizes `v1/Route` through some strange magic but `kubectl` does not. I've updated `route.yaml` to use the full apiVersion (`route.openshift.io/v1`).

config/default/kustomization.yaml
amisevsk: My bad -- I forgot to allow overriding namespace in the `config/devel` kustomization. Should be fixed now.

amisevsk:
I ran into some strange errors running on OpenShift today; I'm looking into the issue.

Makefile
amisevsk: Tested locally and it seems to work fine; I'm not sure what the issue is.

sleshchenko:
> On crc when deployed in cluster, controller loops on creating v1.Role for the workspace. No error is returned but no object is created. Running locally does not reproduce this problem

after adding some additional logs, I got 
```
2020-11-03T12:34:36.481Z        DEBUG   controllers.DevWorkspace        Managing K8s Object     {"Request.Namespace": "devworkspace-controller", "Request.Name": "web-terminal", "kind": "v1.Role", "name": "workspace"}
2020-11-03T12:34:36.481Z        INFO    controllers.DevWorkspace            => Creating v1.Role {"Request.Namespace": "devworkspace-controller", "Request.Name": "web-terminal", "namespace": "devworkspace-controller", "name": "workspace"}
2020-11-03T12:34:36.558Z        DEBUG   controllers.DevWorkspace        created.        {"Request.Namespace": "devworkspace-controller", "Request.Name": "web-terminal", "error": "roles.rbac.authorization.k8s.io \"workspace\" is forbidden: user \"system:serviceaccount:devworkspace-controller:default\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:devworkspace-controller\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"\"], Resources:[\"pods/exec\"], Verbs:[\"create\"]}\n{APIGroups:[\"apps\"], Resources:[\"replicasets\"], Verbs:[\"get\" \"list\" \"watch\"]}\n{APIGroups:[\"extensions\"], Resources:[\"deployments\"], Verbs:[\"get\" \"list\" \"watch\"]}\n{APIGroups:[\"extensions\"], Resources:[\"replicasets\"], Verbs:[\"get\" \"list\" \"watch\"]}"}
```
I've pushed changes to log such error https://github.com/devfile/devworkspace-operator/pull/187/commits/6ea3cff2f24d1907e9ba09908692ec4bbb4979a6 and going to provide fix for roles as well.

Makefile
```diff
@@ -238,54 +101,80 @@ update_devworkspace_api:
...
+
+debug: _print_vars _generate_related_images_env
+	source $(RELATED_IMAGES_FILE)
+	WATCH_NAMESPACE=$(NAMESPACE) dlv debug --listen=:2345 --headless=true --api-version=2 ./main.go --
+
+### install: Install CRDs into a cluster
+install: manifests _kustomize
+	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+
+#### uninstall: Uninstall CRDs from a cluster
```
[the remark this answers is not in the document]
amisevsk: Artifacts of the autogenerated makefile (which does not come with an "uninstall_controller" target at all), I agree here.

Makefile
```diff
@@ -238,54 +101,80 @@ update_devworkspace_api:
...
+uninstall: manifests _kustomize
+	$(KUSTOMIZE) build config/crd | kubectl delete -f -
+
+### deploy: Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: _print_vars _kustomize _create_namespace deploy_registry
```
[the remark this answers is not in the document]
amisevsk: You shouldn't be hitting this issue -- the `manifests` rule updates the CRDs and rbac yaml files, which are committed to the repo. I removed the `manifests` target initially because it takes a few seconds and slows down deployment.

amisevsk:
Rebased on master and squashed most fixups. Thanks @sleshchenko for figuring out the RBAC and testing code. 

I dropped the commit that added workspace roles; instead I've explicitly added the required roles to the main block of [`devworkspace_controller.go`](https://github.com/devfile/devworkspace-operator/pull/187/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R71-R75)

Tested on `crc`:
- Theia-nodejs workspace sample (uses openshift-oauth routing) -- works as expected with a fix for webviews route (container name collision on oauth pods)
- Web Terminal UI flow as regular user (the detection code that enables web terminal works if the controller is installed).
No issues as far as I can see.

sleshchenko (approved):
Tested cloudshell workspace on crc and it seems to work fine. So, I believe it should be good to proceed with it. Good work :+1:

davidfestal (approved):
Great work @amisevsk !

Makefile
```diff
@@ -8,19 +8,21 @@
...
-OPERATOR_SDK_VERSION = v0.17.0
-NAMESPACE ?= devworkspace-controller
-IMG ?= quay.io/devfile/devworkspace-controller:next
-TOOL ?= oc
-ROUTING_SUFFIX ?= 192.168.99.100.nip.io
-PULL_POLICY ?= Always
-WEBHOOK_ENABLED ?= true
-DEFAULT_ROUTING ?= basic
-ADMIN_CTX ?= ""
+# .ONESHELL:
```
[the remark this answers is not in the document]
amisevsk: Yeah we don't need it anymore I think (though it's still useful for suppressing all output with a single `@`)

Makefile
```diff
@@ -8,19 +8,21 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 #
+
+#### TODOS ####
+## - Port over e2e tests.
```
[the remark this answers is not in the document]
amisevsk: Thanks for remembering :)

config/components/rbac/role.yaml
amisevsk: This space is added whenever the file is regenerated :/

Makefile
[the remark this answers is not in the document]
amisevsk: Removed

config/components/manager/manager.yaml
amisevsk: I've removed the labels now and we're using only the ones from before. I still don't know what the purpose of the label is supposed to be, though.

Issue #183, opened 2020-10-22
Implement apply of container component on preStart events


sleshchenko:
### Is your task related to a problem? Please describe.
To support the new plugins model, it's needed to implement apply of container component on preStart events.
Then we'll be able to describe plugins like the following (it's adapted samples from https://github.com/devfile/api/tree/master/samples):

<details>

<summary>Base vsx plugin template</summary>

```yaml
schemaVersion: 2.0.0
metadata:
  publisher: redhat
  name: vsx-template
  type: template
  parameters:
    VSX_LIST # ????
components:
  - name: vsx-installer
    container:
      image: vsx-installer # technically it's adapted artifacts plugin broker which is not in place yet
      volumeMounts:
        - name: vsx
          path: "/vsx"
      env:
        - name: VSX_LIST
          value: ""
    - name: theia-remote-injector
      container: 
       image: "quay.io/eclipse/che-theia-endpoint-runtime-binary:7.20.0"
      volumesMounts:
      - mountPath: "/remote-endpoint"
        name: remote-endpoint
      env:
        - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
          value: /remote-endpoint/plugin-remote-endpoint
        - name: REMOTE_ENDPOINT_VOLUME_NAME
          value: remote-endpoint
  - name: remote-endpoint
    volume:
      emptyDir: {} ? №2
commands:
 - id: copyVsx
   apply:
     component: vsx-installer
 - id: injectRemoteILauncher
   apply:
     component: theia-remote-injector
events:
  preStart:
    - copyVsx
    - injectRemoteILauncher
```
</details>

**? №1** We don't want to get copies of `vsxInstaller` and `injectRemoteILauncher` but the model does not allow to define merging the same components. So, maybe it should be implementation-specific for plugin component - if a different component brings the component with the same name - we try to merge them. Everything except VSX_LIST should be the same.
It may be a bit simpler in terms on interface declaration if we define different env vars in different plugins, like `VSX_JAVA_8`, `VSX_JAVA_DEBUG`, ... Otherwise we should hardcode that only `VSX_LIST` should be merged with appending.

**? №2** empty dir volumes are not implemented yet https://github.com/devfile/api/issues/189

<details>
<summary>Then plugin definition:</summary>

```yaml
schemaVersion: 2.0.0
metadata:
  publisher: redhat
  name: java8
  version: latest
  displayName: Language Support for Java 8
[...]

PR #188, opened 2020-10-28
Init containers

amisevsk:
### What does this PR do?
Add basic functionality for separating out initContainers from regular components in a devfile, based on the rules defined in https://github.com/devfile/api/issues/32

Resolves https://github.com/devfile/devworkspace-operator/issues/183, except code is not used anywhere yet -- it's not clear how to support this with subresources, as the process requires reading the whole devfile. Final implementation will require changing the internal API.

### Is it tested? How?
Test cases are written but it is just library code for now.

### Additional Info
This PR is targetting the `update-sdk-1.x` branch used in PR https://github.com/devfile/devworkspace-operator/pull/187, since otherwise there will be some hassle in rebasing.

### DIFF
--- a/pkg/library/command.go
+++ b/pkg/library/command.go
@@ -1,0 +1,89 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package library
+
+import (
+	"fmt"
+
+	"github.com/devfile/api/pkg/apis/workspaces/v1alpha1"
+)
+
+func getCommandType(command v1alpha1.Command) (v1alpha1.CommandType, error) {
+	err := command.Normalize()
+	if err != nil {
+		return "", err
+	}
+	return command.CommandType, nil
+}
+
+func getCommandsForIds(ids []string, commands []v1alpha1.Command) ([]v1alpha1.Command, error) {
+	var resolvedCommands []v1alpha1.Command
+
+	for _, id := range ids {
+		resolvedCommand, err := getCommandById(id, commands)
+		if err != nil {
+			return nil, err
+		}
+		resolvedCommands = append(resolvedCommands, *resolvedCommand)
+	}
+
+	return resolvedCommands, nil
+}
+
+func getCommandById(id string, commands []v1alpha1.Command) (*v1alpha1.Command, error) {
+	for _, command := range commands {
+		commandKey, err := command.Key()
+		if err != nil {
+			return nil, err
+		}
+		if commandKey == id {
+			return &command, nil
+		}
+	}
+	return nil, fmt.Errorf("no command with key %s is defined", id)
+}
+
+func commandListToComponentKeys(commands []v1alpha1.Command) (map[string]bool, error) {
+	componentKeys := map[string]bool{}
+	for _, command := range commands {
+		commandType, err := getCommandType(command)
+		if err != nil {
+			return nil, err
+		}
+		switch commandType {
+		case v1alpha1.ApplyCommandType:
+			componentKeys[command.Apply.Component] = true
+		case v1alpha1.ExecCommandType:
+			// TODO: This will require special handling (how do we handle prestart exec?)
+			componentKeys[command.Exec.Component] = true
+		case v1alpha1.CompositeCommandType:
+			// TODO: Handle composite commands: what if an init command is composite and refers to other commands
+		default: // Ignore
+		}
+	}
+	return componentKeys, nil
+}
+
+func removeCommandsByKeys(keys []string, commands []v1alpha1.Command) ([]v1alpha1.Command, error) {
+	var filtered []v1alpha1.Command
+	for _, command := range commands {
+		key, err := command.Key()
+		if err != nil {
+			return nil, err
+		}
+		if !listContains(key, keys) {
+			filtered = append(filtered, command)
+		}
+	}
+	return filtered, nil
+}
--- a/pkg/library/command.go
+++ b/pkg/library/command.go
@@ -26,11 +26,11 @@ func getCommandType(command v1alpha1.Command) (v1alpha1.CommandType, error) {
 	return command.CommandType, nil
 }
 
-func getCommandsForIds(ids []string, commands []v1alpha1.Command) ([]v1alpha1.Command, error) {
+func getCommandsForKeys(key []string, commands []v1alpha1.Command) ([]v1alpha1.Command, error) {
 	var resolvedCommands []v1alpha1.Command
 
-	for _, id := range ids {
-		resolvedCommand, err := getCommandById(id, commands)
+	for _, id := range key {
+		resolvedCommand, err := getCommandByKey(id, commands)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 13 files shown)


pkg/library/command.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, command := range commands {
+		commandKey, err := command.Key()
+		if err != nil {
+			return nil, err
+		}
+		if commandKey == id {
+			return &command, nil
+		}
+	}
+	return nil, fmt.Errorf("no command with key %s is defined", id)
```
davidfestal, line 53: In the future I assume that these rules should be implemented in the `validation` package of the `devfile/api` repo.
As soon as we have something working here, we should sync with other teams to move this into the common `validation` code.

pkg/library/lifecycle.go
```diff
@@ -0,0 +1,92 @@
...
+		commandType, err := getCommandType(cmd)
+		if err != nil {
+			return err
+		}
+		switch commandType {
+		case v1alpha1.ApplyCommandType:
+			continue
+		default:
+			// How a prestart exec command should be implemented is undefined currently, so we reject it.
+			return fmt.Errorf("only apply-type commands are supported in the prestart lifecycle binding")
```
davidfestal, line 88: In the future I assume that these rules should be implemented in the `validation` package of the `devfile/api` repo.
As soon as we have something working here, we should sync with other teams to move this into the common `validation` code.

davidfestal (approved):

pkg/library/command.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func removeCommandsByKeys(keys []string, commands []v1alpha1.Command) ([]v1alpha1.Command, error) {
+	var filtered []v1alpha1.Command
+	for _, command := range commands {
+		key, err := command.Key()
+		if err != nil {
+			return nil, err
+		}
+		if !listContains(key, keys) {
```
sleshchenko, line 84: To avoid loop on each iteration alternative could be
```go
func stringListToMap(list []string) map[string]bool {
	res := map[string]bool{}
	for _, item := range list {
		res[item] = true
	}
	return res
}


func removeCommandsByIDs(IDs []string, commands []v1alpha1.Command) ([]v1alpha1.Command, error) {
	toRemove := stringListToMap(IDs)
	var filtered []v1alpha1.Command
	for _, command := range commands {
		key, err := command.Key()
		if err != nil {
			return nil, err
		}
		if !toRemove[key] {
			filtered = append(filtered, command)
		}
	}
	return filtered, nil
}
```
P.S. I understand that this additional loop is typical with a couple of iterations, so you can simply resolve this comment without applying changes )

pkg/library/lifecycle.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	initComponentKeys, err := commandListToComponentKeys(initCommands)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Need to also consider components that are *both* init containers and in the main deployment
+	// Example: component is referenced in both a prestart event and a regular, non-prestart command
+	// TODO: Figure out details of handling postStop commands, since they should not be included in main deployment
+	nonInitCommands, err := removeCommandsByKeys(events.PreStart, commands)
```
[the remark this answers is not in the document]
sleshchenko: P.S. Maybe it even makes sense to generate func like Name(), ID() in addition to Key() but that should not be part of this PR for sure.

pkg/library/testdata/lifecycle/prestart_exec_command.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - exec:
+        id: test_command
+        component: test-container1
+        command: "test_command"
+  events:
+    preStart:
+      - "test_command"
+
+output:
+  initContainers:
```
sleshchenko, line 21: Should not be it ignored if `errRegexp` is specified?

sleshchenko: Adding such condition in test code could help us keep test data clean, but that's not simple, so up to you to add it or not
```go
//both of error and containers should not be configured at the same time
if tt.Output.ErrRegexp != nil && len(*tt.Output.ErrRegexp) > 0 &&
    (len(tt.Output.InitContainers) > 0 || len(tt.Output.MainContainers) > 0) {
    t.Fatal("No init containers or main containers could be configured when error is expected")
}
```

sleshchenko (approved):
LGTM
Nothing to comment apart of minor `Key vs ID` or `id -> ID` stuff

Here is a commit with changes you may consider to apply https://github.com/devfile/devworkspace-operator/commit/f985c576becf7dec485a1ed59bde8dac556a0f0c
I've made them just to understand code better.

pkg/library/lifecycle.go
```diff
@@ -0,0 +1,92 @@
...
+
+// GetInitContainers partitions the components in a devfile's flattened spec into initContainer and non-initContainer lists
+// based off devfile lifecycle bindings and commands. Note that a component can appear in both lists, if e.g. it referred to
+// in a preStart command and in a regular command.
+func GetInitContainers(devfile v1alpha1.DevWorkspaceTemplateSpecContent) (initContainers, mainComponents []v1alpha1.Component, err error) {
+	components := devfile.Components
+	commands := devfile.Commands
+	events := devfile.Events
+	if events == nil || commands == nil {
+		// All components are
```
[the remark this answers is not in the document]
amisevsk: I completed the sentence like four times and then ended up doing a `git reset` for other reasons. Will update :)

pkg/library/lifecycle.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, cmd := range initCommands {
+		commandType, err := getCommandType(cmd)
+		if err != nil {
+			return err
+		}
+		switch commandType {
+		case v1alpha1.ApplyCommandType:
+			continue
+		default:
+			// How a prestart exec command should be implemented is undefined currently, so we reject it.
```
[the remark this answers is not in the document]
amisevsk: Well, I used this phrasing because it's undefined for `exec` but makes sense to ignore for the other types. I'll add something to this doc to make it clearer.

pkg/library/lifecycle.go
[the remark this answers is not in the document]
amisevsk: The main purpose here is that we're integrating with the interfaces in `devfile/api` -- I don't mind returning `"ID"` or `"Name"` in user-facing error messages but internally I'd prefer to use the abstractions of the devfile api. I think it'll become clearer as more library functions are implemented, where `key` is used as a key for merging/overriding, etc.

pkg/library/command.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return nil, err
+		}
+		resolvedCommands = append(resolvedCommands, *resolvedCommand)
+	}
+
+	return resolvedCommands, nil
+}
+
+func getCommandById(id string, commands []v1alpha1.Command) (*v1alpha1.Command, error) {
```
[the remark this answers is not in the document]
amisevsk: I *hate* golint :smile: 

But I'll fix it I guess.

pkg/library/testdata/lifecycle/prestart_exec_command.yaml
amisevsk: In general the convention is you can't make any assumptions about non-error return value when `err != nil`. The leftovers in this test case are due to disabling preStart exec support after having written the tests. I've removed the initContainers/mainContainers section for this file.

pkg/library/lifecycle.go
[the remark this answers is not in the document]
amisevsk: I've updated the code to refer to `keys` by default, since we shouldn't be worried about what the field is named.

pkg/library/command.go
amisevsk: I agree that converting to a set is the more technically efficient way to do this (at one point I had code to convert between the two, etc.) but for lists that we expect to be <10 elements most of the time I feel like it's complexity without benefit.

JPinkney (approved):

PR #190, opened 2020-11-02
Add annotation label for setting the reason why the web terminal was …

JPinkney:
…stopped

### What does this PR do?
This PR (and the che-machine-exec PR that goes with it) add an annotation that shows the reason why the workspace was stopped. Related che-machine-exec PR: https://github.com/eclipse/che-machine-exec/pull/117.

https://issues.redhat.com/projects/WTO/issues/WTO-47

### Is it tested? How?

```
# Update the controller.yaml to point to the new che-machine-exec and make sure that the correct webhook image is deployed
make uninstall
make build
make deploy

# Navigate to the console and open a web terminal in the ui then
oc describe devworkspace
# See that io.devfile.devworkspace.stopped-by is empty

# Wait for the terminal to stop (I change my default timeout to 2m when testing so it was much easier)
oc describe devworkspace
# See that the annotation io.devfile.devworkspace.stopped-by is set to "inactivity"

# Open a new web terminal on the ui
oc describe devworkspace
# See that io.devfile.devworkspace.stopped-by is empty
```

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -59,6 +59,10 @@ const (
 
 	// ControllerServiceAccountNameEnvVar stores the name of the serviceaccount used in the controller.
 	ControllerServiceAccountNameEnvVar = "CONTROLLER_SERVICE_ACCOUNT_NAME"
+
+	// WorkspaceStopReasonAnnotation marks the reason why the workspace was stopped; when a workspace is restarted
+	// this annotation will be cleared
+	WorkspaceStopReasonAnnotation = "controller.devfile.io/stopped-by"
 )
 
 // Constants for che-rest-apis
--- a/pkg/controller/workspace/workspace_controller.go
+++ b/pkg/controller/workspace/workspace_controller.go
@@ -218,6 +218,13 @@ func (r *ReconcileWorkspace) Reconcile(request reconcile.Request) (reconcileResu
 		return reconcile.Result{}, err
 	}
 
+	_, ok := workspace.Annotations[config.WorkspaceStopReasonAnnotation]
+	if ok {
+		delete(workspace.Annotations, config.WorkspaceStopReasonAnnotation)
+		err = r.client.Update(context.TODO(), workspace)
+		return reconcile.Result{Requeue: true}, err
+	}
+
 	immutable := workspace.Annotations[config.WorkspaceImmutableAnnotation]
 	if immutable == "true" && config.ControllerCfg.GetWebhooksEnabled() != "true" {
 		reqLogger.Info("Workspace is configured as immutable but webhooks are not enabled.")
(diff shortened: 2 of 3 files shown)


pkg/controller/workspace/workspace_controller.go
```diff
@@ -218,6 +218,12 @@ func (r *ReconcileWorkspace) Reconcile(request reconcile.Request) (reconcileResu
 		return reconcile.Result{}, err
 	}
 
+	if workspace.Annotations[config.WorkspaceStopReasonAnnotation] != "" {
+		workspace.Annotations[config.WorkspaceStopReasonAnnotation] = ""
```
sleshchenko, line 222: I wonder if it would be better if we fully remove stopReason when workspace is started.

pkg/config/constants.go
```diff
@@ -59,6 +59,10 @@ const (
 
 	// ControllerServiceAccountNameEnvVar stores the name of the serviceaccount used in the controller.
 	ControllerServiceAccountNameEnvVar = "CONTROLLER_SERVICE_ACCOUNT_NAME"
+
+	// WorkspaceStopReasonAnnotation marks the reason why the workspace was stopped; when a workspace is restarted
+	// this annotation will be cleared
+	WorkspaceStopReasonAnnotation = "io.devfile.devworkspace.stopped-by"
```
sleshchenko, line 65: I don't have a strong preference here but something like `controller.devfile.io/stopped-by` would be consistent with other labels/annotations we have. 
That's magic but maybe @amisevsk you have a better vision of what is the controller and what is devworkspace )

pkg/controller/workspace/workspace_controller.go
JPinkney: I'm okay with doing that, @amisevsk wdyt?

pkg/config/constants.go
JPinkney: I knew this was going to be discussed :smile: I just stole the `io.devfile.devworkspace.stopped-by` from the issue until another suggestion came

pkg/controller/workspace/workspace_controller.go
amisevsk: Yeah I think that makes sense -- if it's not stopped we don't necessarily need a stopped-by annotation :)

pkg/config/constants.go
amisevsk: I support the `controller.devfile.io/stopped-by` annotation -- lets finalize the usual naming discussion as quickly as possible :wink: 

For reference, the annotations we're currently using:

```
"controller.devfile.io/workspace_id"
"controller.devfile.io/endpoint_name"
"controller.devfile.io/workspace_name"
"controller.devfile.io/creator"
"controller.devfile.io/restricted-access"
"controller.devfile.io/discoverable-service"
```
so that would be in line with their format.

amisevsk (approved):
LGTM apart from @sleshchenko's comments

pkg/config/constants.go
JPinkney: Makes sense to me, I'll update both PRs

sleshchenko (approved):
Seems to work as expected when I tested manually:
```
kubectl apply -f ./samples/cloud-shell.yaml

 kubectl patch devworkspace cloud-shell --type='merge' \
  -p '{"metadata": {"annotations": {"controller.devfile.io/stopped-by": "sleshche"}},"spec":{"started":false}}'

 kubectl patch devworkspace cloud-shell --type='merge' \
  -p '{"spec":{"started":true}}'
```

PR #193, opened 2020-11-12
Upgrade operator-sdk to 1.1

sleshchenko:
### What does this PR do?
I've discovered that we still depend on operator-sdk v0.17.0.
So, as far as I understand earlier we merged PR that did most of upgrade to OperatoSDK 1.1 which is moving to controller-gen.
This PR upgrades operator-sdk itself to 1.1. Let me know if I missed something and we should not do this upgrade.

Apart from that, it contains some minor improvements as self-descriptive commits.

https://github.com/devfile/devworkspace-operator/issues/180

### Is it tested? How?
I've checked that cloud-shell with openshift-oauth routings works fine on crc 1.18.0(OpenShift 4.6.1 inside).


Makefile
```diff
@@ -124,11 +127,11 @@ debug: _print_vars _generate_related_images_env
 	WATCH_NAMESPACE=$(NAMESPACE) dlv debug --listen=:2345 --headless=true --api-version=2 ./main.go --
 
 ### install: Install CRDs into a cluster
-install: manifests _kustomize
+install_crds: manifests _kustomize _init_devworkspace_crds
```
amisevsk, line 130: Need to also update doc line above (`### install`)

amisevsk (approved):
LGTM, good catch.

JPinkney (approved):
LGTM

Makefile
sleshchenko: but then where will be confusion point :smile: 

Good catch. Will fix

PR #194, opened 2020-11-12
Fix debug and run locally after regenerating with controller-gen

sleshchenko:
### What does this PR do?
This PR fixes debug and run locally after regenerating with controller-gen.

I've moved these changes out from https://github.com/devfile/devworkspace-operator/pull/193 since here is something to discuss:
1. without `.ONESHELL` env vars are not properly propagated to the dvl and eventually to controller. So, we should enable `.ONSHELL` or alernative - create script that will set env vars and run dvl or controller directly.
2. I've figured out that we modify default service account in namespace instead of created devworkspace specific one. I think that it's better to have a dedicated SA (if multiple operators are deployed in the same namespace, they may conflict due default roles modifications) but I did not manage to archive it properly with all deployment,roles,rolesbinding patching via kustomize. It can be solved independently from this PR.

N/A

### Is it tested? How?
```
make deploy
oc patch deployment/devworkspace-controller-manager --patch "{\"spec\":{\"replicas\":0}}" -n devworkspace-controller
make debug
#or make run
#make sure everything works
```
:warning: that's not well tested after the latest changes, but it'll be done before merge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,8 +130,8 @@ debug: _print_vars _generate_related_images_env
 install_crds: manifests _kustomize _init_devworkspace_crds
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
-### deploy: Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: _print_vars _kustomize _init_devworkspace_crds _create_namespace deploy_registry
+### install: Install controller in the configured Kubernetes cluster in ~/.kube/config
+install: _print_vars _kustomize _init_devworkspace_crds _create_namespace deploy_registry
 	mv config/devel/kustomization.yaml config/devel/kustomization.yaml.bak
 	mv config/devel/config.properties config/devel/config.properties.bak
 	mv config/devel/manager_image_patch.yaml config/devel/manager_image_patch.yaml.bak
--- a/Makefile
+++ b/Makefile
@@ -117,14 +117,36 @@ test_e2e: generate fmt vet manifests
 manager: generate fmt vet
 	go build -o bin/manager main.go
 
-### run: Run against the configured Kubernetes cluster in ~/.kube/config
-run: _print_vars _generate_related_images_env
-	source $(RELATED_IMAGES_FILE)
-	WATCH_NAMESPACE=$(NAMESPACE) go run ./main.go
+# it's easier to bump whole kubeconfig instead of grabbing cluster URL from the current context
+_bump_kubeconfig:
+	@mkdir -p $(INTERNAL_TMP_DIR)
+ifndef KUBECONFIG
+	$(eval CONFIG_FILE = ${HOME}/.kube/config)
+else
+	$(eval CONFIG_FILE = ${KUBECONFIG})
+endif
+	cp $(CONFIG_FILE) $(BUMPED_KUBECONFIG)
 
-debug: _print_vars _generate_related_images_env
+_login_with_devworkspace_sa:
+	@$(eval SA_TOKEN := $(shell kubectl get secrets -o=json -n $(NAMESPACE) | jq -r '[.items[] | select (.type == "kubernetes.io/service-account-token" and .metadata.annotations."kubernetes.io/service-account.name" == "default")][0].data.token' | base64 --decode ))
+	@echo "Logging as controller's SA in $(NAMESPACE)"
+	oc login --token=$(SA_TOKEN) --kubeconfig=$(BUMPED_KUBECONFIG)
+
+### run: Run against the configured Kubernetes cluster in ~/.kube/config
+run: _print_vars _generate_related_images_env _bump_kubeconfig _login_with_devworkspace_sa
 	source $(RELATED_IMAGES_FILE)
-	WATCH_NAMESPACE=$(NAMESPACE) dlv debug --listen=:2345 --headless=true --api-version=2 ./main.go --
+	export KUBECONFIG=$(BUMPED_KUBECONFIG)
+	CONTROLLER_SERVICE_ACCOUNT_NAME=default \
+		WATCH_NAMESPACE=$(NAMESPACE) \
+		go run ./main.go
+
+
+debug: _print_vars _generate_related_images_env _bump_kubeconfig _login_with_devworkspace_sa
+	source $(RELATED_IMAGES_FILE)
+	export KUBECONFIG=$(BUMPED_KUBECONFIG)
+	CONTROLLER_SERVICE_ACCOUNT_NAME=default \
+		WATCH_NAMESPACE=$(NAMESPACE) \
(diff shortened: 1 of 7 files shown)


amisevsk (approved):
LGTM, good catch. I used `make run` and `make debug` extensively while working on the operator-sdk 1.1 update and don't know how I didn't run into issues.

sleshchenko:
I continued to pump up this PR, so @amisevsk please review again. The most important new change https://github.com/devfile/devworkspace-operator/pull/194/commits/7f19b2531b55e3e29a364d587d7e07aebc3a1427

sleshchenko:
e2e is failing with error: `kubectl is not found`.

sleshchenko:
From https://docs.openshift.com/container-platform/3.10/cli_reference/differences_oc_kubectl.html#differences-oc-kubectl-using-kubectl it's safe to use oc instead of kubectl. I've checked and the difference we used to have in Makefile for oc/kubectl is not needed anymore(like for make uninstall and make restart). So, to easily fix e2e tests and avoid some users issues - now Makefile detects which k8s cli is available and use it or report an issue.

sleshchenko:
Tests are failing with

```
+ make test_e2e
which: no kubectl in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/go/bin:/tmp/bin:/go/src/github.com/devfile/devworkspace-operator:/go/bin)
which: no controller-gen in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/go/bin:/tmp/bin:/go/src/github.com/devfile/devworkspace-operator:/go/bin)
which: no kustomize in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/go/bin:/tmp/bin:/go/src/github.com/devfile/devworkspace-operator:/go/bin)
make: *** [controller-gen] Error 1
```

sleshchenko:
Prow instance faces: go get: disabled by -mod=vendor due:
```
sh-4.2$ go env
GO111MODULE="on"
GOFLAGS="-mod=vendor"
```
If https://github.com/devfile/devworkspace-operator/pull/194/commits/8094ec939b875ca914058cd56070fa11a6a66873 will fix the issue - we need to find a proper solution for that.

sleshchenko:
Next e2e test barier: `Error patching crds: yq is required`

Makefile
```diff
@@ -76,14 +87,15 @@ _print_vars:
...
+	$(K8S_CLI) create namespace $(NAMESPACE) || true
 
 _generate_related_images_env:
 	@mkdir -p $(INTERNAL_TMP_DIR)
+	echo "export RELATED_IMAGE_devworkspace_webhook_server=$(IMG)" > $(RELATED_IMAGES_FILE)
 	cat ./config/components/manager/manager.yaml \
 		| yq -r \
 			'.spec.template.spec.containers[]?.env[] | select(.name | startswith("RELATED_IMAGE")) | "export \(.name)=\"$${\(.name):-\(.value)}\""' \
-		> $(RELATED_IMAGES_FILE)
+		>> $(RELATED_IMAGES_FILE)
```
amisevsk, line 98: Won't this duplicate `export RELATED_IMAGE_devworkspace_webhook_server=$(IMG)`?

Makefile
```diff
@@ -117,59 +129,81 @@ test_e2e: generate fmt vet manifests
...
+	$(eval CONFIG_FILE = ${HOME}/.kube/config)
+else
+	$(eval CONFIG_FILE = ${KUBECONFIG})
+endif
+	cp $(CONFIG_FILE) $(BUMPED_KUBECONFIG)
+
+_login_with_devworkspace_sa:
+	@$(eval SA_TOKEN := $(shell $(K8S_CLI) get secrets -o=json -n $(NAMESPACE) | jq -r '[.items[] | select (.type == "kubernetes.io/service-account-token" and .metadata.annotations."kubernetes.io/service-account.name" == "default")][0].data.token' | base64 --decode ))
+	@echo "Logging as controller's SA in $(NAMESPACE)"
+	oc login --token=$(SA_TOKEN) --kubeconfig=$(BUMPED_KUBECONFIG)
```
amisevsk, line 145: This now assumes `oc` is installed. Should we do something for `kubectl` as well?

Makefile
```diff
@@ -226,6 +260,8 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
+	# Reset go flags to avoid error caused by set -mod=vendor
+	export GOFLAGS=""
```
amisevsk, line 264: I'd put this on the same line as `go get` to avoid the export (and the dependency on `.ONESHELL`):
```bash
GOFLAGS="" go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 ;\
```

config/devel/manager_image_patch.yaml
```diff
@@ -10,3 +10,6 @@ spec:
       - name: devworkspace-controller
         image: ${IMG}
         imagePullPolicy: Always
+        env:
+        - name: RELATED_IMAGE_devworkspace_webhook_server
+          value: ${IMG}
```
amisevsk, line 15: We had the same idea :) https://github.com/devfile/devworkspace-operator/pull/196

amisevsk (approved):
Generally LGTM, will try to test later.

JPinkney (approved):
Seems to work for me. I had to upgrade my delve from 1.2.0 to 1.5.0 otherwise debugging wouldn't work. Other then Angel's comments it LGTM

Makefile
sleshchenko: Well, fair point but unfortunately, kubectl does not provide such functionality since you how authentication stuff on k8s )
So, instead of relying on kubectl/oc we can generate kubeconfig file by our own, but I think for simplifying we can make oc as prerequisites for make run/debug

Makefile
sleshchenko: Not really. We'll get in the end:

```
export RELATED_IMAGE_devworkspace_webhook_server=sleshchenko/che-workspace-controller:local
export ....
export RELATED_IMAGE_devworkspace_webhook_server="${RELATED_IMAGE_devworkspace_webhook_server:-quay.io/devfile/devworkspace-controller:next}"
```
I think that's should be OK for code simplification where we don't need to analyze `.spec.template.spec.containers[]?.env[].name` to exclude webhookserver.
Does it make sense?

sleshchenko:
Merging since this PR is already pretty huge and no blockers found according to PR checks and reviewers approvals.
Minor stuff could be fixed separately.

PR #196, opened 2020-11-13
Use same image for webhooks deployment when developing

amisevsk:
### What does this PR do?
Makes `make deploy` configure the operator to use the same image for the operator and webhook deployments.

### Is it tested? How?
```
export IMG=<your-image>
export WEBHOOK_ENABLED=true
make deploy
# Verify that `devworkspace-webhook-server uses `<your-image>`
```

### DIFF
--- a/config/devel/manager_image_patch.yaml
+++ b/config/devel/manager_image_patch.yaml
@@ -10,3 +10,6 @@ spec:
       - name: devworkspace-controller
         image: ${IMG}
         imagePullPolicy: Always
+        env:
+          - name: RELATED_IMAGE_devworkspace_webhook_server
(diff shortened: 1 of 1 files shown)


JPinkney (approved):
LGTM, though I see that the change is also available in https://github.com/devfile/devworkspace-operator/pull/194

sleshchenko (approved):

sleshchenko:
I was first :stuck_out_tongue:

sleshchenko:
That's can be safely closed in flavor since https://github.com/devfile/devworkspace-operator/commit/9823f7a3e98eea0db509344cd7a536d5fcb7545b is already in master

Issue #189, opened 2020-10-29
Implement conversion webhooks for devfile/api versions


amisevsk:
> Include the conversion between v1alpha1 et v1alpha2 dans devfile/api

Info: https://book.kubebuilder.io/multiversion-tutorial/conversion.html

Part of https://github.com/devfile/devworkspace-operator/issues/175

amisevsk:
Resolved by https://github.com/devfile/api/pull/213

PR #197, opened 2020-11-17
Migrate to devfile/api v1alpha2 and add conversion webhooks.

amisevsk:
### What does this PR do?
Switches the devworkspace-operator to use `v1alpha2` devworkspaces, and implements conversion webhooks. Currently this is done by updating `go.mod` to point at https://github.com/devfile/api/pull/213, as that PR is not yet merged.

To avoid too many temporary changes in the repo (changing where devworkspace CRDs are downloaded from), you'll have to manually copy in the CRDs generated in https://github.com/devfile/api/pull/213 to `config/crd/bases/`: 
```bash
curl "https://raw.githubusercontent.com/devfile/api/fc53138e/crds/workspace.devfile.io_devworkspaces.yaml" \
    -o config/crd/bases/workspace.devfile.io_devworkspaces.yaml

curl "https://raw.githubusercontent.com/devfile/api/fc53138e/crds/workspace.devfile.io_devworkspacetemplates.yaml" \
    -o config/crd/bases/workspace.devfile.io_devworkspacetemplates.yaml 
```

Resolves https://github.com/devfile/devworkspace-operator/issues/189

### Is it tested? How?
To test this PR (on minikube, for now):
0. Set env vars as usual for development
1. Deploy cert-manager on your cluster: `make install_cert_manager`
2. Deploy controller: `make docker install`
3. Verify that conversion webhooks are created: log message `conversion webhook enabled` appears in `kubectl logs -f deploy/devworkspace-webhook-server`
4. Create a DevWorkspace: `kubectl apply -f samples/theia.yaml`
5. Check that you can get both v1alpha1 and v1alpha2 versions of the devworkspace:
    ```
    kubectl get devworkspaces.v1alpha1.workspace.devfile.io theia -o yaml
    kubectl get devworkspaces.v1alpha2.workspace.devfile.io theia -o yaml
    ```
6. Test usability as normal

### Remaining TODOs
1. I haven't tested/verified that local run/debug still works as expected, but will do that soon.
2. These changes make webhooks mandatory; we likely need to update how `WEBHOOKS_ENABLED` works, perhaps disabling the mutating and validating webhooks when set to false but still deploying the webhooks server
3. The service name for the webhooks server is hard-coded, so it has to be hard-coded into templates as well, we should probably make this configurable
4. Only k8s with cert-manager is supported at this point. The next step is to add support for OpenShift via the [Service CA operator](https://github.com/openshift/service-ca-operator), which supports a similar CRD annotation. We'll need to verify that the certificates secret is correctly propagated to the webhooks server
5. This PR includes a workaround for deploying CRDs when they're already present on the cluster (we first remove them). I've found that on `minikube`, without those changes, the second `make install` fails. You can test this by
    ```bash
    kustomize build config/crd | kubectl apply -f -
    kustomize build config/crd | kubectl apply -f -
    ```
    or by commenting out the delete command in `make install`.
6. This PR updates *existing* CRDs for Components and WorkspaceRoutings. In terms of yaml spec they are identical, except the new version has defaults for some endpoint fields. We need to verify that 
    - Applying a `v1` version of the CRD is not an error when a `v1beta1 CRD with the same GVK is present
    - That upgrading from a previous version will not cause issues for existing devworkspaces.


webhook/main.go
```diff
@@ -49,20 +63,29 @@ func main() {
...
 
-	err = createWebhooks(mgr, cfg)
+	err = createWebhooks(mgr)
 	if err != nil {
 		log.Error(err, "Failed to create webhooks")
 		os.Exit(1)
 	}
 
+	if err := ctrl.NewWebhookManagedBy(mgr).For(&workspacev1alpha1.DevWorkspace{}).Complete(); err != nil {
+		log.Error(err, "failed creating conversion webhook")
```
JPinkney, line 83: Just curious, if the conversion webhooks fail to create should we exit or just continue without them?

JPinkney (commented):
Just some initial comments, <s>my CRC was being a pain so I'll have to look further in depth on monday</s>. I should have read the description better :)

JPinkney:
I just tried the testing instructions and everything was working as expected. For some reason my `make _init_devworkspace_crds` wasn't working but I think that was more of a me issue.

I tried the local run/debug and I get:
```
2020-11-23T15:49:09.458-0500    ERROR   setup   failed to setup webhooks        {"error": "failed to create webhooks deployment: environment variable WEBHOOK_SECRET_NAME is unset"}
```

webhook/workspace/handler/immutable.go
```diff
@@ -72,7 +74,19 @@ func (h *WebhookHandler) HandleImmutableCreate(_ context.Context, req admission.
 	return admission.Denied("Only the workspace controller can create workspace objects.")
 }
 
-func (h *WebhookHandler) handleImmutableWorkspace(oldWksp, newWksp *devworkspace.DevWorkspace, uid string) (allowed bool, msg string) {
+func (h *WebhookHandler) handleImmutableWorkspaceV1alpha1(oldWksp, newWksp *devworkspacev1alpha1.DevWorkspace, uid string) (allowed bool, msg string) {
+	creatorUID := oldWksp.Labels[config.WorkspaceCreatorLabel]
```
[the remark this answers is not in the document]
amisevsk: Nope :) I tried switching based on type and it basically means writing the function body in each `case` statement :)

We could potentially get around this by only considering the metadata -- I don't know if patching it as an interface (runtime.ObjectMeta or whatever it's called) reflects to the cluster correctly.

webhook/main.go
amisevsk: I believe if the webhooks fail to set up you won't be able to create DevWorkspaces -- when you create a DevWorkspace it calls the conversion webhooks (at least for the v1alpha1 case) and if it can't get a response it blocks creation.

Also surprising info about how conversion works: If you specify `conversion: none` in the CRD, you can "get" a v1alpha2-spec workspace that is just the `v1alpha1` and a wrong GVK :shrug:

amisevsk:
Updates to this PR:

- commit 453e5fe adds support for the Serving CA operator when running on OpenShift
- commit 2c8c08a should fix Josh's issue around running/debugging locally (I haven't tested it directly)
- since PR https://github.com/devfile/api/pull/214 was merged in devfile/api, commits 28a8107..b3a6596 are cherry-picks (and updates/modifications where necessary) from @davidfestal's work: https://github.com/amisevsk/devworkspace-operator/compare/use-v1alpha2...devfile:use-v1alpha2-and-new-attributes
  - I squashed some fixups/repeated changes
  - I had to remove an accidentally-committed binary file from the history
   - **NOTE**: These commits also include David's work on implementing initContainers using the library functions merged in https://github.com/devfile/devworkspace-operator/pull/188
I also updated the PR to use the current HEAD of devfile/api, since the conversion PR https://github.com/devfile/api/pull/213 has been merged.

I tested the current PR on `crc` using the `all-in-one-theia-nodejs.devworkspace.yaml` sample David provided and was able to get the Theia UI but didn't have time to test further (all panels are empty in my case). Conversion webhooks also work on `crc` at the moment.

JPinkney (approved):
Code looks good to me, I'm going to test the new changes today

pkg/config/env.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+type ControllerEnv struct{}
+
+const (
+	webHooksSecretNameEnvVar      = "WEBHOOK_SECRET_NAME"
```
sleshchenko, line 23:
```suggestion
	webhooksSecretNameEnvVar      = "WEBHOOK_SECRET_NAME"
```
here and below

pkg/webhook/create.go
```diff
@@ -59,26 +60,31 @@ func SetupWebhooks(ctx context.Context, cfg *rest.Config) error {
...
 		log.Info("Setting up the OpenShift webhook server secure service")
-		err := webhook_openshift.SetupSecureService(client, ctx, namespace)
+		err := webhook_openshift.SetupSecureService(client, ctx, secretName, namespace)
 		if err != nil {
 			return err
 		}
 	} else {
+		secretName, err := config.GetWebhooksSecretName()
+		if err != nil {
+			return fmt.Errorf("failed to create webhooks deployment: %w", err)
```
sleshchenko, line 75: Is it an error message copied from somewhere?
I mean we haven't tried to create deployment yet. Maybe instead - `failed to deploy webhooks due:` or something like that

sleshchenko (commented):
The first round which captures tiny suggestions is done. I will do a deeper review later.

pkg/webhook/openshift/tls.go
```diff
@@ -18,17 +18,15 @@ import (
...
 )
 
 var log = logf.Log.WithName("webhook-openshift")
 
-func SetupSecureService(client crclient.Client, ctx context.Context, namespace string) error {
+func SetupSecureService(client crclient.Client, ctx context.Context, secretName, namespace string) error {
 	log.Info("Attempting to create the secure service")
 	err := service.CreateOrUpdateSecureService(client, ctx, namespace, map[string]string{
-		"service.beta.openshift.io/serving-cert-secret-name": server.WebhookServerTLSSecretName,
+		"service.beta.openshift.io/serving-cert-secret-name": secretName, // TODO
```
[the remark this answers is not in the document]
amisevsk: Nope :) good catch.

Makefile
```diff
@@ -166,23 +173,42 @@ install_crds: _kustomize _init_devworkspace_crds
...
 install: _print_vars _kustomize _init_devworkspace_crds _create_namespace deploy_registry
-	mv config/devel/kustomization.yaml config/devel/kustomization.yaml.bak
-	mv config/devel/config.properties config/devel/config.properties.bak
-	mv config/devel/manager_image_patch.yaml config/devel/manager_image_patch.yaml.bak
-
-	envsubst < config/devel/kustomization.yaml.bak > config/devel/kustomization.yaml
-	envsubst < config/devel/config.properties.bak > config/devel/config.properties
-	envsubst < config/devel/manager_image_patch.yaml.bak > config/devel/manager_image_patch.yaml
-	$(KUSTOMIZE) build config/devel | $(K8S_CLI) apply -f - || true
+	# workaround since applying CRDs when artifacts already exist hangs forever on minikube
```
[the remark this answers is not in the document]
amisevsk: Yeah this is one of the things we need to look into -- it should not be a problem but I can't get any debugging info out of kubectl or the api server :/

pkg/webhook/create.go
amisevsk: Updated the complete message -- it should now be `could not deploy webhooks server: environment variable WEBHOOK_SECRET_NAME is unset`

config/crd/kustomization.yaml
```diff
@@ -5,22 +5,10 @@ resources:
 - bases/controller.devfile.io_components.yaml
 - bases/controller.devfile.io_workspaceroutings.yaml
 - bases/workspace.devfile.io_devworkspaces.yaml
+- bases/workspace.devfile.io_devworkspacetemplates.yaml
 # +kubebuilder:scaffold:crdkustomizeresource
 
 patchesStrategicMerge:
```
sleshchenko, line 11: IDE suggests that it should be array
```suggestion
patchesStrategicMerge: []
```

controllers/controller/workspacerouting/solvers/cluster_solver.go
```diff
@@ -74,17 +74,22 @@ func (s *ClusterSolver) GetExposedEndpoints(
...
-			if endpoint.Attributes[string(controllerv1alpha1.PUBLIC_ENDPOINT_ATTRIBUTE)] != "true" {
+			if endpoint.Exposure != devworkspace.PublicEndpointExposure {
 				continue
 			}
 			url, err := resolveServiceHostnameForEndpoint(endpoint, routingObj.Services)
 			if err != nil {
 				return nil, false, err
 			}
+			endpointAttributes := map[string]string{}
+			err = endpoint.Attributes.Into(&endpointAttributes)
```
sleshchenko, line 85: That's not really related to this PR:
That's going to be not nil if attributes contains any not string value, right? Like int, boolean and so on.
So, we probably need to validate the attributes earlier or skip/try to convert to string not string values.
But again, that's more related to issue about endpoints but not converting webhooks.

controllers/controller/workspacerouting/solvers/common.go
```diff
@@ -35,7 +35,7 @@ func getDiscoverableServicesForEndpoints(endpoints map[string]controllerv1alpha1
 	var services []corev1.Service
 	for _, machineEndpoints := range endpoints {
 		for _, endpoint := range machineEndpoints {
-			if endpoint.Attributes[string(controllerv1alpha1.DISCOVERABLE_ATTRIBUTE)] == "true" {
+			if endpoint.Attributes.GetString(string(controllerv1alpha1.DISCOVERABLE_ATTRIBUTE), nil) == "true" {
```
sleshchenko, line 38:
```suggestion
			if endpoint.Attributes.GetBoolean(string(controllerv1alpha1.DISCOVERABLE_ATTRIBUTE), nil) {
```

pkg/config/env.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+type ControllerEnv struct{}
+
+const (
+	webhooksSecretNameEnvVar      = "WEBHOOK_SECRET_NAME"
```
sleshchenko, line 23:
```suggestion
	webhookSecretNameEnvVar      = "WEBHOOK_SECRET_NAME"
```
just one more nitpick. Here and below

sleshchenko (commented):
The code is good. I start testing

sleshchenko (commented):
I was not able to test on crc since it behaves really strangely.
Will try on minikube and then return to crc )

config/service-ca/crd_webhooks_patch.yaml
```diff
@@ -0,0 +1,41 @@
...
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
+spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      conversionReviewVersions: ["v1"]
+      clientConfig:
+        service:
+          namespace: devworkspace-operator
```
sleshchenko, line 36: Should namespace here be parametrized? Seems hard-coded value is also replaced with the right but I would avoid setting any real value to avoid confusion.

config/cert-manager/manager_certmanager_patch.yaml
```diff
@@ -0,0 +1,15 @@
+# This patch adds an environment variable to the workspace deployment
+# to allow it to figure out which secret to use for the webhooks deployment
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: manager
+  namespace: system
```
sleshchenko, line 7: I'm not sure what the right way to fix it but it seems `namespace: system` here and in other places leads to creation of `devworkspace-controller-system` namespace which then is not used anywhere.
```
 > make install                                                                           Current env vars:
    NAMESPACE=ws-operator
    IMG=sleshchenko/che-workspace-controller:local
namespace/ws-operator created
deployment.apps/che-plugin-registry created
...
namespace/devworkspace-controller-system created
...
```

JPinkney:
Just tested on ocp and it seems the conversion webhooks are working. Using the `all-in-one-theia-nodejs.devworkspace.yaml` all panels were empty (as was mentioned) but those don't look like they're because of this PR :+1:

config/components/certificates/cert-manager/self-signed-certificates.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: selfsigned-issuer
+  namespace: system
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: serving-cert
```
sleshchenko, line 13: I think it makes sense to name it webhook server specific, like
```suggestion
  name: devworkspace-webhooserver-certs
```

Makefile
```diff
@@ -89,9 +93,12 @@ _print_vars:
 _create_namespace:
 	$(K8S_CLI) create namespace $(NAMESPACE) || true
 
-_generate_related_images_env:
-	@mkdir -p $(INTERNAL_TMP_DIR)
+_gen_configuration_env:
+	mkdir -p $(INTERNAL_TMP_DIR)
 	echo "export RELATED_IMAGE_devworkspace_webhook_server=$(IMG)" > $(RELATED_IMAGES_FILE)
+ifeq ($(PLATFORM),kubernetes)
+	echo "export WEBHOOK_SECRET_NAME=devworkspace-operator-webhook-cert" > $(RELATED_IMAGES_FILE)
```
sleshchenko, line 100:
```suggestion
	echo "export WEBHOOK_SECRET_NAME=devworkspace-operator-webhook-cert" >> $(RELATED_IMAGES_FILE)
```

You killed `RELATED_IMAGE_devworkspace_webhook_server` :skull:

sleshchenko (commented):
I'll finish my testing tomorrow

config/components/certificates/cert-manager/self-signed-certificates.yaml
amisevsk: Hah -- I would've assumed the kustomize name-prefixing functionality applied to all resources it creates. I'll fix.

Makefile
amisevsk: Good catch, that's a woopsie :smile:

config/crd/kustomization.yaml
amisevsk: Removed the whole thing, thanks

controllers/controller/component/component_controller.go
```diff
@@ -95,6 +107,22 @@ func (r *ComponentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 	components = append(components, pluginComponents...)
 
+	initContainerComponents, err := adaptor.AdaptDockerimageComponents(instance.Spec.WorkspaceId, initContainers, []devworkspace.Command{})
+
+	for compIndex := range initContainerComponents {
```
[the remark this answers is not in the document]
amisevsk: I've reworked this section, it was pulled from David's changes largely unchanged initially.

controllers/controller/workspacerouting/solvers/cluster_solver.go
amisevsk: I switch it to use `Attributes.Strings()` instead since the commits seem to be out of date with `devfile/api`. This should get around the issue of errant `bool` or `int`s getting in there.

pkg/config/env.go
amisevsk: Snuck this one in a little earlier: 347a49464817bcddca42fe3fe1b94181302a4677

Makefile
```diff
@@ -13,14 +13,18 @@ SHELL := bash
...
+
 export NAMESPACE ?= devworkspace-controller
 export IMG ?= quay.io/devfile/devworkspace-controller:next
 export ROUTING_SUFFIX ?= 192.168.99.100.nip.io
 export PULL_POLICY ?= Always
 export WEBHOOK_ENABLED ?= true
 export DEFAULT_ROUTING ?= basic
 REGISTRY_ENABLED ?= true
-DEVWORKSPACE_API_VERSION ?= v1alpha1
+DEVWORKSPACE_API_VERSION ?= 1103496bda97425017e4f7fb01e4896ecfa2b5d8
```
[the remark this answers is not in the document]
amisevsk: Good point; I've also updated the default to the latest to pick up the `dw` shortname ;)

update_devworkspace_crds.sh
```diff
@@ -38,20 +38,20 @@ cd $TMP_DIR
...
+	git checkout --quiet "${DEVWORKSPACE_API_VERSION}"
 else
 	echo 'DevWorkspace API is specified from revision'
 	git checkout --quiet "${DEVWORKSPACE_API_VERSION}"
 fi
-
-cp deploy/crds/workspace.devfile.io_devworkspaces_crd.yaml \
-   $SCRIPT_DIR/config/crd/bases/workspace.devfile.io_devworkspaces.yaml
+cp crds/workspace.devfile.io_devworkspaces.yaml \
+   crds/workspace.devfile.io_devworkspacetemplates.yaml \
```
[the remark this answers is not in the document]
amisevsk: Good catch, thanks.

config/service-ca/crd_webhooks_patch.yaml
amisevsk: Good point; updated to `system` which is what we use elsewhere.

config/cert-manager/manager_certmanager_patch.yaml
amisevsk: This was due to `manager.yaml` containing an explicit `system` namespace it its yaml. This namespace would get prefixed like everything else and created.

config/components/certificates/cert-manager/self-signed-certificates.yaml
amisevsk: Turns out this issue was due to `config/components/certificates/cert-manager` being used as a base in `/config/cert-manager` and not in `config/base`, where the name prefix is specified. I've moved namePrefix to `config/cert-manager` and `config/service-sa` to compensate.

amisevsk:
All comments should be addressed now. Tested again on Kubernetes (minikube) and everything works. However, we still have an issue with updating existing CRDs:
- Existing resources are specified as `v1beta1` CRDs (`apiVersion: apiextensions.k8s.io/v1beta1`)
- We can't continue using `v1beta1` due to a validation error (we're using defaults now, which is unsupported)
- We can't apply `v1` CRDs due to `preserveUnknownFields`:
   > preserveUnknownFields indicates that object fields which are not specified
     in the OpenAPI schema should be preserved when persisting to storage.
     apiVersion, kind, metadata and known fields inside metadata are always
     preserved. If false, schemas must be defined for all versions. Defaults to
     true in v1beta for backwards compatibility. Deprecated: will be required to
     be false in v1.
    
    if preserveUnknownFields is set to true, we get the error 
    ```
    CustomResourceDefinition.apiextensions.k8s.io "devworkspaces.workspace.devfile.io" is invalid: 
    [spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the 
    schema, spec.conversion.strategy: Invalid value: "Webhook": must be None if 
    spec.preserveUnknownFields is true]
    ```
I'm looking into how to get around this, as it'll likely apply when we try to use these changes in WTO.

sleshchenko:
After pushing a little fix I managed to get working on minikube and crc theia.yaml and all-in-one. On crc additionally theia with openshift-oauth. Panels work just fine for me, the only broken stuff: terminal, which does not display or print in console error. But I'm not sure that it worked before this PR )

P.S. I had to decrease memoryLimits for containers 512 to 256 to get it working on crc. So, maybe to make it easier to test our sample - make 256 as the default.

<details>
<summary>Screenshots</summary>

minikube

minikube

CRC

Screenshot_20201126_173650
Screenshot_20201126_175346


</details>

sleshchenko (approved):
Great job! :muscle: 
I believe that's ready to be merged despite of CRDs issues you described which can be solved separately not to inflate this PR and get a better review then =)

sleshchenko (approved):
Great job! :muscle: 
I believe that's ready to be merged despite of CRDs issues you described which can be solved separately not to inflate this PR and get a better review then =)

amisevsk:
Pushed a couple of last-minute fixups:
- Removed bundle-related makefile rules, since they don't work with our current structure; we'll have to address this later. Operator SDK-generated manifests expect the default folder structure, which we don't/cannot use and don't include devworkspace CRDs
- Reorganized the `config/` directory a bit:
  - Moved `config/components/certificates/cert-manager` up to `config/components/cert-manager`, since I initially thought we might need other types of deployments for certificates (e.g. something for Service CA). Since there's only one subfolder it makes sense to simplify the structure for now
  - Removed namespace from `config/base` as it was hard-coded to devworkspace-operator
  - Removed common labels from `config/base` as they should be applied at the end (otherwise anything not used as a base in `config/base` will not get the labels
- Added a plantuml diagram of the current structure of the `config/` directory with some explanation of what each `kustomize.yaml` does:

Kustomize deployment structure

test/e2e/pkg/deploy/controller.go
```diff
@@ -64,13 +64,3 @@ func (w *Deployment) DeployWorkspacesController() error {
 
 	return nil
 }
-
-func (w *Deployment) CustomResourceDefinitions() error {
```
sleshchenko, line 68: Is it really removed from Makefile https://github.com/amisevsk/devworkspace-operator/blob/add-conversion-v1alpha2/Makefile ?
Or it just does not make sense to execute it separately since e2e tests do make install, which includes crds installing as well?

sleshchenko:
test is failing because it can't initialize CRDs:
```
Downloading devfile/api CRDs to /tmp/tmp.9VboUUlNdl
Initialized empty Git repository in /tmp/tmp.9VboUUlNdl/.git/
aeda60d4361911da85103f224644bfa792498499
DevWorkspace API is specified from branch
fatal: reference is not a tree: aeda60d4361911da85103f224644bfa792498499
make[1]: *** [_init_devworkspace_crds] Error 128
```

sleshchenko:
Something strange happens with tests. PR can be merged without this PR check if `make test_e2e` works locally

sleshchenko:
Found a bug which may cause tests failure on CI, at least it cause my local failure, controller tries to update components forever because of difference in sourceMapping

<details>
<summary>See diff from debug logs</summary>

```json
{
  "wanted": {
    "spec": {
      "workspaceId": "workspace267da45101b44759",
      "components": [
        {
          "name": "dev",
          "container": {
            "image": "quay.io/wto/web-terminal-tooling:latest",
            "env": [
              {
                "name": "PS1",
                "value": "\\[\\e[34m\\]>\\[\\e[m\\]\\[\\e[33m\\]>\\[\\e[m\\]"
              }
            ],
            "memoryLimit": "256Mi",
            "args": [
              "tail",
              "-f",
              "/dev/null"
            ]
          }
        }
      ]
    }
  },
  "clusterComponent": {
    "spec": {
      "workspaceId": "workspace267da45101b44759",
      "components": [
        {
          "name": "dev",
          "container": {
            "image": "quay.io/wto/web-terminal-tooling:latest",
            "env": [
              {
                "name": "PS1",
                "value": "\\[\\e[34m\\]>\\[\\e[m\\]\\[\\e[33m\\]>\\[\\e[m\\]"
              }
            ],
            "memoryLimit": "256Mi",
            "args": [
              "tail",
              "-f",
              "/dev/null"
            ],
            "sourceMapping": "/projects"
          }
        }
      ]
    }
  }
}
```

</details>

I assume component CRD needs to be updated.

amisevsk:
@sleshchenko I tested again on crc and was able to start up a cloud-shell workspace -- no idea why it's failing in CI now. I'll look into the souremapping issue, it may be causing a problem but it's not clear why.

amisevsk:
Merging to unblock all progress on the operator.

PR #203, opened 2020-12-01
Ensure created clusterrole and binding apply to controller deployment

amisevsk:
### What does this PR do?
Update kustomize configs to ensure subjects in clusterrole binding match
serviceaccount used for controller deployment

ClusterRoleBinding created for the controller references the default service account.

### Is it tested? How?
`make install`, but make sure you uninstall from the main branch first to clean up all resources.

### DIFF
--- a/config/components/rbac/role_binding.yaml
+++ b/config/components/rbac/role_binding.yaml
@@ -1,12 +1,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: manager-rolebinding
+  name: rolebinding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: manager-role
+  name: role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: $(CONTROLLER_SERVICE_ACCOUNT)
   namespace: system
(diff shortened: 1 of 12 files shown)


config/components/rbac/role_binding.yaml
```diff
@@ -1,12 +1,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: manager-rolebinding
+  name: rolebinding
```
sleshchenko, line 4: I wonder if it makes any sense to suffix objects' names with their kinds, we typically do it for RBAC only. But that's not only we who did it:
Screenshot_20201201_121849
So, I'm OK with leaving as is

sleshchenko (approved):
Found one more place where SA needs to be tuned, please review my fixup I've added.

LGTM

config/components/rbac/role_binding.yaml
amisevsk: The only reason I do it is because `devworkspace-controller-` is the prefix, and has to prefix something as far as I know :)

I'd love to find an alternative -- the main detail that's important for kustomize is that the name applied to the unpatched yamls can be referenced elsewhere (so we can e.g. get the serviceaccount named "serviceaccount" even though it gets prefixed).

amisevsk:
Good catch on the leader-election role -- I won't approve my own PR but it looks good!

JPinkney (approved):
LGTM

PR #206, opened 2020-12-01
Evaluate plugin registry URL depending on the cluster

sleshchenko:
### What does this PR do?
`make install` sets the right value for plugin registry URL for minikube, but for OpenShift clusters it's needed to be provided manually into configmap. This PR adds evaluating plugin registry URL depending on the cluster. So, for OpenShift users should not set anything, for k8s - ROUTING_SUFFIX is needed (which is evaluated automatically if it's minikube).

N/A

### Is it tested? How?
```bash
make install
#minikube
kc get ingress -n ${NAMESPACE} che-plugin-registry -o=yaml | yq -r '.spec.rules[].host'          
kubectl get configmaps -n ${NAMESPACE} devworkspace-controller-configmap -o=yaml | yq '.data["controller.plugin_registry.url"]'

# crc
# make sure the following values have the same host
kc get route -n ${NAMESPACE} che-plugin-registry -o=yaml | yq -r '.status.ingress[].host'
kubectl get configmaps -n ${NAMESPACE} devworkspace-controller-configmap -o=yaml | yq '.data["controller.plugin_registry.url"]'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,6 +227,12 @@ ifeq ($(PLATFORM),kubernetes)
 else
 	$(K8S_CLI) apply -f config/registry/local/os -n $(NAMESPACE)
 endif
+#evaluate plugin registry URL to use in other targets
+ifeq ($(PLATFORM),kubernetes)
+export PLUGIN_REGISTRY_URL=http://che-plugin-registry.$(ROUTING_SUFFIX)/v3)
+else
+export PLUGIN_REGISTRY_URL=http://$(shell $(K8S_CLI) get route -n $(NAMESPACE) che-plugin-registry -o=yaml | yq -r '.status.ingress[].host')/v3
+endif
 
 ### manifests: Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
--- a/Makefile
+++ b/Makefile
@@ -152,15 +152,15 @@ _login_with_devworkspace_sa:
 	oc login --token=$(SA_TOKEN) --kubeconfig=$(BUMPED_KUBECONFIG)
 
 ### run: Run against the configured Kubernetes cluster in ~/.kube/config
-run: _print_vars _gen_configuration_env _bump_kubeconfig _login_with_devworkspace_sa
+run: _print_vars _gen_configuration_env _bump_kubeconfig _login_with_devworkspace_sa _eval_plugin_registry_url
 	source $(RELATED_IMAGES_FILE)
 	export KUBECONFIG=$(BUMPED_KUBECONFIG)
 	CONTROLLER_SERVICE_ACCOUNT_NAME=default \
 		WATCH_NAMESPACE=$(NAMESPACE) \
 		go run ./main.go
 
-
-debug: _print_vars _gen_configuration_env _bump_kubeconfig _login_with_devworkspace_sa
+### debug: Run controller locally with debugging enabled, watching cluster defined in ~/.kube/config
+debug: _print_vars _gen_configuration_env _bump_kubeconfig _login_with_devworkspace_sa _eval_plugin_registry_url
 	source $(RELATED_IMAGES_FILE)
 	export KUBECONFIG=$(BUMPED_KUBECONFIG)
 	CONTROLLER_SERVICE_ACCOUNT_NAME=default \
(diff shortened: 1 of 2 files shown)


JPinkney (commented):
Should test-ctrl be replaced with `$(NAMESPACE)`?

Afterward when I ran `k get route -n test-ctrl che-plugin-registry -o=yaml | yq -r '.status.ingress[].host'` I got `che-plugin-registry-devworkspace-controller.apps-crc.testing` as expected

But for some reason the correct URL didn't get added in the configmap. I got: `http:///v3`

sleshchenko:
> Should test-ctrl be replaced with $(NAMESPACE)?

It must be! And now it is)

JPinkney:
I tried again this time on ocp and
`k get route -n devworkspace-controller che-plugin-registry -o=yaml | yq -r '.status.ingress[].host'` evaulates successfully but the `controller.plugin_registry.url` in the configmap still gets set to http:///v3

sleshchenko:
@amisevsk you're our Makefile master. Does it work for you? Any ideas why PLUGIN_REGISTRY_URL is propagated for me but not for Josh?

JPinkney:
Super strange. It failed the first few times but now its working

Makefile
```diff
@@ -227,6 +227,12 @@ ifeq ($(PLATFORM),kubernetes)
 else
 	$(K8S_CLI) apply -f config/registry/local/os -n $(NAMESPACE)
 endif
+#evaluate plugin registry URL to use in other targets
+ifeq ($(PLATFORM),kubernetes)
+export PLUGIN_REGISTRY_URL=http://che-plugin-registry.$(ROUTING_SUFFIX)/v3)
+else
+export PLUGIN_REGISTRY_URL=http://$(shell $(K8S_CLI) get route -n $(NAMESPACE) che-plugin-registry -o=yaml | yq -r '.status.ingress[].host')/v3
+endif
```
amisevsk, line 235: All lines that aren't part of a rule (since these are unindented they're not part of `deploy_registry`) are evaluated before any rules, so if the plugin registry is not already deployed this will export `PLUGIN_REGISTRY_URL=http:///v3`. E.g. using the Makefile

```makefile
default:        
        @echo default
$(info 'hi')
```
will output
```
❯ make
'hi'
default
```

One (extremely ugly) way to work around this would be to use `eval`:
```makefile
default: set_var
        @echo $(SOME_VAR) # works without export below
        @echo ${SOME_VAR} # works without export below
        @cat myfile.txt | envsubst # works *only* with export below

set_var:
        $(eval export SOME_VAR := "hi")
```

amisevsk (commented):
Make is so old and idisyncratic, it's _probably_ not the best tool for how complex our stuff is getting :/

In short, make will go through the file and evaluate all unindented lines (`ifdef`, etc. are more like preprocessor directives than shell conditionals) and then begin evaluating rules. There's no difference in terms of execution whether 
```makefile
export PLUGIN_REGISTRY_URL=http://$(shell $(K8S_CLI) get route -n $(NAMESPACE) che-plugin-registry -o=yaml | yq -r '.status.ingress[].host')/v3
```
is at line 234 or at the top of the file.

Makefile
amisevsk: We also might still run into an edge-case here, there's a chance the plugin registry route will not have been reconciled across the cluster and so may not have a URL set (I don't recall the exact mechanics here -- maybe it's always set).

amisevsk: Looking into the close-timing parts of it: `.status.ingress[].host` gets set once the route is ready, but `.spec.host` gets set when the route is created. If you check too soon after creating the route, there's a chance (I can reproduce it fairly often) that `.status.ingress` will be `null`.

amisevsk (approved):
Tested on crc and with the additional changes it seems to be in good shape. To test:
```bash
make uninstall # make sure nothing is currently installed
export NAMESPACE=test-namespace
make install
kubectl get cm devworkspace-controller-configmap -o yaml | yq .data # ensure registry is set
```
I also took the liberty of adding `_eval_registry_url` to the `run` and `debug` rules, but now I'm not sure it's needed.

sleshchenko:
works fine for me on crc and minikube

Issue #198, opened 2020-11-20
Verify if mkdir container can be safely removed 


sleshchenko:
DevWorkspace controller from POC stage run mkdir containers as init containers but that's not going to help solve files permissions issues if other init containers mounts some subfolders, because (as Che used to do it) subfolders must be initialized before they are mounted.

But since it works somehow with the current approach, and Angel heard that it might be resolved on k8s side(we don't have a good referene), WE MUST make sure it works on K8s/OpenShift clusters and remove mkdir init container at all,
OR such permissions issue still exists on some k8s/openshift clusters - we should rework it in the proper way, where subfolders are inited from a separate pod before they are mounted into devworkspace pod.

For more see https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L338

sleshchenko:
I'm not sure if it can be removed safely. Despite the fact that currently, DevWorkspace Operator has kind of no-op implementation and we don't have any issues on crc or minikube users still reports issues, like
> k8s 1.17 via eks as well
> https://github.com/kubernetes/kubernetes/issues/41638#issuecomment-680666823

Che (and I suppose CRW) has it enabled by default: 
[github.com/eclipse/che/...che.properties#che.infra.kubernetes.pvc.precreate_subpaths](https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L332)
We can drop it for now since the current implementation does not work as it's supposed to work but I think we should return back to it later, and implement a pluggable mechanism that would pre-create subpaths in the right way, with a dedicated pod (not init container). But that should be smarter that Che has currently and it does not need to be run on each workspace start. Volumes should be analyzed and precreated subpaths should be stored somewhere in PVC annotations to avoid unneeded runs.

sleshchenko:
It's proven that the current implementation with initContainer is not able to handle permissions issues in the proper way, so this functionality is cleaned up https://github.com/devfile/devworkspace-operator/pull/211

Init subpaths should be reimplemented in the proper way withing https://github.com/devfile/devworkspace-operator/issues/214

PR #211, opened 2020-12-02
Remove no-op mkdir init container

sleshchenko:
### What does this PR do?
Since we mount subpaths to other init containers, like https://github.com/devfile/devworkspace-operator/blob/master/samples/all-in-one-theia-nodejs.devworkspace.yaml#L112, this mkdir init containers is no-op and does nothing. See example from real Theia-Next pod
```json
{
  "image": "quay.io/eclipse/che-theia-endpoint-runtime-binary:next",
  "imagePullPolicy": "Always",
  "name": "remote-runtime-injector",
  "volumeMounts": [
    {
      "mountPath": "/remote-endpoint",
      "name": "claim-devworkspace",
      "subPath": "workspace0dbcaccab5834328/remote-endpoint/"
    }
  ]
}
{
  "args": [
    "-p",
    "-v",
    "-m",
    "777",
    "/tmp/devworkspaces/workspace0dbcaccab5834328" !!! <- it's no op because kubelet already should init this for remote-endpoint
  ],
  "command": [
    "/usr/bin/mkdir"
  ],
  "image": "registry.access.redhat.com/ubi8/ubi-minimal",
  "imagePullPolicy": "Always",
  "name": "precreate-subpaths",
  "volumeMounts": [
    {
      "mountPath": "/tmp/devworkspaces",
      "name": "claim-devworkspace"
    },
  ]
}

```

So, this PR just simply removes that init container and if we really need precreate subpaths on some of K8s clusters, we need to implement it with a dedicated pod that will be run before main workspace Deployment. 

https://github.com/devfile/devworkspace-operator/issues/198

### Is it tested? How?
It's not tested yet.

### DIFF
--- a/controllers/workspace/provision/deployment.go
+++ b/controllers/workspace/provision/deployment.go
@@ -201,7 +201,6 @@ func getSpecDeployment(
 	}
 
 	if IsPVCRequired(components) {
-		deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, precreateSubpathsInitContainer(workspace.Status.WorkspaceId))
 		deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, getPersistentVolumeClaim())
 	}
 
@@ -319,29 +318,4 @@ func getPersistentVolumeClaim() corev1.Volume {
 		},
 	}
 	return pvcVolume
-}
-
-func precreateSubpathsInitContainer(workspaceId string) corev1.Container {
-	initContainer := corev1.Container{
-		Name:    "precreate-subpaths",
-		Image:   "registry.access.redhat.com/ubi8/ubi-minimal",
-		Command: []string{"/usr/bin/mkdir"},
-		Args: []string{
-			"-p",
-			"-v",
-			"-m",
-			"777",
-			"/tmp/devworkspaces/" + workspaceId,
-		},
-		ImagePullPolicy: corev1.PullPolicy(config.ControllerCfg.GetSidecarPullPolicy()),
-		VolumeMounts: []corev1.VolumeMount{
-			{
-				MountPath: "/tmp/devworkspaces",
-				Name:      config.ControllerCfg.GetWorkspacePVCName(),
-				ReadOnly:  false,
-			},
-		},
-		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
-	}
-	return initContainer
 }


amisevsk (commented):
I'm generally for removing unnecessary code, but it's worth noting that the original issue the precreate-subpaths init container was made for still exists. I had to poke around old Che issues/PRs, but one of the original issues is in cleaning up workspace storage when using something like the common PVC strategy (sharing one PVC between workspaces). Using the yaml spec
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    openshift.io/scc: restricted
  name: test-pod
spec:
  containers:
  - image: quay.io/eclipse/che-theia:7.22.1
    imagePullPolicy: Always
    name: theia-ide
    securityContext:
      runAsUser: 1000710000
    volumeMounts:
    - mountPath: /dir1/
      name: claim-devworkspace
      subPath: test/path1
    - mountPath: /dir2/
      name: claim-devworkspace
      subPath: test/path1/path2/path3/path4
  restartPolicy: Always
  terminationGracePeriodSeconds: 1
  volumes:
  - name: claim-devworkspace
    persistentVolumeClaim:
      claimName: claim-devworkspace
```
you'll end up with directories `path2` and `path3` that can only be deleted by the root user:
```
ls -lR /dir1
/dir1:
total 0
drwxr-x---    3 0        root            19 Dec  2 17:41 path2

/dir1/path2:
total 0
drwxr-x---    3 0        root            19 Dec  2 17:41 path3

/dir1/path2/path3:
total 0
drwxrwx---    2 0        root             6 Dec  2 17:41 path4
```

sleshchenko:
@amisevsk Thanks for figuring out and reminding stuff with cleaning up files from PVC.

So, kind of proof that the current mkdir helps nothing with solving the permissions issue. Two workspaces created, `workspacecfcf8477f79945d7` one without mkdir. `workspace256986cb19144cc2` one with mkdir (master branch)
Trying to remove workspace files on PVC, with precreate and without works in the same way, despite of a bit different permissions:
Screenshot_20201203_122304
Screenshot_20201203_122323
Screenshot_20201203_122339


Pod that I used to try clean up workspace files:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: mount-root
spec:
  containers:
  - image: quay.io/eclipse/che-theia:7.22.1
    imagePullPolicy: Always
    name: theia-ide
    volumeMounts: 
        -
          "mountPath": "/tmp/devworkspaces"
          "name": "claim-devworkspace"
  restartPolicy: Always
  terminationGracePeriodSeconds: 1
  volumes:
  - name: claim-devworkspace
    persistentVolumeClaim:
      claimName: claim-devworkspace
```

sleshchenko:
I've created a separate issue to implement a proper mechanism to precreate and clean up subpaths once workspace is removed.

sleshchenko:
So, I would merge this PR once I retested that it does not break any devworkspace on minikube/crc/clusterbot platform to avoid confusing that this mkdir initContainers brings

sleshchenko:
The research above did not give me a clear understanding if mkdir as initContainer is really redundant or not :confused:

amisevsk:
Yeah I'm not entirely sure -- as far as I know our usage of PVCs shouldn't be impacted by the original issue, since I think the original problem was due to mounting something like `/workspaceId/workspace-logs/log-for-theia`, which would result in `workspace-logs` not having permissions to properly remove files.

Also I've noticed when running in `crc` and `minikube` that the UID behaviour is slightly different from production clusters I've used -- I'm able to set UID=0 on `crc` if I manually create a pod, and in that case I can delete anything I want.

amisevsk (approved):
I think we should merge this PR as-is. We don't need this functionality now, and whether we need it or not will be decided by how we implement persistent storage. It would not be too difficult to re-add it later if it turns out to be crucial.

sleshchenko:
On the minikube we have the issue with permissions but that preCreate init container does not help at all:
```bash
!!!Without Precreate

> kc exec -ti devworkspaces-pvc-5c57d86ff6-h7ls8 -- /bin/bash
bash-5.0$ cd /devworkspaces/
bash-5.0$ ls -la
total 32
drwxrwxrwx    8 0        root          4096 Dec  4 11:21 .
drwxr-xr-x    1 0        root          4096 Dec  4 11:28 ..
drwxr-xr-x    6 0        root          4096 Dec  4 11:21 workspace217a195d4747498d
bash-5.0$ rm -rf workspace217a195d4747498d
rm: can't remove 'workspace217a195d4747498d/remote-endpoint': Permission denied
rm: can't remove 'workspace217a195d4747498d/plugins': Permission denied
rm: can't remove 'workspace217a195d4747498d/plugins-node-debug-plugin': Permission denied
rm: can't remove 'workspace217a195d4747498d/projects': Permission denied

!!!With Precreate

> kc exec -ti devworkspaces-pvc-5c57d86ff6-mswp8 -- /bin/bash
bash-5.0$ rm -rf /devworkspaces/workspace8d2e5cfbf78b4b91
rm: can't remove '/devworkspaces/workspace8d2e5cfbf78b4b91/remote-endpoint': Permission denied
rm: can't remove '/devworkspaces/workspace8d2e5cfbf78b4b91/plugins': Permission denied
rm: can't remove '/devworkspaces/workspace8d2e5cfbf78b4b91/plugins-node-debug-plugin': Permission denied
rm: can't remove '/devworkspaces/workspace8d2e5cfbf78b4b91/projects': Permission denied

bash-5.0$ cd /devworkspaces
bash-5.0$ ls -la
total 36
drwxrwxrwx    9 0        root          4096 Dec  4 11:30 .
drwxr-xr-x    1 0        root          4096 Dec  4 11:32 ..
drwxr-xr-x    6 0        root          4096 Dec  4 11:21 workspace217a195d4747498d
drwxr-xr-x    6 0        root          4096 Dec  4 11:31 workspace8d2e5cfbf78b4b91

```
Screenshot_20201204_133326

After playing with mkdir pod command, I think it's clear that initContainer can not handle the permissions issue as I initially supposed because subfolders are already created with the wrong permissions and init container is not permitted to change it (without root access):
```go
		Args: []string{
			"-c",
			"ls -la /tmp/devworkspaces/ | grep " + workspaceId +
				"&& ls -la /tmp/devworkspaces/" + workspaceId +
				"&& mkdir -p -v -m 777 /tmp/devworkspaces/" + workspaceId +
				"&& chmod -R 777 /tmp/devworkspaces/" + workspaceId +
				"&& ls -la /tmp/devworkspaces/" + workspaceId,
		},
```
```bash
> kc logs workspace89407d97442642bd-7c6b94bb86-pwxkg -c precreate-subpaths
drwxr-xr-x  4 root root 4096 Dec  4 11:54 workspace89407d97442642bd
total 16
drwxr-xr-x  4 root root 4096 Dec  4 11:54 .
drwxrwxrwx 15 root root 4096 Dec  4 11:54 ..
drwxrwxrwx  3 root root 4096 Dec  4 11:54 plugins
drwxrwxrwx  2 root root 4096 Dec  4 11:54 remote-endpoint
chmod: changing permissions of '/tmp/devworkspaces/workspace89407d97442642bd': Operation not permitted
chmod: changing permissions of '/tmp/devworkspaces/workspace89407d97442642bd/remote-endpoint': Operation not permitted
chmod: changing permissions of '/tmp/devworkspaces/workspace89407d97442642bd/plugins': Operation not permitte
```

Issue #165, opened 2020-09-29
Workspaces should fail when we can't find a plugin


amisevsk:
Currently, if a workspace specifies a plugin that cannot be retrieved, the controller retries on a loop (with backoff). Instead we should fail workspace startup with an understandable message.

PR #212, opened 2020-12-02
Fail devworkspace startup if we fail to download plugin metadata

amisevsk:
### What does this PR do?
Add structure to Components subresource to allow unrecoverable errors to be propagated to the DevWorkspace controller, and make the DevWorkspace controller fail DevWorkspace startup if an unrecoverable error occurs

Fixes https://github.com/devfile/devworkspace-operator/issues/165

### Is it tested? How?
Tested on `crc`:
1. Deploy controller as normal
2. Apply theia devworkspace with nonexisting plugin:
    ```yaml
    kind: DevWorkspace
    metadata:
      name: theia
    spec:
      started: true
      routingClass: 'basic'
      template:
        projects:
          - name: project
            git:
              remotes:
        components:
          - plugin:
              id: eclipse/che-does-not-exist/latest
          - plugin:
        commands:
          - exec:
              id: say hello
              component: plugin
    ```
3. DevWorkspace status should be "Failed", and failed condition should mention that the plugin failed to download:
    ```
    ❯ kc get dw theia -o yaml | yq -y .status
    conditions:
      - lastTransitionTime: '2020-12-02T20:16:30Z'
        message: Failed to download plugin meta.yaml for eclipse/che-does-not-exist/latest
        status: 'True'
        type: FailedStart
    phase: Failed
    workspaceId: workspace1aff901509704461
    ```

### DIFF
--- a/pkg/adaptor/errors.go
+++ b/pkg/adaptor/errors.go
@@ -1,0 +1,27 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package adaptor
+
+// DownloadMetasError represents an error that occurs while downloading plugin meta.yamls
+// This error wraps the underlying error that caused the failure.
+type DownloadMetasError struct {
+	Plugin string
+	Err    error
+}
+
+var _ error = (*DownloadMetasError)(nil)
+
+func (e *DownloadMetasError) Error() string {
+	return "Failed to download plugin meta.yaml for " + e.Plugin
+}
+func (e *DownloadMetasError) Unwrap() error { return e.Err }
--- a/apis/controller/v1alpha1/component_types.go
+++ b/apis/controller/v1alpha1/component_types.go
@@ -33,10 +33,16 @@ type WorkspaceComponentSpec struct {
 // ComponentStatus defines the observed state of Component
 // +k8s:openapi-gen=true
 type WorkspaceComponentStatus struct {
-	// Whether the component has finished processing its spec
+	// Ready indicates whether the component has finished processing its spec
 	Ready bool `json:"ready"`
-	// Descriptions of processed components from spec
-	ComponentDescriptions []ComponentDescription `json:"componentDescriptions"`
+	// Failed indicates that an unresolvable problem prevents this component
+	// from being ready.
+	Failed bool `json:"failed,omitempty"`
+	// Message stores additional context about the Component's current state (e.g.
+	// reason for failure)
+	Message string `json:"message,omitempty"`
+	// ComponentDescriptions of processed components from spec
+	ComponentDescriptions []ComponentDescription `json:"componentDescriptions,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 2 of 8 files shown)


JPinkney (approved):
LGTM

sleshchenko:
Nitpick: For easier/faster reviewing, it's better to have the full command for execution, like for
> Apply theia devworkspace with nonexisting plugin:
```yaml
cat <<EOF | oc apply -f -
kind: DevWorkspace
metadata:
  name: theia
spec:
  started: true
  routingClass: 'basic'
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - plugin:
          id: eclipse/che-does-not-exist/latest
      - plugin:
    commands:
      - exec:
          id: say hello
          component: plugin
EOF
```
You still can use yaml as block language and it does not hurt a lot.

pkg/adaptor/errors.go
```diff
@@ -0,0 +1,27 @@
...
+// This error wraps the underlying error that caused the failure.
+type DownloadMetasError struct {
+	Plugin string
+	Err    error
+}
+
+var _ error = (*DownloadMetasError)(nil)
+
+func (e *DownloadMetasError) Error() string {
+	return "Failed to download plugin meta.yaml for " + e.Plugin
```
sleshchenko, line 25: Operator admin could check logs for more details:
```
2020-12-03T13:12:38.626+0200    INFO    controllers.Component   Failed to download plugin metas {"Request.Namespace": "devworkspace-controller", "Request.Name": "components-workspace827d4c63d6bb4955-plugins", "err": "failed to fetch plugin meta.yaml from URL 'http://che-plugin-registry-devworkspace-controller.apps.ci-ln-27h2cmk-f76d1.origin-ci-int-gce.dev.openshift.com/v3/plugins/eclipse/che-does-not-exist/latest/meta.yaml': Downloading http://che-plugin-registry-devworkspace-controller.apps.ci-ln-27h2cmk-f76d1.origin-ci-int-gce.dev.openshift.com/v3/plugins/eclipse/che-does-not-exist/latest/meta.yaml failed. Status code 404. Response body: <!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">\n<html><head>\n<title>404 Not Found</title>\n</head><body>\n<h1>Not Found</h1>\n<p>The requested URL was not found on this server.</p>\n</body></html>\n"}
```
But from a user-perspective, it may be really difficult to figure out the cause, like if default plugin registry is misconfigured, operator has bug and construct wrong URL. So
maybe it makes sense to add here at least plugin URL, like `Failed to download plugin registry meta.yaml for e.Plugin. Please check e.URL`

sleshchenko (approved):
Works as designed. Just one small proposal to be more user-oriented.

amisevsk:
> Nitpick: For easier/faster reviewing, it's better to have the full command for execution

Yeah I thought about that but I've been historically too lazy to memorize the 
```bash
cat <<EOF | oc apply -f -
//...
EOF
```
incantation :)

I'll do it in the future.

pkg/adaptor/errors.go
amisevsk: This is actually tricky -- plugin resolution is handled completely by the plugin broker, which currently does not offer a convenient way of finding the URL we'll eventually use to try and download the plugin. We could potentially grab it from the underlying error message, but it's tricky as that tends to log the full response, which in the case of OpenShift is an HTML page sometimes.

The correct way to surface the URL we're trying would be to break [this section](https://github.com/eclipse/che-plugin-broker/blob/92c3b00094475b3a157de6dd892ec28119de763f/utils/metadata_resolver.go#L36-L46) of the broker out into a separate function, release the broker, and then use that here.

amisevsk:
I'm going to merge this PR as-is, as we're likely not going to be using the plugin broker for much longer (as we move to a different plugin mechanism in devfile/api v1alpha2) and fixing Serhii's suggestion (though it's a good idea) would be a fair amount of work + a plugin broker release.

sleshchenko:
> Yeah I thought about that but I've been historically too lazy to memorize the

I always search for EOF in my bash history (Ctrl+E) to find this command each time I need it, it's why it asked you to provide it in the PRs. Who is lazy? :-D

PR #213, opened 2020-12-03
Single Host + TLS on OpenShift fixup

sleshchenko:
### What does this PR do?
:man_facepalming: it brings changes which I made but did not push for https://github.com/devfile/devworkspace-operator/pull/201 before merging

### DIFF
--- a/internal/map/map.go
+++ b/internal/map/map.go
@@ -1,0 +1,21 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package maputils
+
+func Append(target map[string]string, key, value string) map[string]string {
+	if target == nil {
+		target = map[string]string{}
+	}
+	target[key] = value
+	return target
+}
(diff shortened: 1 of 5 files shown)


internal/map/map.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package maputils
+
+func Append(target map[string]string, key, value string) map[string]string {
```
amisevsk, line 15: One minor detail in this implementation: this approach has a side-effect if you pass in a non-nil map, which could be a little confusing:
```go
	var m map[string]string
	Append(m, "one", "two")
	fmt.Printf("%+v\n", m) // Prints nil map
```
```go
	m := map[string]string{"one": "two"}
	Append(m, "one", "three")
	fmt.Printf("%+v\n", m) // Prints map[one:three]
```
compare this to the built-in append, which does not modify the original argument.

amisevsk (commented):
One minor comment

internal/map/map.go
sleshchenko: You're right.

I misunderstood the documentation:

> // If it has sufficient capacity, the destination is resliced to accommodate the
> // new elements.

I also considered naming like `Put(...)` or `SafelyPut(...)`
@amisevsk Any ideas on how we can avoid confusion? Do you think it's better to use any of the names above?

sleshchenko: P.S. I wish golang allows enable compiler check they have for build-in append, where return value must be used or explicitly ignored.
Screenshot_20201204_110759

PR #215, opened 2020-12-04
Fix type in Makefile when evaluating plugin registry URL

amisevsk:
### What does this PR do?
Fix a typo that causes `make  _eval_plugin_registry_url` to fail on Kubernetes

My mistake

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,7 +298,7 @@ install_cert_manager:
 # export plugin registry url for other rules to use
 _eval_plugin_registry_url: deploy_registry
 ifeq ($(PLATFORM),kubernetes)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -298,7 +298,7 @@ install_cert_manager:
 # export plugin registry url for other rules to use
 _eval_plugin_registry_url: deploy_registry
 ifeq ($(PLATFORM),kubernetes)
-	$(eval export PLUGIN_REGISTRY_URL=http://che-plugin-registry.$(ROUTING_SUFFIX)/v3))
+	$(eval export PLUGIN_REGISTRY_URL=http://che-plugin-registry.$(ROUTING_SUFFIX)/v3)
```
sleshchenko, line 301: How it's possible?!!! I'd tested it! :man_facepalming:

PR #216, opened 2020-12-04
Add support for workspacerouting managed by external controller

amisevsk:
### What does this PR do?
Instead of failing a workspace when the routingClass is unsupported, apply an annotation (`controller.devfile.io/external_routing`) to signify that we expect something else to reconcile the object.

This is to allow an implementation-specific controller to reconcile a workspace routing and have it still be used by the devworkspace controller.

Resolves https://github.com/eclipse/che/issues/18520

### Is it tested? How?
```bash
cat <<EOF | oc apply -f -
kind: DevWorkspace
metadata:
  name: theia-failed
spec:
  routingClass: does-not-exist
  started: true
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - name: theia
        plugin:
          id: eclipse/che-theia/next
      - name: terminal
        plugin:
          id: eclipse/che-machine-exec-plugin/nightly
    commands:
      - id: say hello
        exec:
          component: plugin
EOF
```

1. Message `Could not find solver for routing class` is logged by workspacerouting controller
2. Workspace is stuck in "Starting" phase (as no controller for `routingClass: does-not-exist` exists)
3. Workspace has annotation `controller.devfile.io/external_routing`:
    ```
    $ kc get workspacerouting -o yaml | yq '.items[].metadata.annotations'
    {controller.devfile.io/external_routing
      "controller.devfile.io/external_routing": "true"
    }
    ```

### DIFF
--- a/controllers/controller/workspacerouting/predictes.go
+++ b/controllers/controller/workspacerouting/predictes.go
@@ -1,0 +1,57 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package workspacerouting
+
+import (
+	"errors"
+	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var routingPredicates = predicate.Funcs{
+	CreateFunc: func(ev event.CreateEvent) bool {
+		obj, ok := ev.Object.(*controllerv1alpha1.WorkspaceRouting)
+		if !ok {
+			return true
+		}
+		if _, err := getSolverForRoutingClass(obj.Spec.RoutingClass); errors.Is(err, externalRoutingError) {
+			return false
+		}
+		return true
+	},
+	DeleteFunc: func(_ event.DeleteEvent) bool {
+		// Return true to ensure finalizers are removed
+		return true
+	},
+	UpdateFunc: func(ev event.UpdateEvent) bool {
+		newObj, ok := ev.ObjectNew.(*controllerv1alpha1.WorkspaceRouting)
+		if !ok {
+			return true
+		}
+		if _, err := getSolverForRoutingClass(newObj.Spec.RoutingClass); errors.Is(err, externalRoutingError) {
+			return false
+		}
+		return true
+	},
+	GenericFunc: func(ev event.GenericEvent) bool {
+		obj, ok := ev.Object.(*controllerv1alpha1.WorkspaceRouting)
+		if !ok {
+			return true
+		}
+		if _, err := getSolverForRoutingClass(obj.Spec.RoutingClass); errors.Is(err, externalRoutingError) {
+			return false
+		}
+		return true
+	},
+}
--- a/controllers/controller/workspacerouting/predictes.go
+++ b/controllers/controller/workspacerouting/predictes.go
@@ -14,6 +14,7 @@ package workspacerouting
 
 import (
 	"errors"
+
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
--- a/controllers/controller/workspacerouting/workspacerouting_controller.go
+++ b/controllers/controller/workspacerouting/workspacerouting_controller.go
@@ -74,10 +74,6 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 		return reconcile.Result{}, err
 	}
 
-	if instance.Annotations[config.WorkspaceExternalRoutingAnnotation] != "" {
-		return reconcile.Result{}, nil
-	}
-
 	// Check if the WorkspaceRouting instance is marked to be deleted, which is
 	// indicated by the deletion timestamp being set.
 	if instance.GetDeletionTimestamp() != nil {
--- a/controllers/controller/workspacerouting/workspacerouting_controller.go
+++ b/controllers/controller/workspacerouting/workspacerouting_controller.go
@@ -16,6 +16,8 @@ import (
 	"context"
 	"fmt"
 
+	maputils "github.com/devfile/devworkspace-operator/internal/map"
+
 	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers"
 	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/google/go-cmp/cmp"
@@ -71,6 +73,10 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 		return reconcile.Result{}, err
 	}
 
+	if instance.Annotations[config.WorkspaceExternalRoutingAnnotation] != "" {
+		return reconcile.Result{}, nil
+	}
+
 	// Check if the WorkspaceRouting instance is marked to be deleted, which is
 	// indicated by the deletion timestamp being set.
 	if instance.GetDeletionTimestamp() != nil {
@@ -78,16 +84,20 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 		return reconcile.Result{}, r.finalize(instance)
 	}
 
-	if instance.Status.Phase == controllerv1alpha1.RoutingFailed {
-		return reconcile.Result{}, nil
-	}
-
 	solver, err := getSolverForRoutingClass(instance.Spec.RoutingClass)
 	if err != nil {
-		reqLogger.Error(err, "Could not get solver for routingClass")
-		instance.Status.Phase = controllerv1alpha1.RoutingFailed
-		statusErr := r.Status().Update(ctx, instance)
-		return reconcile.Result{}, statusErr
+		// No solver in this controller for this routingClass
+		reqLogger.Info("Could not find solver for routing class; assuming external routing", "routingClass", instance.Spec.RoutingClass)
+		instance.Annotations = maputils.Append(instance.Annotations, config.WorkspaceExternalRoutingAnnotation, "true")
+		updateErr := r.Update(ctx, instance)
+		if updateErr != nil && errors.IsConflict(err) {
+			return reconcile.Result{Requeue: true}, nil
+		}
+		return reconcile.Result{}, updateErr
+	}
+
+	if instance.Status.Phase == controllerv1alpha1.RoutingFailed {
+		return reconcile.Result{}, nil
 	}
 
 	// Add finalizer for this CR if not already present
@@ -245,7 +255,7 @@ func getSolverForRoutingClass(routingClass controllerv1alpha1.WorkspaceRoutingCl
 		}
 		return &solvers.ClusterSolver{TLS: true}, nil
 	default:
-		return nil, fmt.Errorf("routing class %s not supported", routingClass)
+		return nil, fmt.Errorf("routing class %s not supported in this controller", routingClass)
 	}
 }
 
--- a/controllers/controller/workspacerouting/workspacerouting_controller.go
+++ b/controllers/controller/workspacerouting/workspacerouting_controller.go
@@ -14,27 +14,28 @@ package workspacerouting
 
 import (
 	"context"
+	"errors"
 	"fmt"
-
-	maputils "github.com/devfile/devworkspace-operator/internal/map"
 
 	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers"
 	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/google/go-cmp/cmp"
+
+	"github.com/go-logr/logr"
 	routeV1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	"github.com/go-logr/logr"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 )
+
+var externalRoutingError = errors.New("routingclass not supported by this controller")
 
 const workspaceRoutingFinalizer = "workspacerouting.controller.devfile.io"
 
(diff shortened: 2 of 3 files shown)


controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -71,23 +73,31 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
...
 		return reconcile.Result{}, r.finalize(instance)
 	}
 
-	if instance.Status.Phase == controllerv1alpha1.RoutingFailed {
-		return reconcile.Result{}, nil
-	}
-
 	solver, err := getSolverForRoutingClass(instance.Spec.RoutingClass)
 	if err != nil {
-		reqLogger.Error(err, "Could not get solver for routingClass")
```
metlos, line 87: This is not correct, because there can be other error conditions than just an unknown solver.

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -71,23 +73,31 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 		return reconcile.Result{}, err
 	}
 
+	if instance.Annotations[config.WorkspaceExternalRoutingAnnotation] != "" {
```
metlos, line 76: Do we really have to have this annotation set? "External" is a little bit loosely defined here and it doesn't really bring much of a performance advantage.

sleshchenko: I'm not sure how much this annotation provides info for the user.
How many devworkspace operators are going to provide embedded routing?
At times Che and WebTerminal operators will depend on the DevWorkspace Operator, we may even consider moving webterminal routing to a dedicated controller.
So, I'm a bit more for just ignoring unsupported classes.

A good example of an external controller K8s has: Ingresses. So, when you create ingress with an unknown ingress class - just nothing happens.
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: not-existing
  creationTimestamp: "2020-12-04T09:20:58Z"
  generation: 1
  labels:
    app.kubernetes.io/name: che-plugin-registry
    app.kubernetes.io/part-of: devworkspace-operator
  name: che-plugin-registry
spec:
  rules:
  - host: che-plugin-registry.192.168.39.170.nip.io
    http:
      paths:
      - backend:
          serviceName: che-plugin-registry
          servicePort: 8080
        path: /
        pathType: ImplementationSpecific
status:
  loadBalancer: {}
```
Also, interesting how it's handled that multiple controllers listen to the same class. Probably it's not handled and two controllers just race for CR instance.

metlos: If we want to add some information about what controller handles a certain routing what about calling it `controller.devfile.io/handled-by` with the namespaced name of the handling controller? Each controller would be in charge of adding this info though and it would be purely informational.

controllers/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: Good catch.

controllers/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: > controller.devfile.io/handled-by with the namespaced name of the handling controller

then WorkspaceRouting CRs could get in trouble if the operator is reinstalled in another namespace, which is not a common case but still complicates things.

So, if we want to expose to the end-user that some operator is handling routing or not, it could be handled maybe better with status conditions.

If there is a condition - Initializing, somebody is handling CR, 
if conditions are empty - nobody works with your routingClass. It can be propagated then to WorkspaceConditions.

sleshchenko, line 76: If the user for some reason(occasionally) annotates the embedded routing with it, then such CR is going to hang on removal since finalizers won't be handled.

amisevsk: The main reason for the annotation is to stop the reconcile early. Otherwise the devworkspace operator logs would be full of "could not find solver for routing class" as the other controller reconciles the workspaceRouting.

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: That's what I get for throwing together a PR at the end of the day :smile: 

I'll fix it.

controllers/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: > The main reason for the annotation is to stop the reconcile early. Otherwise the devworkspace operator logs would be full of "could not find solver for routing class" as the other controller reconciles the workspaceRouting.


How then multiple external workspaceRouting controllers is supposed to early stop reconcile?

amisevsk: In general we'll have a tricky situation in switching between a supported routingclass and an external one -- if we initially set e.g. `basic`, and then switch to `che-workspace`, the controller can't know to remove old routing objects unless it also risks removing routing objects created by the external controller. I'll look into it more.

Should we make routingclass immutable once a workspace is created? I think it might make things cleaner here.

sleshchenko: > Should we make routingclass immutable once a workspace is created? I think it might make things cleaner here.

Sounds good to me

davidfestal: > So, I'm a bit more for just ignoring unsupported classes.
> 
> A good example of an external controller K8s has: Ingresses. So, when you create ingress with an unknown ingress class - just nothing happens.
>
> ...
> 
> Also, interesting how it's handled that multiple controllers listen to the same class. Probably it's not handled and two controllers just race for CR instance.

That was exactly how I initially thought it when designing the `WorkspaceRouting` (initially `WorkspaceExposure` :-) ) in the POC.
And Ingresses being managed by external controllers has been the model that drove this design.

If possible, I'd also prefer just doing nothing special when a routing class is not supported.
However we may want to define a timeout.
Let me explain: when the main controller created / updated a `WorkspaceRouting` object, its routing class is not supported by the main controller, and its reconciliation didn't start after a give timeout, just make it fail.
And possibly the error log about "solver not found" would only be added at this time.

Do you think it would be possible ?

davidfestal: Not sure I'm a bit fan of this limitation, unless it cannot be avoided

davidfestal: Can't this be fixed by what @sleshchenko proposed in comment https://github.com/devfile/devworkspace-operator/pull/216#discussion_r535979658 ?
If in the status you add the `routingClass` that was used during the last reconciliation, then that means that each controller would also be able to read this value and perform any required cleaning, no ?

davidfestal (commented):
It would be great if as much code as possible could be imported and reused from the devorkspace controller GH repo, when writing an external workspace routing controller.

Even the main reconciliation loop seem to be reusable, if fed by a list of supported solver instances.

Being able to "only write the solver" and reuse the rest by imports would be great when writing external routing controllers.

I assume some (light ?) refactoring could allow that. Do you think it could be part of the PR ?

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -85,15 +86,11 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
...
-	if err != nil {
-		// No solver in this controller for this routingClass
-		reqLogger.Info("Could not find solver for routing class; assuming external routing", "routingClass", instance.Spec.RoutingClass)
-		instance.Annotations = maputils.Append(instance.Annotations, config.WorkspaceExternalRoutingAnnotation, "true")
-		updateErr := r.Update(ctx, instance)
-		if updateErr != nil && errors.IsConflict(err) {
-			return reconcile.Result{Requeue: true}, nil
-		}
-		return reconcile.Result{}, updateErr
+	if err != nil && !errors.Is(err, externalRoutingError) {
```
davidfestal, line 85: The `externalRoutingError` should never happen right ?

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: I've moved to using predicates to filter unsupported routingclass from ever being reconciled, so the annotation is removed. I'll look into a timeout function (though it might be tricky to implement) and potentially add it in a separate PR.

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Removed

amisevsk:
Update on the current state of this PR:
1. I've removed the external routing annotation; instead events are filtered so workspaceroutings with unsupported routingclass are never reconciled by our controller. Deletion events always reach the controller so that finalizers can be cleared
2. It looks like switching routingClass will not be a problem; we already have it set up so that routingclass in workspacerouting is never changed -- if the devworkspace has an updated routingClass, the old workspacerouting is deleted first.

I'm looking into the timeout and reusability questions separately and will likely file those as a separate PR.

sleshchenko (commented):
1. Are we going to propagate DevWorkspace Operator ServiceAccount to external routing controllers? If no, then possible mutatingWebhooks are going to block external routing controllers. Possible ways I see to solve it:
1.1. DevWorkspace Operator denies to modify only restricted-access workspace related objects which are related to embedded routing classes; 
1.2. we declare that workspaceroutings controllers should fail restricted access workspace if they do not brings their specific webhooks;

2. Seems it works in the wrong way when we modify routing class on the instance which is in starting state:
```bash
oc apply -f ./samples/theia-next.yaml
sleep 2
oc patch dw theia --patch '{ "spec": { "routingClass": "non-existing" } }' --type=merge
sleep 2
oc get dw theia -o=yaml | yq '.status'
```
Expected: dw is in starting for a while
Actual: dw is failed to start with `Failed to install network objects required for devworkspace`

controllers/controller/workspacerouting/predictes.go
```diff
@@ -0,0 +1,58 @@
...
+	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var routingPredicates = predicate.Funcs{
+	CreateFunc: func(ev event.CreateEvent) bool {
+		obj, ok := ev.Object.(*controllerv1alpha1.WorkspaceRouting)
+		if !ok {
+			return true
```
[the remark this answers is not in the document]
amisevsk: This could be come a reconcile issue -- we could apply an annotation saying `routingClass: basic` or something, but we'd still have to trigger reconciles for resources with no annotation. Otherwise, if the annotation is removed for some reason, it would no longer trigger reconciles on modification, potentially breaking routing. We could make some sort of agreement to add a "managed-by" annotation to externally-created routing objects, but I don't know if it makes sense to handle that now.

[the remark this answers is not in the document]
sleshchenko: That should not block that PR

sleshchenko, line 27: Please add a comment that describes that it's a change related to Service/Route/Ingress related to routing. So, triggering reconcile loop which will be processed or ignored by controller...

Will it be really ignored?... I just don't understand in which place, so I need to play with it.

controllers/controller/workspacerouting/predictes.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+		if _, err := getSolverForRoutingClass(obj.Spec.RoutingClass); errors.Is(err, externalRoutingError) {
+			return false
+		}
+		return true
+	},
+	DeleteFunc: func(_ event.DeleteEvent) bool {
+		// Return true to ensure finalizers are removed
```
sleshchenko, line 35: That's also invoked when Service or Ingress is removed, right?
If so - this comment is a bit misleading, so - we should process:
- all Workspace Routing independent from their ingresses routes 
- all other routing-related objects (Service, Ingress, .. ) to recreate object if needed.

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -14,26 +14,29 @@ package workspacerouting
...
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 )
 
+var externalRoutingError = errors.New("routingclass not supported by this controller")
```
sleshchenko, line 38: I think `routingIsNotSupported` would be a bit clearer

sleshchenko (commented):
LGTM
I assume I just need to play with changes to understand better how event filters work. Will do soon.

controllers/controller/workspacerouting/predictes.go
sleshchenko: I got a failed workspace once I manually created Service with manually set ownerRef, I believe it means that external routing is not going to work even with these changes.

To reproduce:
```
# do make run in another tab

# Change it when you run in another namespace

kubectl apply -f ./samples/cloud-shell.yaml
sleep 2
DW_ID=$(kubectl get dw cloud-shell -o=yaml | yq '.status.workspaceId' -r)

# kubectl patch dw/cloud-shell --patch '{"spec":{"routingClass":"external"}}'
# ^ for some reason returns Error from server (UnsupportedMediaType): the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml
# as well as 
#kubectl patch dw/cloud-shell --patch '
#spec:
#  routingClass: openshift-oauth
#'

# I'm smart!!!
kubectl get dw cloud-shell -o=yaml | sed 's/routingClass: openshift-oauth/routingClass: external/' | kubectl apply -f -
# -D

ROUTING_UID=$(kubectl get workspaceroutings.controller.devfile.io routing-${DW_ID} -o=yaml | yq '.metadata.uid' -r)


cat <<EOF | KUBECONFIG=/tmp/devworkspace-controller/kubeconfig kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  labels:
    controller.devfile.io/workspace_id: ${DW_ID}
  name: ${DW_ID}-service
  ownerReferences:
  - apiVersion: controller.devfile.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: WorkspaceRouting
    name: routing-${DW_ID}
    uid: ${ROUTING_UID}
spec:
  ports:
  - name: cloud-shell-proxy
    port: 4400
    protocol: TCP
    targetPort: 4400
  selector:
    controller.devfile.io/workspace_id: workspace7126ed1fe35c47f5
EOF
sleep 1
kc get dw cloud-shell -o=yaml | yq '.status'
#expected workspace is in starting phase
#actual it's failed
```

sleshchenko (changes requested):
That PR does not seem to solve the issue with external routing.

I think it makes sense to proceed with one PR (this + https://github.com/devfile/devworkspace-operator/pull/219) does bring external routing support

sleshchenko:
This does not seem to be addressed or answered as well.

> Are we going to propagate DevWorkspace Operator ServiceAccount to external routing controllers? If no, then possible mutatingWebhooks are going to block external routing controllers. Possible ways I see to solve it:
1.1. DevWorkspace Operator denies to modify only restricted-access workspace related objects which are related to embedded routing classes;
1.2. we declare that workspaceroutings controllers should fail restricted access workspace if they do not brings their specific webhooks;

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Ha, turns out it *can* happen, as we still trigger events for workspace services regardless of routingClass

amisevsk:
I'm closing this PR in favor of https://github.com/devfile/devworkspace-operator/pull/219; I initially split them because I thought this PR was near merging but if more work needs to be done we should just use https://github.com/devfile/devworkspace-operator/pull/219

controllers/controller/workspacerouting/predictes.go
amisevsk: Fixed by https://github.com/devfile/devworkspace-operator/pull/219/commits/0c70133bee407b27fa4277bec3348ea80069c086 -- turns out my reasoning was wrong in this thread: https://github.com/devfile/devworkspace-operator/pull/216#discussion_r536228115

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Fixed in #219

controllers/controller/workspacerouting/predictes.go
amisevsk: Updated the comment in #219

PR #219, opened 2020-12-04
Add support for workspacerouting managed by external controller and make routing controller reusable

amisevsk:
### What does this PR do?
Enables support for alternate controllers for a routingClass in the workspacerouting controller.

* Make the sync-to-cluster functionality of the WorkspaceRouting controller reusable. This is done by making the `getSolverForRoutingClass` function a field in the `WorkspaceRoutingController` struct, allowing a custom solver to be provided for a given routingClass.
* Configure controllers to propagate the `restricted-access` annotation to *all* created objects
* Update webhooks to only enforce immutability on objects with the `restricted-access` annotation, to allow for an external controller to update/create resources.

Resolves eclipse/che#18520

### Is it tested? How?
For our normal samples, nothing should be different.

```bash
cat <<EOF | oc apply -f -
kind: DevWorkspace
metadata:
  name: theia-external
spec:
  routingClass: external-sample
  started: true
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - name: theia
        plugin:
          id: eclipse/che-theia/next
      - name: terminal
        plugin:
          id: eclipse/che-machine-exec-plugin/nightly
    commands:
      - id: say hello
        exec:
          component: plugin
EOF
```

1. Workspace deployment does not fail, but hangs with no URL in status field. Logs in devworkspace operator state "waiting on routings to be ready". 
2. Deploy ["external" sample operator](https://github.com/amisevsk/devworkspace-controller-external-routing) which is designed to provide routing for `routingClass: external-sample`
3. Routing becomes ready, a service is created
4. Workspace "starts" (but cannot be accessed since sample routing controller implementation is just a test)

### DIFF
--- a/controllers/controller/workspacerouting/predicates.go
+++ b/controllers/controller/workspacerouting/predicates.go
@@ -25,6 +25,9 @@ func getRoutingPredicatesForSolverFunc(getSolver WorkspaceRoutingSolverFunc) pre
 		CreateFunc: func(ev event.CreateEvent) bool {
 			obj, ok := ev.Object.(*controllerv1alpha1.WorkspaceRouting)
 			if !ok {
+				// If object is not a WorkspaceRouting, it must be a service/ingress/route related to the workspace
+				// The safe choice here is to trigger a reconcile to ensure that all resources are in sync; it's the job
+				// of the controller to ignore WorkspaceRoutings for other routing classes.
 				return true
 			}
 			if _, err := getSolver(obj.Spec.RoutingClass); errors.Is(err, RoutingNotSupported) {
@@ -33,12 +36,16 @@ func getRoutingPredicatesForSolverFunc(getSolver WorkspaceRoutingSolverFunc) pre
 			return true
 		},
 		DeleteFunc: func(_ event.DeleteEvent) bool {
-			// Return true to ensure finalizers are removed
+			// Return true to ensure objects are recreated if needed, and that finalizers are
+			// removed on deletion.
 			return true
 		},
 		UpdateFunc: func(ev event.UpdateEvent) bool {
 			newObj, ok := ev.ObjectNew.(*controllerv1alpha1.WorkspaceRouting)
 			if !ok {
+				// If object is not a WorkspaceRouting, it must be a service/ingress/route related to the workspace
+				// The safe choice here is to trigger a reconcile to ensure that all resources are in sync; it's the job
+				// of the controller to ignore WorkspaceRoutings for other routing classes.
 				return true
 			}
 			if _, err := getSolver(newObj.Spec.RoutingClass); errors.Is(err, RoutingNotSupported) {
--- a/controllers/controller/workspacerouting/predicates.go
+++ b/controllers/controller/workspacerouting/predicates.go
@@ -13,14 +13,14 @@
 package workspacerouting
 
 import (
-	"errors"
+	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers"
 
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-func getRoutingPredicatesForSolverFunc(getSolver WorkspaceRoutingSolverFunc) predicate.Funcs {
+func getRoutingPredicatesForSolverFunc(solverProvider solvers.RoutingSolverGetter) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(ev event.CreateEvent) bool {
 			obj, ok := ev.Object.(*controllerv1alpha1.WorkspaceRouting)
@@ -30,7 +30,7 @@ func getRoutingPredicatesForSolverFunc(getSolver WorkspaceRoutingSolverFunc) pre
 				// of the controller to ignore WorkspaceRoutings for other routing classes.
 				return true
 			}
-			if _, err := getSolver(obj.Spec.RoutingClass); errors.Is(err, RoutingNotSupported) {
+			if !solverProvider.HasSolver(obj.Spec.RoutingClass) {
 				return false
 			}
 			return true
@@ -48,7 +48,9 @@ func getRoutingPredicatesForSolverFunc(getSolver WorkspaceRoutingSolverFunc) pre
 				// of the controller to ignore WorkspaceRoutings for other routing classes.
 				return true
 			}
-			if _, err := getSolver(newObj.Spec.RoutingClass); errors.Is(err, RoutingNotSupported) {
+			if !solverProvider.HasSolver(newObj.Spec.RoutingClass) {
+				// Future improvement: handle case where old object has a supported routingClass and new project does not
+				// to allow for cleanup when routingClass is switched.
 				return false
 			}
 			return true
@@ -58,7 +60,7 @@ func getRoutingPredicatesForSolverFunc(getSolver WorkspaceRoutingSolverFunc) pre
 			if !ok {
 				return true
 			}
-			if _, err := getSolver(obj.Spec.RoutingClass); errors.Is(err, RoutingNotSupported) {
+			if !solverProvider.HasSolver(obj.Spec.RoutingClass) {
 				return false
 			}
 			return true
--- a/controllers/controller/workspacerouting/predicates.go
+++ b/controllers/controller/workspacerouting/predicates.go
@@ -20,7 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-func getRoutingPredicatesForSolverFunc(solverProvider solvers.RoutingSolverGetter) predicate.Funcs {
+func getRoutingPredicatesForSolverFunc(solverGetter solvers.RoutingSolverGetter) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(ev event.CreateEvent) bool {
 			obj, ok := ev.Object.(*controllerv1alpha1.WorkspaceRouting)
@@ -30,7 +30,7 @@ func getRoutingPredicatesForSolverFunc(solverProvider solvers.RoutingSolverGette
 				// of the controller to ignore WorkspaceRoutings for other routing classes.
 				return true
 			}
-			if !solverProvider.HasSolver(obj.Spec.RoutingClass) {
+			if !solverGetter.HasSolver(obj.Spec.RoutingClass) {
 				return false
 			}
 			return true
@@ -48,8 +48,8 @@ func getRoutingPredicatesForSolverFunc(solverProvider solvers.RoutingSolverGette
 				// of the controller to ignore WorkspaceRoutings for other routing classes.
 				return true
 			}
-			if !solverProvider.HasSolver(newObj.Spec.RoutingClass) {
-				// Future improvement: handle case where old object has a supported routingClass and new project does not
+			if !solverGetter.HasSolver(newObj.Spec.RoutingClass) {
+				// Future improvement: handle case where old object has a supported routingClass and new object does not
 				// to allow for cleanup when routingClass is switched.
 				return false
 			}
@@ -60,7 +60,7 @@ func getRoutingPredicatesForSolverFunc(solverProvider solvers.RoutingSolverGette
 			if !ok {
 				return true
 			}
-			if !solverProvider.HasSolver(obj.Spec.RoutingClass) {
+			if !solverGetter.HasSolver(obj.Spec.RoutingClass) {
 				return false
 			}
 			return true
--- a/controllers/controller/workspacerouting/solvers/solver.go
+++ b/controllers/controller/workspacerouting/solvers/solver.go
@@ -13,12 +13,19 @@
 package solvers
 
 import (
+	"errors"
+	"fmt"
+
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"github.com/devfile/devworkspace-operator/pkg/config"
 	oauthv1 "github.com/openshift/api/oauth/v1"
 	routeV1 "github.com/openshift/api/route/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+var RoutingNotSupported = errors.New("routingclass not supported by this controller")
 
 type RoutingObjects struct {
 	Services     []v1.Service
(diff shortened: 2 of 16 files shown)


controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -293,7 +302,10 @@ func (r *WorkspaceRoutingReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if config.ControllerCfg.IsOpenShift() {
 		bld.Owns(&routeV1.Route{})
 	}
-	bld.WithEventFilter(routingPredicates)
+	if r.GetSolverFunc == nil {
```
sleshchenko, line 305: What is the reason to set up the default get solver func? It may lead to set up two controllers that listen to "default"/"embedded" routing classes.
What is instead we require it and explicitly specify it on our main.go, see like in the following https://github.com/sleshchenko/devworkspace-operator/commit/0a86eac28422cd3dd19cf997e5717cfeaaf21d73

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -35,15 +35,24 @@ import (
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 )
 
-var externalRoutingError = errors.New("routingclass not supported by this controller")
+var ExternalRoutingError = errors.New("routingclass not supported by this controller")
```
sleshchenko, line 38: From the code, it would be clearer if it's named `RoutingNotSupported` or something like that. Then the doc says that controllers should throw it when they do not support the configured routing, so, it's expected to be handled by other controllers.

sleshchenko (commented):
Overall LGTM

davidfestal (commented):
I'm OK with the approach. The more we can reuse from the code in external routing controllers, the better it is.

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: Ah I somehow missed that I could return an error here -- that makes much more sense.

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -14,33 +14,49 @@ package workspacerouting
...
+)
+
 const workspaceRoutingFinalizer = "workspacerouting.controller.devfile.io"
 
+// WorkspaceRoutingSolverFunc is a function that obtains a Solver
+// (see github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers)
+// for a particular WorkspaceRouting instance. This function should return an RoutingNotSupported if
+// the routingClass is not recognized, and any other error if the routingClass is invalid (e.g. an OpenShift-only
+// routingClass on a vanilla Kubernetes platform).
+type WorkspaceRoutingSolverFunc func(routingClass controllerv1alpha1.WorkspaceRoutingClass) (solver solvers.RoutingSolver, err error)
```
metlos, line 51: Would it make sense to pass a `Client` to this function as well so that the solvers can use it for querying/manipulating the cluster or is it ok for the solvers to obtain the client by other means?

metlos (commented):
I like the way how this can be simply reused. Just one small question below...

controllers/controller/workspacerouting/workspacerouting_controller.go
davidfestal: I assume passing the client could make sense if we also want to reuse the main reconciliation loop code. wdyt @amisevsk ?

amisevsk: We could pass the client, but I don't see why it should be needed; this function is basically a mapping from `routingClass -> Solver`. Is there a reason we would need to reach out to the cluster here?

The Solver interface in general does not include any cluster-related actions -- it simply returns a list of objects that we want to exist on the cluster. The main reconcile loop is responsible for making what's on the cluster match what's returned by the solver.

amisevsk: Thinking about specific use-cases, do we foresee needing to extend the `RoutingObjects` struct? For a given workspace, the workspacerouting reconciler will make sure the services/routes/ingresses returned by `GetSpecObjects` exist on the cluster, and any other objects are removed. 

We can discuss this for a separate PR as the need becomes clearer.

amisevsk: Of course, the detail I forgot here is that `GetSolverFunc` has to work in the predicates as well, to properly filter events from triggering reconciles. I've removed the change from the PR; we can work to add it later if we find it necessary.

amisevsk:
I've set up a quick sample repo showing how the workspacerouting controller can be reused: https://github.com/amisevsk/devworkspace-controller-external-routing

If a workspace is created with routingclass `external-sample`, it will define a plain service and allow the workspace start to continue (the workspace is unaccessible, of course).

amisevsk:
Updated PR description with current testing info.

controllers/workspace/provision/routing.go
```diff
@@ -122,13 +122,20 @@ func getSpecRouting(
...
+	}
+
 	routing := &v1alpha1.WorkspaceRouting{
 		ObjectMeta: v1.ObjectMeta{
 			Name:      fmt.Sprintf("routing-%s", workspace.Status.WorkspaceId),
 			Namespace: workspace.Namespace,
 			Labels: map[string]string{
 				config.WorkspaceIDLabel: workspace.Status.WorkspaceId,
 			},
+			Annotations: annotations,
```
sleshchenko, line 138: Seems since we set annotations, we need to adapt routingDiffOpts, otherwise, resricted-access annotation changes are not propagated from DevWorkspace to its routing.

webhook/workspace/handler/immutable.go
```diff
@@ -39,9 +39,17 @@ var ImmutableWorkspaceDiffOptions = []cmp.Option{
 }
 
 func (h *WebhookHandler) HandleImmutableMutate(_ context.Context, req admission.Request) admission.Response {
```
sleshchenko, line 41: If we take into account restricted annotation related changes, I think it makes sense to rework(at least rename) this HandleImmutable since it's not much anymore.

Maybe something like `HandleRestrictedAccessedObjOnUpdate` would suit a bit more now, but I'm not sure.

controllers/workspace/provision/routing.go
amisevsk: Good catch! Will update.

controllers/controller/workspacerouting/workspacerouting_controller.go
sleshchenko: it's going to be breaking change, which should be harder to do later, right?

Also, I wonder if returning an object that is not used in predicate influences performance?

Maybe include RoutingSolverFactory (probably wrong naming for go):
```
  isSupported()
  getSolver() <-  currently Get Solvers returns a new instance... even though it's not used at all.
```

sleshchenko (approved):
Works fine in the described scenario.
LGTM
An ability to pass client to solvers could be solved separately

controllers/workspace/provision/routing.go
amisevsk: I put a bit of work into this but it's pretty ugly as is. Another option might be to use `cmp.FilterPath` but I couldn't get it to work as expected.

metlos:
One thing that came up from my discussions with @davidfestal is the need to pass down additional information down to the (external) routing class solvers. The best way I could think of is to pass down (prefixed?) annotations from the workspace CR down to workspace routing CRs so that the solver can pick up this additional info when producing the routing objects..

E.g.

```yaml
kind: DevWorkspace
metadata:
   annotations:
      "routingClass.che/gateway": my-gateway
...
``` 
would make the annotation appear on the `WorkspaceRouting` object as well:

```yaml
kind: WorkspaceRouting
metadata:
  annotations:
    "routingClass.che/gateway": my-gateway
```

This would require passing down the whole `WorkspaceRouting` object down to `GetSpecObjects()` impls, not just the spec...

Do you think it'd make sense to do that in the scope of this PR or would you rather handle this separately?

amisevsk:
> The best way I could think of is to pass down (prefixed?) annotations from the workspace CR down to workspace routing CRs so that the solver can pick up this additional info when producing the routing objects..

This may have to be a separate PR; currently passing *all* annotations is a bit problematic since we add e.g. timing-related annotations, and that could result in looping reconciles as the routing is updated every time we reconcile the devworkspace, and I don't want to hard-code a specific prefix prematurely.

Perhaps we can define an attribute on DevWorkspaces that we handle specifically? Something like `routingAdditionalAttributes` and add a similar attributes field to WorkspaceRoutings? cc: @davidfestal 

> This would require passing down the whole WorkspaceRouting object down to GetSpecObjects() impls, not just the spec...

I actually already slipped this change in, on commit [a068d39](https://github.com/devfile/devworkspace-operator/pull/219/commits/a068d39b4ddd63da670a82608108f0e2cab3c09d#diff-cf7df7a74bac109a870d75f37f3faf6c3663a34db5b1344db8256b82e186e09fR33) (the short-form commit message is a little unclear here)

amisevsk:
Rebased to clean up the history a bit; also I've updated the sample repo with external routing (https://github.com/amisevsk/devworkspace-controller-external-routing) so it should be in sync with the current setup.

metlos (approved):
This looks good. The https://github.com/che-incubator/devworkspace-che-operator has been updated to the latest version of the code and seems to integrate nicely.

I'm looking forward to having the ability to pass down the annotations in the upcoming PR :)

controllers/workspace/provision/routing.go
sleshchenko: It was just an hour of tries and failures and seems I made it work, I still don't understand how it's going to work.

https://github.com/devfile/devworkspace-operator/commit/456d09a5ea151ee2bd4f406d600f8fa7e383f473

How to test:
```bash
make run
#1
kind: DevWorkspace
metadata:
  name: theia
spec:
  started: true
  routingClass: non-supported-yet
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - name: theia
        plugin:
          id: eclipse/che-theia/next
      - name: terminal
        plugin:
          id: eclipse/che-machine-exec-plugin/nightly
    commands:
      - id: say hello
        exec:
          component: plugin
EOF
#2
DW_ID=$(kubectl get dw theia -o=yaml | yq '.status.workspaceId' -r)

#3
kubectl patch workspaceroutings.controller.devfile.io routing-$DW_ID --type=merge --patch '
metadata:
  finalizers: ["my-custom"]
'
# ^ operator logs contains only  We are not interested in made routing update

#4
kubectl patch workspaceroutings.controller.devfile.io routing-$DW_ID --type=merge --patch '
metadata:
  labels:
    any: any
'
# ^ operator logs contains:
#  We are not interested in made routing update 3x or even more +
#  We are interested in made routing update 1x +
```

sleshchenko (approved):
Nothing critical to comment found, just a few typos

controllers/workspace/provision/routing.go
amisevsk: This is similar to what I had when I tried, but I was having trouble verifying it worked exactly as I expected. I think using filterPath is kind of harder to read/parse, even though the huge list of fields is ugly.

Eventually I'd like to move away from the `cmp` package -- it's listed as explicitly not intended for production and we've been using it as a crutch. In working on https://github.com/devfile/devworkspace-operator/pull/229 I came across the k8s `equality` package, which looks like it'll do what we want in a cleaner way and I'd like to work on migrating to that sometime soon.

controllers/controller/workspacerouting/predicates.go
```diff
@@ -0,0 +1,69 @@
...
+		UpdateFunc: func(ev event.UpdateEvent) bool {
+			newObj, ok := ev.ObjectNew.(*controllerv1alpha1.WorkspaceRouting)
+			if !ok {
+				// If object is not a WorkspaceRouting, it must be a service/ingress/route related to the workspace
+				// The safe choice here is to trigger a reconcile to ensure that all resources are in sync; it's the job
+				// of the controller to ignore WorkspaceRoutings for other routing classes.
+				return true
+			}
+			if !solverProvider.HasSolver(newObj.Spec.RoutingClass) {
+				// Future improvement: handle case where old object has a supported routingClass and new project does not
```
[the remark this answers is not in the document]
amisevsk: :+1: woops.

controllers/controller/workspacerouting/predicates.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers"
+
+	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+func getRoutingPredicatesForSolverFunc(solverProvider solvers.RoutingSolverGetter) predicate.Funcs {
```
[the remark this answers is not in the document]
amisevsk: Because I'm bad at Go naming conventions :)

Aligned.

controllers/controller/workspacerouting/solvers/solver.go
```diff
@@ -30,11 +37,71 @@ type RoutingObjects struct {
...
+
+var _ RoutingSolverGetter = (*SolverGetter)(nil)
+
+func (_ *SolverGetter) HasSolver(routingClass controllerv1alpha1.WorkspaceRoutingClass) bool {
+	if routingClass == "" {
+		// Special case for built-in: empty routing class returns the default solver for the DevWorkspace controller.
+		return true
+	}
+	switch routingClass {
+	case controllerv1alpha1.WorkspaceRoutingDefault,
```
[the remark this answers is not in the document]
amisevsk: Yep, that's leftovers from an earlier design on routing. I'll update.

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -14,33 +14,40 @@ package workspacerouting
...
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 )
 
+var (
+	NoSolversEnabled = errors.New("reconciler does not define GetSolverFunc")
```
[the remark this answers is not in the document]
amisevsk: :+1:

webhook/workspace/mutate.go
```diff
@@ -65,7 +65,7 @@ func (m *ResourcesMutator) Handle(ctx context.Context, req admission.Request) ad
 			case handler.V1ServiceKind, handler.V1beta1IngressKind, handler.V1RouteKind,
 				handler.V1alpha1ComponentKind, handler.V1alpha1WorkspaceRoutingKind:
 
-				return m.HandleImmutableMutate(ctx, req)
+				return m.HandleRestrictedAccessMutate(ctx, req)
```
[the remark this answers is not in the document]
amisevsk: Updated

controllers/workspace/provision/routing.go
sleshchenko: > it's listed as explicitly not intended for production and we've been using it as a crutch

Good catch!
Before I found 
> Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal.

I played and came solution where we have not bad readability https://github.com/devfile/devworkspace-operator/commit/20e92d3d3be6efa276eb95cd7bbec20d7ae68822
But it does not matter because it's not production-ready ) One day we can use it for tests only :-D

So, I'm OK with `equality` I just don't understand in details how the same would look like there.

sleshchenko (approved):
Approving again since it fully solves the original issue with external routing.
The reworking to `equality` instead of cmp should be done separetely

davidfestal (approved):

Issue #217, opened 2020-12-04
Contribute the flattened devfile 2.0 with needed plugins and tools


sleshchenko:
We have a goal to dogfood, so develop DevWorkpace Operator with DevWorkspace 2.0

This issue is about to contribute initial Devfile 2.0 we can start with, that will have needed tools (kubectl, oc, kustomize, controller-gen, ...) and commands to:
- [ ] install the Operator on the cluster(not the one where we dogfood);
- [ ] run(+in debug mode) devworkspace that connects to another cluster;
- [ ] build docker image (it requires a rootless tool, like buildah?)
- [ ] run e2e tests;

Issue #210, opened 2020-12-01
DevWorkspace dogfooding


l0rd:
## What problem do we want to address?

We want to demonstrate/validate that a Theia based DevWorkspace can be used to work on a real world, cloud-native project (this one).

## How are we going to address it?

Add a devfile 2.0 (initially flattened?) at the root of the repository with
- golang/operator-sdk/kubebuilder support
- vscode kubernetes
- commands to run / debug a (second?) devworkspace controller

PR #220, opened 2020-12-04
WIP: initial implementation of support for developing the controller in-cluster

amisevsk:
### What does this PR do?
Adds the beginning of an implementation that enables us to use a DevWorkspace to develop the DevWorkspace controller. This is done by 
1. Adding a label that can be applied to stop controllers from reconciling a workspace once it is running
2. Add a dockerfile for building the "dev" image
3. Add a sample devworkspace CR spec for the development workspace

There are still some todos:
- [ ] Prompted to install go-pls on start, even though it is installed
- [ ] Errors logged for go plugin on start
    ```
    ERROR:Remote plugin in vscode-go: promise rejection is not handled in two seconds: TypeError: Cannot read property 'id' of undefined
    ERROR:Remote plugin in vscode-go: promise rejection stack trace: TypeError: Cannot read property 'id' of undefined
    ```
- [ ] No easy mechanism for logging in to cluster when using k8s
- [ ] No podman/docker support, as podman package isn't available in Debian-stretch AFAICT.

Solves https://github.com/devfile/devworkspace-operator/issues/217
Part of https://github.com/devfile/devworkspace-operator/issues/210

### Is it tested? How?
```
make docker install && kubectl apply -f samples/devworkspace-controller.yaml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -324,6 +324,22 @@ ifneq ($(shell operator-sdk version | cut -d , -f 1 | cut -d : -f 2 | cut -d \" 
 	@echo 'WARN: Please use the recommended operator-sdk if you face any issue.'
 endif
 
+inject_kubeconfig:
+	@{ \
+	MACHINE_EXEC_ENDPOINT=$$($(K8S_CLI) get ingress -o json | jq -r '.items[] | select(.metadata.name | test("che-mach-exec")).spec.rules[].host') ;\
+	SA_TOKEN=$$($(K8S_CLI) get secrets -o json | jq -r '.items[] | select(.metadata.name | test("^devworkspace-controller")).data.token' | base64 -d) ;\
+	curl --location -X POST $${MACHINE_EXEC_ENDPOINT}/exec/init \
+	  -H "Content-Type: application/json" \
+		--data-raw '{
+			"container": "go-cli",
+			"kubeconfig": {
+				"username": "developer"
+			}
+		}'
+	} ;\
+	POD_NAME=$$($(K8S_CLI) get po -l 'controller.devfile.io/workspace_name=devworkspace-operator-dev' -o=jsonpath="{.items[*].metadata.name}") ;\
+	$(K8S_CLI) exec $${POD_NAME} -c go-cli -- kubectl config set 'users.developer.token' "$${SA_TOKEN}" --set-raw-bytes=true
+
 .PHONY: help
 ### help: print this message
 help: Makefile
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -37,6 +37,11 @@ const (
 	SidecarDefaultMemoryLimit = "128M"
 	PVCStorageSize            = "1Gi"
 
+	// WorkspaceStartOnlyLabel is used to specify that the DevWorkspace should not be reconciled further once
+	// it has entered the "Running" phase. This is to enable development of the DevWorkspace operator from within
+	// a DevWorkspace. This label is ignored when `devworkspace.experimental_features_enabled` is set to false.
+	WorkspaceStartOnlyLabel = "controller.devfile.io/development.start_only"
+
 	// WorkspaceIDLabel is label key to store workspace identifier
 	WorkspaceIDLabel = "controller.devfile.io/workspace_id"
 
--- a/samples/devworkspace-controller.yaml
+++ b/samples/devworkspace-controller.yaml
@@ -49,7 +49,6 @@ spec:
               protocol: http
               attributes:
                 type: ide
-                cookiesAuthEnabled: "true"
             - name: "webviews"
               exposure: public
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -324,6 +324,22 @@ ifneq ($(shell operator-sdk version | cut -d , -f 1 | cut -d : -f 2 | cut -d \"
 	@echo 'WARN: Please use the recommended operator-sdk if you face any issue.'
 endif
 
+inject_kubeconfig:
```
sleshchenko, line 327: It's here just for test purpose or we are going to use it during development?

pkg/config/constants.go
```diff
@@ -37,6 +37,11 @@ const (
 	SidecarDefaultMemoryLimit = "128M"
 	PVCStorageSize            = "1Gi"
 
+	// WorkspaceStartOnlyLabel is used to specify that the DevWorkspace should not be reconciled further once
+	// it has entered the "Running" phase. This is to enable development of the DevWorkspace operator from within
+	// a DevWorkspace. This label is ignored when `devworkspace.experimental_features_enabled` is set to false.
+	WorkspaceStartOnlyLabel = "controller.devfile.io/development.start_only"
```
sleshchenko, line 43: I'm not sure I understand how we are going to use start_only in development. Start workspace, run local controller, expect that local controller is going to reconcile the workspace where we work?
How then we will test webhooks/workspace start?

I see some annotations or even labels (to easier exclude it from webhooks, but annotations are designed for such things, like we have restricted-access) that allows excluding devworkspace CR from attention all namespaces operator.
Then workflow is like the following:
1. We create a workspace with devworkspace operator project.
2. It's reconciled by all namespace operator.
3. We do changes;
4. We run local operator.
5. We test our changes on the `managed_by_local_operator` DevWorkspace CR.

samples/devworkspace-controller.yaml
```diff
@@ -0,0 +1,209 @@
...
+            - controller.devfile.io/workspace_id=$(CHE_WORKSPACE_ID)
+          endpoints:
+            - name: "che-mach-exec"
+              exposure: public
+              targetPort: 4444
+              protocol: ws
+              secure: true
+              attributes:
+                type: terminal
+                cookiesAuthEnabled: "true"
```
sleshchenko, line 144: Just a side comment, that was provided by jwtproxy. Not sure if we need it here now, or later with new authentication.

pkg/config/constants.go
sleshchenko: I think I came to workflow you had in mind with start_only:
1. make install
2. kc apply -f ./samples/devworkspace-controller -< with start_only
3. make inject_kubeconfig
4. kc patch deployment/devworkspace-controler-manager --patch "{\"spec\":{\"replicas\":0}}"`
5. Make changes in the workspace.
6. make run now from the workspace.

Such development workspace is like a space shuttle that leaves unneeded engines and fuel tanks )

Makefile
amisevsk: Ideally it'll be used during development, provided I can get it to work. The idea is you would use che-machine-exec to log in the workspace terminal as the devworkspace-controller serviceaccount, allowing you to run/debug "locally" and have it run in the cluster like a normal controller.

pkg/config/constants.go
amisevsk: That's pretty much my idea at the moment -- you still have to bootstrap your initial workspace load, sadly.

One flow we could also enable, if we figured out building images within the cluster, would be to not patch the controller to stop it and instead rollout new deployments from within the cluster itself. The annotation is mostly here as a safety net -- we don't want a development bug in our controller killing the workspace being used to develop it.

samples/devworkspace-controller.yaml
amisevsk: This is mostly copy-paste :)

If it's not needed we should remove it.

samples/devworkspace-controller.yaml
```diff
@@ -0,0 +1,206 @@
...
+          image: "quay.io/eclipse/che-sidecar-go:1.14.4-e8a574b"
+          memoryLimit: '512Mi'
+          mountSources: true
+          env:
+            - name: GOPATH
+              value: /go
+            - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
+              value: /remote-endpoint/plugin-remote-endpoint 
+            - name: THEIA_PLUGINS
+              value: local-dir:///plugins/sidecars/vscode-typescript
```
JPinkney, line 162:
```suggestion
              value: local-dir:///plugins/sidecars/vscode-go
```
This is the part I was talking about the standup, I did the same thing in vscode-go

amisevsk:
Closing PR as woefully out-of-date. We'll need to revisit this later.

PR #223, opened 2020-12-08
Place Che API Sidecar as the first container

sleshchenko:
### What does this PR do?
This PR places Che API Sidecar as the first container to workaround an issue where Che Theia fails to access it just after startup.

Also, the next Theia is used here since the webview with 7.20 does not work due bug that is fixed on the next version:
Screenshot_20201208_112632

The workaround proposed here https://github.com/devfile/devworkspace-operator/pull/201#issuecomment-736759534

### Is it tested? How?
```bash
# 1. Deploy controller as usual
#export IMG=mylocal:local
make docker
make install

# 2. Run all in one sample
kubectl apply -f samples/all-in-one-theia-nodejs.devworkspace.yaml

# 3. Once workspace is started. Open it and make sure Welcome Page powered by Theia Webview is displayed at least sometimes :-D I did not manage to get it working stable, I assume it's because of cluster-ip issue but I may be wrong and maybe that PR did not work properly. In any case, it should increase the chance to get working Webview
```
Screenshot_20201208_161143

More often it's not displayed:
Screenshot_20201208_111927

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -192,7 +192,8 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		}
 		// TODO: first half of provisioning rest-apis
 		cheRestApisComponent := restapis.GetCheRestApisComponent(workspace.Name, workspace.Status.WorkspaceId, workspace.Namespace)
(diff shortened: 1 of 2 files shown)


amisevsk (approved):
Haven't tested directly, but LGTM. I love a sneaky fix :)

PR #224, opened 2020-12-09
Cert manager in README & fix a typo in Makefile

metlos:
### What does this PR do?

Update the README.md to mention the need to install
the cert manager prior to installing the operator.

Also, fix the syntax error in the Makefile preventing
installation on Kubernetes.

none

### Is it tested? How?
`make install` now works on kubernetes (tested with minikube).

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,9 @@ metadata:
 
 ## Running the controller in a cluster
 
-The controller requires [cert-manager](https://cert-manager.io) running in the cluster. You can install it using `make install_cert_manager` if you 
-don't run it already. The minimum version of cert-manager is `v1.0.4`.
+When deployed to Kubernetes, the controller requires [cert-manager](https://cert-manager.io) running in the cluster. 
+You can install it using `make install_cert_manager` if you don't run it already. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,9 @@ metadata:
 
 ## Running the controller in a cluster
 
+The controller requires [cert-manager](https://cert-manager.io) running in the cluster. You can install it using `make install_cert_manager` if you
```
sleshchenko, line 31: That's needed for Kubernetes cluster only. On Openshift we use openshift out-of-the-box mechanism to get the certificates for the webhook server.

README.md
```diff
@@ -76,6 +79,7 @@ Some of the rules supported by the makefile:
 | install | install controller to cluster |
 | restart | restart cluster controller deployment |
 | install_crds | update CRDs on cluster |
+| install_cert_manager | installs the cert-manager to the cluster |
```
sleshchenko, line 82:
```suggestion
| install_cert_manager | installs the cert-manager to the K8s cluster |
```

sleshchenko (approved):
Thanks!

Issue #214, opened 2020-12-03
Clean up files belonging to removed workspaces


sleshchenko:
Currently, DevWorkspace Operator uses one PVC for all workspace and provides isolation with subpaths.
And these subpaths are never clean up. It means that if user recreates workspace - they probably will exceed file system quota.

It's needed to clean up subpaths that belong to removed workspaces. I see it's implemented with an additional finalizer, that set up a dedicated Deployment/Pod/Job that mounts / of the workspace PVC and cleans up the needed files. Note that this Pod can be blocked if another workspace is running at the same time (Since PVC has RWO). To avoid unneeded runs of such pre-create deployment, we may store initialized subpaths in the PVC annotations.

Also, pay attention to be able to remove subpaths - it may be needed to initialize them in the right way, where we mount `/`, initialize subpaths, and only then we are able to mount subpaths which we should be able to clean up without permissions issues. More see https://github.com/devfile/devworkspace-operator/pull/211

PR #230, opened 2020-12-18
Clean up DevWorkspaces PVC when a DevWorkspace is deleted.

amisevsk:
### What does this PR do?
Runs a similar job to Eclipse Che, which removes workspace files from the common PVC when the workspace is deleted.

~~**Note:** It's expected that this will likely fail on OpenShift as-is, due to the PVC subpath problem (https://github.com/devfile/devworkspace-operator/issues/198). On minikube, it seems pods can run as root and so the problem doesn't manifest.~~ This should be resolved now.

Resolves https://github.com/devfile/devworkspace-operator/issues/214

### Is it tested? How?
```bash
# Deploy controller as normal
make docker install
# Create Theia DevWorkspace
kubectl apply -f samples/theia-next
# Delete DevWorkspace
kubectl delete dw theia
# Repeat create and delete a few times for good measure
# Check that workspace files are cleaned up
apiVersion: v1
kind: Pod
metadata:
  name: cleanup
  labels:
    app: cleanup
spec:
  containers:
    - image: quay.io/libpod/busybox:1.30.1
      name: cleanup
      volumeMounts:
        - mountPath: /tmp/claim-devworkspace
          name: claim-devworkspace
      command: ["sleep"]
      args: ["1d"]
  volumes:
    - name: claim-devworkspace
      persistentVolumeClaim:
        claimName: claim-devworkspace
EOF
# Output should be just `.` and `..`
kubectl exec -it cleanup -- ls -al /tmp/claim-devworkspace
kubectl delete po cleanup
```
There *may* be a weird race condition where the job completes successfully but paths still exist in the PVC -- I can't reproduce it consistently. Interestingly, kubernetes seems to create subpaths in a mounted volume even if it's mounted to another pod, which may be related.

### DIFF
--- a/controllers/workspace/finalize.go
+++ b/controllers/workspace/finalize.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"time"
 
 	"github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/controllers/workspace/provision"
@@ -68,10 +69,33 @@ func (r *DevWorkspaceReconciler) finalize(ctx context.Context, log logr.Logger, 
 	if wait {
 		return reconcile.Result{Requeue: true}, nil
 	}
+
+	terminating, err := r.namespaceIsTerminating(ctx, workspace.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	} else if terminating {
+		//Namespace is terminating, it's redundant to clean PVC files since it's going to be removed
+		log.Info("Namespace is terminating; clearing storage finalizer")
+		clearFinalizer(workspace)
+		return reconcile.Result{}, r.Update(ctx, workspace)
+	}
+
+	pvcExists, err := r.pvcExists(ctx, workspace)
+	if err != nil {
+		return reconcile.Result{}, err
+	} else if !pvcExists {
+		//PVC does not exist. nothing to clean up
+		log.Info("PVC does not exit; clearing storage finalizer")
+		clearFinalizer(workspace)
+		// job will be clean up by k8s garbage collector
+		return reconcile.Result{}, r.Update(ctx, workspace)
+	}
+
 	specJob, err := r.getSpecCleanupJob(workspace)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
 	clusterJob, err := r.getClusterCleanupJob(ctx, workspace)
 	if err != nil {
 		return reconcile.Result{}, err
@@ -111,7 +135,8 @@ func (r *DevWorkspaceReconciler) finalize(ctx context.Context, log logr.Logger, 
 			return r.updateWorkspaceStatus(workspace, r.Log, failedStatus, reconcile.Result{}, nil)
 		}
 	}
-	return reconcile.Result{}, nil
+	// Requeue at least each 10 seconds to check if PVC is not removed by someone else
+	return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 }
 
 func (r *DevWorkspaceReconciler) getSpecCleanupJob(workspace *v1alpha2.DevWorkspace) (*batchv1.Job, error) {
@@ -221,4 +246,34 @@ func clearFinalizer(workspace *v1alpha2.DevWorkspace) {
 		}
 	}
 	workspace.SetFinalizers(newFinalizers)
+}
+
+func (r *DevWorkspaceReconciler) namespaceIsTerminating(ctx context.Context, namespace string) (bool, error) {
+	namespacedName := types.NamespacedName{
+		Name: namespace,
+	}
+	n := &corev1.Namespace{}
+
+	err := r.Get(ctx, namespacedName, n)
+	if err != nil {
+		return false, err
+	}
+
+	return n.Status.Phase == corev1.NamespaceTerminating, nil
+}
+
+func (r *DevWorkspaceReconciler) pvcExists(ctx context.Context, workspace *v1alpha2.DevWorkspace) (bool, error) {
+	namespacedName := types.NamespacedName{
+		Name:      config.ControllerCfg.GetWorkspacePVCName(),
+		Namespace: workspace.Namespace,
+	}
+	err := r.Get(ctx, namespacedName, &corev1.PersistentVolumeClaim{})
+	if err != nil {
+		if k8sErrors.IsNotFound(err) {
+			return false, nil
+		}
+
+		return false, err
+	}
+	return true, nil
 }
--- a/controllers/workspace/finalize.go
+++ b/controllers/workspace/finalize.go
@@ -44,6 +44,8 @@ const (
 var (
 	cleanupJobCompletions  = int32(1)
 	cleanupJobBackoffLimit = int32(3)
+	// PVCCleanupPodMemoryLimit is the memory limit used for PVC clean up pods
+	PVCCleanupPodMemoryLimit = resource.MustParse("32Mi")
 )
 
 func (r *DevWorkspaceReconciler) setFinalizer(ctx context.Context, workspace *v1alpha2.DevWorkspace) (ok bool, err error) {
@@ -109,6 +111,7 @@ func (r *DevWorkspaceReconciler) finalize(ctx context.Context, log logr.Logger, 
 func (r *DevWorkspaceReconciler) getSpecCleanupJob(workspace *v1alpha2.DevWorkspace) (*batchv1.Job, error) {
 	workspaceId := workspace.Status.WorkspaceId
 	pvcName := config.ControllerCfg.GetWorkspacePVCName()
+
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      common.PVCCleanupJobName(workspaceId),
@@ -122,7 +125,8 @@ func (r *DevWorkspaceReconciler) getSpecCleanupJob(workspace *v1alpha2.DevWorksp
 			BackoffLimit: &cleanupJobBackoffLimit,
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
-					RestartPolicy: "Never",
+					RestartPolicy:   "Never",
+					SecurityContext: provision.GetDevWorkspaceSecurityContext(),
 					Volumes: []corev1.Volume{
 						{
 							Name: pvcName,
@@ -144,7 +148,7 @@ func (r *DevWorkspaceReconciler) getSpecCleanupJob(workspace *v1alpha2.DevWorksp
 							},
 							Resources: corev1.ResourceRequirements{
 								Limits: corev1.ResourceList{
-									corev1.ResourceMemory: resource.MustParse(config.PVCCleanupPodMemoryLimit),
+									corev1.ResourceMemory: PVCCleanupPodMemoryLimit,
 								},
 							},
 							VolumeMounts: []corev1.VolumeMount{
@@ -185,7 +189,7 @@ func (r *DevWorkspaceReconciler) getClusterCleanupJob(ctx context.Context, works
 	return clusterJob, nil
 }
 
-func isFinalizerNecessary(workspace *v1alpha2.DevWorkspace) bool {
+func isFinalizerNecessary(_ *v1alpha2.DevWorkspace) bool {
 	// TODO: Implement checking whether persistent storage is used (once other choices are possible)
 	// Note this could interfere with cloud-shell until this TODO is resolved.
 	return true
--- a/controllers/workspace/finalize.go
+++ b/controllers/workspace/finalize.go
@@ -1,0 +1,208 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"path"
+
+	"github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/controllers/workspace/provision"
+	"github.com/devfile/devworkspace-operator/pkg/config"
+
+	"github.com/go-logr/logr"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	pvcClaimMountPath   = "/tmp/devworkspaces/"
+	cleanupCommandFmt   = "rm -rf %s"
+	pvcCleanupFinalizer = "storage.controller.devfile.io"
+)
+
+var (
+	cleanupJobCompletions  = int32(1)
+	cleanupJobBackoffLimit = int32(3)
+)
+
+func (r *DevWorkspaceReconciler) setFinalizer(ctx context.Context, workspace *v1alpha2.DevWorkspace) (ok bool, err error) {
+	if !isFinalizerNecessary(workspace) || hasFinalizer(workspace) {
+		return true, nil
+	}
+	workspace.SetFinalizers(append(workspace.Finalizers, pvcCleanupFinalizer))
+	return false, r.Update(ctx, workspace)
+}
+
+func (r *DevWorkspaceReconciler) finalize(ctx context.Context, log logr.Logger, workspace *v1alpha2.DevWorkspace) (reconcile.Result, error) {
+	if !hasFinalizer(workspace) {
+		return reconcile.Result{}, nil
+	}
+	// Need to make sure Deployment is cleaned up before starting job to avoid mounting issues for RWO PVCs
+	wait, err := provision.DeleteWorkspaceDeployment(ctx, workspace, r.Client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if wait {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	specJob, err := r.getSpecCleanupJob(workspace)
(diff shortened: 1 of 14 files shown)


amisevsk:
Looking into the issue of successful cleaning that leaves files on the PVC, it seems that deleting a workspace within ~7 seconds of its creation (on my machine) results in the cleanup job completing successfully without ultimately removing files from the PVC.

It seems like minikube also allows for multiple mounts, so I'll need to update this PR to ensure the Deployment is removed before the cleanup job runs to support cases where PVCs are RWO. I suspect this'll fix the issue above as well.

snippet for testing the timing:
```bash
for i in {1..30}; do
  sleep $((0.5 * $i))
  kubectl get dw theia -o yaml | yq '.status.workspaceId'
  kubectl delete dw theia
done
```

controllers/workspace/finalize.go
```diff
@@ -0,0 +1,202 @@
...
+		if condition.Status != corev1.ConditionTrue {
+			continue
+		}
+		switch condition.Type {
+		case batchv1.JobComplete:
+			log.Info("PVC clean up job successful; clearing finalizer")
+			clearFinalizer(workspace)
+			return r.Update(ctx, workspace)
+		case batchv1.JobFailed:
+			log.Error(fmt.Errorf("PVC clean up job failed: message %q", condition.Message),
```
sleshchenko, line 91: Well, I suppose it may fail due to PVC mounting issue if there is another DevWorkspace is running (workspace that is removed was stopped).
And I would say we need to wait until PVC is available to be bound to CleanUp Job.

Also, we should not fail or even run Job if PVC is already removed (it can be implemented even now in https://github.com/devfile/devworkspace-operator/pull/230/files#diff-55c79087cb84a6e155de791d4a71a4506191f8c5eae13cc1ca17ba62ebca6bf8R179)

controllers/workspace/finalize.go
```diff
@@ -0,0 +1,202 @@
...
+			Namespace: workspace.Namespace,
+			Labels: map[string]string{
+				config.WorkspaceIDLabel: workspaceId,
+			},
+		},
+		Spec: batchv1.JobSpec{
+			Completions:  &cleanupJobCompletions,
+			BackoffLimit: &cleanupJobBackoffLimit,
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
```
sleshchenko, line 115: Is not it necessary to reuse the same security context as we use for deployment workspace deployment https://github.com/devfile/devworkspace-operator/blob/master/controllers/workspace/provision/deployment.go#L137

Or missing security context here leads to using root to avoid permissions issues. If so - it would be better to set it explicitely

sleshchenko (commented):
Looked through. In general looks good, will test and review it in details later

controllers/workspace/finalize.go
amisevsk: In general, PVC cleanup comes with a ton of corner cases; I think we should run the job whenever the finalizer is present (as it is set up currently) since it being set implies that something was stored in the PVC at some point. If the directory does not exist, `rm` will still return a successful exit code. 

We do need to cover edge cases in mounting PVCs -- I'll look into that.

controllers/workspace/finalize.go
amisevsk: I added the security context in, and then also realized that we were setting it somewhat incorrectly -- GID should be 0 (at least, that's what most of our images assume).

amisevsk:
After adding in securityContext, I realized that we were going to run into the PVC subpath problem on every workspace, since we mount `<workspace-id>/[plugins|projects|remote-runtime]`, and the `<workspace-id>` subdir would have 755 permissions. I finally figured out a workaround -- add a dummy mount for `<workspace-id>` to the first container in the spec, as the first volume mount.

controllers/workspace/finalize.go
```diff
@@ -0,0 +1,213 @@
...
+func (r *DevWorkspaceReconciler) getSpecCleanupJob(workspace *v1alpha2.DevWorkspace) (*batchv1.Job, error) {
+	workspaceId := workspace.Status.WorkspaceId
+	pvcName := config.ControllerCfg.GetWorkspacePVCName()
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.PVCCleanupJobName(workspaceId),
+			Namespace: workspace.Namespace,
+			Labels: map[string]string{
+				config.WorkspaceIDLabel: workspaceId,
```
sleshchenko, line 118: I think it makes sense to propagate restricted access as well here if workspace has it

controllers/workspace/finalize.go
```diff
@@ -0,0 +1,213 @@
...
+							Name:    common.PVCCleanupJobName(workspaceId),
+							Image:   images.GetPVCCleanupJobImage(),
+							Command: []string{"/bin/sh"},
+							Args: []string{
+								"-c",
+								fmt.Sprintf(cleanupCommandFmt, path.Join(pvcClaimMountPath, workspaceId)),
+							},
+							Resources: corev1.ResourceRequirements{
+								Limits: corev1.ResourceList{
+									corev1.ResourceMemory: resource.MustParse(config.PVCCleanupPodMemoryLimit),
```
sleshchenko, line 149: Is it really a good idea to panic here if mistake took a place in source code?

Maybe better to initialize this var in init func, or just use var instead of const and then

```
// PVCCleanupPodMemoryLimit is the memory limit used for PVC clean up pods
var PVCCleanupPodMemoryLimit = resource.MustParse("32Mi")
```

controllers/workspace/finalize.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if clusterJob == nil {
+		err := r.Create(ctx, specJob)
+		if err != nil && !k8sErrors.IsAlreadyExists(err) {
+			return err
+		}
+		return nil
+	}
+	if !equality.Semantic.DeepDerivative(specJob.Spec, clusterJob.Spec) {
+		err := r.Patch(ctx, specJob, client.Merge)
```
sleshchenko, line 84: please double check if we're able to patch job, I may be wrong but I suppose we must recreate job instead.

controllers/workspace/finalize.go
```diff
@@ -0,0 +1,213 @@
...
+		if condition.Status != corev1.ConditionTrue {
+			continue
+		}
+		switch condition.Type {
+		case batchv1.JobComplete:
+			log.Info("PVC clean up job successful; clearing finalizer")
+			clearFinalizer(workspace)
+			return r.Update(ctx, workspace)
+		case batchv1.JobFailed:
+			log.Error(fmt.Errorf("PVC clean up job failed: message %q", condition.Message),
```
sleshchenko, line 100: Does the message is really detailed? Or it just says - `container exited with 1 code`
If so, please consider propagating job container logs into devworkspace logs

controllers/workspace/provision/deployment.go
```diff
@@ -189,19 +217,29 @@ func getSpecDeployment(
 					Volumes:                       podAdditions.Volumes,
 					RestartPolicy:                 "Always",
 					TerminationGracePeriodSeconds: &terminationGracePeriod,
-					SecurityContext: &corev1.PodSecurityContext{
-						RunAsUser: user,
-						FSGroup:   user,
-					},
-					ServiceAccountName:           saName,
-					AutomountServiceAccountToken: nil,
+					SecurityContext:               GetDevWorkspaceSecurityContext(),
```
sleshchenko, line 220: Seems it lead Devworkspace to an endless reconcile loop.
The only way to catch the cause:
```
	if !cmp.Equal(specDeployment, clusterDeployment, deploymentDiffOpts) {
		clusterAPI.Logger.Info("deployment needs to be updated", "diff",  cmp.Diff(specDeployment, clusterDeployment, deploymentDiffOpts))
		clusterAPI.Logger.Info("Updating deployment...")
```
probably we need something like that on debug log level always...
So, 
```
&v1.Deployment{
...
- \t\t\t\tSecurityContext:       nil,
+ \t\t\t\tSecurityContext:       s\"&PodSecurityContext{SELinuxOptions:nil,RunAsUser:nil,RunAsNonRoot:nil,SupplementalGroups:[],FSGroup:nil,RunAsGroup:nil,Sysctls:[]Sysctl{},WindowsOptions:nil,FSGroupChangePolicy:nil,}\",
  \t\t\t\tImagePullSecrets:      nil,
  \t\t\t\tHostname:              \"\",
  \t\t\t\t... // 1 ignored and 13 identical fields
  \t\t\t},
  \t\t},
  \t\tStrategy:        {Type: \"Recreate\"},
  \t\tMinReadySeconds: 0,
  \t\t... // 2 ignored and 1 identical fields
  \t},
  \t... // 1 ignored field
  }
```

sleshchenko (commented):
It does not fully work on OpenShift since DevWorkspace is stuck in STARTING phase.

Continue to review other aspects

controllers/workspace/finalize.go
amisevsk: Good catch, thanks. Should I also update the webhooks to block changes if the annotation is set?

controllers/workspace/finalize.go
amisevsk: Makes sense, thanks.

controllers/workspace/finalize.go
amisevsk: The pod gets a message like `exited with 1 code` but the job gets something referring to "backoff limit reached". I don't know a good way to propagate useful information -- the logs will generally just be something like "permission denied" from the pods, and I don't know if there's an easy way to get relevant statuses (if a pod fails for any other reason).

I'll poke at it and see if I can figure something out.

controllers/workspace/provision/deployment.go
amisevsk: Defaulting hits us again -- I'll fix it for the PR but this is an issue that should be resolved by moving to the `equality` package instead of `cmp`.

controllers/workspace/finalize.go
amisevsk: It's some fields are patchable, others seem to not be:
- The standard rules apply for the pod template (can only change a few of the fields)
- I can update restartCount but not completions
- etc.

It's probably safer to just delete the job and recreate it.

amisevsk: Set it up to delete Job if it doesn't match spec.

controllers/workspace/finalize.go
amisevsk: Added propagation of restricted-access and also made webhooks cover jobs in the same way as everything else.

controllers/workspace/provision/deployment.go
amisevsk: Fixed for now.

controllers/workspace/finalize.go
amisevsk: My current workaround is to not clear finalizers when they fail so that you can look at the pods/their logs directly to figure out what went wrong.

controllers/workspace/finalize.go
amisevsk: I haven't found an easy way of testing RWO volumes, but I'm fairly certain that the only result when another workspace has the PVC mounted will be pods stuck in a pending state until they can grab it. This shouldn't cause the job to fail.

controllers/workspace/finalize.go
sleshchenko: >  Should I also update the webhooks to block changes if the annotation is set?

There is no strong need to do that, but I don't see any reason why we should not do it.

amisevsk:
@sleshchenko I've added code to update the devworkspace status when PVC cleanup fails, however:
- We don't have a phase for `Error` -- only `Failed`
- We don't have a condition to represent a controller-side error, only `FailedStart`

To work around this for now, I'm just using strings to define "new" conditions/phases -- I'll open an issue in devfile/api to add them since I think they generally make sense to represent error states that don't come from the devfile itself.

With the current changes, a failed cleanup looks like:
- controller logs:
    ```
    2020-12-23T15:50:20.119Z	INFO	controllers.DevWorkspace	Reconciling Workspace	{"Request.Namespace": "dw", "Request.Name": "theia"}
    2020-12-23T15:50:20.119Z	INFO	controllers.DevWorkspace	Finalizing DevWorkspace	{"Request.Namespace": "dw", "Request.Name": "theia"}
    2020-12-23T15:50:20.119Z	ERROR	controllers.DevWorkspace	Failed to clean PVC on workspace deletion	{"Request.Namespace": "dw", "Request.Name": "theia", "error": "PVC clean up job failed: message \"Job has reached the specified backoff limit\""}
    ```
- cluster:
    ```
    ❯ kc get dw
    NAME    WORKSPACE ID                PHASE   URL
    theia   workspace3ca3c725d7884820   Error   http://workspace3ca3c725d7884820-theia-3100.192.168.49.2.nip.io
    
    ❯ kc get dw theia -o yaml | yq -y '.status'
    conditions:
      - lastTransitionTime: '2020-12-23T15:50:20Z'
        status: Unknown
        type: ServiceAccountReady
      - lastTransitionTime: '2020-12-23T15:50:20Z'
        status: Unknown
        type: RoutingReady
      - lastTransitionTime: '2020-12-23T15:50:20Z'
        status: Unknown
        type: Ready
      - lastTransitionTime: '2020-12-23T15:50:20Z'
        message: Failed to clean PVC on deletion. See logs for job "cleanup-workspace3ca3c725d7884820"
          for details
        status: 'True'
        type: Error
    ideUrl: http://workspace3ca3c725d7884820-theia-3100.192.168.49.2.nip.io
    phase: Error
    workspaceId: workspace3ca3c725d7884820
    ```
    users can then look at the job and its pods for more details.

WDYT?

Also: I've rebased on master and squashed the history somewhat.

sleshchenko:
Locally e2e tests fail due try to create job in terminating workspace:
```
2020-12-29T10:42:08.042Z        ERROR   controller      Reconciler error        {"reconcilerGroup": "workspace.devfile.io", "reconcilerKind": "DevWorkspace", "controller": "devworkspace", "name": "web-terminal", "namespace": "test-terminal-namespace", "error": "jobs.batch \"cleanup-workspace86c313d280504851\" is forbidden: unable to create new content in namespace test-terminal-namespace because it is being terminated"}
```

controllers/workspace/finalize.go
sleshchenko: >  but I'm fairly certain that the only result when another workspace has the PVC mounted will be pods stuck in a pending state until they can grab it. This shouldn't cause the job to fail.

That's true. can be reproduced when you remove worksspace PVC

sleshchenko:
On the PROW devworkspace hangs before removing
Screenshot_20201229_174154
I think PVC is cleaned up before job is created but after we check that PVC exists ) Added requeuing at least each 10 seconds

sleshchenko:
BTW provided logs from prow does not show any clue. Info-dumping from cluster should be improved.

sleshchenko:
Something is wrong with cleaning up after tests:
```
~/pro/devworkspace-operator pvc-cleanup-on-delete *10 > kc get pod -n devworkspace-controller                      kube admin 19:26:11
NAME                                               READY   STATUS    RESTARTS   AGE
che-plugin-registry-98b8fb5b6-7whsl                1/1     Running   0          7m44s
devworkspace-controller-manager-59dc5c6895-t9tss   1/1     Running   0          7m36s
devworkspace-webhook-server-76bccc5d67-cwlld       1/1     Running   0          7m14s


+ make uninstall
which: no kubectl in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/go/bin:/tmp/bin)
which: no controller-gen in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/go/bin:/tmp/bin)
bash: operator-sdk: command not found
2020/12/29 17:20:04 well-defined vars that were never replaced: OPERATOR_NAMESPACE
customresourcedefinition.apiextensions.k8s.io "components.controller.devfile.io" deleted
customresourcedefinition.apiextensions.k8s.io "devworkspaces.workspace.devfile.io" deleted
customresourcedefinition.apiextensions.k8s.io "devworkspacetemplates.workspace.devfile.io" deleted
customresourcedefinition.apiextensions.k8s.io "workspaceroutings.controller.devfile.io" deleted
clusterrole.rbac.authorization.k8s.io "devworkspace-controller-edit-workspaces" deleted
clusterrole.rbac.authorization.k8s.io "devworkspace-controller-metrics-reader" deleted
clusterrole.rbac.authorization.k8s.io "devworkspace-controller-proxy-role" deleted
clusterrole.rbac.authorization.k8s.io "devworkspace-controller-role" deleted
clusterrole.rbac.authorization.k8s.io "devworkspace-controller-view-workspaces" deleted
clusterrolebinding.rbac.authorization.k8s.io "devworkspace-controller-proxy-rolebinding" deleted
clusterrolebinding.rbac.authorization.k8s.io "devworkspace-controller-rolebinding" deleted

E1229 17:27:22.829229       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.RoleBinding: rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:devworkspace-controller:devworkspace-controller-serviceaccount" cannot list resource "rolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope
2020-12-29T17:27:32.450Z        INFO    controllers.DevWorkspace        Reconciling Workspace   {"Request.Namespace": "test-terminal-namespace", "Request.Name": "web-terminal"}
2020-12-29T17:27:32.450Z        INFO    controllers.DevWorkspace        Finalizing DevWorkspace {"Request.Namespace": "test-terminal-namespace", "Request.Name": "web-terminal"}
E1229 17:27:34.442742       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.RoleBinding: rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:devworkspace-controller:devworkspace-controller-serviceaccount" cannot list resource "rolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope
E1229 17:27:36.316731       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:devworkspace-controller:devworkspace-controller-serviceaccount" cannot list resource "configmaps" in API group "" at the cluster scope
E1229 17:27:37.302697       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:devworkspace-controller:devworkspace-controller-serviceaccount" cannot list resource "deployments" in API group "apps" at the cluster scope

```

sleshchenko (approved):
LGTM

I'm not going to merge this PR to let @amisevsk review the changes I've added.

amisevsk (commented):
Updates LGTM, thanks @sleshchenko for catching the edge cases (somehow I didn't test *only* creating a non-pvc workspace? :smile:)

PR #234, opened 2020-12-22
Improve logging in controller

amisevsk:
### What does this PR do?
* Add env var ($DEVELOPMENT_MODE) to deployment to enable/disable development mode; default to true for now
* Set dev mode in default logging according to $DEVELOPMENT_MODE
* Log workspace ID with all log lines in controller, to allow matching reconciles for e.g. workspaceRouting with the actual DevWorkspace
* Move controller config initialization into main.go to ensure it is completed before starting any controllers

At some point, my IDE modified whitespace in `import` blocks, causing our imports to shuffle around meaninglessly.

Minor frustrations I had with our logs:
- We had dev-mode logging hardcoded to enabled (now it's hard-coded, but in yaml! :smile:)
- It wasn't straightforward to match a log for a component/routing with the workspace that owned it

### Is it tested? How?
Tested on minikube

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,28 +17,28 @@ import (
 	"flag"
 	"os"
 
-	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers"
-
 	"github.com/devfile/devworkspace-operator/controllers/controller/component"
 	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting"
+	"github.com/devfile/devworkspace-operator/controllers/controller/workspacerouting/solvers"
 	"github.com/devfile/devworkspace-operator/internal/cluster"
+	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/pkg/webhook"
-
-	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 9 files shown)


sleshchenko (approved):

sleshchenko (commented):
We have another place that is interested in consuming Development Mode env var https://github.com/amisevsk/devworkspace-operator/blob/log-workspace-id/webhook/main.go#L50

main.go
```diff
@@ -85,6 +85,11 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	setupLog.Info("This is what logs look like by default")
```
sleshchenko, line 88: What is the purpose of this line since the logger is set up before and `setupControllerConfig` is not going to change anything? Or it was used locally and it should be removed now?

Screenshot_20201229_114437

amisevsk: It's a leftover from testing that I forgot to remove -- sorry (was testing to see if we can update the logger to add workspace_id after its created -- we can't)

amisevsk: Fixed.

JPinkney (approved):

Issue #185, opened 2020-10-22
Implement Volume component


sleshchenko:
It's needed to implement Volume component that will define PVC or emptyDir with size configuration.
```yaml
components:
  - name: maven
    container:
      image: registry.redhat.io/codeready-workspaces/stacks-java-rhel8:2.1
      mountSources: true
      volumeMounts:
        - name: my-storage
          path: /home/jboss/.settings
  - name: my-storage
    volume:
      size: 500Mi
```
If component references non-declared volume - devworkspace should fail to start or failed to be created?
Probably failed to start since such validation needs external resources to be fetched like in plugins case. See below.

? `mountSource: true` should be converted to volume with name projects?

Volume can be reused/configured from plugin. Like here https://github.com/devfile/api/blob/master/samples/devfiles/spring-boot-http-booster-devfile.yaml.

```yaml
...
components:
  - name: java-support
    plugin:
      id: redhat/java8/latest
      components:
        - name: vscode-java            
          container:
            memoryLimit: 2Gi
        - name: m2 # it already has volume define we just configure it
          volume:
            size: 2G
...
  - name: maven-tooling
    container:
      image: registry.redhat.io/codeready-workspaces/stacks-java-rhel8:2.1
      mountSources: true
      memoryLimit: 768Mi
      volumeMounts:
        - name: m2 # using volume from plugin definition
          path: /home/jboss/.m2
```

Which PVC strategy we should support in devworkspace? Probably common only - one PVC for one namespace. Should we implement the same isolation mechanism as for Che?
```
- PVC structure:
/workspaceId1
  /volumeName1
  /volumeName2
/workspaceId2
  /volumeName1
  /volumeName2
```

? Since plugins need volumes in initContainers, we can face issues with initSubpaths from initContainer https://github.com/devfile/devworkspace-operator/blob/1b1bc9e374d079e1090dad9967c719ac5915b1a2/pkg/controller/workspace/provision/deployment.go#L312 and probably we need to run a separate job to do it

PR #240, opened 2021-01-12
Plugin flattening

amisevsk:
(Don't worry about line count, 1100 lines are from test cases and 400 are from new devworkspace samples)

### What does this PR do?
Readds support for Plugins in DevWorkspaces by implementing basic plugin resolution and merging.

Commits:
- Add a basic implementation of plugins based on k8s `NamespacedName` references, as described in the devfile API spec.
- Use this new implementation to resolve plugins in the main reconcile loop (could potentially be moved to a subresource in the future)
- Some minor fixups to the reconcile loop (since we now have to be careful to not update the workspace accidentally)
- Add samples to support new functionality (`samples/plugins` contains plugin definitions; `samples/with-k8s-ref` contains devworkspaces that reference these plugins)
- Add support for plugin overriding (components and commands)
- Add checking for plugin import cycles (plugin a -> plugin b -> plugin a) to avoid killing the controller due to stack overflows
- Add *really basic* support for plugin and editor compatibility. We add labels to plugins when we create them and check those labels to report when a devworkspace uses incompatible plugins and editors
- Add test cases for the library functions (97% coverage)

In a couple of cases, I had to work around devfile/api issues: https://github.com/devfile/api/issues/296 and https://github.com/devfile/api/issues/295; these are referenced in the code with TODOs.

Hopefully commit messages are clear enough to explain the purpose of each change.

https://github.com/devfile/devworkspace-operator/issues/249; also a followup to https://github.com/devfile/devworkspace-operator/pull/237 and issue #185 

### Is it tested? How?
This PR adds a makefile rule `install_plugin_templates`, which deploys new DevWorkspaceTemplates to the `devworkspace-plugins` namespace (namespace is hardcoded so we can easily refer to them). After this is done, DevWorkspaces can be deployed with plugins included from the `samples/with-k8s-ref/` directory.

## TODO:
- [ ] Reimplement defaulting for the web-terminal container component
- [ ] Re-enable default Web Terminal DW in e2e tests.
- [x] Merge https://github.com/devfile/devworkspace-operator/pull/237 and rebase this PR


samples/with-k8s-ref/theia-next.yaml
```diff
@@ -0,0 +1,29 @@
...
+            namespace: devworkspace-plugins
+      - name: terminal
+        plugin:
+          kubernetes:
+            name: machine-exec
+            namespace: devworkspace-plugins
+    commands:
+      - id: say-hello
+        exec:
+          component: plugin
```
sleshchenko, line 27: Seems it's not a valid devfile because command references non-existing component.

samples/with-k8s-ref/theia-nodejs.yaml
```diff
@@ -0,0 +1,81 @@
...
+    components:
+      - name: che-theia
+        plugin:
+          kubernetes:
+            name: theia-next
+            namespace: devworkspace-plugins
+      - name: machine-exec
+        plugin:
+          kubernetes:
+            name: machine-exec
```
sleshchenko, line 22:
```
      - name: che-theia
        plugin:
          kubernetes:
            name: theia-next
```
already includes machine-exec, do we need to specify it here again?

Makefile
```diff
@@ -199,6 +199,11 @@ endif
 	mv config/base/config.properties.bak config/base/config.properties
 	mv config/base/manager_image_patch.yaml.bak config/base/manager_image_patch.yaml
 
+### install_plugin_templates: Deploy sample plugin templates to namespace devworkspace-plugins:
+install_plugin_templates: _print_vars
+	$(K8S_CLI) create namespace devworkspace-plugins || true
+	$(K8S_CLI) apply -f samples/plugins -n devworkspace-plugins
```
sleshchenko, line 205: I suppose we need to grant all users read-only access DevWorkspaceTemplates in this namespace.
And probably even all ServiceAccounts, otherwise how theia or dashboard will list them.

sleshchenko (commented):
Amazing work. Seems you did it faster than I need to review )

sleshchenko (commented):
Seems to work as expected.

We need to agree on the plan regards devfile's plugin registry, if we are going to support them or not (ones we're able to reference with ID, as we used to do).
if no - there is no a reason to keep it in the model.
Actually URI could work with the registry as well, it's just a question if we need some defaulting for the registry or not.

Makefile
sleshchenko: I see RBAC for SAs are updated, but usual users did not get such permissions yet. Like we have for view and edit for devworkspaces

https://github.com/amisevsk/devworkspace-operator/blob/fcad74e80920ca16b3dd432e280c79db1172b964/config/components/rbac/edit-workspaces-cluster-role.yaml

pkg/library/flatten/flatten.go
```diff
@@ -118,6 +119,10 @@ func resolvePluginComponent(
 	switch {
 	// TODO: Add support for plugin ID and URI
 	case plugin.Kubernetes != nil:
+		// Search in devworkspace's namespace if namespace ref is unset
```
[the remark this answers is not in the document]
amisevsk: :+1:

pkg/library/flatten/compatibility.go
```diff
@@ -0,0 +1,83 @@
...
+package flatten
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	EditorNameLabel          = "devworkspace.devfile.io/editor-name"
+	EditorCompatibilityLabel = "devworkspace.devfile.io/editor-compatibility"
```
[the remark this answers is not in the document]
amisevsk: Yeah this part is mostly a proof of concept for how attributes could be used to provide this functionality -- it's kind of tagged on to the cycle-checking portion.

amisevsk:
> We need to agree on the plan regards devfile's plugin registry, if we are going to support them or not (ones we're able to reference with ID, as we used to do).

I agree -- my current understanding is that `ID` would be used in reference to a [devfile/registry](https://github.com/devfile/registry) which is a devfile/api-side spec. I think it's easy enough to support that we shouldn't worry about getting rid of it, especially for backwards compatibility purposes (WTO uses plugin IDs)

@sleshchenko WDYT about:
- ID specified without `plugin.registryURL`: check "internal" registry (allows us to still support WTO use case)
- ID specified with `plugin.registryURL`: check registry at `registryURL` and assume it follows devfile/registry spec

samples/with-k8s-ref/theia-nodejs.yaml
amisevsk: I'll fix up the samples :) 

I mostly copy-pasted from scratch files I had for testing so they're sloppy.

samples/with-k8s-ref/theia-next.yaml
amisevsk: That's what I get for being careless :)

Makefile
amisevsk: Good catch -- these files are written manually (not generated) so I'll need to add DevWorkspaceTemplates here.

samples/with-k8s-ref/theia-nodejs.yaml
amisevsk: Currently, theia-next.yaml doesn't include `machine-exec` as far as I know -- if it does applying this DevWorkspace should result in failure to start.

amisevsk:
I re-squashed and PR #237 and rebased onto it, which broke any commit refs, sorry for any confusion.

samples/with-k8s-ref/theia-next.yaml
amisevsk: I squashed this fix into the commit that adds samples

pkg/library/flatten/flatten.go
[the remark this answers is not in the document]
amisevsk: Updated to `InstanceNamespace` and squashed into `aae830c - Look for plugins in the current namespace if namespace is empty`

amisevsk:
#237 is merged and this PR has been rebased onto the new master branch.

Makefile
```diff
@@ -199,6 +199,11 @@ endif
 	mv config/base/config.properties.bak config/base/config.properties
 	mv config/base/manager_image_patch.yaml.bak config/base/manager_image_patch.yaml
 
+### install_plugin_templates: Deploy sample plugin templates to namespace devworkspace-plugins:
+install_plugin_templates: _print_vars
```
sleshchenko, line 203: I'm OK with it as temporary solution until Theia did not provide DWT for its plugins. But generally I'm not sure we should have a dedicated namespace for devworkspace-plugins, and to avoid security issues we should disable cross-namespace DTW usage at this point.

sleshchenko (approved):
Approving it, since I'm OK with the proposed devworkspace startup flow.
Did not re-review or retest, but everything was fine I tried it.

The TODOs are listed in https://github.com/devfile/devworkspace-operator/issues/249 as separate items and for me it's OK to handle them in further PRs.

Makefile
amisevsk: Yeah, this is a development feature, since we need something to replace plugin registry until something else is implemented. this rule deploys sample plugins that are compatible with devfile samples.

pkg/library/flatten/compatibility.go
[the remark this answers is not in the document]
amisevsk: This bit has been removed from the current PR.

Makefile
amisevsk: Done in 83f5133

amisevsk:
Reminding myself of why those TODOs are there (a couple of weeks later :)) -- I think it makes sense to do that as a separate PR. This PR implements "plugins via kubernetes reference" but WTO depends on "plugins via plugin ID". I don't expect it to be too complicated but it's probably cleaner than adding to this already-large PR. 

I'll retest everything on minikube and `crc` and merge once all tests are passing.

amisevsk:
Tested current changes:
| Platform | Workspace | Works? | Notes |
|---|---|---|---|
| Kubernetes | with-k8s-ref/web-terminal-dev.yaml | Starts, no UI | Expected since on k8s web terminal is not supported |
| | with-k8s-ref/theia-next.yaml | Works | Note this works with Theia importing its own plugins (e.g. vsx-installer) |
| | with-k8s-ref/theia-nodejs.yaml | Works | |
| OpenShift | with-k8s-ref/web-terminal-custom-tooling.yaml | Starts, unable to start web terminal due to error: `Error Loading OpenShift command line terminal: Could not retrieve $KUBECONFIG` (see below) | Note: the custom tooling variant is required since we don't have defaulting for the dev container | 
| | with-k8s-ref/theia-next.yaml | Works | |
| | with-k8s-ref/theia-nodejs.yaml | Works, sort of | Using tasks fails with `Error launching task 'download-dependencies': Request runTask failed with message: Failed to execute Che command: self signed certificate in certificate chain` |

The default use case for web terminal on OpenShift fails because the web-terminal plugin tries to read `$KUBECONFIG` from the first container in the pod. This functionality depends on the tooling container being first in the list, and fails for the current PR as web-terminal ends up being the first container due to the way merging works in devfile/api. I'm not blocking merge of this PR as-is because this failure mode should be handled in web-terminal, IMO. I've created https://github.com/devfile/devworkspace-operator/issues/258 to track this

Makefile
amisevsk: As for cross-namespace dwt use, I'm not sure I see the security risk necessarily. We end up running into an issue of cleanup if we're spreading dwts across the cluster, for statically created resources as we currently use.

amisevsk:
Merging PR as-is to unblock other work -- it seems fairly stable with known issues.

PR #245, opened 2021-01-15
Add support for configuring the workspace routing controllers through annotations on DevWorkspace.

metlos:
### What does this PR do?
It copies annotations with certain format to the workspace routing objects expecting them to contain the configuration of the external routing controller in question.

#243

### DIFF
--- a/controllers/workspace/provision/routing.go
+++ b/controllers/workspace/provision/routing.go
@@ -149,7 +149,7 @@ func getSpecRouting(
 	}
 
 	// copy the annotations for the specific routingClass from the workspace object to the routing
-	expectedAnnotationPrefix := workspace.Spec.RoutingClass + config.RoutingClassAnnotationInfix
+	expectedAnnotationPrefix := workspace.Spec.RoutingClass + config.RoutingAnnotationInfix
 	for k, v := range workspace.GetAnnotations() {
 		if strings.HasPrefix(k, expectedAnnotationPrefix) {
 			annotations[k] = v
--- a/controllers/workspace/provision/routing.go
+++ b/controllers/workspace/provision/routing.go
@@ -15,6 +15,7 @@ package provision
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	devworkspace "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
@@ -145,6 +146,14 @@ func getSpecRouting(
 	if val, ok := workspace.Annotations[config.WorkspaceRestrictedAccessAnnotation]; ok {
 		annotations = map[string]string{}
 		annotations[config.WorkspaceRestrictedAccessAnnotation] = val
+	}
+
+	// copy the annotations for the specific routingClass from the workspace object to the routing
+	expectedAnnotationPrefix := workspace.Spec.RoutingClass + config.RoutingClassAnnotationInfix
+	for k, v := range workspace.GetAnnotations() {
+		if strings.Index(k, expectedAnnotationPrefix) == 0 {
+			annotations[k] = v
+		}
 	}
 
 	routing := &v1alpha1.WorkspaceRouting{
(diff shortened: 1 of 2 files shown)


pkg/config/constants.go
```diff
@@ -70,6 +70,10 @@ const (
 
 	// PVCCleanupPodMemoryLimit is the memory limit used for PVC clean up pods
 	PVCCleanupPodMemoryLimit = "32Mi"
+
+	// RoutingClassAnnotationInfix is the infix of the annotations of DevWorkspace that are passed down as annotation to the WorkspaceRouting objects.
+	// The full annotation name is supposed to be "<routingClass>.routingclass.controller.devfile.io/<anything>"
+	RoutingClassAnnotationInfix = ".routingclass.controller.devfile.io/"
```
sleshchenko, line 76: this annotation is propagated to routing (technically to workspacerouting) but not to routingclass.
Ingress implementation use the following format:
```
nginx.ingress.kubernetes.io/...
traefik.ingress.kubernetes.io/...
```
So, I think we can simplify annotation name without harm, like
`.routing.controller.devfile.io/`
or longer `.workspacerouting.controller.devfile.io/` (the CRD identifier is `workspacerouting.controller.devfile.io`). I prefer the first.

controllers/workspace/provision/routing.go
```diff
@@ -147,6 +148,14 @@ func getSpecRouting(
 		annotations[config.WorkspaceRestrictedAccessAnnotation] = val
 	}
 
+	// copy the annotations for the specific routingClass from the workspace object to the routing
+	expectedAnnotationPrefix := workspace.Spec.RoutingClass + config.RoutingClassAnnotationInfix
+	for k, v := range workspace.GetAnnotations() {
+		if strings.Index(k, expectedAnnotationPrefix) == 0 {
```
sleshchenko, line 154: `strings.Index(k, expectedAnnotationPrefix) == 0`
 -> 
 `strings.HasPrefix(k, expectedAnnotationPrefix)`
 ?

sleshchenko (approved):
LGTM
please wait @amisevsk approval

JPinkney (approved):

amisevsk (approved):
LGTM.

Does it make sense to extend this in the future so that e.g. any annotations with the `.routing.controller.devfile.io/` infix are propagated to the WorkspaceRouting subresource? 

Alternatively (and more long term) we could use the attributes field from the devfile to pass more structured data (e.g. we could pass `json` instead of a string). Currently, raw devfiles [support top-level attributes](https://github.com/devfile/api/blob/2c77856341ea89fc554890aac7462467e5768d93/pkg/devfile/header.go#L32) but DevWorkspaces do not, so it's not an easy 1-1 translation.

We can improve this later if we find the need for it.

davidfestal:
> Does it make sense to extend this in the future so that e.g. any annotations with the `.routing.controller.devfile.io/` infix are propagated to the WorkspaceRouting subresource?

+1
It does to me

> 
> Alternatively (and more long term) we could use the attributes field from the devfile to pass more structured data (e.g. we could pass `json` instead of a string). Currently, raw devfiles [support top-level attributes](https://github.com/devfile/api/blob/2c77856341ea89fc554890aac7462467e5768d93/pkg/devfile/header.go#L32) but DevWorkspaces do not, so it's not an easy 1-1 translation.
> 
> We can improve this later if we find the need for it.

I don't think Devfile attributes should be used to convey information about how a DevWorkspace instance will behave. So everything related to routing or more generally the setup of a workspace instance should probably live in the `DevWorkspace` K8S resource annotations.

davidfestal (approved):

metlos:
> Does it make sense to extend this in the future so that e.g. any annotations with the .routing.controller.devfile.io/ infix are propagated to the WorkspaceRouting subresource?

Wait, that's what I thought I was doing here :) This should propagate all the annos with the appropriate prefix (composed of the concrete routing class and the infix) to the workspace routing subresource.

Have I missed some place where this needs to be done, too?

amisevsk:
@metlos You've done what you think (e.g. `che.routing.controller.devfile.io/*` is propagated -- I was moreso thinking along the lines of also propagating `my-field.routing.controller.devfile.io/*` or even `routing.controller.devfile.io/anything`. I.e. `routing.controller.devfile.io/*` defines an annotation we propagate in all cases.

metlos:
I'm merging this as is and we can revisit later if we find the need to propagate more loosely as suggested by @amisevsk .

PR #246, opened 2021-01-15
Add CI check that generated files are up-to-date

amisevsk:
### What does this PR do?
Add a step to the GitHub CI jobs to check that `make generate manifests` does not change repo (ensure that generated yaml manifests and deepcopy functions are up to date).

Update annotations on devworkspace reconciler to make sure we generate the RBAC we currently are using.

Running `make manifests` changes role.yaml

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -58,6 +58,19 @@ jobs:
           exit 1
         fi
     -
+      name: Check manifests
+      run: |
+        git reset HEAD --hard
+        pip install yq
+        # Note: fmt is necessary after generate since generated sources will
+        #       fail format check by default.
+        make generate fmt manifests
+        if [[ ! -z $(git status -s) ]]
+        then
+          echo "generated sources are not up to date"
+          exit 1
+        fi
+    -
       name: Run Go Tests
       run: |
         python -m pip install --upgrade pip yq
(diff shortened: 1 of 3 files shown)


sleshchenko (approved):

.github/workflows/pr.yml
```diff
@@ -57,6 +57,19 @@ jobs:
...
           exit 1
         fi
+    -
+      name: Check manifests
+      run: |
+        git reset HEAD --hard
+
+        # Note: fmt is necessary after generate since generated sources will
+        #       fail format check by default.
+        make generate fmt manifests
```
sleshchenko, line 67: it fails with `Error patching crds: yq is required`

amisevsk: Ah bummer. That's probably why we didn't include it it in the first place.

amisevsk:
Fixed CI (:tada:) and also improved Makefile slightly to not print warnings needlessly ("current-context is not set" and "operator-sdk is not installed")

.github/workflows/pr.yml
amisevsk: Added `pip install yq` and it seems to work without problems

amisevsk:
I also added a commit (since dropped) that changed a manifest file to verify that CI will fail as we expect.

PR #248, opened 2021-01-21
Allow RoutingSolver.GetSpecObjects to return an error

metlos:
### What does this PR do?
This adds the ability for the RoutingSolver.GetSpecObjects to also return an error.

This seems to be required in `devworkspace-che-operator` which needs to contact k8s API
during this call.

eclipse/che#18153

### Is it tested? How?
No tests added as this just changes the method signature and no existing impl needs to
return the errors.

### DIFF
--- a/controllers/controller/workspacerouting/workspacerouting_controller.go
+++ b/controllers/controller/workspacerouting/workspacerouting_controller.go
@@ -97,9 +97,7 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 			return reconcile.Result{}, nil
 		}
 		reqLogger.Error(err, "Invalid routing class for workspace")
-		instance.Status.Phase = controllerv1alpha1.RoutingFailed
-		statusErr := r.Status().Update(ctx, instance)
-		return reconcile.Result{}, statusErr
+		return reconcile.Result{}, r.markRoutingFailed(instance)
 	}
 
 	// Add finalizer for this CR if not already present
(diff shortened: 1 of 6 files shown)


controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -114,7 +114,11 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 	}
 
 	restrictedAccess, setRestrictedAccess := instance.Annotations[config.WorkspaceRestrictedAccessAnnotation]
-	routingObjects := solver.GetSpecObjects(instance, workspaceMeta)
+	routingObjects, err := solver.GetSpecObjects(instance, workspaceMeta)
+	if err != nil {
+		return reconcile.Result{}, err
```
amisevsk, line 119: One thing to think about here is whether we should support failing the workspace start from this error (either via an error type or via another return value). This can come as a separate PR, though.

amisevsk (approved):
LGTM :+1:

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: The general flows we should think about are
1. Unexpected error -> return error
2. Expected, transient error (something's not ready) -> print message, return `reconcile.Result{RequeueAfter: 1 * time.Second}, nil`
3. Expected, fatal error (something's not configured) -> fail workspace startup; print message.

JPinkney (approved):

sleshchenko (approved):

controllers/controller/workspacerouting/workspacerouting_controller.go
metlos: I like the idea very much. How does one fail workspace startup though?

metlos: I've added new changes that implement at least the "controller not ready, please retry" part..

amisevsk: The additions look great so far! If you want to fail workspace startup, you can do what we do if e.g. `RoutingNotSupported` is returned above:

```
		instance.Status.Phase = controllerv1alpha1.RoutingFailed
		statusErr := r.Status().Update(ctx, instance)
		return reconcile.Result{}, statusErr
```
This will be read by the DevWorkspace controller on the next reconcile, which will cause the overall DW to fail to start. We also have a TODO to propagate some sort of status message from the routing to specify what the issue is; I'll get to that sometime soon hopefully.

amisevsk (approved):
Additional changes look good :+1:

Great addition!

PR #251, opened 2021-01-26
Make the service definition for endpoints public and make it slightly more versatile.

metlos:
### What does this PR do?
`$TITLE`.

The functions commonly used in the solvers contained within the main operator prove to be useful also outside in the external operators. This PR makes these functions public and adds a little bit more versatility to them.

Also, it makes sure that no service is created for a discoverable endpoint that also has "none" exposure (and hence no service should exist for it).

https://github.com/eclipse/che/issues/18583

### Is it tested? How?
Manually by making sure a workspace can still be deployed after the changes.

### DIFF
--- a/controllers/controller/workspacerouting/solvers/common.go
+++ b/controllers/controller/workspacerouting/solvers/common.go
@@ -13,6 +13,7 @@
 package solvers
 
 import (
+	"github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 	devworkspace "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/pkg/common"
@@ -31,10 +32,16 @@ type WorkspaceMetadata struct {
 	RoutingSuffix string
 }
 
-func getDiscoverableServicesForEndpoints(endpoints map[string]controllerv1alpha1.EndpointList, meta WorkspaceMetadata) []corev1.Service {
+// GetDiscoverableServicesForEndpoints converts the endpoint list into a set of services, each corresponding to a single discoverable
+// endpoint from the list. Endpoints with the NoneEndpointExposure are ignored.
+func GetDiscoverableServicesForEndpoints(endpoints map[string]controllerv1alpha1.EndpointList, meta WorkspaceMetadata) []corev1.Service {
 	var services []corev1.Service
 	for _, machineEndpoints := range endpoints {
 		for _, endpoint := range machineEndpoints {
+			if endpoint.Exposure == devworkspace.NoneEndpointExposure {
+				continue
+			}
+
 			if endpoint.Attributes.GetBoolean(string(controllerv1alpha1.DISCOVERABLE_ATTRIBUTE), nil) {
 				// Create service with name matching endpoint
 				// TODO: This could cause a reconcile conflict if multiple workspaces define the same discoverable endpoint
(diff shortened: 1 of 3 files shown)


controllers/controller/workspacerouting/solvers/common.go
```diff
@@ -68,50 +75,69 @@ func getDiscoverableServicesForEndpoints(endpoints map[string]controllerv1alpha1
...
+	// "set" of ports that are still left for exposure
+	ports := map[int]struct{}{}
+	for _, es := range endpoints {
+		for _, endpoint := range es {
+			ports[endpoint.TargetPort] = struct{}{}
+		}
+	}
+
+	// "set" of exposure types that are allowed
+	validExposures := map[v1alpha2.EndpointExposure]struct{}{}
```
amisevsk, line 103: Canonically, a set can be defined as `map[v1alpha2.EndpointExposure]bool` -- then you can use it as expected:
```golang
// Add to set:
validExposures[myExposure] = true
// Check membership:
if validExposures[myExposure] {
  // set contains myExposure
}
```
This uses the helpful fact that trying to read a non-existent key from a map returns the types zero value (which is `false` for bool)

amisevsk (commented):
Generally LGTM from a read through; I'll test the code soon.

controllers/controller/workspacerouting/solvers/common.go
metlos: Interesting. The code is updated now.

amisevsk (approved):
LGTM

JPinkney (approved):

controllers/controller/workspacerouting/solvers/common.go
```diff
@@ -68,50 +75,69 @@ func getDiscoverableServicesForEndpoints(endpoints map[string]controllerv1alpha1
...
-				servicePorts = append(servicePorts, servicePort)
+				})
 			}
 		}
 	}
 
+	return service
+}
+
+func getServicesForEndpoints(endpoints map[string]controllerv1alpha1.EndpointList, meta WorkspaceMetadata) []corev1.Service {
```
sleshchenko, line 134: Is this func needed?
Is so, should it be public?

controllers/controller/workspacerouting/solvers/common.go
```diff
@@ -68,50 +75,69 @@ func getDiscoverableServicesForEndpoints(endpoints map[string]controllerv1alpha1
 	return services
 }
 
-func getServicesForEndpoints(endpoints map[string]controllerv1alpha1.EndpointList, meta WorkspaceMetadata) []corev1.Service {
-	if len(endpoints) == 0 {
-		return nil
+// GetServiceForEndpoints returns a single service that exposes all endpoints of given exposure types, possibly also including the discoverable types.
+func GetServiceForEndpoints(endpoints map[string]controllerv1alpha1.EndpointList, meta WorkspaceMetadata, includeDiscoverable bool, exposureType ...devworkspace.EndpointExposure) corev1.Service {
```
sleshchenko, line 79: Should it return (no-op/redundant) service if none of endpoints satisfies the specified exposureType?

sleshchenko (approved):
LGTM, just a few minor questions

controllers/controller/workspacerouting/solvers/common.go
metlos: This is the original "shape" of the function and as such is used in a number of places in the codebase. l didn't want to introduce more change than required so I just reimplemented this existing method using the new one for consistency but left it here.

I don't think it should be public because it hides too many "knobs" and as such functions more as a utility method for the code in the package than a library method to be exposed.

controllers/controller/workspacerouting/solvers/common.go
metlos: Good point!

PR #252, opened 2021-01-27
Remove plugin registry

amisevsk:
### What does this PR do?
Removes old code related to devfile v1 and plugin meta.yamls:
- Remove internal registry and related packages
- Remove plugin registry from deployment (and the requirement to set plugin registry in configmap)
- Remove the component controller (the spec for `Components` is not helpful for what we need to do with DevWorkspaces.)
- Remove `pkg/adaptor` since that was used in the component controller

Also has minor fixes for issues I encountered while cleaning up the code:
- Avoid make logging `The connection to the server xxxx was refused - did you specify the right host or port?` when not connected to a cluster
- Fix a nil map warning

This PR leaves the Component CRD in place as we should figure out how to deprecate it in WTO.

Follow-up to https://github.com/devfile/devworkspace-operator/pull/237, which removes components from the reconcile process and disables support for plugins. Plugins support is re-implemented in https://github.com/devfile/devworkspace-operator/pull/240, without the use of components or related code.

### Is it tested? How?
Tested on `minikube` and `crc`.

### Additional Information
This PR also makes it easier to support cases like using `chectl` to deploy this operator, as setting the plugin registry URL in the controller's configmap could be complicated to do at build time.


JPinkney (approved):
LGTM

controllers/workspace/provision/routing.go
```diff
@@ -144,15 +145,14 @@ func getSpecRouting(
 
 	var annotations map[string]string
 	if val, ok := workspace.Annotations[config.WorkspaceRestrictedAccessAnnotation]; ok {
-		annotations = map[string]string{}
-		annotations[config.WorkspaceRestrictedAccessAnnotation] = val
+		maputils.Append(annotations, config.WorkspaceRestrictedAccessAnnotation, val)
```
sleshchenko, line 148:
```suggestion
		annotations = maputils.Append(annotations, config.WorkspaceRestrictedAccessAnnotation, val)
```
you're caught on the confusion you predicted )

controllers/workspace/provision/routing.go
```diff
@@ -144,15 +145,14 @@ func getSpecRouting(
...
 	}
 
 	// copy the annotations for the specific routingClass from the workspace object to the routing
 	expectedAnnotationPrefix := workspace.Spec.RoutingClass + config.RoutingAnnotationInfix
 	for k, v := range workspace.GetAnnotations() {
 		if strings.HasPrefix(k, expectedAnnotationPrefix) {
-			annotations[k] = v
+			maputils.Append(annotations, k, v)
```
sleshchenko, line 155: Technically it should work but as long as map is initialized before. So, safer option is to do the following, which is forced by common Append for slices.
```suggestion
			annotations = maputils.Append(annotations, k, v)
```

sleshchenko (approved):
Approved assuming that WTO related stuff will be restored in https://github.com/devfile/devworkspace-operator/pull/252 or in further PR:
- default WTO container
- using WTO plugin by id (at least for time being until we did not find better way)
- RELATED_IMAGE_ stuff

controllers/workspace/provision/routing.go
amisevsk: Squashed into b40bf35eb061c12ae61b1825b072ef2778a44886

controllers/workspace/provision/routing.go
amisevsk: Squashed into b40bf35eb061c12ae61b1825b072ef2778a44886

amisevsk:
I've rebased on master. To make rebasing https://github.com/devfile/devworkspace-operator/pull/263 on top of this easier, I've undone some of the changes (remove meta.yamls from the internal registry and re-add some of the image-handling functions.

PR #253, opened 2021-01-28
Enable the solvers to do the finalization of a workspace routing themselves.

metlos:
### What does this PR do?
The motivation for doing this is the need of the devworkspace-che-operator
to do a custom finalization of the routing - it needs to delete
the configmaps for the gateway configuration which need to live in
a different namespace than the workspace.

For this to make sense, the openshift oauth solver is now in charge of
managing the oauthclient for the workspace. As such the oauthclient is no
longer part of the ResolvedObjects struct.

eclipse/che#18583

### DIFF
--- a/controllers/controller/workspacerouting/solvers/solver.go
+++ b/controllers/controller/workspacerouting/solvers/solver.go
@@ -17,7 +17,6 @@ import (
 
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/pkg/config"
-	oauthv1 "github.com/openshift/api/oauth/v1"
 	routeV1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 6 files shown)


controllers/controller/workspacerouting/solvers/solver.go
```diff
@@ -29,14 +28,25 @@ type RoutingObjects struct {
 	Ingresses    []v1beta1.Ingress
 	Routes       []routeV1.Route
 	PodAdditions *controllerv1alpha1.PodAdditions
-	OAuthClient  *oauthv1.OAuthClient
```
amisevsk, line 32: +1000

amisevsk (approved):
Haven't tested directly, but these changes are a great idea. I always hated the special treatment of OAuthClients.

JPinkney (approved):

sleshchenko (approved):
Did not test but changes looks good.
:+1:

PR #255, opened 2021-02-02
Propagate devworkspace CR reference as env vars

sleshchenko:
### What does this PR do?
This PR moves every Che API sidecar env vars to the corresponding file + add propagating devworkspace CR reference as env vars: Name, Namespace, ID.

No particular issue is created, but it's needed to help Theia access devworkspace related objects.

### Is it tested? How?
not yet

### DIFF
--- a/controllers/workspace/env/common_env_vars.go
+++ b/controllers/workspace/env/common_env_vars.go
@@ -20,39 +20,16 @@ import (
 func CommonEnvironmentVariables(workspaceName, workspaceId, namespace, creator string) []corev1.EnvVar {
 	return []corev1.EnvVar{
 		{
-			Name: "CHE_MACHINE_TOKEN",
+			Name:  "DEVWORKSPACE_NAMESPACE",
+			Value: namespace,
 		},
 		{
-			Name:  "CHE_PROJECTS_ROOT",
-			Value: config.DefaultProjectsSourcesRoot,
-		},
-		{
-			Name:  "CHE_API",
-			Value: config.DefaultApiEndpoint,
-		},
-		{
-			Name:  "CHE_API_INTERNAL",
-			Value: config.DefaultApiEndpoint,
-		},
-		{
-			Name:  "CHE_API_EXTERNAL",
-			Value: config.DefaultApiEndpoint,
-		},
-		{
-			Name:  "CHE_WORKSPACE_NAME",
+			Name:  "DEVWORKSPACE_NAME",
 			Value: workspaceName,
 		},
 		{
-			Name:  "CHE_WORKSPACE_ID",
+			Name:  "DEVWORKSPACE_ID",
 			Value: workspaceId,
-		},
-		{
-			Name:  "CHE_AUTH_ENABLED",
-			Value: config.AuthEnabled,
-		},
-		{
-			Name:  "CHE_WORKSPACE_NAMESPACE",
-			Value: namespace,
 		},
 		{
 			Name:  "DEVWORKSPACE_CREATOR",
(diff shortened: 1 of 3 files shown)


controllers/workspace/env/common_env_vars.go
```diff
@@ -20,40 +20,17 @@ import (
...
-		},
-		{
-			Name:  "CHE_API_INTERNAL",
-			Value: config.DefaultApiEndpoint,
-		},
-		{
-			Name:  "CHE_API_EXTERNAL",
-			Value: config.DefaultApiEndpoint,
+			Name:  "DEVWORKSPACE_NAMESPACE",
+			Value: namespace,
```
amisevsk, line 24: This is kind of redundant since namespace can be read from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`, but I see how this would be useful so :+1:

amisevsk (approved):
LGTM

benoitf:
don't know if it's me, but trying with image built on top of this PR, samples/flattened_theia-next.yaml is not deploying

I've a CrashLoopBack pod
```
che                       workspace37f11e3c952a4542-754d85c657-vsvrw         0/3     Init:CrashLoopBackOff   1          19s
```

benoitf:
Looking at theia-next vsx-installer command
it's invoking https://kubernetes.default.svc/apis/workspace.devfile.io/v1alpha2/namespaces//devworkspaces/ (meaning that CHE_WORKSPACE_NAMESPACE and CHE_WORKSPACE_NAME are not there
in "https://kubernetes.default.svc/apis/workspace.devfile.io/v1alpha2/namespaces/${CHE_WORKSPACE_NAMESPACE}/devworkspaces/${CHE_WORKSPACE_NAME}"

pkg/library/shim/component.go
```diff
@@ -26,12 +26,57 @@ import (
...
 
-func FillDefaultEnvVars(podAdditions *v1alpha1.PodAdditions, workspace devworkspace.DevWorkspaceTemplateSpec) {
+func FillDefaultEnvVars(podAdditions *v1alpha1.PodAdditions, workspace devworkspace.DevWorkspace) {
 	for idx, mainContainer := range podAdditions.Containers {
-		podAdditions.Containers[idx].Env = append(mainContainer.Env, corev1.EnvVar{
+		podAdditions.Containers[idx].Env = append(mainContainer.Env, defaultEnvVars(mainContainer.Name, workspace)...)
+	}
+
+	for idx, mainContainer := range podAdditions.InitContainers {
+		podAdditions.Containers[idx].Env = append(mainContainer.Env, defaultEnvVars(mainContainer.Name, workspace)...)
```
benoitf, line 35:
```suggestion
		podAdditions. InitContainers[idx].Env = append(mainContainer.Env, defaultEnvVars(mainContainer.Name, workspace)...)
```

?

benoitf (changes requested):
please fix typo else we don't have env var in init containers

sleshchenko:
> please fix typo else we don't have env var in init containers

sorry for bad quality of this PR. I'll fix it and retest that everything works as expected!

pkg/library/shim/component.go
amisevsk: Good catch, I skimmed too quickly.

benoitf:
@sleshchenko not a big deal  :-)
 it works fine once typo is fixed
Thx

benoitf (approved):

controllers/workspace/env/common_env_vars.go
sleshchenko: That's true that it duplicates already available info, but typically it's easier to read env var than file, it's why I added it as well, assuming that it should not hurt anyone

Issue #250, opened 2021-01-21
Begin publishing flattened, ready-to-deploy yaml templates of the controller


amisevsk:
## Description

Currently, the DevWorkspace operator is only deployable by setting environment variables and running `kustomize`. To support deploying the operator via [`chectl`](https://github.com/che-incubator/chectl), we need to publish processed yamls that can be referenced in `chectl` without the need to run kustomization. This could potentially be done via GitHub releases to avoid filling the repo with thousands of lines of yaml.

PR #257, opened 2021-02-03
Add script for generating deployment yamls and static deployment files

amisevsk:
### What does this PR do?
This PR is a rework of #254. See that PR for context.

- Add script `config/generate-templates.sh` which outputs the yaml formerly used in `make install`
    - Script supports `--use-defaults` option to not respect current shell environment and use default values instead, though this does copy values from the Makefile (which we can maybe avoid)
- Add makefile rules `generate_deployment` and `generate_default_deployment`. The former generates yamls as before and is intended for development use; the latter generates static files that can be applied to any cluster (after routing suffix is set)
- Add PR check that `make generate_default_deployment` does not change files, as those now need to be kept in-sync with templates.

As in #254, the output of the script is
```
config/{static,current}
├── kubernetes
│   ├── objects
│   │   ├── <object-name>.<object-kind.yaml
│   │   └── (...)
│   └── combined.yaml
└── openshift
    ├── objects
    │   ├── <object-name>.<object-kind.yaml
    │   └── (...)
    └── combined.yaml
```

Closes https://github.com/devfile/devworkspace-operator/issues/250

### Is it tested? How?
Tested deployment manually, it's basically just moving code around.


config/generate-templates.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+#
+# Copyright (c) 2012-2018 Red Hat, Inc.
```
sleshchenko, line 3: I don't know if it's copied/pasted or generated but it's time to update years )

config/generate-templates.sh
```diff
@@ -0,0 +1,125 @@
...
+#
+# Note: The configmap generated when using `--use-defaults` will have an empty
+# value for devworkspace.routing.cluster_host_suffix as there is no suitable
+# default.
+
+set -e
+
+SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
+
+DEVWORKSPACE_BRANCH=${DEVWORKSPACE_BRANCH:-master}
```
sleshchenko, line 29: Seems this var is not used anymore

Makefile
```diff
@@ -176,26 +176,20 @@ install_crds: _kustomize _init_devworkspace_crds
...
+	$(K8S_CLI) apply -f config/current/openshift/combined.yaml || true
 endif
 
-	mv config/cert-manager/kustomization.yaml.bak config/cert-manager/kustomization.yaml
-	mv config/service-ca/kustomization.yaml.bak config/service-ca/kustomization.yaml
-	mv config/base/config.properties.bak config/base/config.properties
-	mv config/base/manager_image_patch.yaml.bak config/base/manager_image_patch.yaml
+### generate_deployment: Generate files used for deployment from kustomize templates, using environment variables
+generate_deployment:
+	config/generate-templates.sh
```
sleshchenko, line 188: Generate deployments invokes generate templates )
Maybe it makes sense to rename the script to generate_deployment as well?

.github/workflows/pr.yml
```diff
@@ -70,6 +70,15 @@ jobs:
           echo "generated sources are not up to date"
           exit 1
         fi
+    -
+      name: Check deployment files
+      run: |
+        git reset HEAD --hard
```
sleshchenko, line 76: Why do we need it? I see that we need it for check format, since `go get -u` modifies go.mod, but here if there is any change happened - we should fail earlier (the same question for the step above)

sleshchenko (approved):
LGTM

Please consider renaming `generate-templates.sh`

JPinkney (approved):
LGTM other than Serhii's comments

Makefile
amisevsk: Yeah, I don't have a good term for "all of the yaml required to deploy the DevWorkspaceOperator" :)

.github/workflows/pr.yml
amisevsk: I'm just following the format we have for other tasks :)

Ideally, CI would fallthrough and report all issues but I understand why it doesn't.

Makefile
sleshchenko: Is it necessary to have kustomize templates in config? It always confused me, so, if possible (which should be I think) I propose the following structure:
```
/deploy # at least cert-manager, eclipse-che use deploy folder to keep deployment (not k8s deployment) files
        # nginx ingress controller uses deployments https://github.com/nginxinc/kubernetes-ingress/tree/master/deployments
  /templates # it's where our kustomize templates live
  /deployment
    /openshift
    /kubernetes
  /current # it's where Makefile's static file live
```
WDYT?

amisevsk: That would work, just gotta change some paths.

amisevsk: It is done :tada:

.github/workflows/pr.yml
amisevsk: Removed these lines

config/generate-templates.sh
amisevsk: Removed

config/generate-templates.sh
amisevsk: I don't know where I copy-pasted from :)

amisevsk: Fixed all license years

.gitattributes
```diff
@@ -0,0 +1,2 @@
+deploy/deployment             linguist-generated=true
```
sleshchenko, line 1: :heart: for idea

but it does not seem to work.
Seems to should be
```
deploy/deployment/*             linguist-generated
deploy/templates/crd/bases/*    linguist-generated
```
https://github.com/devfile/devworkspace-operator/compare/master...sleshchenko:static-deploy

sleshchenko (approved):

.gitattributes
amisevsk: Why would the [github docs](https://docs.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github) lie to me like that :D

amisevsk: Cherry-picked your commit and squashed it.

amisevsk:
Rebased on master; it was somewhat complicated due to https://github.com/devfile/devworkspace-operator/pull/252

sleshchenko:
Tests are failed due `yq` lack: `ERROR: Program yq is required for this script`
I tried but did not find really minimal `yq` container. The best I found is `jshimko/kube-tools-aws` 351MB.
I see the following approaches:
- migrate to `yq` from container
- install yq on the PROW
- add an ability to deploy DWO from generated templates, where we just sed IMG (I don't see e2e tests needs anything else https://github.com/devfile/devworkspace-operator/blob/master/.ci/openshift_e2e.sh#L38 )

Please go either of proposed or your way to fix e2e tests.

amisevsk:
/test v5-devworkspaces-operator-e2e

Had to rebase to ensure all license years are up to date.

.gitattributes
sleshchenko: `linguist-generated=true` should work as well.
The cause it did not work I think missed `*`, so files inside folders are not considered as generated by default.

PR #259, opened 2021-02-03
Let the solver getter set up the controller manager

metlos:
### What does this PR do?
This enables the solver to reconcile different kinds of objects that it might be managing.

I specifically allowed full control over the manager builder, because in the `devworkspace-che-operator` I need to react on objects in different namespace than the one where workspace-routing objects live and so `Owns()` would not work for me.

eclipse/che#18583

### DIFF
--- a/controllers/controller/workspacerouting/solvers/solver.go
+++ b/controllers/controller/workspacerouting/solvers/solver.go
@@ -21,6 +21,7 @@ import (
 	routeV1 "github.com/openshift/api/route/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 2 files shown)


sleshchenko (approved):

JPinkney (approved):

amisevsk (approved):
I've no real issues with this PR, but I am having some trouble imagining how this will be used. The reconcile loop will trigger for `WorkspaceRouting` resources, so which routing will be enqueued for the external resource? If this doesn't make sense, feel free to merge anyways.

Edit: Reviewed this before seeing https://github.com/che-incubator/devworkspace-che-operator/pull/9 :smile:

PR #268, opened 2021-02-13
Support poststart events in devworkspaces

amisevsk:
### What does this PR do?
Add support for `events.postStart` in devfiles by converting the command to a postStart lifecycle hook on the relevant container.

One rough spot in the implementation is that there can only be one postStart command per container. I get around this by using `"/bin/bash" "-c"` and concatenating the command with `&&`. If we can restrict postStart commands to one per container, then this could be cleaner.

There's a known issue here too: Theia doesn't handle loading plugins if the vsix isn't downloaded when Theia starts. To get e.g. the vim plugin to be recognized, I had to stop and start the workspace. This is a Theia issue, IMO.

Draft PR until:
- [ ] Implement preStop as well, since it's basically the same logic
- [ ] Add test cases to cover new code

I can't seem to find it

### Is it tested? How?
```bash
make docker install install_plugin-templates
kind: DevWorkspace
metadata:
  name: theia-nodejs
spec:
  started: true
  template:
    projects:
      - name: project
        git:
          remotes:
    components:
      - name: che-theia
        plugin:
          kubernetes:
            name: theia-next
      - name: machine-exec
        plugin:
          kubernetes:
            name: machine-exec
      - name: typescript
        plugin:
          kubernetes:
            name: vscode-typescript
          components:
            - name: sidecar-typescript
              container:
                memoryLimit: 512Mi
      - name: vscode-vim
        plugin:
          kubernetes:
            name: vscode-vim
      - name: nodejs
        container:
          image: quay.io/eclipse/che-nodejs10-ubi:nightly
          memoryLimit: 512Mi
          endpoints:
            - name: nodejs
              protocol: http
              targetPort: 3000
          mountSources: true
EOF
# wait until pod is running, open Theia in browser and open terminal in e.g. Theia container:
ls -alR /plugins # should show two downloaded vsixes, one in /plugins and one in /plugins/sidecar/<plugin-name>
```

### DIFF
--- a/pkg/library/lifecycle/lifecycle.go
+++ b/pkg/library/lifecycle/lifecycle.go
@@ -15,6 +15,8 @@ package lifecycle
 import (
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
+
 	"github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/library/lifecycle/lifecycle.go
```diff
@@ -71,6 +73,50 @@ func GetInitContainers(devfile v1alpha2.DevWorkspaceTemplateSpecContent) (initCo
...
+}
+
+func writePostStartLifecycle(container *corev1.Container, cmd string) error {
+	if container.Lifecycle == nil {
+		container.Lifecycle = &corev1.Lifecycle{}
+	}
+	if container.Lifecycle.PostStart == nil {
+		container.Lifecycle.PostStart = &corev1.Handler{
+			Exec: &corev1.ExecAction{
+				Command: []string{
```
sleshchenko, line 104: According to https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion

> The Container's status is not set to RUNNING until the postStart handler completes.

which means, it can't serve the traffic because service just won't pick such pod up.

Install vsx should be relatively quick operation but depends on the network.
Also, I see other postStart operations which developer may want to start on workspace start to speed up processes - like make yarn, mvn, go mod download dependencies. And the editor should be able to listen to output or result of such operations.

So, I'm not sure who should manage postStart commands, editor or DevWorkspace controller.
If DevWorkspace controller, it makes sense to do it out from K8s API and add API to follow command status/output.

amisevsk: Makes sense.

PR #270, opened 2021-02-17
Update devfile/api dependency and message with info about failures

amisevsk:
### What does this PR do?
Update the devfile/api dependency to latest [283b0c54946e9fea9872c25e1e086c303688f0e8](https://github.com/devfile/api/commit/283b0c54946e9fea9872c25e1e086c303688f0e8), including a fix for the new package path (see: https://github.com/devfile/api/pull/280)

Add support for propagating status info to devworkspace status, reusing the `message` field, now that `ideUrl` has been removed from printColumns:
```
❯ kc get dw
NAME           WORKSPACE ID                PHASE     INFO
theia          workspace5eacae28f63149c4   Running   http://workspace5eacae28f63149c4-theia-3100.192.168.49.2.nip.io
web-terminal   workspace548b4c42f1394436   Failed    Failed to install network objects required for devworkspace
```

Closes https://github.com/devfile/devworkspace-operator/issues/232

### Is it tested? How?
Test usual flows, functionality should not be changed except for messages

cc: @metlos I suspect you will also have to update your code to adapt to https://github.com/devfile/api/pull/280


pkg/library/container/mountSources.go
```diff
@@ -43,7 +43,7 @@ func handleMountSources(k8sContainer *corev1.Container, devfileContainer *devwor
 	var sourceMapping string
 	if vm := getProjectsVolumeMount(k8sContainer); vm != nil {
 		// Container already mounts projects volume; need to set env vars according to mountPath
-		// TODO: see issue https://github.com/devfile/api/issues/290
+		// TODO: see issue https://github.com/devfile/api/v2/issues/290
```
sleshchenko, line 46: I suspect you used find/replace :-D

controllers/workspace/status.go
```diff
@@ -138,3 +140,20 @@ func getIdeUrl(exposedEndpoints map[string]v1alpha1.ExposedEndpointList) string
...
+		case devworkspace.WorkspaceError:
+			return conditionMessage
+		case devworkspace.WorkspaceFailedStart:
+			failedStartMsg = conditionMessage
+		}
+	}
+	if failedStartMsg != "" {
+		return failedStartMsg
+	}
+	return workspace.Status.IdeUrl
```
sleshchenko, line 158: Should we return IdeURL when it's not ready yet?

sleshchenko (approved):
The PR LGTM and improves a lot UX.

I have a proposal for another improvement, maybe we should report the progressing condition into INFO, like the following:

```go
//conditionOrders contains ordered conditions in way the should become ready
// first not ready should be propagated to info
var conditionOrders = map[devworkspace.WorkspaceConditionType]string {
  devworkspace.WorkspaceComponentsReady: "Waiting for components",
  devworkspace.WorkspaceServiceAccountReady: "Waiting for service account",
  devworkspace.WorkspaceRoutingReady: "Waiting for routing",
  //devworkspace.WorkspaceDeploymentReady
  //devworkspace.WorkspaceIDEReady
}

func getInfoMessage(workspace *devworkspace.DevWorkspace, conditions map[devworkspace.WorkspaceConditionType]string) string {
	var failedStartMsg string
	for conditionType, conditionMessage := range conditions {
		switch conditionType {
		// Take error condition message as overriding failed start message
		case devworkspace.WorkspaceError:
			return conditionMessage
		case devworkspace.WorkspaceFailedStart:
			failedStartMsg = conditionMessage
		case devworkspace.WorkspaceReady:
			return workspace.Status.IdeUrl
		}
	}
	if failedStartMsg != "" {
		return failedStartMsg
	}

	for conditionType, progress := range conditionOrders {
		if _, present := conditions[conditionType]; !present {
			return progress
		}
	}

	return workspace.Status.IdeUrl
}
```

^ it's not tested

pkg/library/storage/commonStorage.go
```diff
@@ -18,8 +18,8 @@
 // - Devfile API spec is unclear on how mountSources should be handled -- mountPath is assumed to be /projects
 //   and volume name is assumed to be "projects"
 //   see issues:
-//     - https://github.com/devfile/api/issues/290
-//     - https://github.com/devfile/api/issues/291
+//     - https://github.com/devfile/api/v2/issues/290
```
JPinkney, line 21: I think this might be another find and replace error :wink:

JPinkney (approved):

pkg/library/container/mountSources.go
amisevsk: :smile: I'll admit to a non-judicious use of `sed` here.

controllers/workspace/status.go
amisevsk: That's what we currently do -- personally I like it because I can copy it and refresh a few times to load a workspace, rather than repeatedly using `kubectl get`, but I'm happy to do it either way (and with your suggestion above of reporting latest condition, it makes sense to not put ideURL until it's started anyways)

sleshchenko: Probably it's another issue, but I saw we return outdated URL, after routing class is changed but workspace is stopped.

and with proposal to report the latest condition, I just wondered how we can avoid confusion we got  when workspace is started but not started, because Che Routing is not available yet.

There we may also report finalizing a workspace, currently the way is to check removal timestamp )

amisevsk:
Updated the status message to show what the startup process is waiting on, as suggested by Serhii:

status-animation

amisevsk:
Added support for `Stopping`/`Stopped` and also the message `Cleaning up resources for deletion` when a workspace is deleted and we're finalizing.

Issue #271, opened 2021-02-17
Default routing class not propagated into the workspacerouting object


metlos:
If the devworkspace operator is configured with a default routing class, this routing class is not persisted in the workspace routing object. This then makes it impossible for the external solvers to reconcile such routings.

PR #272, opened 2021-02-17
Explicitly set the routing class in the workspace routings so that external solvers can be used if specified as the default routing class

metlos:
### What does this PR do?
This makes sure that the workspace routing always contains an explicit routing class. This in turn makes sure that the default routing class configured in the devworkspace-operator's configuration can be handled by an external solver.

Without this, the external solver cannot know the routing class unless it reads the configuration of the main devworkspace-operator, which seems quite hacky.

#271

### DIFF
--- a/controllers/controller/workspacerouting/solvers/solver.go
+++ b/controllers/controller/workspacerouting/solvers/solver.go
@@ -98,9 +98,6 @@ func (_ *SolverGetter) HasSolver(routingClass controllerv1alpha1.WorkspaceRoutin
 }
 
 func (_ *SolverGetter) GetSolver(client client.Client, routingClass controllerv1alpha1.WorkspaceRoutingClass) (RoutingSolver, error) {
-	if routingClass == "" {
-		routingClass = controllerv1alpha1.WorkspaceRoutingClass(config.ControllerCfg.GetDefaultRoutingClass())
-	}
 	switch routingClass {
 	case controllerv1alpha1.WorkspaceRoutingBasic:
 		return &BasicSolver{}, nil
--- a/controllers/controller/workspacerouting/workspacerouting_controller.go
+++ b/controllers/controller/workspacerouting/workspacerouting_controller.go
@@ -80,6 +80,11 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 	reqLogger = reqLogger.WithValues(config.WorkspaceIDLoggerKey, instance.Spec.WorkspaceId)
 	reqLogger.Info("Reconciling WorkspaceRouting")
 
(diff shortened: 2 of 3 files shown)


sparkoo (approved):

controllers/controller/workspacerouting/solvers/solver.go
```diff
@@ -98,9 +98,6 @@ func (_ *SolverGetter) HasSolver(routingClass controllerv1alpha1.WorkspaceRoutin
 }
 
 func (_ *SolverGetter) GetSolver(client client.Client, routingClass controllerv1alpha1.WorkspaceRoutingClass) (RoutingSolver, error) {
-	if routingClass == "" {
-		routingClass = controllerv1alpha1.WorkspaceRoutingClass(config.ControllerCfg.GetDefaultRoutingClass())
```
sleshchenko, line 102: If we don't support it, then used config `GetDefaultRoutingClass()` and related should be removed.

Or, instead defaulting can be moved to other level, let's say DevWorkpace controller, can update CR with default routing if it's missing.

I'm not sure it's really needed feature, but during development it was handy to switch between basic for k8s, and openshift-oauth for OpenShift, and all examples worked without change.

@amisevsk WDYT?

amisevsk: I'm fine removing empty routingClass (or moving the defaulting into the main controller) but I also don't see the need for the change necessarily; I don't think DWO should be configured to use a default routingClass that it does not support, as that could lead to confusion (and also make integrating multiple dependants of DWO tricky).

I think this PR does what you suggest though, in still supporting devworkspaces without a routingClass, by setting the routingClass in the routing provision function.

amisevsk (approved):
I'm +½ on these changes.

On one hand, I can see that it makes sense to split the configuration as proposed and make routingClass mandatory in WorkspaceRoutings. 

On the other hand, the default routingClass should not be something the DevWorkspace operator can't handle internally, IMO. If we want a DevWorkspace to show special functionality, we should have to explicitly specify it.

Either way, I don't feel too strongly one way or another, so if it makes life easier for alternate implementations (even as a development feature), I'm in approval. The utility of this feature is mostly for development, anyways.

I do hope once Che starts making DevWorkspaces, that it will specify `routingClass: che` explicitly, though.

controllers/controller/workspacerouting/solvers/solver.go
sleshchenko: Ah, I must scroll down :-D

Actually I did but as I see not, did not read carefully enough

sleshchenko (approved):
tested and I think it improves UX a bit in sense that user can figure out the run routing that is applied by default, by checking workspace routing.
So, +1

PR #275, opened 2021-02-18
WIP Change default namespace to openshift-operators

sleshchenko:
### What does this PR do?
This PR changes default namespace to openshift-operators.
I did it since chectl start depends on our default namespace.
Because patching namespace on chectl side is tricky part, since it's needed to be updated in Certificate, CRDs annotations, CRDs spec, RoleBinding.

And I think openshift-operators is a better option, since it's always
all-cluster watching + WTO supports only it.
So, potentially chectl can rely on WTO objects(if they already are created) if it's updated to v1alpha2.
So, we have less conflicts.

It's related to https://github.com/che-incubator/chectl/pull/1100#issuecomment-780679764

### Is it tested? How?
Not yet but changes are simple

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ ifndef VERBOSE
 MAKEFLAGS += --silent
 endif
 
-export NAMESPACE ?= devworkspace-controller
+export NAMESPACE ?= openshift-operators
 export IMG ?= quay.io/devfile/devworkspace-controller:next
 export ROUTING_SUFFIX ?= 192.168.99.100.nip.io
 export PULL_POLICY ?= Always
--- a/deploy/generate-deployment.sh
+++ b/deploy/generate-deployment.sh
@@ -69,7 +69,7 @@ done
 
 if $USE_DEFAULT_ENV; then
   echo "Using defaults for environment variables"
-  export NAMESPACE=devworkspace-controller
+  export NAMESPACE=openshift-operators
   export IMG=quay.io/devfile/devworkspace-controller:next
   export PULL_POLICY=Always
   export DEFAULT_ROUTING=basic
--- a/deploy/deployment/kubernetes/combined.yaml
+++ b/deploy/deployment/kubernetes/combined.yaml
@@ -3426,7 +3426,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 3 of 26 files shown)


Makefile
```diff
@@ -17,7 +17,7 @@ ifndef VERBOSE
 MAKEFLAGS += --silent
 endif
 
-export NAMESPACE ?= devworkspace-controller
+export NAMESPACE ?= openshift-operators
```
sleshchenko, line 20: If it makes any sense we can stick to devworkspace-controller for development, but after chectl depends on it, I think it makes sense to unify default.

sparkoo:
what will happen on pure kubernetes?

Makefile
sleshchenko: marked as WIP since if openshift-operator is the default, then we must not remove that namespace on uninstall

sleshchenko:
> what will happen on pure kubernetes?

It's such a good catch!
I need to take a look if k8s has a dedicated namespace for one-namespace.

deploy/generate-deployment.sh
```diff
@@ -69,7 +69,7 @@ done
 
 if $USE_DEFAULT_ENV; then
   echo "Using defaults for environment variables"
-  export NAMESPACE=devworkspace-controller
+  export NAMESPACE=openshift-operators
```
amisevsk, line 72: IMO this is the only change that should be made; the namespace in the Makefile will be ignored when the `deployment/deploy` templates are generated anyways.

sleshchenko: I would argue on that.
I mean I would be happy if we can reuse Makefile to work with deployment run with chectl. And Makefile reuse NAMESPACE from there to restart deployments at least.

But as we discussed on standup, probably we should stick to `devworkspace-controller` namespace as only available, that would work for K8s and Openshift and easier to clean up.

amisevsk: Using the Makefile in `chectl` means having `chectl` depend on `make`, which is a strange choice. If we need a development version of DWO to run with Che, it should be deployed manually from our repo.

sleshchenko: I meant you deploy it with chectl, but then you need to update/debug DevWorkpaceOperator with makefile.
if we unify values, it would be possible to deploy/update in any case you like.

sleshchenko:
Closing since we think it's better to stick to devworkspace-operator or any other value that works for K8s and OpenShift + easy to clean up by removing the whole namespace.

PR #276, opened 2021-02-18
Add Stopped By Annotation to ignored immutable fields

JPinkney:

### What does this PR do?
This PR makes it so that the stopped-by annotation is ignored when evaluating an update in the webhook server

No issue created, I just noticed it when creating the WTO 1.2 blog

### Is it tested? How?

```
make docker deploy
# create web terminal
# web-terminal should terminate
```

### DIFF
--- a/webhook/workspace/handler/immutable.go
+++ b/webhook/workspace/handler/immutable.go
@@ -33,6 +33,9 @@ var ImmutableWorkspaceDiffOptions = []cmp.Option{
 	// field managed by cluster and should be ignored while comparing
 	cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ManagedFields", "Finalizers", "DeletionTimestamp"),
 	cmpopts.IgnoreFields(devworkspace.DevWorkspaceSpec{}, "Started"),
+	cmpopts.IgnoreMapEntries(func(k string, v string) bool {
+		return k == config.StoppedByAnnotation
+	}),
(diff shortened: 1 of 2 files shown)


amisevsk (approved):
Much simpler than I had in mind :)

webhook/workspace/handler/immutable.go
```diff
@@ -33,6 +33,9 @@ var ImmutableWorkspaceDiffOptions = []cmp.Option{
 	// field managed by cluster and should be ignored while comparing
 	cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ManagedFields", "Finalizers", "DeletionTimestamp"),
 	cmpopts.IgnoreFields(devworkspace.DevWorkspaceSpec{}, "Started"),
+	cmpopts.IgnoreMapEntries(func(k string, v string) bool {
+		return k == config.StoppedByAnnotation
```
sleshchenko, line 37: does it mean that it will be ignored in all possible maps? Labels, annotations, devfile attributes?

There is no security leak, but that's not accurate check.
I'm fully OK with proceeding with it, but we must consider moving to production-ready approach, which Angel proposed some time ago

sleshchenko: Now I checked the master branch, and it's just the same, apart from formatting )

sleshchenko (approved):

Issue #277, opened 2021-02-19
Workspace routing objects for the current workspace are not available within the workspace pod.


benoitf:
I created a DevWorkspace using devWorkspace Operator on minikube (no Eclipse Che there)

Then, from che-theia container, I tried to get workspacerouting objects and it failed

```bash
/projects/tmp $ ./kubectl get workspaceroutings.controller.devfile.io/routing-workspaceeb55021d3cff42e0 -n che

Error from server (Forbidden): workspaceroutings.controller.devfile.io "routing-workspaceeb55021d3cff42e0" is forbidden: User "system:serviceaccount:che:workspaceeb55021d3cff42e0-sa" cannot get resource "workspaceroutings" in API group "controller.devfile.io" in the namespace "che"
```

but I can access the dev workspace object:
```
/projects/tmp $ ./kubectl  get devworkspaces/theia -n che
NAME    WORKSPACE ID                PHASE     URL
theia   workspaceeb55021d3cff42e0   Running   http://workspaceeb55021d3cff42e0-theia-3100.192.168.64.31.nip.io
```

or the pod object
```
/projects/tmp $ ./kubectl  get pods -n che
NAME                                         READY   STATUS    RESTARTS   AGE
workspaceeb55021d3cff42e0-77f7bd767f-tld2s   3/3     Running   3          16d
```

Trying on the host( where minikube is launched, the command is successful)
```
$ kubectl get workspaceroutings.controller.devfile.io/routing-workspaceeb55021d3cff42e0 -n che                                                        NAME                                AGE
routing-workspaceeb55021d3cff42e0   16d
```

PR #278, opened 2021-02-19
Allow read for workspacerouting and watch/list for dw

sleshchenko:
### What does this PR do?
Allow read for workspacerouting and watch/list for dw

Fixes https://github.com/devfile/devworkspace-operator/issues/277

### Is it tested? How?
not yet
but the image with this is built as 
`export IMG=sleshchenko/che-workspace-controller:extended-rbac`

### DIFF
--- a/controllers/workspace/provision/rbac.go
+++ b/controllers/workspace/provision/rbac.go
@@ -57,7 +57,7 @@ func generateRBAC(namespace string) []runtime.Object {
 				{
 					Resources: []string{"devworkspaces"},
 					APIGroups: []string{"workspace.devfile.io"},
-					Verbs:     []string{"patch", "get", "update", "watch"},
+					Verbs:     []string{"get", "watch", "list", "patch", "update"},
 				},
 				{
 					Resources: []string{"workspaceroutings"},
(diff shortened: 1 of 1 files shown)


benoitf:
I was able to fetch workspacerouting object from a container inside workspace pod with this image

```
/projects/tmp $ ./kubectl get workspaceroutings.controller.devfile.io/routing-workspaceeb55021d3cff42e0 -n che
NAME                                AGE
routing-workspaceeb55021d3cff42e0   16d
```

while before I had
```
/projects/tmp $ ./kubectl get workspaceroutings.controller.devfile.io/routing-workspaceeb55021d3cff42e0 -n che
Error from server (Forbidden): workspaceroutings.controller.devfile.io "routing-workspaceeb55021d3cff42e0" is forbidden: User "system:serviceaccount:che:workspaceeb55021d3cff42e0-sa" cannot get resource "workspaceroutings" in API group "controller.devfile.io" in the namespace "che"
```

benoitf:
I think it would be nice to list as well the dev workspaces objects

```
/kubectl get devworkspaces.workspace.devfile.io -n che
Error from server (Forbidden): devworkspaces.workspace.devfile.io is forbidden: User "system:serviceaccount:che:workspaceeb55021d3cff42e0-sa" cannot list resource "devworkspaces" in API group "workspace.devfile.io" in the namespace "che"
```

benoitf:
it's because in the IDE we display a list of other user workspaces:
https://recordit.co/g3buGBrRuk

(but it can be discussed in another issue)

benoitf:
tested image and it's working as expected. Thanks !

amisevsk (approved):

PR #280, opened 2021-02-23
Initialize annotations if they are nil before setting the timer

JPinkney:

### What does this PR do?
Apparently, if you create a devworkspace through the rest api annotations aren't automatically initialized (they are when you're using kubectl/oc though) and that causes `Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)` when trying to set the timing information. This PR makes sure that `workspace.Annotations` is initialized before using

None

### Is it tested? How?

I'm not sure an easy way to test this without requiring setting everything up. I can probably think of one tomorrow though

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -372,6 +372,9 @@ func (r *DevWorkspaceReconciler) stopWorkspace(workspace *devworkspace.DevWorksp
 func (r *DevWorkspaceReconciler) syncTimingToCluster(
 	ctx context.Context, workspace *devworkspace.DevWorkspace, timingInfo map[string]string, reqLogger logr.Logger) {
 	if timing.IsEnabled() {
+		if workspace.Annotations == nil {
+			workspace.Annotations = map[string]string{}
+		}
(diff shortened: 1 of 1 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -372,6 +372,9 @@ func (r *DevWorkspaceReconciler) stopWorkspace(workspace *devworkspace.DevWorksp
 func (r *DevWorkspaceReconciler) syncTimingToCluster(
 	ctx context.Context, workspace *devworkspace.DevWorkspace, timingInfo map[string]string, reqLogger logr.Logger) {
 	if timing.IsEnabled() {
+		if workspace.Annotations == nil {
```
sleshchenko, line 375: Could you make check that we don't have another places where we try to write into workspace labels/annotations without nil check?

I just wonder if we won't get the same error on further steps if we disable timing.

sleshchenko (approved):
> I'm not sure an easy way to test this without requiring setting everything up. I can probably think of one tomorrow though

The described issue is pretty clear, so that specific testing is not really needed.

amisevsk (approved):

PR #281, opened 2021-02-23
Add support for ephemeral volumes

amisevsk:
### What does this PR do?
Adds support for volume components with `ephemeral: true`, which are provisioned as emptyDir volumes in the workspace pod.

This PR updates the logic for determining whether a container needs storage. The new logic is:
- If any non-ephemeral volumes are defined, we need storage
- If any container component has `mountSources: true`, we need storage *unless* there's an explicit projects volume with `ephemeral: true`
this new check does not check that any container has mountSources as before, since that is covered by the first case, or an error (a container mounting a volume that is not defined)

I had to update the relevant test to support this change, and also added tests to cover the ephemeral volume use cases.

I've also added a sample that showcases some of the ways to use ephemeral volumes:
- We can override the default projects volume to be ephemeral
- We can override volumes in plugins to be ephemeral (we make the theia `plugins` volume ephemeral)

Closes https://github.com/devfile/api/issues/310

### Is it tested? How?
1. Deploy controller, *make sure CRDs are updated* to include support for ephemeral volumes (I made that mistake...) (see: https://github.com/devfile/devworkspace-operator/pull/270)
2. Deploy plugin templates (`make install_plugin_templates`)
    - Note some of the templates have been updated to e.g. use an ephemeral volume for remote runtime
3. Deploy sample that uses emptyDir
    ```bash
    kubectl apply -f samples/with-k8s-ref/emptyDir-theia.yaml
    ```
4. Verify:
    - All volumes in workspace pod are emptyDir (`kubectl get po <workspace> -o yaml | yq -y '.spec.volumes')
    - All volumeMounts in containers refer to these emptyDir containers
    - No finalizer is set on the workspace
    - The workspace runs normally

I had a strange issue testing locally that I'm currently blaming minikube for: when I initially apply the sample, the volumes in the devworkspace are simply
```
      - name: projects
        volume: {}
```
but applying twice correctly shows the ephemeral field. No idea why that's happening and praying it doesn't reproduce elsewhere.

### DIFF
--- a/pkg/library/storage/commonStorage.go
+++ b/pkg/library/storage/commonStorage.go
@@ -25,6 +25,8 @@ package storage
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"github.com/devfile/devworkspace-operator/pkg/library/constants"
 	containerlib "github.com/devfile/devworkspace-operator/pkg/library/container"
 	corev1 "k8s.io/api/core/v1"
@@ -39,16 +41,18 @@ import (
 //
 // Also adds appropriate k8s Volumes to PodAdditions to accomodate the rewritten VolumeMounts.
 func RewriteContainerVolumeMounts(workspaceId string, podAdditions *v1alpha1.PodAdditions, workspace devworkspace.DevWorkspaceTemplateSpec) error {
-	if !NeedsStorage(workspace) {
-		return nil
-	}
 	devfileVolumes := map[string]devworkspace.VolumeComponent{}
+	var ephemeralVolumes []devworkspace.Component
+
 	for _, component := range workspace.Components {
 		if component.Volume != nil {
 			if _, exists := devfileVolumes[component.Name]; exists {
 				return fmt.Errorf("volume component '%s' is defined multiple times", component.Name)
 			}
 			devfileVolumes[component.Name] = *component.Volume
+			if component.Volume.Ephemeral {
+				ephemeralVolumes = append(ephemeralVolumes, component)
+			}
 		}
 	}
 	if _, exists := devfileVolumes[constants.ProjectsVolumeName]; !exists {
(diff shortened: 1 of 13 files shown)


amisevsk:
@sleshchenko I might be seeing your issue from https://github.com/devfile/devworkspace-operator/pull/269#pullrequestreview-592960021 -- using `che-theia:next` I'm stuck at the loading spinner, but switching to `7.26.0` I have no issues.

SkorikSergey:
Selection_140

pkg/library/storage/commonStorage.go
```diff
@@ -80,34 +87,56 @@ func RewriteContainerVolumeMounts(workspaceId string, podAdditions *v1alpha1.Pod
...
+			if err != nil {
+				return fmt.Errorf("failed to parse size for Volume %s: %w", component.Name, err)
+			}
+			vol.EmptyDir.SizeLimit = &sizeResource
+		}
+		podAdditions.Volumes = append(podAdditions.Volumes, vol)
+	}
+
+	if NeedsStorage(workspace) {
+		podAdditions.Volumes = append(podAdditions.Volumes, corev1.Volume{
```
sleshchenko, line 108: Do we need to do L67-88 if workspace does not need a storage? Maybe it makes sense to extend this condition body?

pkg/library/storage/testdata/error-unparseable-ephemeral-size.yaml
```diff
@@ -0,0 +1,26 @@
...
+          image: testing-image-1
+          sourceMapping: "/plugins-mountpath"
+          mountSources: true
+      - name: projects
+        volume:
+          ephemeral: true
+          size: 512XX
+
+output:
+  errRegexp: "failed to parse size for Volume projects.*"
```
sleshchenko, line 26:
```suggestion
  errRegexp: "failed to parse size for Volume projects.*"

```

sleshchenko (approved):
Tested and the described scenario works as expected, apart from fact that I needed to patch theia to be 7.26.0 which is know issue.

pkg/library/storage/commonStorage.go
amisevsk: Technically no, but we do also check for invalid volume references there (`volume, ok := devfileVolumes[vm.Name]`). However, it probably makes sense to separate out that check since at some point we'll just use some validation library from devfile/api to not have to worry about it.

pkg/library/storage/testdata/error-unparseable-ephemeral-size.yaml
amisevsk: Thought I caught all of these :)

pkg/library/storage/commonStorage.go
amisevsk: Reworked this section

JPinkney (approved):
Tested and working as expected :+1:

amisevsk:
From Serhii:

DevWorkspace with 
```yaml
spec:
  started: true
  template:
    components:
    - name: theia-next
      plugin:
        kubernetes:
          name: theia-next
    - name: machine-exec
      plugin:
        kubernetes:
          name: machine-exec
```
has PVC mounted but no finalizer set; looking into it.

sleshchenko (approved):
Still LGTM, additional thanks for handling the reported bug.

Issue #264, opened 2021-02-05
Implement the mechanism to configure dockercfg secret as imagePullSecrets for devworkspace


sleshchenko:
In the scope of https://github.com/eclipse/che/issues/18990 Dashboard is going to create dockercfg in the DevWorkspace namespace.

It's needed to implement the mechanism to use such secret as imagePullSecret.

Possible alternatives:
- secret is labeled with something like `controller.devfile.io/secret-kind: imagePullSecret`
- devworkspace operator lists and mounts all available dockercfg secrets to the devworkspace pods.
- something else

sleshchenko:
PR is in place https://github.com/devfile/devworkspace-operator/pull/284

PR #284, opened 2021-02-25
Attach dockercfg secrets as pull image secret

sleshchenko:
### What does this PR do?
This PR add an ability to label secrets with types `kubernetes.io/dockercfg` and `kubernetes.io/dockerconfigjson` which `controller.devfile.io/workspace_pull_secret: true` and it will be used as pull secret for DevWorkspace-related deployment.

fixes https://github.com/devfile/devworkspace-operator/issues/264

### Is it tested? How?
I created two robots and mirror che theia in my private repo, so, they can be used for testing:

```bash
# 1. Created patched sample with private image
sed -e 's|image: "quay.io/eclipse/che-theia:next"|image: "quay.io/sleshche/che-theia:latest"|g' ./samples/flattened_theia-next.yaml \
  | kubectl apply -f -

# 2. it won't be started due
#         message: 'rpc error: code = Unknown desc = Error response from daemon: unauthorized:
#         access to the requested resource is not authorized'


# 3. Create tests robots
# Robot without any permissions, just to check that it's ignored since there is no needed label
apiVersion: v1
data:
  .dockerconfigjson: ewoJImF1dGhzIjogewoJCSJxdWF5LmlvIjogewoJCQkiYXV0aCI6ICJjMnhsYzJoamFHVXJhVjlqWVc1ZmJtOTBhR2x1WnpwUE4xcFVSRVExVUZvMFFWVlNWRGxEVlU5VVNEUlZSVXBhUWxBMVRVUXdOVGd5UVRGR1RFWk5NbGxKVkZGQ1R6QTNSMXBEUmtrelZUYzBVRmxTTWpNMCIKCQl9Cgl9Cn0=
kind: Secret
metadata:
  name: blank-robot
  labels: {}
type: kubernetes.io/dockerconfigjson
EOF

# Robot with read access to private quay.io/sleshche/che-theia
apiVersion: v1
data:
  .dockerconfigjson: ewoJImF1dGhzIjogewoJCSJxdWF5LmlvIjogewoJCQkiYXV0aCI6ICJjMnhsYzJoamFHVXJZMmhsWDNSb1pXbGhYM0psWVdSdmJteDVPa1ZOV1ZJeE5sRlRORXBVVVUweFdUZFROVkZHUVZaWlJFazNOME5KVWtaSFVsaGFRMFJVVms0MlUwVXpXRWRSVTFZNVZqTktTVEJXUVZoT09GTTBTVkk9IgoJCX0KCX0KfQ==
kind: Secret
metadata:
  name: che-theia-readonly
  labels:
    controller.devfile.io/devworkspace_pullsecret: "true"
type: kubernetes.io/dockerconfigjson
EOF

# 4. Restart a workspace 

kubectl patch dw theia --type merge --patch '                                     
spec:
  started: false
'
kubectl patch dw theia --type merge --patch '                                     
spec:
  started: true
'

# 5. Workspace now is started.
```

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -304,7 +304,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	if !pullSecretStatus.Continue {
 		return reconcile.Result{Requeue: pullSecretStatus.Requeue}, pullSecretStatus.Err
 	}
-	podAdditions = append(podAdditions, pullSecretStatus.PodAdditions)
+	allPodAdditions = append(allPodAdditions, pullSecretStatus.PodAdditions)
 	reconcileStatus.Conditions[PullSecretsReadyCondition] = ""
 
 	// Step six: Create deployment and wait for it to be ready
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -305,6 +305,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: pullSecretStatus.Requeue}, pullSecretStatus.Err
 	}
 	podAdditions = append(podAdditions, pullSecretStatus.PodAdditions)
+	reconcileStatus.Conditions[PullSecretsReadyCondition] = ""
 
 	// Step six: Create deployment and wait for it to be ready
 	timing.SetTime(timingInfo, timing.DeploymentCreated)
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -300,6 +300,12 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	serviceAcctName := serviceAcctStatus.ServiceAccountName
 	reconcileStatus.Conditions[devworkspace.WorkspaceServiceAccountReady] = ""
 
+	pullSecretStatus := provision.PullSecrets(clusterAPI)
+	if !pullSecretStatus.Continue {
+		return reconcile.Result{Requeue: pullSecretStatus.Requeue}, pullSecretStatus.Err
+	}
+	podAdditions = append(podAdditions, pullSecretStatus.PodAdditions)
+
 	// Step six: Create deployment and wait for it to be ready
 	timing.SetTime(timingInfo, timing.DeploymentCreated)
 	deploymentStatus := provision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI)
--- a/controllers/workspace/provision/pull_secret.go
+++ b/controllers/workspace/provision/pull_secret.go
@@ -1,0 +1,67 @@
+//
+// Copyright (c) 2019-2021 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package provision
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"github.com/devfile/devworkspace-operator/pkg/config"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type PullSecretsProvisioningStatus struct {
+	ProvisioningStatus
+	v1alpha1.PodAdditions
+}
+
+func PullSecrets(clusterAPI ClusterAPI) PullSecretsProvisioningStatus {
+	labelSelector, err := labels.Parse(fmt.Sprintf("%s=%s", config.DevWorkspacePullSecretLabel, "true"))
+	if err != nil {
+		return PullSecretsProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{
+				Err:     err,
+				Requeue: true,
+			},
+		}
+	}
+
+	secrets := corev1.SecretList{}
+	err = clusterAPI.Client.List(context.TODO(), &secrets, &client.ListOptions{LabelSelector: labelSelector})
+	if err != nil {
+		return PullSecretsProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{
+				Err:     err,
+				Requeue: true,
+			},
+		}
+	}
+
+	var dockerCfgs []corev1.LocalObjectReference
+	for _, s := range secrets.Items {
+		if s.Type == corev1.SecretTypeDockercfg || s.Type == corev1.SecretTypeDockerConfigJson {
+			dockerCfgs = append(dockerCfgs, corev1.LocalObjectReference{Name: s.Name})
+		}
+	}
+	return PullSecretsProvisioningStatus{
+		ProvisioningStatus: ProvisioningStatus{
+			Continue: true,
+		},
+		PodAdditions: v1alpha1.PodAdditions{
+			PullSecrets: dockerCfgs,
+		},
+	}
+}
--- a/controllers/workspace/provision/pull_secret.go
+++ b/controllers/workspace/provision/pull_secret.go
@@ -33,8 +33,8 @@ func PullSecrets(clusterAPI ClusterAPI) PullSecretsProvisioningStatus {
 	if err != nil {
 		return PullSecretsProvisioningStatus{
 			ProvisioningStatus: ProvisioningStatus{
-				Err:     err,
-				Requeue: true,
+				Err:         err,
+				FailStartup: true,
 			},
 		}
 	}
@@ -44,8 +44,7 @@ func PullSecrets(clusterAPI ClusterAPI) PullSecretsProvisioningStatus {
 	if err != nil {
 		return PullSecretsProvisioningStatus{
 			ProvisioningStatus: ProvisioningStatus{
-				Err:     err,
-				Requeue: true,
+				Err: err,
 			},
 		}
 	}
--- a/controllers/workspace/status.go
+++ b/controllers/workspace/status.go
@@ -32,6 +32,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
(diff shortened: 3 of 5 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -307,6 +307,12 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	serviceAcctName := serviceAcctStatus.ServiceAccountName
 	reconcileStatus.Conditions[devworkspace.WorkspaceServiceAccountReady] = ""
 
+	pullSecretStatus := provision.PullSecrets(clusterAPI)
+	if !pullSecretStatus.Continue{
+		return reconcile.Result{Requeue: pullSecretStatus.Requeue}, pullSecretStatus.Err
```
sleshchenko, line 312: Note that it will be exposed in Workspace status as WAITING FOR DEPLOYMENT, not sure if we should update devfile/api or introduce our own conditions instead

samples/flattened_theia-next.yaml
```diff
@@ -37,60 +37,11 @@ spec:
...
-              value: remote-endpoint
       - name: theia-ide
         attributes:
           "app.kubernetes.io/name": che-theia.eclipse.org
           "app.kubernetes.io/part-of": che.eclipse.org
           "app.kubernetes.io/component": editor
-
-          # Added by Che-theia at start when detecting, after cloning, that the extensions.json in the repo
-          # contains the vscode-pull-request-github vscode plugin.
-          "che-theia.eclipse.org/vscode-extensions":
```
sleshchenko, line 92: plugins are removed from simple theia-next sample, since it fails for me on minikube even from https://github.com/devfile/devworkspace-operator/pull/286 with error:
```
downloading "https://github.com/microsoft/vscode-pull-request-github/releases/download/v0.8.0/vscode-pull-request-github-0.8.0.vsix" to /plugins
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   645  100   645    0     0   4640      0 --:--:-- --:--:-- --:--:--  4640
100  468k  100  468k    0     0   960k      0 --:--:-- --:--:-- --:--:-- 5002k
mkdir: can't create directory 'for': Permission denied

downloading for to for
/bin/sh: can't create for/for: nonexistent directory
mkdir: can't create directory 'deployment': Permission denied
```
So, I just don't won't to invest time in that vsinstaller, but it may be a sing that something is wrong with PVC on DevWorkspace Operator level.

controllers/workspace/devworkspace_controller.go
sleshchenko: I've added https://github.com/devfile/devworkspace-operator/pull/284/commits/a8cfd710ce0070eb3689da84bec4df210c91f3f6, but note that it will be exposed on workspace status only in case pull secrets failed from first try.

samples/flattened_theia-next.yaml
amisevsk: :thinking: I'm 90% sure the issue should be fixed (the problem was that we do a for loop over a big string and by default bash splits on `' '`. But I agree, we can remove vsx installer.

sleshchenko: It's fixed now. But I still removing vsx installer from Theia Next to keep to simple.

controllers/workspace/provision/pull_secret.go
```diff
@@ -0,0 +1,67 @@
...
+	v1alpha1.PodAdditions
+}
+
+func PullSecrets(clusterAPI ClusterAPI) PullSecretsProvisioningStatus {
+	labelSelector, err := labels.Parse(fmt.Sprintf("%s=%s", config.DevWorkspacePullSecretLabel, "true"))
+	if err != nil {
+		return PullSecretsProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{
+				Err:     err,
+				Requeue: true,
```
amisevsk, line 37: If we're returning `err != nil`, `Requeue: true` is not necessary, as returning an error always triggers a requeue.
```suggestion
```

controllers/workspace/status.go
```diff
@@ -165,6 +167,9 @@ func getInfoMessage(workspace *devworkspace.DevWorkspace, conditions map[devwork
 	if _, ok := conditions[devworkspace.WorkspaceServiceAccountReady]; ok {
 		return "Waiting for deployment to be ready"
 	}
+	if _, ok := conditions[PullSecretsReadyCondition]; ok {
+		return "Waiting for workspace pull secrets"
+	}
 	if _, ok := conditions[devworkspace.WorkspaceRoutingReady]; ok {
 		return "Waiting for workspace serviceaccount"
 	}
```
amisevsk, line 175: Are we sure this is the right logic? I would assume that if the `PullSecretsReady` condition is true, then the pull secrets have been prepared and we're not actually waiting on them. 

The general approach to this progress message I've taken is once a condition is set, we're waiting on the next condition, i.e.
```suggestion
	if _, ok := conditions[devworkspace.WorkspaceServiceAccountReady]; ok {
		return "Waiting for deployment to be ready"
	}
	if _, ok := conditions[PullSecretsReadyCondition]; ok {
		return "Waiting for workspace service account"
	}
	if _, ok := conditions[devworkspace.WorkspaceRoutingReady]; ok {
		return "Waiting for workspace pull secrets"
	}
```
Each condition gets set in sequence, so the idea is once a condition is set, the workspace is waiting on the _next_ condition in the list -- If `conditions[devworkspace.WorkspaceServiceAccountReady]` is set, then we know `conditions[PullSecretsReadyCondition]` is set, etc.

pkg/config/constants.go
```diff
@@ -46,6 +46,11 @@ const (
 	// WorkspaceNameLabel is label key to store workspace name
 	WorkspaceNameLabel = "controller.devfile.io/workspace_name"
 
+	// PullSecretLabel marks the intention that secret should be used as pull secret for devworkspaces withing namespace
+	// Only secrets with 'true' value will be mount as pull secret
+	// Should be assigned to secrets with type docker config types (kubernetes.io/dockercfg and kubernetes.io/dockerconfigjson)
+	DevWorkspacePullSecretLabel = "controller.devfile.io/workspace_pull_secret"
```
amisevsk, line 52: Here we go with naming again :)

Should this be `devworkspace_pull_secret`?

amisevsk (approved):
LGTM with a few comments

controllers/workspace/provision/pull_secret.go
amisevsk: Note also that this condition (which we should never hit) is a failure state, so it should return `FailStartup: true`

pkg/config/constants.go
sleshchenko: yes and no. 

If yes, we must change `controller.devfile.io/workspace_name` and `controller.devfile.io/workspace_id` which probably makes sense.

I'll do it

controllers/workspace/status.go
sleshchenko: Good catch.
it's downside of multitasking :-D

I'll fix it

pkg/config/constants.go
sleshchenko: changed to devworkspace_pull_secret. workspace_id and name will be changed in another PR.

maybe even devworkspace_pullsecret would be even better.

JPinkney (approved):
Other than Angel's comments it looks good to me

controllers/workspace/provision/pull_secret.go
sleshchenko: Should be fixed now, on parsing labels we fails startup, on error during list secrets, return error without requeue: true

pkg/config/constants.go
sleshchenko: we agreed and now it's changed to `controller.devfile.io/deworkspace_pullsecret`

controllers/workspace/status.go
sleshchenko: I refactored the way how we expose info message https://github.com/devfile/devworkspace-operator/pull/284/commits/d208482adb745eb279aee3fb99747c70519d3672

And it seems to work fine but I will need to take a look why 
waiting for editor did not appeared, maybe it just passed to fast

```
> kc get dw theia -w
NAME    WORKSPACE ID                PHASE   INFO
theia   workspace60395b9042b94129           
theia   workspace60395b9042b94129           
theia   workspace60395b9042b94129           
theia   workspace60395b9042b94129   Starting   Processing DevWorkspace components
theia   workspace60395b9042b94129   Starting   Processing DevWorkspace components
theia   workspace60395b9042b94129   Starting   Waiting for workspace routing objects
theia   workspace60395b9042b94129   Starting   Waiting for workspace routing objects
theia   workspace60395b9042b94129   Starting   Waiting for workspace routing objects
theia   workspace60395b9042b94129   Starting   Waiting for workspace serviceaccount
theia   workspace60395b9042b94129   Starting   Waiting for workspace serviceaccount
theia   workspace60395b9042b94129   Starting   Waiting for workspace serviceaccount
theia   workspace60395b9042b94129   Starting   Waiting for deployment to be ready
theia   workspace60395b9042b94129   Starting   Waiting for deployment to be ready
theia   workspace60395b9042b94129   Starting   Waiting for deployment to be ready
theia   workspace60395b9042b94129   Running    https://workspace60395b9042b94129.apps.cluster-4237.4237.sandbox422.opentlc.com/theia/
theia   workspace60395b9042b94129   Running    https://workspace60395b9042b94129.apps.cluster-4237.4237.sandbox422.opentlc.com/theia/
theia   workspace60395b9042b94129   Running    https://workspace60395b9042b94129.apps.cluster-4237.4237.sandbox422.opentlc.com/theia/
theia   workspace60395b9042b94129   Running    https://workspace60395b9042b94129.apps.cluster-4237.4237.sandbox422.opentlc.com/theia/
```

amisevsk: Sometimes it starts quickly :)

controllers/workspace/status.go
```diff
@@ -143,6 +145,36 @@ func getIdeUrl(exposedEndpoints map[string]v1alpha1.ExposedEndpointList) string
...
+	},
+	{
+		PullSecretsReadyCondition,
+		"Waiting for workspace pull secrets",
+	},
+	{
+		devworkspace.WorkspaceReady,
+		"Waiting for deployment to be ready",
+	},
+}
```
amisevsk, line 176: :+1: good idea!

We should consider (in a separate PR) using `WorkspaceReady` to be the final condition (i.e. after health checks) and add our own `DeploymentReady` condition to use in its place.

amisevsk (approved):

JPinkney (approved):
New changes LGTM

amisevsk:
I also checked that workspaces enter a terminating phase correctly -- we don't have a `phase` for terminating/deleting, so that is unchanged but the devworkspace shows `Cleaning up resources for deletion` in the `message`.

Issue #283, opened 2021-02-25
Set up release for DevWorkspace Operator


sleshchenko:
Since Che 7.27 is going to have initial integration with DWO https://github.com/eclipse/che/milestone/136

It's needed to set up release, which will:
- create v0.$VERSION.x branch, like v0.1.x
- build and push released version image
- regenerate default templates with the released image, https://github.com/devfile/devworkspace-operator/blob/main/deploy/generate-deployment.sh#L73
- set version in VERSION or somewhere else
- commit made changes
- do git tag (deployment files should be used from there)

- bump version in main branch  to v0.${VERSION+1}.0, like v0.2.0

nickboldt:
Please provide more details:
* how to build nightly
* how to build release
* image name & repo in quay to which to publish image (is this already set up or does it need to be done?)
* bump version in main branch in what file(s)? a VERSION file or something else?
* multiarch considerations - do we need to worry about z, p, aarch, arm at this stage, or punt that down the line to phase 4?
* is the version pattern for image tags, or for CRs? Do you expect that there'll be a new release every 3 weeks, and the CR version will also be incremented for each release?

amisevsk:
>Please provide more details:
>
>* how to build nightly, how to build release
>    -  given there are [2 dockerfiles which both produce an image based on a ubi8 container](https://github.com/devfile/devworkspace-operator/tree/main/build), why do we need two different builder images? Could we remove the alpine-builder-based Dockerfile and rename rhel.Dockerfile to Dockerfile for simplicity?
>

Looking at the dockerfiles, yeah we can probably just use the rhel one for everything

>* image name & repo in quay to which to publish image (is this already set up or does it need to be done?)
>    - something new under https://quay.io/organization/devfile ?
>

We've currently got [quay.io/devfile/devworkspace-controller](https://quay.io/repository/devfile/devworkspace-controller) receiving nightlies, though the naming might be a bit strange (the repo was created before we renamed to `devworkspace-operator`)

>* bump version in main branch in what file(s)? a VERSION file or something else?
>    - https://github.com/devfile/devworkspace-operator/blob/main/version/version.go ?
>    - should that version be aligned to the CR version, or are they completely different?
>

No sure what to do here; afaik we don't have a VERSION file, as go versions are defined by git tags. We should decide a branching strategy (probably what WTO uses)

[...]

PR #292, opened 2021-03-03
Add release workflow

mkuznyetsov:
### What does this PR do?
Add make-release.sh script, that can perform releases of devworkspace-operator. Also add GitHub Actions workflow, that uses this script 

https://github.com/devfile/devworkspace-operator/issues/283

### Is it tested? How?
no automated tests, only running locally make-release.sh

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: Release Devworkspace Operator
+name: Release DevWorkspace Operator
 
 on:
   workflow_dispatch:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,42 @@
+name: Release Devworkspace Operator
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'release version in format 0.y.z'
+        required: true
+
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone source code
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Login to quay.io
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+          registry: quay.io
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Setup Kustomize & yq
+        run: |
+          python -m pip install --upgrade pip
+          pip install yq
+
+          cd /usr/bin
+          curl -s "https://github.com/kubernetes-sigs/kustomize/blob/kustomize/v4.0.4/hack/install_kustomize.sh 4.0.4"  | bash
+          cd ${GITHUB_WORKSPACE}
+      - name: Build image, generate deployments and push git tag
+        run: |
+          git config --global user.name "Mykhailo Kuznietsov"
+          git config --global user.email "mkuznets@redhat.com"
+
+          export GITHUB_TOKEN=${{ secrets.CHE_INCUBATOR_BOT_GITHUB_TOKEN }}
+          ./make-release.sh --version ${{ github.event.inputs.version}}
--- a/make-release.sh
+++ b/make-release.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+# Copyright (c) 2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+set -e
+#!/bin/bash	
+QUAY_REPO="quay.io/devfile/devworkspace-controller:${VERSION}"	
+MAIN_BRANCH="main"
+
+while [[ "$#" -gt 0 ]]; do	
+  case $1 in	
+    '-v'|'--version') VERSION="$2"; shift 1;;		
+  esac	
+  shift 1	
+done	
+
+bump_version () {	
+  CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)	
+
+  NEXT_VERSION=$1	
+  BUMP_BRANCH=$2	
+
+  git checkout "${BUMP_BRANCH}"	
+
+  echo "Updating project version to ${NEXT_VERSION}"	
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+          cd ${GITHUB_WORKSPACE}
+      - name: Build image, generate deployments and push git tag
+        run: |
+          git config --global user.name "Mykhailo Kuznietsov"
+          git config --global user.email "mkuznets@redhat.com"
+          
+          export GITHUB_TOKEN=${{ secrets.CHE_INCUBATOR_BOT_GITHUB_TOKEN }}
+          ./make-release.sh --version ${{ github.event.inputs.version}}
+
```
sleshchenko, line 43:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'release version in format 0.y.z'
+        required: true
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
```
benoitf, line 12: should be a fixed version (20.04 or 18.04)
```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: Clone source code
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Login to Quay.io
+        uses: azure/docker-login@v1
```
benoitf, line 19: better to use the more 'official' `docker/login-action@v1`

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Setup Kustomize & yq
+        run: |
+          python -m pip install --upgrade pip
+          pip install yq
+
+          cd /usr/bin
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
```
benoitf, line 34: should stick to a fixed version of the script (not master)

make-release.sh
```diff
@@ -0,0 +1,146 @@
...
+
+# change VERSION file	
+echo "${VERSION}" > VERSION
+
+QUAY_REPO="quay.io/devfile/devworkspace-controller:${VERSION}"
+docker build -t "${QUAY_REPO}" -f ./build/Dockerfile .
+docker push "${QUAY_REPO}"
+
+# replace image version in default environment
+sed -i "s/IMG=quay.io\/devfile\/devworkspace-controller:.*/IMG=quay.io\/devfile\/devworkspace-controller:${VERSION}/" ./deploy/generate-deployment.sh
```
sleshchenko, line 116: I think `./deploy/generate-deployment.sh --use-defaults` should be adapted to make the work without sed

sleshchenko (approved):
LGTM

Some minor stuff can be improved(like avoid sed in generate-deployment.sh) but it's not critical, I'm OK with merging if it works and then improve that.

make-release.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/bash
```
benoitf, line 1: the script is not passing shellcheck

sleshchenko:
please rebase against main branck to fix validate PR check

.github/workflows/release.yml
mkuznyetsov: ok, will lock it down to the latest 4.0.4 version

make-release.sh
mkuznyetsov: sure, later we could probably make it so that default env variables could be overriden by user, if it wouldn't break anything

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
+name: Release Devworkspace Operator
```
benoitf, line 1:
```suggestion
name: Release DevWorkspace Operator
```

make-release.sh
```diff
@@ -0,0 +1,146 @@
...
+
+# change VERSION file + commit change into ${BASEBRANCH} branch	
+if [[ "${BASEBRANCH}" != "${BRANCH}" ]]; then	
+  # bump the y digit, if it is a major release	
+  [[ $BRANCH =~ ^([0-9]+)\.([0-9]+)\.x ]] && BASE=${BASH_REMATCH[1]}; NEXT=${BASH_REMATCH[2]}; (( NEXT=NEXT+1 )) # for BRANCH=7.10.x, get BASE=7, NEXT=11	
+  NEXT_VERSION_Y="${BASE}.${NEXT}.0-SNAPSHOT"	
+  bump_version "${NEXT_VERSION_Y}" "${BASEBRANCH}"	
+fi	
+# bump the z digit	
+[[ $VERSION =~ ^([0-9]+)\.([0-9]+)\.([0-9]+) ]] && BASE="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"; NEXT="${BASH_REMATCH[3]}"; (( NEXT=NEXT+1 )) # for VERSION=7.7.1, get BASE=7.7, NEXT=2
```
benoitf, line 137: just minor: maybe better to have example of version with VERSION scheme of devworkspace and not che
like 0.1.1 / 0.1

make-release.sh
```diff
@@ -0,0 +1,146 @@
...
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+set -e
+#!/bin/bash
```
amisevsk, line 14: Looks like a typo
```suggestion
```

make-release.sh
amisevsk: This would also commit the modified `generate-deployment.sh` file, is that something we want to do? I'm fine with this for now but agree we should maybe make `generate-deployment.sh` take a `--tag` parameter or similar.

amisevsk (approved):

make-release.sh
mkuznyetsov: ok

make-release.sh
sleshchenko: seems to be fixed now

sleshchenko (approved):
Seems to be good to proceed with

sleshchenko:
The made changes must not affect e2e tests in any way, merging without waiting for them

Issue #288, opened 2021-03-02
Get rid of Che Rest API sidecar


sleshchenko:
Che Theia already use k8s api to get everything needed so it's time to get rid of Che Rest API sidecar on DevWorkspace Operator level.

amisevsk:
I have some related WIP changes I need to push to make this easier

sleshchenko:
Both PRs are merged, closing the issue.

PR #295, opened 2021-03-03
removing che-rest-api-sidecar

tinakurian:
### What does this PR do?
Removes Che Rest API sidecar

https://github.com/devfile/devworkspace-operator/issues/288

### Is it tested? How?

This is an internal refactor. Ran unit tests


amisevsk (approved):
LGTM :+1:

Thanks for catching some weird mistakes along the way (double imports, etc.)

sleshchenko (commented):
I'm late on the party :tada: 
LGTM

Issue #161, opened 2020-09-29
workspace routings objects name


benoitf:
### Is your enhancement related to a problem? Please describe.
workspaces resources are available with devworkspaces
but workspace routings are with workspaceroutings

### Describe the solution you'd like
if instead of using Workspace we use DevWorkspace
we should have WorkspaceRoutings renamed to DevWorkspaceRoutings

amisevsk:
I don't think it matters too much what internal bookkeeping structs are called, but at this point it's hard to change names since they're used in the web terminal operator.

benoitf:
can be filtered using
```
$ kubectl get WorkspaceRoutings --all-namespaces --selector="controller.devfile.io/workspace_id=<id>"
```

PR #297, opened 2021-03-04
Rename workspacerouting to DevWorkspaceRouting (dwr)

sleshchenko:
### What does this PR do?
This PR renames WorkspaceRouting to DevWorkspaceRouting(also adds shortname dwr).

Before merging it we need to sync with Florent and Platform team since Theia and DevWorkspaceChe operator depends on it.
It fully renamed workspace routing to devworkspace routing:
Screenshot_20210315_123734

All other workspace -> devworkspace renaming is moved to https://github.com/devfile/devworkspace-operator/issues/321

It fixes https://github.com/devfile/devworkspace-operator/issues/161

### Is it tested? How?
**The proper update testing:**
```
git checkout main
make uninstall

gh pr checkout 297
export DWO_IMG=sleshchenko/che-workspace-controller:dwr
make install
# play with DevWorkspace make sure they works and their DWR are created.
```

**Try to test without cleaning up workspacerouting**:
1. Install the main branch state:
```bash
git checkout main
export DWO_IMG=quay.io/devfile/devworkspace-controller:next
make install
```

<details>

<summary> 2.Create devworkspaces with different routings (Basic, Web Terminal, Openshift OAuth</summary>

```bash
oc new-project dev-ws
# Web Terminal Routing
kubectl apply -f ./samples/web-terminal.yaml
# Basic Routing
kubectl apply -f ./samples/flattened_theia-next.yaml
# OpenShift OAuth
cat << EOF | kubectl apply -f -
kind: DevWorkspace
metadata:
  name: theia-oauth
spec:
  routingClass: openshift-oauth
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
    ### BEGIN Contributions from machine-exec plugin ###
      - name: terminal
        attributes:
          "app.kubernetes.io/name": che-terminal.eclipse.org
          "app.kubernetes.io/part-of": che.eclipse.org
          "app.kubernetes.io/component": terminal
        container:
          image: "quay.io/eclipse/che-machine-exec:nightly"
          command: ['/go/bin/che-machine-exec']
          args:
            - '--url'
            - '0.0.0.0:4444'
            - '--pod-selector'
            - controller.devfile.io/workspace_id=$\(DEVWORKSPACE_ID)
          endpoints:
            - name: "che-mach-exec"
              exposure: public
              targetPort: 4444
              protocol: ws
              secure: true
              attributes:
                type: terminal
    ### END Contributions from machine-exec plugin ###
    ### BEGIN Contributions from Theia plugin ###
      - name: plugins
        volume: {}
      - name: theia-ide
        attributes:
          "app.kubernetes.io/name": che-theia.eclipse.org
          "app.kubernetes.io/part-of": che.eclipse.org
          "app.kubernetes.io/component": editor
        container:
          image: "quay.io/eclipse/che-theia:next"
          env:
            - name: THEIA_PLUGINS
              value: local-dir:///plugins
            - name: HOSTED_PLUGIN_HOSTNAME
              value: 0.0.0.0
            - name: HOSTED_PLUGIN_PORT
              value: "3130"
            - name: THEIA_HOST
              value: 0.0.0.0
          volumeMounts:
            - path: "/plugins"
              name: plugins
          mountSources: true
          memoryLimit: "512M"
          endpoints:
            - name: "theia"
              exposure: public
              targetPort: 3100
              secure: true
              protocol: http
              attributes:
                type: ide
            - name: "webviews"
              exposure: public
              targetPort: 3100
              protocol: http
              secure: true
              attributes:
                type: webview
                unique: "true"
            - name: "theia-dev"
              exposure: public
              targetPort: 3130
              protocol: http
              attributes:
                type: ide-dev
            - name: "theia-redir-1"
              exposure: public
              targetPort: 13131
              protocol: http
            - name: "theia-redir-2"
              exposure: public
              targetPort: 13132
              protocol: http
            - name: "theia-redir-3"
              exposure: public
              targetPort: 13133
              protocol: http
    commands:
      - id: say-hello
        exec:
          component: plugin
EOF

```

</details>

3. Make sure all of them are running ^.
4. Update DWO with PR state.
```
gh pr checkout 297
export DWO_IMG=sleshchenko/che-workspace-controller:dwr
make install
```

5. Make sure all of devworkspaces are still running with working URLs (at least Theia)
6. Check `kubectl get dwr`
7. Make sure namespace is not hanging in TERMINATING state: `kubectl delete namespace dev-ws`.
:warning: It is!


sleshchenko:
It should wait until we achieve milestone 1, planned v0.1.1

benoitf:
From che-theia side I could check if a CRD exists and use either WorkspaceRouting or DevWorkspaceRouting stuff

benoitf:
but if you merge it I can adjust quickly so I'm fine to merge it as soon as possible

deploy/templates/base/config.properties
```diff
@@ -1,9 +1,9 @@
 controller.plugin_artifacts_broker.image=quay.io/eclipse/che-plugin-artifacts-broker:v3.4.0
 controller.webhooks.enabled=true
-devworkspace.routing.cluster_host_suffix=${ROUTING_SUFFIX}
-devworkspace.default_routing_class=${DEFAULT_ROUTING}
+devworkspace.routing.cluster_host_suffix=
+devworkspace.default_routing_class=basic
 # image pull policy that is applied to all workspace's containers
-devworkspace.sidecar.image_pull_policy=${PULL_POLICY}
+devworkspace.sidecar.image_pull_policy=Always
```
amisevsk, line 6: Accidental commit?

amisevsk (approved):
LGTM pending tests (I'll try find time today/tomorrow)

Looks like some deploy/templates change were committed accidentally (ctrl-c during make install?)

deploy/templates/base/config.properties
sleshchenko: it's only true pull policy :-D

deploy/templates/base/manager_image_patch.yaml
```diff
@@ -9,8 +9,8 @@ spec:
     spec:
       containers:
       - name: devworkspace-controller
-        image: ${IMG}
+        image: quay.io/devfile/devworkspace-controller:next
```
JPinkney, line 12: Was this one an accident as well? I assume we still want the templates

deploy/templates/cert-manager/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: ${NAMESPACE}
+namespace: devworkspace-controller
```
JPinkney, line 2: And maybe this one too

deploy/templates/service-ca/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: ${NAMESPACE}
+namespace: devworkspace-controller
```
JPinkney, line 2: Also potentially this one :laughing:

deploy/templates/components/manager/manager.yaml
```diff
@@ -46,4 +46,4 @@ spec:
           - name: RELATED_IMAGE_default_tls_secrets_creation_job
             value: "quay.io/eclipse/che-tls-secret-creator:alpine-3029769"
           - name: RELATED_IMAGE_pvc_cleanup_job
-            value: "quay.io/libpod/busybox:1.30.1"
+            value: "registry.access.redhat.com/ubi8-minimal:8.3-291"
```
JPinkney, line 49: Should we seperate this out into a seperate commit so that if we ever have to go looking for when it was changed it might be easier?

deploy/templates/base/manager_image_patch.yaml
sleshchenko: true, probably I just did ctrl+c when make generate as Angel assumed. I'll revert

deploy/templates/components/manager/manager.yaml
sleshchenko: Good catch. I did not even plan to change it in the current PR, but seems it was made and I accidentally committed it.

After we do milestone 1, I'll update PR and make sure generated stuff are separate from other and test if it works. And then only re-request review.

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -41,32 +41,32 @@ var (
 	NoSolversEnabled = errors.New("reconciler does not define SolverGetter")
 )
 
-const workspaceRoutingFinalizer = "workspacerouting.controller.devfile.io"
+const devWorkspaceRoutingFinalizer = "devworkspacerouting.controller.devfile.io"
```
sleshchenko, line 44: :warning: It means old DevWorkspaces with such finalizers will hangs in terminating state, but I think we're safe since it's only OpenShift OAuth which uses it so far.

But we also can leave some safe double check if somebody thinks it worths it.

deploy/templates/base/config.properties
sleshchenko: should be fixed now

Makefile
```diff
@@ -208,7 +208,7 @@ uninstall: generate_deployment
 # leave workspaces in a hanging state if we add finalizers.
 	$(K8S_CLI) delete devworkspaces.workspace.devfile.io --all-namespaces --all --wait || true
 	$(K8S_CLI) delete devworkspacetemplates.workspace.devfile.io --all-namespaces --all || true
-	$(K8S_CLI) delete workspaceroutings.controller.devfile.io --all-namespaces --all --wait || true
+	$(K8S_CLI) delete devworkspaceroutings.controller.devfile.io --all-namespaces --all --wait || true
```
amisevsk, line 211: I'm not sure what we should do here -- with this change there seems to be no automated way to delete `workspaceroutings` from the cluster. I suppose it's not really a problem since we generally require a clean uninstall + reinstall, but it's worth noting that the actual process would be
```
git checkout <before this commit>
make uninstall
git checkout <after this commit>
make install
```

amisevsk (approved):
Tested installing current `main` branch commit, creating workspaces, and updating. Everything seems to work fine (except that the workspaceroutings CRD is left on the cluster).

Makefile
sleshchenko: > we generally require a clean uninstall + reinstall

Yes and not. I don't think Che operator is going to reinstall everything for upgrading from v0.1.1 to v0.2.0. 
I think that we agreed that at this point it's OK if we just leave workspaceroutings on the cluster... But it makes sense to keep removing of workspacerouting in out `make uninstall` and on  Che Operator side for a couple of releases.

Maybe then it's a good time to remove components CRDs if we don't need them anymore.

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
sleshchenko: One more example with finalizers - Che routing =(
Which mean there will be hanging workspaceroutings, after Che 7.27 -> 7.28 update.

The only safe option - let DWO for a few releases clean up workspacerouting CRs (Che DWO should listen to workspacerouting and devworkspacerouting...), then Che Operator can remove CRD.

Or just assume nobody used devworkspaces with Che 7.27.1 and just rename.

Makefile
sleshchenko: make uninstall
make install
the only safe option =(
Is Che Operator supposed to drop every DevWorkspace instance before upgrade?...

sleshchenko: Alternative: we assume that workspacerouting and devworkspacerouting has the same stuff to finalize.
So, Che Operator updated DWO, update WorkspaceRouting to has [] as finalizers, remove all WorkspaceRoutings CRs, remove CRD.

metlos: I think at this point in time we should not worry about leaving garbage behind in the Che-related usecases -  we're in the alpha's in Che so this kind of stuff is expected IMHO and can be simply documented in the release notes or somewhere.

I think the situation is different for webterminal though where I assume we need to make sure that the upgrade of the operator is not going to cause any stuck devworkspaces...

Issue #199, opened 2020-11-25
Every time devworkspace operator is restarted, a new serviceaccount token is created


amisevsk:
In cases where the devworkspace operator deployment is stuck in a crashloop, every time it starts it will manage the webhooks server serviceaccount in such a way that a new sa-token is created. 

```
❯ kc get secrets
NAME                                      TYPE                                  DATA   AGE
default-token-tcwqm                       kubernetes.io/service-account-token   3      31m
devworkspace-operator-webhook-cert        kubernetes.io/tls                     3      31m
devworkspace-webhook-server-token-52mhn   kubernetes.io/service-account-token   3      29m
devworkspace-webhook-server-token-5427w   kubernetes.io/service-account-token   3      24m
devworkspace-webhook-server-token-gbgjc   kubernetes.io/service-account-token   3      4m16s
devworkspace-webhook-server-token-hj8jt   kubernetes.io/service-account-token   3      19m
devworkspace-webhook-server-token-lhlvg   kubernetes.io/service-account-token   3      27m
devworkspace-webhook-server-token-mb6dk   kubernetes.io/service-account-token   3      31m
devworkspace-webhook-server-token-mdljr   kubernetes.io/service-account-token   3      9m23s
devworkspace-webhook-server-token-ns25v   kubernetes.io/service-account-token   3      14m
devworkspace-webhook-server-token-r67bb   kubernetes.io/service-account-token   3      30m
devworkspace-webhook-server-token-v92b7   kubernetes.io/service-account-token   3      30m
devworkspace-webhook-server-token-xlh7g   kubernetes.io/service-account-token   3      30m
devworkspace-webhook-server-token-zbstn   kubernetes.io/service-account-token   3      83s
```

PR #300, opened 2021-03-04
Avoid updating webhooks service account unless required.

amisevsk:
### What does this PR do?
* Remove checks for whether webhooks are enabled, since they're now required (for conversion at least)
* Only update the webhooks serviceAccount when necessary, as otherwise our update wipes out the `.secrets` field and causes k8s to create a new secret for the SA.

Fixes https://github.com/devfile/devworkspace-operator/issues/199

### Is it tested? How?
1. Deploy operator
2. Run `make restart` a few times; after first run, log should contain `Webhook server service account up to date` instead of `Updated webhook server service account` and only one secret named `devworkspace-controller-serviceaccount-token-xxxxx` should exist.

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -143,11 +143,13 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return r.updateWorkspaceStatus(clusterWorkspace, reqLogger, &reconcileStatus, reconcileResult, err)
 	}()
 
-	msg, err := r.validateCreatorLabel(clusterWorkspace)
-	if err != nil {
-		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
-		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = msg
-		return reconcile.Result{}, err
+	if workspace.Annotations[config.WorkspaceRestrictedAccessAnnotation] == "true" {
+		msg, err := r.validateCreatorLabel(clusterWorkspace)
+		if err != nil {
+			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
+			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = msg
+			return reconcile.Result{}, err
+		}
 	}
 
 	if _, ok := clusterWorkspace.Annotations[config.WorkspaceStopReasonAnnotation]; ok {
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -143,7 +143,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return r.updateWorkspaceStatus(clusterWorkspace, reqLogger, &reconcileStatus, reconcileResult, err)
 	}()
 
-	msg, err := r.validateCreatorTimestamp(clusterWorkspace)
+	msg, err := r.validateCreatorLabel(clusterWorkspace)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/webhook/service_account.go
```diff
@@ -40,12 +41,16 @@ func CreateWebhookSA(client crclient.Client,
 		if err != nil {
 			return err
 		}
-		serviceAccount.ResourceVersion = existingCfg.ResourceVersion
-		err = client.Update(ctx, serviceAccount)
-		if err != nil {
-			return err
+		if needsUpdate(serviceAccount, existingCfg) {
+			serviceAccount.ResourceVersion = existingCfg.ResourceVersion
+			err = client.Update(ctx, serviceAccount)
```
sleshchenko, line 46: Does it mean that after we update labels on SA, new token will be created? Can we avoid it with patching or it's by design for k8s?

controllers/workspace/validation.go
```diff
@@ -20,15 +20,12 @@ import (
...
+// validateCreatorLabel checks that a devworkspace was created after workspace-related mutating webhooks
 // and ensures a creator ID label is applied to the workspace. If webhooks are disabled, validation succeeds by
 // default.
 //
 // If error is not nil, a user-readable message is returned that can be propagated to the user to explain the issue.
-func (r *DevWorkspaceReconciler) validateCreatorTimestamp(workspace *devworkspace.DevWorkspace) (msg string, err error) {
-	if config.ControllerCfg.GetWebhooksEnabled() != "true" {
-		return "", nil
-	}
+func (r *DevWorkspaceReconciler) validateCreatorLabel(workspace *devworkspace.DevWorkspace) (msg string, err error) {
```
sleshchenko, line 28: Does it make sense to disable that check for non-restricted-access workspaces?

controllers/workspace/devworkspace_controller.go
```diff
@@ -156,17 +156,6 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: true}, err
 	}
 
-	restrictedAccess := clusterWorkspace.Annotations[config.WorkspaceRestrictedAccessAnnotation]
```
sleshchenko, line 159: would it be a good idea to check if webhooks really exists on the cluster here? Or when they absent workspace can't be reconciled due `validateCreatorLabel` check? 
I think the second.

sleshchenko (approved):
LGTM

Left some comments but they can be addressed separately

pkg/webhook/service_account.go
amisevsk: The `.secrets` field in a SA object gets populated with a reference to the secret that should be used. When we apply the spec SA as an update, it wipes out that field and k8s fills it again... with a new secret. 

We could potentially work around this by copying the current secrets field over to the spec before updating, but I'm not sure if that causes any issues in other ways, so I opted to avoid updating unless necessary (which realistically, should not be needed 99% of the time)

controllers/workspace/validation.go
amisevsk: Probably, but all workspaces get a creator label now; the check itself is doing something kind of different from before. The only way we would have a DW without the label would be if it was created before webhooks, in which case it may make sense to fail the workspace anyways since it might have missed conversion, etc.

controllers/workspace/devworkspace_controller.go
amisevsk: I think it's probably safer to do some check that webhooks are enabled (i.e. actually checking that webhooks exist on the cluster), but I would separate that out into another issue.

amisevsk: Created https://github.com/devfile/devworkspace-operator/issues/305

Issue #302, opened 2021-03-05
ExposedEndpoint.Attributes should handle free-form JSON as values


metlos:
Currently `ExposedEndpoint.Attributes` is `map[string]string` but it is derived from devfile API `Endpoint.Attributes` which is `map[string]apiext.JSON`.

When converting a devworkspace that contains endpoint attributes with non-string values, the workspace routing fails with a JSON marshaling error.  This happens for basic routing as well as openshift oauth routing.

We probably should use the same type for `ExposedEndpoints.Attributes` as for `Endpoint.Attributes`. 

The other solution would be to somehow do some clever type conversions in `Attributes.Into()` but that seems very prone to error or unexpected behavior IMHO.

PR #303, opened 2021-03-05
Make routing attributes allign with devfile attributes

sleshchenko:
### What does this PR do?
Make routing attributes align with devfile attributes

It fixes https://github.com/devfile/devworkspace-operator/issues/302

### Is it tested? How?
Tested as usual `kc apply -f ./samples/flattened_theia-next.yaml`

### DIFF
--- a/apis/controller/v1alpha1/workspacerouting_types.go
+++ b/apis/controller/v1alpha1/workspacerouting_types.go
@@ -14,6 +14,7 @@ package v1alpha1
 
 import (
 	devworkspace "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	devfileAttr "github.com/devfile/api/v2/pkg/attributes"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -68,7 +69,8 @@ type ExposedEndpoint struct {
 	// Public URL of the exposed endpoint
 	Url string `json:"url"`
 	// Attributes of the exposed endpoint
-	Attributes map[string]string `json:"attributes"`
+	// +optional
+	Attributes devfileAttr.Attributes `json:"attributes,omitempty"`
 }
 
 type EndpointList []devworkspace.Endpoint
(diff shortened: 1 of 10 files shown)


apis/controller/v1alpha1/workspacerouting_types.go
```diff
@@ -68,7 +69,8 @@ type ExposedEndpoint struct {
 	// Public URL of the exposed endpoint
 	Url string `json:"url"`
 	// Attributes of the exposed endpoint
-	Attributes map[string]string `json:"attributes"`
+	// +optional
```
metlos, line 72: What's a default value for an optional map? Because we're not doing null checks on this property and afaik the default value for a map is `nil`.

sleshchenko: Good catch. Then we must do nil checks

sleshchenko: But it's not map, it's Attributes. Need to check

sleshchenko: At least Che Theia Next sample has some endpoints with missing attributes and nothing failed during my testing )

metlos: Seems I was both wrong and right.. https://play.golang.org/p/aBUm_t0rwzo 🤯

sleshchenko: Yeap
Nil map is safe to read, but must be initialized before putting entry

amisevsk: The default value for maps is `nil`, but reading from a nil map is the same as reading from an empty map (i.e. doesn't panic). The only issue with nil maps is adding an item, which would require a nil check (but we don't have any writes to the map, only copies of the whole map in the codebase).

amisevsk (commented):
LGTM, but I worry about how changes are propagated when a cluster is "upgraded" with the changed CRDs. I'll test this out some, but it may also be a non-issue if we'll have the breaking change of https://github.com/devfile/devworkspace-operator/pull/297 anyways.

apis/controller/v1alpha1/workspacerouting_types.go
amisevsk: Oops sorry, had this conversation open since this morning and didn't refresh the page to get the later comments.

amisevsk (approved):
Tested updating manually on OpenShift and didn't find any issues. Also tested using a non-string attribute on an endpoint and everything works as expected.

Issue #305, opened 2021-03-06
Check that required webhooks exist on the cluster when a DW has restricted-access annotation


amisevsk:
### Description

As a follow-up to https://github.com/devfile/devworkspace-operator/pull/300#discussion_r588127296, webhooks are enabled by default but the controller only sets up the webhooks deployment at startup. Should we also do an explicit check for webhooks when we encounter the restricted-access annotation?

PR #307, opened 2021-03-08
Fail any DevWorkspace when webhooks are not present on cluster

amisevsk:
**Note**: I'm not sure this PR should even be merged. The added checks may catch some simple configuration errors (e.g. webhooks accidentally get deleted) but does nothing to address the main reason we do any checking for webhooks (i.e. preventing unauthorized access). I'm opening this PR for discussion in case there's something I'm missing, but it seems kind of pointless IMO.

### What does this PR do?
* Validate that webhooks exist on cluster when reconciling a DevWorkspace, and fail the workspace if they're not present
* Rework validateCreatorID check to check that both mutating and validating webhooks exist and that they're older than the DevWorkspace. 
* Extract two functions: `killWorkspace`, which deletes a workspace's deployment and marks it as failed, and `failWorkspace` which just cuts out the four lines we use to fail workspaces everywhere in the main reconcile loop.

The PR is split into two commits:
- The first commit is a half-way solution, doing the detailed check when a workspace has the `restricted-access` annotation set.
- The second commit extends the check above to all DevWorkspaces and does some light refactoring

Fixes https://github.com/devfile/devworkspace-operator/issues/305

### Is it tested? How?
1. Create a DevWorkspace on the cluster
2. Delete the mutating webhook: `kc delete mutatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io`
3. Edit the DevWorkspace created in step 1
4. Workspace should be failed, workspace deployment should be deleted.

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -144,7 +144,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	}()
 
 	if workspace.Annotations[config.WorkspaceRestrictedAccessAnnotation] == "true" {
-		msg, err := r.validateCreatorLabel(clusterWorkspace)
+		msg, err := r.validateRestrictedWorkspace(clusterWorkspace)
 		if err != nil {
 			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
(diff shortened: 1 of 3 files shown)


amisevsk:
Closing this PR as it's out of date with the current repo and I don't think there's much value here.

Issue #298, opened 2021-03-04
Separate workspace routing into a separate library


metlos:
Because the controllers/workspacerouting  package is reused in 2 operators now, it would be good if it was separated out of the main devworkspace-operator code into a separate codebase.

This would improve the code isolation and chances of introducing dependencies on code/configuration that is not common to all the operators that use the workspace routing code.

In particular, it would make it possible to not rely on `config.ControllerConfig` in the workspace routing codepaths, because the configuration mechanism differs between the operators.

amisevsk:
I agree that the code used in the workspaceRouting controller shouldn't depend on any mutable state from elsewhere (i.e. no config settings), but would prefer to keep it in this codebase so that we can share constants, attributes, annotations, etc.

PR #308, opened 2021-03-08
Make workspaceRoutings controller not depend on pkg/config

amisevsk:
### What does this PR do?
1. Move constants out of `pkg/config` to `pkg/constants`
2. Remove the dependency on `pkg/config` from workspaceRoutings controller. This is comes with a few caveats:
    - We set whether we're in openshift on the reconciler struct during setup
    - We have to propagate this value down through all solvers (this means changing some interfaces cc:@metlos)
    - For (deprecated?) openshift-oauth routing, we just assume a pull policy of "always" for the openshift-oauth sidecar.

Fixes https://github.com/devfile/devworkspace-operator/issues/298

### Is it tested? How?
Should be no functional changes (i.e. usual test flow)


controllers/controller/workspacerouting/solvers/basic_solver.go
```diff
@@ -41,19 +41,19 @@ type BasicSolver struct{}
...
 	return false
 }
 
-func (s *BasicSolver) Finalize(routing *controllerv1alpha1.WorkspaceRouting) error {
+func (s *BasicSolver) Finalize(*controllerv1alpha1.WorkspaceRouting) error {
 	return nil
 }
 
-func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.WorkspaceRouting, workspaceMeta WorkspaceMetadata) (RoutingObjects, error) {
+func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.WorkspaceRouting, workspaceMeta WorkspaceMetadata, isOpenShift bool) (RoutingObjects, error) {
```
sleshchenko, line 52: since BasicSolver already mention dependency on platform, what do you think if we push isOpenshift into BasicSolver as a field?
```
// Kubernetes: use Ingresses without TLS
// OpenShift: use Routes with TLS enabled
type BasicSolver struct{}
```

controllers/controller/workspacerouting/workspacerouting_controller.go
```diff
@@ -183,17 +186,20 @@ func (r *WorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
...
-	}
-
 	clusterRoutingObj := solvers.RoutingObjects{
 		Services:  clusterServices,
 		Ingresses: clusterIngresses,
-		Routes:    clusterRoutes,
 	}
+
+	if r.IsOpenShift {
+		routesInSync, clusterRoutes, err := r.syncRoutes(instance, routes)
```
sleshchenko, line 195: Do you see any sense in syncing Ingresses on OpenShift?
I know they work somehow but how much we expect routings create ingresses on OpenShift?

pkg/library/container/mountSources.go
```diff
@@ -14,9 +14,10 @@ package container
 
 import (
 	devworkspace "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
-	"github.com/devfile/devworkspace-operator/pkg/config"
-	"github.com/devfile/devworkspace-operator/pkg/library/constants"
 	corev1 "k8s.io/api/core/v1"
+
+	constants2 "github.com/devfile/devworkspace-operator/pkg/constants"
```
sleshchenko, line 19: we need more constants, I won't approve unless we have constants3 as well ;-D

I know, it's just refactoring tools._
```suggestion
	devfileConstants "github.com/devfile/devworkspace-operator/pkg/constants"
```

or any other shorter name which makes clear it's from devfile/api

amisevsk: Yeah, that's the downside of IDE import magic :D

metlos (commented):
👍🏼 for moving constants from `config` to `constants`.

👍🏼 👎🏼 for `isOpenShift` being passed down - I would rather see the detection separated out as suggested in the comments below.

controllers/controller/workspacerouting/workspacerouting_controller.go
amisevsk: My idea was to not really worry about what we get back and just sync all the objects we can, which would mean that e.g. a solver that only generates ingresses would still work on OpenShift (hopefully). Otherwise, all solvers have to support routes instead of ingresses on OpenShift.

It's not terribly important, though, so I'll make the change.

controllers/controller/workspacerouting/solvers/basic_solver.go
amisevsk: That's the idea I couldn't think of yesterday as I was doing this. Thanks!

sleshchenko:
rebased against main branch to fix the issue with pull_secret labels that was used from config.

sleshchenko (approved):
Looks better after the last changes.

I'm a bit +1 to have infrastructure that caches the evaluated value for isOpenShift.
I vote this way since we have 4 places cluster.IsOpenShift (which do 4 requests instead of 1) and then we have config.IsOpenShift which is called 4 times as well.
But that's not strong opinion.

> if the behaviour of our code depends on the cluster state it should explicitly be visible somewhere.

true, that makes a bit less visible that Routing depends on cluster, but I don't see it a big issue

amisevsk:
I started implementing the suggestion in https://github.com/che-incubator/devworkspace-che-operator/blob/main/pkg/infrastructure/infrastructure.go and came across a few issues:

1. Error handling becomes more difficult, as our best option in an `init` method is to panic. Otherwise, the equivalent of `config.ControllerConfig.isOpenShift()` has to handle a potential error case (which will require error handling all the way up from each use-site). The provided sample also hides errors behind setting `undetected`, which is an error case in most situations (you'd e.g. end up creating ingresses instead of routes on OpenShift).
2. We're treating potentially transient (e.g. network) errors as fatal, as we only set our info about the cluster once. It's a problem present in parts of our current codebase, so not a huge concern, but it seems strange.
3. Not a direct impact, but using a global var in this way means that any code that imports the package is dependent on running in a context where there is some cluster to look at; we can't mock out an init method (at least, not nicely).

Some WIP changes are available here: https://github.com/amisevsk/devworkspace-operator/commit/7eb790a8c7a9024be8d63f92af37e0af36e01218; I can push these to this PR if it seems to be worth it.

sleshchenko:
> We're treating potentially transient (e.g. network) errors as fatal, as we only set our info about the cluster once. It's a problem present in parts of our current codebase, so not a huge concern, but it seems strange.

Another example of such issue:
```
{"error": "unable to retrieve the complete list of server APIs: packages.operators.coreos.com/v1: the server is currently unable to handle the request"}
```
Which I tried to solve but did not managed to finalize yet.
And in my case, only cluster restart helped, and other basic operation, like remove namespace, could not be finished due the same error.
So, with this approach or the new proposed, we will get the same broken controller, I think.
We can try several times on failure in the proposed solution to fix potentially transient issues.

> Not a direct impact, but using a global var in this way means that any code that imports the package is dependent on running in a context where there is some cluster to look at; we can't mock out an init method (at least, not nicely).

That's the issue.
What if infrastructure would have `initState()` method which would be needed to call before other methods, like during controller setting up.
Is openshift panic if it's called before `initState()`, `evaluate`, ...

metlos:
Let me summarize the great observations you had:

* **handle transient errors** - I think the only good response to this is to just error out and let kubernetes restart the operator. If we cannot determine what kind of cluster we are running in, it makes no sense to continue.

* **testing, initialization** - I can see the trouble with initializing `infrastructure.Current` during `init()` in test environments where there is no kubernetes context available. Therefore I like the idea of explicit `infrastructure.Initialize()` or some such that would be called during operator startup and could return an error with which the operator could ordinarily exit and let k8s restart it. In the test code, we could initialize `infastructure.Current` "manually" to whatever value we need in the tests.

amisevsk:
> What if infrastructure would have `initState()` method which would be needed to call before other methods, like during controller setting up.
> Is openshift panic if it's called before `initState()`, `evaluate`, ...

This gets into the real issues with global variables though -- there's a global, mutable state, that must be initialized or your code panics, meaning you have to trace the whole execution up to where you try to use it to make sure it's been initialized. This also brings us back pretty close to where we started -- controllerCfg is a very similar construct (except it also depends on a configmap).

metlos:
> > What if infrastructure would have `initState()` method which would be needed to call before other methods, like during controller setting up.
> > Is openshift panic if it's called before `initState()`, `evaluate`, ...
> 
> This gets into the real issues with global variables though -- there's a global, mutable state, that must be initialized or your code panics, meaning you have to trace the whole execution up to where you try to use it to make sure it's been initialized. This also brings us back pretty close to where we started -- controllerCfg is a very similar construct (except it also depends on a configmap).

What is the alternative though? IMHO this is quite similar to `AddToScheme()` or logger initialization - just some setup you need to make sure is done at startup...

metlos (approved):

sleshchenko:
All PRs has default templates check broken. I'll fix in separately.

Issue #236, opened 2020-12-29
Add readiness and liveness probe for controller and webhook server


sleshchenko:
Add readiness and liveness probe for controller and webhook server.

It came from issues.redhat.com/browse/CRW-683
issues.redhat.com/browse/CRW-916

sleshchenko:
The related PR on Che Operator side https://github.com/eclipse/che-operator/pull/610/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R116

sleshchenko:
Is implemented in https://github.com/devfile/devworkspace-operator/pull/309

PR #309, opened 2021-03-09
readiness and liveness probes

tinakurian:
### What does this PR do?
Adds the readiness and liveness probes for devworkspaces

https://github.com/devfile/devworkspace-operator/issues/236

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
 	ctrlconfig "sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	// +kubebuilder:scaffold:imports
 )
@@ -82,11 +83,12 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseDevMode(config.GetDevModeEnabled())))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:             scheme,
-		MetricsBindAddress: metricsAddr,
-		Port:               9443,
-		LeaderElection:     enableLeaderElection,
-		LeaderElectionID:   "8d217f93.devfile.io",
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: ":6789",
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "8d217f93.devfile.io",
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
@@ -126,6 +128,18 @@ func main() {
 	setupLog.Info("setting up webhooks")
 	if err := webhook.SetupWebhooks(context.Background(), cfg); err != nil {
 		setupLog.Error(err, "failed to setup webhooks")
+		os.Exit(1)
+	}
+
+	// Setup health check
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		setupLog.Error(err, "Unable to set up health check")
+		os.Exit(1)
+	}
+
+	// Setup ready check
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		setupLog.Error(err, "Unable to set up ready check")
 		os.Exit(1)
 	}
 
--- a/pkg/webhook/deployment.go
+++ b/pkg/webhook/deployment.go
@@ -15,20 +15,19 @@ package webhook
 import (
 	"context"
 
+	"github.com/devfile/devworkspace-operator/internal/images"
+	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
 	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
 	"github.com/devfile/devworkspace-operator/webhook/server"
-
-	"github.com/devfile/devworkspace-operator/internal/images"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/devfile/devworkspace-operator/pkg/config"
 )
 
 func CreateWebhookServerDeployment(
@@ -104,6 +103,34 @@ func getSpecDeployment(webhooksSecretName, namespace string) (*appsv1.Deployment
 							Image:           images.GetWebhookServerImage(),
 							Command:         []string{"/usr/local/bin/entrypoint", "/usr/local/bin/webhook-server"},
 							ImagePullPolicy: corev1.PullAlways,
+							LivenessProbe: &corev1.Probe{
+								Handler: corev1.Handler{
+									HTTPGet: &corev1.HTTPGetAction{
+										Path:   "/healthz",
+										Port:   intstr.FromString("liveness-port"),
+										Scheme: "HTTP",
+									},
+								},
+								InitialDelaySeconds: 15,
+								TimeoutSeconds:      5,
+								PeriodSeconds:       10,
+								SuccessThreshold:    1,
+								FailureThreshold:    5,
+							},
+							ReadinessProbe: &corev1.Probe{
+								Handler: corev1.Handler{
+									HTTPGet: &corev1.HTTPGetAction{
+										Path:   "/readyz",
+										Port:   intstr.FromString("liveness-port"),
+										Scheme: "HTTP",
+									},
+								},
+								InitialDelaySeconds: 10,
+								TimeoutSeconds:      5,
+								PeriodSeconds:       10,
+								SuccessThreshold:    1,
+								FailureThreshold:    3,
+							},
 							VolumeMounts: []corev1.VolumeMount{
 								{
 									Name:      server.WebhookServerCertsVolumeName,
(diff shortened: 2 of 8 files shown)


webhook/main.go
```diff
@@ -67,6 +68,7 @@ func main() {
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Namespace: namespace,
 		Scheme:    scheme,
+		HealthProbeBindAddress: ":8443",
```
amisevsk, line 71: I think this will cause a conflict since the webhook server listens on this port so the health service can't bind:
```json
{
  "msg":"Manager exited non-zero","error":"listen tcp 0.0.0.0:8443: bind: address already in use",
  "stacktrace":"main.main\n\t/devworkspace-operator/webhook/main.go:109\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"
}
```

tinakurian: Thanks @amisevsk! changed to 6789 :-)

pkg/webhook/deployment.go
```diff
@@ -15,6 +15,8 @@ package webhook
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/util/intstr"
```
amisevsk, line 18: Tedious nitpick: Could you group the imports in this file? E.g.
```go
import (
  "<library-imports>"

  "<local-repo-imports>"

  "<dependency-imports>"
)
```

Once there's blank lines introduced between imports, editors tend to continue to mangle them.

webhook/main.go
```diff
@@ -16,6 +16,8 @@ import (
 	"os"
 	"os/signal"
 
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
```
amisevsk, line 19: Same as above on imports (also: optional)

amisevsk (approved):
LGTM :+1:  

Tested on `crc` and `minikube` and readiness/liveness probes are added and succeed.

pkg/webhook/deployment.go
```diff
@@ -101,6 +103,34 @@ func getSpecDeployment(webhooksSecretName, namespace string) (*appsv1.Deployment
...
+								Handler: corev1.Handler{
+									HTTPGet: &corev1.HTTPGetAction{
+										Path:   "/healthz",
+										Port:   intstr.FromString("liveness-port"),
+										Scheme: "HTTP",
+									},
+								},
+								InitialDelaySeconds: 1,
+								TimeoutSeconds:      1,
+								PeriodSeconds:       1,
```
sleshchenko, line 116: The tricky part with probes, choose meaningful settings for timeouts. Controller and webhook server seems to have different values currently.
Webhook server has 1s delay + 1 second period * 3 times FailureThreshold. Does it mean that webhook server will be restarted in 4 seconds if probe don't get available for this pretty short time?

For liveness probe we have 30 times FailureThreshold with 1 second period, which means webhooks server will receive a traffic 30 seconds after it hangs or crashed (not the process but server itself is not able to handle traffic)
^ that it's not really true and just speculation since readiness probe and liveness probe use the same endpoint which means container will be restarted after 3 seconds of unavailability(1s period * 3 times FailureThreshold). But it indicates we are able to find better values.

I did not think carefully enough to propose good numbers.

webhook/main.go
```diff
@@ -90,6 +93,18 @@ func main() {
...
 
+	// Setup health checks
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		log.Error(err, "Unable to set up health check")
+		os.Exit(1)
+	}
+
+	// Setup health checks
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		log.Error(err, "Unable to set up health check")
```
sleshchenko, line 104:
```suggestion
		log.Error(err, "Unable to set up ready check")
```

webhook/main.go
```diff
@@ -90,6 +93,18 @@ func main() {
 	var shutdownChan = make(chan os.Signal, 1)
 	signal.Notify(shutdownChan, syscall.SIGTERM)
 
+	// Setup health checks
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		log.Error(err, "Unable to set up health check")
+		os.Exit(1)
+	}
+
+	// Setup health checks
```
sleshchenko, line 102:
```suggestion
	// Setup ready checks
```

main.go
```diff
@@ -121,6 +123,18 @@ func main() {
...
 
+	// Setup health checks
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		setupLog.Error(err, "Unable to set up health check")
+		os.Exit(1)
+	}
+
+	// Setup health checks
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		setupLog.Error(err, "Unable to set up health check")
```
sleshchenko, line 134:
```suggestion
		setupLog.Error(err, "Unable to set up ready check")
```

main.go
```diff
@@ -121,6 +123,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup health checks
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		setupLog.Error(err, "Unable to set up health check")
+		os.Exit(1)
+	}
+
+	// Setup health checks
```
sleshchenko, line 132:
```suggestion
	// Setup ready checks
```

sleshchenko (approved):
LGTM in general, just a few places to think a bit more about.

pkg/webhook/deployment.go
tinakurian: good catch! sorry about that, didn't notice it - fixed here: https://github.com/devfile/devworkspace-operator/pull/309/files#diff-ab898919aca276514502fbe2f2c75745364e965107af8d3f497519260891e271R20

webhook/main.go
tinakurian: good catch! fixed here: https://github.com/devfile/devworkspace-operator/pull/309/files#diff-a078bce520bdda0f38cdd11ca1fc03fe464c7260312a6f9d1d6fbe6b5ea12ff1R23

pkg/webhook/deployment.go
tinakurian: Nice find and good point! I've changed it to be more generous and made sure that both the numbers match for webhook server and the controller: https://github.com/devfile/devworkspace-operator/pull/309/files#diff-4585a303265f378412beec910445bdbc4cc40b71d51978eb4be67c4712faa2c6R18

webhook/main.go
tinakurian: :woman_facepalming: good catch :smile: fixed! (same with the ones below)

pkg/webhook/deployment.go
```diff
@@ -101,6 +101,34 @@ func getSpecDeployment(webhooksSecretName, namespace string) (*appsv1.Deployment
...
+										Path:   "/readyz",
+										Port:   intstr.FromString("liveness-port"),
+										Scheme: "HTTP",
+									},
+								},
+								InitialDelaySeconds: 15,
+								TimeoutSeconds:      5,
+								PeriodSeconds:       10,
+								SuccessThreshold:    1,
+								FailureThreshold:    10,
```
sleshchenko, line 130: If I'm not mistaked, with this settings container will be restart due liveness probe failure before readiness probe fails.

tinakurian: Yeah, that's correct, is this okay?

sleshchenko: We agreed on the call to make that it makes readiness probe useless.

The one I propose:
```
Liveness 
InitialDelaySeconds: 15, #start checking if container needs to be restarted is started in 15 seconds
TimeoutSeconds:      5, 
PeriodSeconds:       10,
SuccessThreshold:    1,  # once one request succeed, it does not need to be restared.
FailureThreshold:    5,  # Restart container if it does not respond 5*10 seconds
```
^ 
```
Readinsess
InitialDelaySeconds: 10, # start checking if container can receive a traffic in 10 seconds.
TimeoutSeconds:      5,
PeriodSeconds:       10,
SuccessThreshold:    1, # once one request succeed, it can receive external traffic
FailureThreshold:    3, # Temporary do not sent traffic to container if it's not available for 3*10 seconds
```

tinakurian: Thanks again Serhii! I've updated the code to reflect these changes!

amisevsk: Not terribly important, but we generally expect the devworkspace servers to start _fairly_ quickly. The readiness values above mean that either the controller starts 10 seconds after the pod is created, or 20 seconds after the pod is created. Since the readiness probes are only run once per boot, it might make sense to use more frequent checking:
```
Readiness
InitialDelaySeconds: 3
TimeoutSeconds: 3
PeriodSeconds: 3
SuccessThreshold: 1
FailureThreshold: 10 # meaning 10*3 = 30 seconds to start before it's considered failed.
```

It doesn't have a big effect on production deployments, but for development use-cases it might slow things down when re-deploying to only check for readiness 10 seconds after the pod is up.

IMO the first liveness check should happen after the maximum time for readiness checks (30 seconds). With the setup above, we could have liveness probes succeed before the readiness probe (if the readiness probe misses the first window)

sleshchenko: > Since the readiness probes are only run once per boot

Really??? I need revise my knowledge, probably I misunderstood things. Probes were always confusing stuff

JPinkney (approved):
I've just tested and the readiness and liveness probes seem to be there and working. I'll defer to Serhii and Angel as to what the values should be since they know much more about that stuff then I do :smile:

pkg/webhook/deployment.go
sleshchenko: from https://v1-18.docs.kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

> Note: Readiness probes runs on the container during its whole lifecycle.


> Readiness and liveness probes can be used in parallel for the same container. Using both can ensure that traffic does not reach a container that is not ready for it, and that containers are restarted when they fail.

sleshchenko: But stuff that I did not know before: startup probes. But they seems to be related only to liveness probes(not readiness), it does not seem to help us in the case described by Angel.

I'm OK with the proposed:
```
Readiness
InitialDelaySeconds: 3
TimeoutSeconds: 3
PeriodSeconds: 3
SuccessThreshold: 1
FailureThreshold: 10 # meaning 10*3 = 30 seconds to start before it's considered failed.
```
but not sure how much 3s period is going to generate workload for K8s or controller. Probably little enough that we don't feel difference.

amisevsk: Ah maybe I misunderstood readiness probes :upside_down_face:

sleshchenko: You just prove my statement

> Probes were always confusing stuff

despite their simple idea they tend to be difficult to understand for some reason

PR #306, opened 2021-03-08
Remove trailing whitespace from make-release.sh

amisevsk:
### What does this PR do?
Trim trailing whitespace from make-release.sh, which I noticed while rebasing https://github.com/devfile/devworkspace-operator/pull/296

PR #311, opened 2021-03-09
Rework generate-deployment to take a default image parameter

amisevsk:
### What does this PR do?
Since `generate-deployment.sh` is used to generate deploy yamls for releases, it's necessary to be able to specify an image to use when generating the "default" deployment.

This PR adds the `--default-image` parameter for generate-deployment.sh, to avoid the requirement of using sed in make-release.sh

One thing I didn't test is whether this breaks CI on version branches -- previously, make-release commits updated `generate-deployment.sh` so that rerunning it without parameters would generate the same yamls. With the current changes, running `make generate_default_deployment` on the tagged release would result in changes to `deploy/deployment`. We could potentially work around this by making the tag _not_ configurable again, but this would require setting up and maintaining a VERSION file.

https://github.com/devfile/devworkspace-operator/pull/306#pullrequestreview-607416057

### Is it tested? How?
Tested by running `./deploy/generate-deployment.sh --use-defaults --default-image test:test`

### DIFF
--- a/deploy/generate-deployment.sh
+++ b/deploy/generate-deployment.sh
@@ -27,18 +27,25 @@ set -e
 SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
 
 function print_help() {
-  echo "Usage: generate-deployment.sh [ARGS]"
-  echo "Arguments:"
-  echo "  --use-defaults"
-  echo "      Output deployment files to deploy/deployment, using default"
-  echo "      environment variables rather than current shell variables."
-  echo "      Implies '--split yaml'"
-  echo "  --split-yaml"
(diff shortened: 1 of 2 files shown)


amisevsk:
Did we get a definitive answer on 

> One thing I didn't test is whether this breaks CI on version branches -- previously, make-release commits updated `generate-deployment.sh` so that rerunning it without parameters would generate the same yamls. With the current changes, running `make generate_default_deployment` on the tagged release would result in changes to `deploy/deployment`. We could potentially work around this by making the tag _not_ configurable again, but this would require setting up and maintaining a VERSION file.

I think it will be fine, but it's something to keep in mind going forward.

PR #312, opened 2021-03-10
Increase resource limits for controller to avoid OOMKilled

sleshchenko:
Increase resource limits for controller to avoid OOMKilled

### Is it tested? How?
It's extracted from https://github.com/devfile/devworkspace-operator/pull/296
and it fixes or just workaround https://github.com/devfile/devworkspace-operator/issues/313

### DIFF
--- a/deploy/deployment/kubernetes/combined.yaml
+++ b/deploy/deployment/kubernetes/combined.yaml
@@ -16935,8 +16935,8 @@ spec:
         name: devworkspace-controller
         resources:
           limits:
-            cpu: 100m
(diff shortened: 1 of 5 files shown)


JPinkney (approved):

amisevsk (approved):
LGTM, though to be clear, the updated limits are just a guess :)

sleshchenko:
> LGTM, though to be clear, the updated limits are just a guess :)

It's just more stable than currently, but it seems to be a good compromise before requesting 1Gi for controller )

PR #315, opened 2021-03-11
Ask confirmation before pushing next image

sleshchenko:
### What does this PR do?
Someone pushed next tag occasionally from the PR, which broke all new installations which tried to use next image.
With make docker it's really easy to override occasionally next image which I also did a few times and had to quickly fix.
Now if you have locally set IMG with your own repo - everything will work as previously.
But when you try to push quay.io/devfile/devworkspace-controller:next, you'll need to confirm it.

I also thought about env var which would allow to avoid confirmation, but so far CI does not use make install.

It also helps to prevent another mistake, when I deployed https://github.com/che-incubator/devworkspace-che-operator I did not success since I have `IMG=sleshchenko/che-workspace-controller:local` in my bashrc, so, not to conflict, I renamed `IMG` to `DWO_IMG` to use unique name for that parameter.

no issue is created

### Is it tested? How?
```
unset IMG
make docker
...
 ---> Using cache
 ---> 28aface1225a
Successfully built 28aface1225a
Successfully tagged quay.io/devfile/devworkspace-controller:next
Are you sure we want to push quay.io/devfile/devworkspace-controller:next? [y/N] n
Makefile:286: recipe for target 'docker-push' failed
make: *** [docker-push] Error 1
```

### DIFF
--- a/.ci/openshift_e2e.sh
+++ b/.ci/openshift_e2e.sh
@@ -35,7 +35,7 @@ export CI_COMPONENT="devworkspace-operator"
 # DEVWORKSPACE_OPERATOR env var exposed by Openshift CI in e2e test pod. More info about how images are builded in Openshift CI: https://github.com/openshift/ci-tools/blob/master/TEMPLATES.md#parameters-available-to-templates
 # Dependencies environment are defined here: https://github.com/openshift/release/blob/master/ci-operator/config/devfile/devworkspace-operator/devfile-devworkspace-operator-master__v5.yaml#L36-L38
 
-export IMG=${DEVWORKSPACE_OPERATOR}
+export DWO_IMG=${DEVWORKSPACE_OPERATOR}
 
 # Pod created by openshift ci don't have user. Using this envs should avoid errors with git user.
 export GIT_COMMITTER_NAME="CI BOT"
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ MAKEFLAGS += --silent
 endif
 
 export NAMESPACE ?= devworkspace-controller
-export IMG ?= quay.io/devfile/devworkspace-controller:next
+export DWO_IMG ?= quay.io/devfile/devworkspace-controller:next
 export ROUTING_SUFFIX ?= 192.168.99.100.nip.io
 export PULL_POLICY ?= Always
 export DEFAULT_ROUTING ?= basic
@@ -61,7 +61,7 @@ endif
 VERSION ?= 0.0.1
 OPERATOR_SDK_VERSION = v1.1.0
 # Default bundle image tag
-BUNDLE_IMG ?= controller-bundle:$(VERSION)
+DWO_BUNDLE_IMG ?= controller-bundle:$(VERSION)
 # Options for 'bundle-build'
 ifneq ($(origin CHANNELS), undefined)
 BUNDLE_CHANNELS := --channels=$(CHANNELS)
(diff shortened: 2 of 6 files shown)


.ci/openshift_e2e.sh
```diff
@@ -35,7 +35,7 @@ export CI_COMPONENT="devworkspace-operator"
 # DEVWORKSPACE_OPERATOR env var exposed by Openshift CI in e2e test pod. More info about how images are builded in Openshift CI: https://github.com/openshift/ci-tools/blob/master/TEMPLATES.md#parameters-available-to-templates
 # Dependencies environment are defined here: https://github.com/openshift/release/blob/master/ci-operator/config/devfile/devworkspace-operator/devfile-devworkspace-operator-master__v5.yaml#L36-L38
 
-export IMG=${DEVWORKSPACE_OPERATOR}
+export DWO_IMG=${DEVWORKSPACE_OPERATOR}
```
sleshchenko, line 38: Technically it's controller not operator (since DWO is DevWorkspace Operator), but I did not come to good short name for Controller (DWO_CTRL_IMG, DWC???...)

So, in my opinion DWO_IMG is good enough and for operator we'll be able to use DWO_INDEX_IMG, DWO_BUNDLE_IMG

Makefile
```diff
@@ -327,8 +332,8 @@ _operator_sdk:
 		else \
 			SDK_VER=$$(operator-sdk version | cut -d , -f 1 | cut -d : -f 2 | cut -d \" -f 2) && \
 			if [ "$${SDK_VER}" != $(OPERATOR_SDK_VERSION) ]; then \
-				echo "WARN: operator-sdk $(OPERATOR_SDK_VERSION) is expected to be used for this target but $${SDK_VER} found"
-				echo "WARN: Please use the recommended operator-sdk if you face any issue"
+				echo "WARN: operator-sdk $(OPERATOR_SDK_VERSION) is expected to be used for this target but $${SDK_VER} found" \
```
sleshchenko, line 335: I just assume it should be done

Makefile
```diff
@@ -277,11 +277,16 @@ docker: _print_vars docker-build docker-push
...
 docker-build:
-	docker build . -t ${IMG} -f build/Dockerfile
+	docker build . -t ${DWO_IMG} -f build/Dockerfile
 
 ### docker-push: Push the controller image
 docker-push:
-	docker push ${IMG}
+ifneq ($(INITIATOR),CI)
+ifeq ($(DWO_IMG),quay.io/devfile/devworkspace-controller:next)
+	@echo -n "Are you sure we want to push $(DWO_IMG)? [y/N] " && read ans && [ $${ans:-N} = y ]
```
amisevsk, line 286: It's strange to me that this occurred in the first place -- I've never been able to push to quay.io/devfile/devworkspace-controller directly (AFAIK). Pushing to this repo should be restricted so that only CI can do it.

Makefile
```diff
@@ -344,7 +349,7 @@ help: Makefile
 	@sed -n 's/^### /    /p' $< | awk 'BEGIN { FS=":" } { printf "%-30s -%s\n", $$1, $$2 }'
 	@echo ''
 	@echo 'Supported environment variables:'
-	@echo '    IMG                        - Image used for controller'
+	@echo '    DWO_IMG                        - Image used for controller'
```
amisevsk, line 352: The joys of manual spacing :D 
```suggestion
	@echo '    DWO_IMG                    - Image used for controller'
```

amisevsk (approved):

Makefile
sleshchenko: > I've never been able to push to quay.io/devfile/devworkspace-controller directly (AFAIK).

Haha.
Controller team(You, Josh and I) are admin for that repository and quay.io does not provide an ability to maintain repo but not be able to push there.

I might revoke your permissions there and then you would need to ask me do the stuff, but at this point, CI is set up and I'm not sure which maintaining is expected for that repo.

But in any case this PR would help me to prevent occasional push, or I need separate accounts for administrations and personal testing stuff.

amisevsk: :+1: makes sense. I also just realized that I'm never logged in to quay.io, so that would explain why I can't just push there :smile:

JPinkney (approved):

PR #317, opened 2021-03-11
using coputil

tinakurian:
### What does this PR do?
This PR replaces some of the functions in finalize.go with the equivalent functions found in github.com/redhat-cop/operator-utils/pkg/util

N/A
### Is it tested? How?
e2e-tests and unit tests were run

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/onsi/ginkgo v1.12.1
 	github.com/onsi/gomega v1.10.1
 	github.com/openshift/api v0.0.0-20200205133042-34f0ec8dab87
+	github.com/redhat-cop/operator-utils v0.1.0
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/zap v1.16.0 // indirect
 	golang.org/x/lint v0.0.0-20191125180803-fdd1cda4f05f // indirect
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -19,22 +19,20 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/onsi/ginkgo v1.12.1
 	github.com/onsi/gomega v1.10.1
 	github.com/openshift/api v0.0.0-20200205133042-34f0ec8dab87
+	github.com/redhat-cop/operator-utils v0.0.0-20190827162636-51e6b0c32776
```
sleshchenko, line 14: is it version from 2019?
Maybe better to depends on released v1.1.2 (Jan 30) https://github.com/redhat-cop/operator-utils/releases/tag/v1.1.2 ?

sleshchenko (approved):
Thanks for cleaning that part :+1:

go.mod
tinakurian: Yeah, it seems it causes dependency issues, the furthest i can go is v0.1.0 so i've updated to that version - is this okay?

sleshchenko: that's OK, I just wondered if we can use the one that corresponds to our operator-sdk.

amisevsk (approved):

go.mod
amisevsk: The issue is controller-runtime -- we use `0.6.3`, the latest coputil uses `0.7.0`, and the latest version is `0.8.3`. I've never understood go dependency management well enough to figure out these issues though.

amisevsk: `0.7.0` features quite a few [breaking changes](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0), also

JPinkney (approved):

Issue #162, opened 2020-09-29
Links to external resources from a devworkspace


benoitf:
### Is your enhancement related to a problem? Please describe.
When we've a dev workspace, we can ask resources like with

```
$ kubectl get devworkspaces --all-namespaces
```

but runtime information is missing as well as the components from these objects.
That information is stored in other custom resources

### Describe the solution you'd like
It would be nice that from a DevWorkspace we could have pointers to the runtime info custom resources and component custom resources

There is like a pattern-rule with workspace-id suffix but would be more convenient to know the related source names from the DevWorkspace resource

### According to 2nd October call:
We could use labels for all workspace related objects with an identifier of the dev workspace.

It may be simpler to use workspace name but not workspaceId that we have since namespace/name fully identifies devworkspace CR instance.

Most of workspace-related objects are cluster-scoped (so namespace label seems to be redundant), but some of them could be cluster-scoped, like OpenShift Oauth client. Then to grab all workspace-related objects we need to have namespace label as well.

Example of label: `io.devfile.dev_workspace.name` + `io.devfile.dev_workspace.namespace`

Che-Theia knows the identifier of the dev workspace and then will be able to do the queries.

\* name/namespace could be combined in one label if it makes sense + `namespace format/name format` satisfies label value format, like we won't exceed the maximum length.
Having different labels for name and namespace could be more straightforward since namespace is optional when you do a namespace-scoped query (which should 100% case of Che Theia).

amisevsk:
The additional components are internal on purpose; runtime information is stored in the configmap defined for each workspace for now, and mounted in the `che-rest-apis` container.

benoitf:
we could use labels with identifier of the dev workspace 
example of label:`io.devfile.devworkspace.name`

Che-Theia know the identifier of the dev workspace and then will be able to do the queries

need to check with cluster admin roles/namespaces

benoitf:
I've moved comment to the description

sleshchenko:
Closing it, what may be considered as left: link to flattened devfile from devworkspace status. But we agreed we won't add it until we see necessity and a dedicated issue should exist on devfile/api side.

PR #318, opened 2021-03-11
Mount flattened devfile in devworkspace containers

amisevsk:
### What does this PR do?
Mount files to each container in a devworkspace, via a configmap:
```
/devworkspace-metadata/flattened.devworkspace.yaml - the flattened DevWorkspaceTemplateSpec used to create the actual deployment
/devworkspace-metadata/original.devworkspace.yaml - the original DevWorkspaceTemplateSpec that exists on the cluster
```

Additionally, this PR annotates the flattened DevWorkspace with some additional info:
- Components/Commands/Projects added by a plugin are annotated with `controller.devfile.io/imported-by=<plugin-id>`, where `<plugin-id>` is the ID of the plugin component that resolved to that Component/etc.
- Endpoints on Container components are annotated with `controller.devfile.io/endpoint-url=<url>` to more easily match created routes/ingresses on the cluster with devfile endpoints.

I don't know that we created an issue for this, but the closest is probably https://github.com/devfile/devworkspace-operator/issues/162

### Is it tested? How?
Tested as usual. Once a DevWorkspace is started, you can issue commands like `cat $DEVWORKSPACE_METADATA/flattened.devworkspace.yaml` to check the mounted files.

cc: @benoitf I can't add you as a reviewer here for some reason but your comments are welcome.

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/devfile/devworkspace-operator/controllers/workspace/provision"
 	"github.com/devfile/devworkspace-operator/pkg/common"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
+	"github.com/devfile/devworkspace-operator/pkg/library/annotate"
 	containerlib "github.com/devfile/devworkspace-operator/pkg/library/container"
 	"github.com/devfile/devworkspace-operator/pkg/library/flatten"
 	registry "github.com/devfile/devworkspace-operator/pkg/library/flatten/internal_registry"
@@ -199,23 +200,6 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{}, nil
 	}
 	shimlib.FillDefaultEnvVars(devfilePodAdditions, *workspace)
-	err = metadata.ProvisionWorkspaceMetadata(devfilePodAdditions, clusterWorkspace, workspace, &clusterAPI)
-	if err != nil {
-		switch provisionErr := err.(type) {
-		case *metadata.NotReadyError:
-			reqLogger.Info(provisionErr.Message)
-			return reconcile.Result{Requeue: true, RequeueAfter: provisionErr.RequeueAfter}, nil
-		case *metadata.ProvisioningError:
-			reqLogger.Info(fmt.Sprintf("DevWorkspace start failed: %s", provisionErr))
-			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
-			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning storage: %s", provisionErr)
-			return reconcile.Result{}, nil
-		default:
-			return reconcile.Result{}, provisionErr
-		}
-	}
-
-	allPodAdditions := []controllerv1alpha1.PodAdditions{*devfilePodAdditions}
 
 	reconcileStatus.Conditions[devworkspace.WorkspaceComponentsReady] = ""
 	timing.SetTime(timingInfo, timing.ComponentsReady)
@@ -258,7 +242,27 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	annotate.AddURLAttributesToEndpoints(&workspace.Spec.Template, routingStatus.ExposedEndpoints)
+
+	// Step three: provision a configmap on the cluster to mount the flattened devfile in deployment containers
+	err = metadata.ProvisionWorkspaceMetadata(devfilePodAdditions, clusterWorkspace, workspace, &clusterAPI)
+	if err != nil {
+		switch provisionErr := err.(type) {
+		case *metadata.NotReadyError:
+			reqLogger.Info(provisionErr.Message)
+			return reconcile.Result{Requeue: true, RequeueAfter: provisionErr.RequeueAfter}, nil
+		case *metadata.ProvisioningError:
+			reqLogger.Info(fmt.Sprintf("DevWorkspace start failed: %s", provisionErr))
+			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
+			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning metadata configmap: %s", provisionErr)
+			return reconcile.Result{}, nil
+		default:
+			return reconcile.Result{}, provisionErr
+		}
+	}
+
 	// Step four: Collect all workspace deployment contributions
+	allPodAdditions := []controllerv1alpha1.PodAdditions{*devfilePodAdditions}
 	routingPodAdditions := routingStatus.PodAdditions
 	if routingPodAdditions != nil {
 		allPodAdditions = append(allPodAdditions, *routingPodAdditions)
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -28,6 +28,7 @@ import (
 	registry "github.com/devfile/devworkspace-operator/pkg/library/flatten/internal_registry"
 	shimlib "github.com/devfile/devworkspace-operator/pkg/library/shim"
 	storagelib "github.com/devfile/devworkspace-operator/pkg/library/storage"
+	"github.com/devfile/devworkspace-operator/pkg/provision/metadata"
 	"github.com/devfile/devworkspace-operator/pkg/timing"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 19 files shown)


benoitf:
hello, just a minor thing but `original.devworkspace.yaml` looks like a package/namespace naming and not a 'user friendly name' like `devworkspace-original.yaml`

maybe you can remove anyway the devworkspace name from it `/devworkspace-metadata/flattened.yaml` or use a subfolder 


On a more 'high level' feedback, I would not need to fetch workspace-routings/workspace-pods and dev-workspace template information

is there 1-1 relationship between a component name and the container name ? (for component/container)
if not, probably there should be an annotation with the container name like `controller.devfile.io/container-name=<container-name>`

and can we have information like workspace-id as annotation as well ?

sleshchenko:
> cc: @benoitf I can't add you as a reviewer here for some reason but your comments are welcome.

Florent is not devfile member yet :man_shrugging: 
Mario or David should be able to add him.

benoitf:
@sleshchenko 🦸

amisevsk:
> is there 1-1 relationship between a component name and the container name ? (for component/container)
if not, probably there should be an annotation with the container name like `controller.devfile.io/container-name=<container-name>`

In short -- yes. For many fields, the devfile/api spec tries to be compatible with the analogous k8s field to allow just using the same value. 

> hello, just a minor thing but `original.devworkspace.yaml` looks like a package/namespace naming and not a 'user friendly name' like `devworkspace-original.yaml`
>
>maybe you can remove anyway the devworkspace name from it `/devworkspace-metadata/flattened.yaml` or use a subfolder 

I can definitely change it -- my assumption was that the files were to be used by tooling and not directly user-facing (since they edit the devfile, not the DevWorkspace). Another option I considered was setting environment variables (e.g. `RESOLVED_DEVWORKSPACE` or something) to point at the mounted filepaths and avoid the ambiguity.

benoitf:
@amisevsk env variable make it also convenient for 'old school users' wanting to see what is being brought listing env variables
new users look at mounted path as well
 (for che-theia usecase it can be anything)

but definitely if we have a all-in-one file with everything being annotated, it's very helpful for clients (no need to maintain several structs with routing or status, pods informations) we just filter out the content of this file to grab everything we need to know about the current workspace.

controllers/workspace/devworkspace_controller.go
```diff
@@ -240,7 +242,27 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
...
+	err = metadata.ProvisionWorkspaceMetadata(devfilePodAdditions, clusterWorkspace, workspace, &clusterAPI)
+	if err != nil {
+		switch provisionErr := err.(type) {
+		case *metadata.NotReadyError:
+			reqLogger.Info(provisionErr.Message)
+			return reconcile.Result{Requeue: true, RequeueAfter: provisionErr.RequeueAfter}, nil
+		case *metadata.ProvisioningError:
+			reqLogger.Info(fmt.Sprintf("DevWorkspace start failed: %s", provisionErr))
+			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
+			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning storage: %s", provisionErr)
```
sleshchenko, line 257: it's not storage, right? )

pkg/provision/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	flattenedYaml, err := yaml.Marshal(flattened.Spec.Template)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal flattened DevWorkspace yaml: %w", err)
+	}
+
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-metadata", original.Status.WorkspaceId), // TODO
```
sleshchenko, line 116: What does this TODO mean and when you plan to address it?

sleshchenko (approved):

sleshchenko (changes requested):
I assume it's a side effect but my DW is modified with flattened devfile.
I got it with Che integration. Rechecking and providing steps to reproduce

controllers/workspace/devworkspace_controller.go
```diff
@@ -240,7 +242,27 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	annotate.AddURLAttributesToEndpoints(&workspace.Spec.Template, routingStatus.ExposedEndpoints)
```
sleshchenko, line 245: it seems to have side effects:

<details>

<summary>Create DW with local plugin</summary>

cat EOF >> kubectl apply -f

---
kind: DevWorkspace
metadata:
  name: theia
  namespace: user1-che
spec:
  started: true
  template:
    components:
    - name: nodejs
      container:
        image: quay.io/eclipse/che-nodejs10-ubi:nightly
        memoryLimit: 512Mi
        endpoints:
        - name: nodejs
          protocol: http
          targetPort: 3000
        mountSources: true
    - name: theia-ide-workspace05c266f85ef741bd
      plugin:
        kubernetes:
          name: theia-ide-local-plugin
          namespace: user1-che
    commands:
    - id: download-dependencies
      exec:
        component: nodejs
        commandLine: npm install
        workingDir: "${PROJECTS_ROOT}/project/app"
    - id: run-the-app
      exec:
        component: nodejs
        commandLine: nodemon app.js
        workingDir: "${PROJECTS_ROOT}/project/app"
    - id: run-the-app-with-debugging-enabled
      exec:
        component: nodejs
        commandLine: nodemon --inspect app.js
        workingDir: "${PROJECTS_ROOT}/project/app"
    - id: stop-the-app
      exec:
        component: nodejs
        commandLine: 'node_server_pids=$(pgrep -fx ''.*nodemon (--inspect )?app.js''
          | tr "\\n" " ") && echo "Stopping node server with PIDs: ${node_server_pids}"
          && kill -15 ${node_server_pids} &>/dev/null && echo ''Done.'''
---
kind: DevWorkspaceTemplate
metadata:
  name: theia-ide-local-plugin
  namespace: user1-che
spec:
  commands:
  - id: init-container-command
    apply:
      component: remote-runtime-injector
  events:
    preStart:
    - init-container-command
  components:
  - name: remote-endpoint
    volume:
      ephemeral: true
  - name: remote-runtime-injector
    container:
      image: quay.io/eclipse/che-theia-endpoint-runtime-binary:next
      env:
      - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
        value: "/remote-endpoint/plugin-remote-endpoint"
      - name: REMOTE_ENDPOINT_VOLUME_NAME
        value: remote-endpoint
      volumeMounts:
      - name: remote-endpoint
        path: "/remote-endpoint"

EOF

</details>
once it's started, check that `kubectl get dw theia` has `remote-runtime-injector` component in it's spec.template.

sleshchenko: also, IDE url is not propagate to status for some reason, not sure if it's related to this PR. Checking on main branch 
```
theia   workspaced091e53ca11849ad   Running   Workspace is running
```

pkg/provision/metadata/metadata.go
amisevsk: I knew I had left one in there somewhere. I'll fix.

controllers/workspace/devworkspace_controller.go
amisevsk: https://github.com/devfile/devworkspace-operator/pull/326

sleshchenko: should be fixed after you rebase

controllers/workspace/devworkspace_controller.go
amisevsk: Yup -- copy paste :)

Fixed in d1b4a9a1

pkg/provision/metadata/metadata.go
amisevsk: Done -- moved name generation to naming.go

sleshchenko (approved):
LGTM and seems to work as expected.

The only comment regards:

> Components/Commands/Projects added by a plugin are annotated with controller.devfile.io/imported-by=<plugin-id>, where <plugin-id> is the ID of the plugin component that resolved to that Component/etc.

I've tested it with k8s component which references DWT near DW.
In the DW, it's:

```
      - name: theia-ide-workspace05c266f85ef741bd
        plugin:
          kubernetes:
            name: theia-ide-local-plugin
            namespace: user1-che
```

and in flattened devfile it's:
```
- attributes:
    controller.devfile.io/imported-by: theia-ide-workspace05c266f85ef741bd
```

So, then I assume it's just a typo and it's always `controller.devfile.io/imported-by=<plugin-name>`.

We do support nested plugins, right?
If so, then: like DW -> 1DWT -> 2DWT -> 3DWT. In such case the component name from DW will be filled in, but not like plugin name from 2DWT?

amisevsk:
> So, then I assume it's just a typo and it's always controller.devfile.io/imported-by=<plugin-name>.

Yeah, this is unclear language on my part -- I mean the id as in id-of-the-devfile-component (should have said name or key, as you point out). Otherwise it's hard to define what a reference looks like, since it could be URI, ID, k8s ref, etc.

> We do support nested plugins, right?
> If so, then: like DW -> 1DWT -> 2DWT -> 3DWT. In such case the component name from DW will be filled in, but not like plugin name from 2DWT?

Yes, my idea was to annotate based on the visible plugin (so the original component name should be used). Otherwise, it's hard to actually use the annotation, since the flattened devfile has no reference to intermediate steps. I'll add a test case to make this clearer.

PR #320, opened 2021-03-15
Print changes files on PR check

sleshchenko:
### What does this PR do?
This PR adds printing of made changes on PR check, since sometimes the cause may be in tools versions difference which are used and it's tricky to figure out the cause. Like https://github.com/devfile/devworkspace-operator/pull/320/checks?check_run_id=2111552072
where changes are format lines.

After I did `make generate_default_template` I had nothing locally, to fix it I removed kustomize and let make install a new for me.

P.S. I love how kustomize installed with go get reports its version:
```
{Version:unknown GitCommit:$Format:%H$ BuildDate:1970-01-01T00:00:00Z GoOs:linux GoArch:amd64}
```
:heart_eyes: 

No issue is created, fixes PR check on all PRs.

### Is it tested? How?
Just check that PR checks passed.


JPinkney (approved):

sleshchenko:
pffff!!!! PR check is still failing but now it has less conflicts. It's due the fields order

sleshchenko:
I seems to got the right generate deployment locally after I install:
```
curl -s "https://raw.githubusercontent.com/\
kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
```

sleshchenko:
I tried to print the used version in Makefile but then it cames to something more difficult that just execute a command in Makefile, I feel to be a child :-D

sleshchenko:
Github actions has:
```
GitHub actions has /usr/local/bin/kustomize @ {Version:kustomize/v4.0.5 GitCommit:9e8e7a7fe99ec9fbf801463e8607928322fc5245 BuildDate:2021-03-08T20:53:03Z GoOs:linux GoArch:amd64}
```
which is from ubuntu-latest

amisevsk (commented):
We should figure out what's going on with kustomize. I tested three different versions and got differently formatted results for generate_default_deployment (none of which match this PR :upside_down_face:)

- Tested v4.0.4 (recent release) -- does not match
- Tested v3.5.4 downloaded from GitHub -- does not match
- Deleted kustomize and let makefile install -- does not match (but does match 3.5.4 at least)

sleshchenko:
> v4.0.4 (recent release)

Please try with latest but not recent ) github action has 4.0.5

amisevsk:
> Please try with latest but not recent ) github action has 4.0.5

... 4.0.5 works :exploding_head:

amisevsk (approved):

Issue #321, opened 2021-03-15
Finalize workspace -> devworkspace renaming


sleshchenko:
We renamed workspace CR to devworkspace CR but there are places which still use workspace:
- workspace_id, name labels;
- .status.workspaceId;
- devworkspace CR docs references to workspace
- maybe even more.

sleshchenko:
The dependents components which needs to be updated:
- Devfile API
- DevWorkspace Client Library
- Controller itself
- Che Theia
- Che machine exec
- DevWorkspace Che Operator (probably Che operator at time we'll provide changes)

sleshchenko:
Postponing to next milestone since we don't have enough time to do it before 0.2.0, and it should be pretty safe to postpone and finish without a rush.

sleshchenko:
All PRs I created are merged now.

PR #322, opened 2021-03-15
Finalize workspace -> devWorkspace renaming

sleshchenko:
### What does this PR do?
This PR is supposed to finalize workspace -> devWorkspace renaming but it does not do it fully.

It's for https://github.com/devfile/devworkspace-operator/issues/321

Che Machine exec PR https://github.com/eclipse-che/che-machine-exec/pull/138

### Is it tested? How?
not yet.

### DIFF
--- a/apis/controller/v1alpha1/devworkspacerouting_types.go
+++ b/apis/controller/v1alpha1/devworkspacerouting_types.go
@@ -13,7 +13,7 @@
 package v1alpha1
 
 import (
-	devworkspace "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	devfileAttr "github.com/devfile/api/v2/pkg/attributes"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -21,15 +21,15 @@ import (
 // DevWorkspaceRoutingSpec defines the desired state of DevWorkspaceRouting
 // +k8s:openapi-gen=true
 type DevWorkspaceRoutingSpec struct {
-	// WorkspaceId for the workspace being routed
-	WorkspaceId string `json:"workspaceId"`
-	// Class of the routing: this drives which Workspace Routing controller will manage this routing
+	// Id for the DevWorkspace being routed
+	DevWorkspaceId string `json:"devworkspaceId"`
+	// Class of the routing: this drives which DevWorkspaceRouting controller will manage this routing
 	RoutingClass DevWorkspaceRoutingClass `json:"routingClass,omitempty"`
 	// Routing suffix for cluster
 	RoutingSuffix string `json:"routingSuffix"`
 	// Machines to endpoints map
 	Endpoints map[string]EndpointList `json:"endpoints"`
-	// Selector that should be used by created services to point to the workspace Pod
+	// Selector that should be used by created services to point to the devworkspace Pod
 	PodSelector map[string]string `json:"podSelector"`
 }
 
@@ -46,7 +46,7 @@ const (
 // DevWorkspaceRoutingStatus defines the observed state of DevWorkspaceRouting
 // +k8s:openapi-gen=true
 type DevWorkspaceRoutingStatus struct {
-	// Additions to main workspace deployment
+	// Additions to main devworkspace deployment
 	PodAdditions *PodAdditions `json:"podAdditions,omitempty"`
 	// Machine name to exposed endpoint map
 	ExposedEndpoints map[string]ExposedEndpointList `json:"exposedEndpoints,omitempty"`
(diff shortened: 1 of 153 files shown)


pkg/constants/constants.go
```diff
@@ -35,30 +35,30 @@ const (
 	SidecarDefaultMemoryLimit = "128M"
 	PVCStorageSize            = "1Gi"
 
-	// WorkspaceIDLabel is label key to store workspace identifier
-	WorkspaceIDLabel = "controller.devfile.io/workspace_id"
+	// DevWorkspaceIDLabel is label key to store workspace identifier
+	DevWorkspaceIDLabel = "controller.devfile.io/devworkspace_id"
```
benoitf, line 39: I think `id` would be enough

pkg/constants/constants.go
```diff
@@ -35,30 +35,30 @@ const (
...
+
+	// DevWorkspaceNameLabel is label key to store workspace name
+	DevWorkspaceNameLabel = "controller.devfile.io/devworkspace_name"
```
benoitf, line 42: only `name` : `controller.devfile.io/name` ?

pkg/library/flatten/testdata/k8s-ref/already-flattened.yaml
```diff
@@ -19,7 +19,7 @@ input:
           command: ["/go/bin/che-machine-exec",
                     "--authenticated-user-id", "$(DEVWORKSPACE_CREATOR)",
                     "--idle-timeout", "$(DEVWORKSPACE_IDLE_TIMEOUT)",
-                    "--pod-selector", "controller.devfile.io/workspace_id=$(DEVWORKSPACE_ID)",
+                    "--pod-selector", "controller.devfile.io/devworkspace_id=$(DEVWORKSPACE_ID)",
```
benoitf, line 22: `controller.devfile.io/id=`

pkg/constants/constants.go
sleshchenko: but this label is propagate to all devworkspace related stuff, pod, service, route.... It may be not clear enough to see `controller.devfile.io/name` label.

pkg/constants/constants.go
sleshchenko: it's the time to make this label shorter if we wish, but I don't think we win anything with such change, just make things less clearer.
It would be clearer if it would be `devworkspace.devfile.io/name` but it's `controller.devfile.io/` group, and just name is not clear IMHO

apis/controller/v1alpha1/devworkspacerouting_types.go
```diff
@@ -13,23 +13,23 @@
...
 )
 
 // DevWorkspaceRoutingSpec defines the desired state of DevWorkspaceRouting
 // +k8s:openapi-gen=true
 type DevWorkspaceRoutingSpec struct {
-	// WorkspaceId for the workspace being routed
-	WorkspaceId string `json:"workspaceId"`
-	// Class of the routing: this drives which Workspace Routing controller will manage this routing
+	// Id for the DevWorkspace being routed
+	DevWorkspaceId string `json:"devWorkspaceId"`
```
sleshchenko, line 25: devworkspaceId vs devWorkspaceId

sleshchenko: Agreed on DevWorkspaceId as go field, but json version is devworkspaceId

pkg/constants/constants.go
sleshchenko: duplicated

pkg/library/flatten/testdata/k8s-ref/already-flattened.yaml
sleshchenko: duplicated

amisevsk (approved):
LGTM. Tested updating from main to this PR (manually) and saw no issues except from some strangeness that I think was my cluster rather than the operator.

sleshchenko:
Tested failed because I forget to rename `workspace_name` to `devworkspace_name` label.
Now it should be fine. Ran e2e against RHPDS cluster.

sleshchenko:
Found a bug. Deployment must be recreated if selector is changed.
Otherwise workspace won't start due:

```
2021-04-08T11:45:29.354Z        ERROR   controller      Reconciler error        {"reconcilerGroup": "workspace.devfile.io", "reconcilerKind": "DevWorkspace", "controller": "devworkspace", "name": "spring-petclinic", "namespace": "opentlc-mgr-che", "error": "Deployment.apps \"workspace8313386c181c4e11\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\"controller.devfile.io/devworkspace_id\":\"workspace8313386c181c4e11\", \"controller.devfile.io/devworkspace_name\":\"spring-petclinic\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable"}
```

sleshchenko:
I tested all changed components together: DWO, DWCO, Che Theia, Che Machine Exec, Che Dashboard and it seems to work

sleshchenko:
I tested and it works, merging but @amisevsk @JPinkney you have any comments to the last commits - let me know and I'll fix it.

Issue #294, opened 2021-03-03
makefile should check if kustomize is installed


benoitf:
After pulling recent changes and performing `$ make install`
I got

```
deploy/generate-deployment.sh: line 128: kustomize: command not found
make: *** [generate_deployment] Error 127
```

it should check if the tool exists and for example on macos propose to do: `brew install kustomize`

amisevsk:
It should already do this, from the `_kustomize` rule:
```
_kustomize:
ifeq (, $(shell which kustomize))
	@{ \
	set -e ;\
	KUSTOMIZE_GEN_TMP_DIR=$$(mktemp -d) ;\
	cd $$KUSTOMIZE_GEN_TMP_DIR ;\
	go mod init tmp ;\
	GOFLAGS="" go get sigs.k8s.io/kustomize/kustomize/v3@v3.5.4 ;\
	rm -rf $$KUSTOMIZE_GEN_TMP_DIR ;\
	}
KUSTOMIZE=$(GOBIN)/kustomize
else
KUSTOMIZE=$(shell which kustomize)
endif
```
I suspect the issue may be in how `$PATH` is configured on macOS -- it should include `$GOPATH/bin`

amisevsk:
Fixed by https://github.com/devfile/devworkspace-operator/pull/324

PR #324, opened 2021-03-15
Download required version of kustomize to bin/kustomize/

amisevsk:
### What does this PR do?
Make the `_kustomize` rule download kustomize to `.kustomize/kustomize`, and use this version of kustomize for all Makefile-related calls. This is required since different versions of kustomize output differently formatted yaml templates, resulting in spurious changes to files in deploy/.

Fixes https://github.com/devfile/devworkspace-operator/issues/294

### Is it tested? How?
Tested three cases: 
1. Nothing in `.kustomize` or folder does not exist -> download kustomize v4.0.5 there
2. Wrong version kustomize in `.kustomize/kustomize` -> delete current and redownload
3. Kustomize present and correct version -> do nothing

cc: @benoitf Could you also test these changes to make sure it works as expect on macOS?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,13 +25,11 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.6
-      - name: Setup Kustomize & yq
+      - name: Setup yq
         run: |
           python -m pip install --upgrade pip
           pip install yq
 
-          cd /usr/bin
-          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v4.0.5/hack/install_kustomize.sh" | bash
           cd ${GITHUB_WORKSPACE}
       - name: Build image, generate deployments and push git tag
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
           pip install yq
 
           cd /usr/bin
-          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v4.0.4/hack/install_kustomize.sh"  | bash
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v4.0.5/hack/install_kustomize.sh" | bash
           cd ${GITHUB_WORKSPACE}
       - name: Build image, generate deployments and push git tag
         run: |
@@ -39,4 +39,5 @@ jobs:
           git config --global user.email "mkuznets@redhat.com"
 
           export GITHUB_TOKEN=${{ secrets.CHE_INCUBATOR_BOT_GITHUB_TOKEN }}
+          export KUSTOMIZE=/usr/bin/kustomize
           ./make-release.sh --version ${{ github.event.inputs.version}}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,5 +37,4 @@ jobs:
           git config --global user.email "mkuznets@redhat.com"
 
           export GITHUB_TOKEN=${{ secrets.CHE_INCUBATOR_BOT_GITHUB_TOKEN }}
-          export KUSTOMIZE=/usr/bin/kustomize
           ./make-release.sh --version ${{ github.event.inputs.version}}
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,6 @@ DEVWORKSPACE_CTRL_SA=devworkspace-controller-serviceaccount
 INTERNAL_TMP_DIR=/tmp/devworkspace-controller
 BUMPED_KUBECONFIG=$(INTERNAL_TMP_DIR)/kubeconfig
 RELATED_IMAGES_FILE=$(INTERNAL_TMP_DIR)/environment
-KUSTOMIZE_VER=4.0.5
-export KUSTOMIZE_DIR=./bin/kustomize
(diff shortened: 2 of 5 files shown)


sleshchenko (approved):
LGTM

I'm just not really happy with having many folders in the projects root. 
So, I would suggest we download everything we need to one folder, like `.bin`.
If it makes sense, it can then has subfolders.

[testbin](https://github.com/devfile/devworkspace-operator/blob/main/Makefile#L126) may go there as well

amisevsk:
Fixed an issue where cached kustomize isn't used in generate_deployment.sh, and moved binaries into `bin/` instead of separate folders.

JPinkney (approved):
Ah, the beauty of working with kustomize :upside_down_face:

benoitf (approved):
```
$  make install
...
{Version:kustomize/v4.0.5 GitCommit:9e8e7a7fe99ec9fbf801463e8607928322fc5245 BuildDate:2021-03-08T20:53:03Z GoOs:darwin GoArch:amd64}
kustomize installed to devworkspace-operator/bin/kustomize/kustomize
Checking for devfile/api CRDs
CRDs from devfile/api are already initialized
Generating config for Kubernetes
File saved to devworkspace-operator/deploy/current/kubernetes/combined.yaml
...
```

deploy/generate-deployment.sh
```diff
@@ -105,7 +105,16 @@ for var in "${required_vars[@]}"; do
   fi
 done
 
-required_bin=(kustomize envsubst csplit yq)
+if [ -z ${KUSTOMIZE} ]; then
+  echo "Required env var KUSTOMIZE not set. Set KUSTOMIZE to point to a kustomize v4.0.5 binary"
```
sleshchenko, line 109: Seems it's going to break release script.

amisevsk: :disappointed: I should've known!

I'll look into it. I assume we'll need to do something similar during release to avoid the same versioning issue we're facing everywhere else.

sleshchenko: but don't duplicate. Let's move kustomize install into script or make release uses makefile.

sleshchenko (approved):

.github/workflows/release.yml
```diff
@@ -31,12 +31,13 @@ jobs:
           pip install yq
 
           cd /usr/bin
-          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v4.0.4/hack/install_kustomize.sh"  | bash
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v4.0.5/hack/install_kustomize.sh" | bash
```
sleshchenko, line 34: well, I would let makefile install the kustomize. Otherwise, released deployment can be different from what we have in main branch after we upgrade kustomize in one place

Makefile
```diff
@@ -310,19 +313,16 @@ install_cert_manager:
 	kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v1.0.4/cert-manager.yaml
 
 _kustomize:
```
sleshchenko, line 315: Since it's only generate-deployment.sh who needs kustomize https://github.com/devfile/devworkspace-operator/blob/main/Makefile#L185, maybe it makes sense to move installing there ?

.github/workflows/release.yml
amisevsk: True, but currently make-release.sh doesn't depend on make at all and calls the script directly :/ 

I wonder if make is even installed in the ubuntu runner

Makefile
amisevsk: Good idea -- done.

.github/workflows/release.yml
amisevsk: Fixed by moving download of kustomize into generate-deployment.sh

.github/workflows/release.yml
```diff
@@ -25,18 +25,17 @@ jobs:
...
-          cd /usr/bin
-          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v4.0.4/hack/install_kustomize.sh"  | bash
           cd ${GITHUB_WORKSPACE}
       - name: Build image, generate deployments and push git tag
         run: |
           git config --global user.name "Mykhailo Kuznietsov"
           git config --global user.email "mkuznets@redhat.com"
 
           export GITHUB_TOKEN=${{ secrets.CHE_INCUBATOR_BOT_GITHUB_TOKEN }}
+          export KUSTOMIZE=/usr/bin/kustomize
```
sleshchenko, line 40:
```suggestion
```

README.md
```diff
@@ -29,12 +29,13 @@ metadata:
 ## Prerequisites
 - go
 - git
-- kustomize.io
 - sed
 - jq
 - yq (python-yq from https://github.com/kislyuk/yq#installation, other distributions may not work)
 - docker
 
+Note: kustomize `v4.0.5` is required for most tasks. It is downloaded automatically to the `.kustomize` folder in this repo when required. This downloaded version is used regardless of whether or not kustomize is already installed on the system.
```
sleshchenko, line 37: Not really for most but it's not really matter.

sleshchenko (approved):

README.md
amisevsk: By number, maybe not, but it's needed for `make install` and `make uninstall` :)

tinakurian (approved):

PR #325, opened 2021-03-16
Sync makefile and default deployment CRDs versions

sleshchenko:
### What does this PR do?
It's just a quick fix to update default deployment to the desired, but we would need to rework how we define devfile CRDs to avoid such desync in future.

No issue is created but it fixes:

```
error: error validating "samples/plugins/theia-next.yaml": error validating data: ValidationError(DevWorkspaceTemplate.spec.components[1].volume): unknown field "ephemeral" in io.devfile.workspace.v1alpha2.DevWorkspaceTemplate.spec.components.volume; if you choose to ignore these errors, turn validation off with --validate=false
```
when you try to create devworkspace with ephemeral volume used on the default deployment.

### DIFF
--- a/deploy/deployment/kubernetes/combined.yaml
+++ b/deploy/deployment/kubernetes/combined.yaml
@@ -7074,6 +7074,11 @@ spec:
                                     description: Configuration overriding for a Volume
                                       component in a plugin
                                     properties:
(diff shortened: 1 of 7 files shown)


amisevsk (approved):
:facepalm:

PR #326, opened 2021-03-16
Fix accidentally applying flattened DW to cluster in reconcile

amisevsk:
### What does this PR do?
We're accidentally applying the flattened DevWorkspace content to the cluster, overwriting the original file. This PR is a minor fixup to avoid that issue, but we should improve this in the future to better delineate the 'working copy' flattened DevWorkspace from the DevWorkspace that exists on the cluster.

### Is it tested? How?
1. `kubectl apply -f samples/with-k8s-ref/theia-next.yaml`
2. Wait for workspace to be waiting for deployment and check that resource on cluster is unchanged (e.g. `kubectl apply -f samples/with-k8s-ref/theia-next.yaml` a second time should report `devworkspace.workspace.devfile.io/theia unchanged`


sleshchenko (approved):
Tested and works fine.

There is an issue with not propagate IDE url, but it's not really related.

PR #329, opened 2021-03-17
Introduce latest web terminal plugin version

sleshchenko:
### What does this PR do?
It's possible way to fix https://issues.redhat.com/browse/WTO-65, I'm a bit tired to see 4.5.0 on the all  OpenShift versions. But we can't simply remove 4.5.0 versions not to break existing instances.

It will be needed to cherry-pick in to v1alphax branch if we decide it's a good idea.

https://issues.redhat.com/browse/WTO-65

### Is it tested? How?
tested that web terminal started.


amisevsk (approved):
:+1:

JPinkney (approved):
Makes sense to me

sleshchenko:
Adding volume to custom terminal tooling in moved to https://github.com/devfile/devworkspace-operator/pull/360

PR #330, opened 2021-03-17
Permit changes to labels/annotations in restricted-access DevWorkspaces

amisevsk:
### What does this PR do?
To support automatically-set labels/annotations on resources, allow metadata changes as long as they do not change creator ID and restricted-access.

There's a bit of code duplication, since `metav1.ObjectMeta` and `metav1.Object` aren't the same thing :shrug: 

https://issues.redhat.com/browse/WTO-75

### Is it tested? How?
1. Create web-terminal as user X
2. Wait until pod is running
3. Log in as user Y
4. Add annotation `test="true"` to devworkspace -> change succeeds
5. Try to change restricted-access or creator-id annotation/label -> change blocked by webhook.

### DIFF
--- a/webhook/workspace/handler/immutable.go
+++ b/webhook/workspace/handler/immutable.go
@@ -123,20 +123,14 @@ func (h *WebhookHandler) handleImmutableObj(oldObj, newObj runtime.Object, uid s
 	if uid == h.ControllerUID {
 		return true, ""
 	}
-	if changePermitted(oldObj, newObj) {
-		return true, ""
-	}
-	return false, "object is owned by workspace and cannot be updated."
+	return changePermitted(oldObj, newObj)
 }
 
 func (h *WebhookHandler) handleImmutableRoute(oldObj, newObj runtime.Object, username string) (allowed bool, msg string) {
 	if username == h.ControllerSAName {
 		return true, ""
 	}
-	if changePermitted(oldObj, newObj) {
-		return true, ""
-	}
-	return false, "object is owned by workspace and cannot be updated."
+	return changePermitted(oldObj, newObj)
 }
 
 func (h *WebhookHandler) handleImmutableService(oldObj, newObj runtime.Object, uid, username string) (allowed bool, msg string) {
@@ -145,10 +139,7 @@ func (h *WebhookHandler) handleImmutableService(oldObj, newObj runtime.Object, u
 	if uid == h.ControllerUID || username == serviceCAUsername {
 		return true, ""
 	}
-	if changePermitted(oldObj, newObj) {
-		return true, ""
-	}
-	return false, "object is owned by workspace and cannot be updated."
+	return changePermitted(oldObj, newObj)
 }
 
 func (h *WebhookHandler) checkRestrictedAccessAnnotation(req admission.Request) (restrictedAccess bool, err error) {
@@ -159,14 +150,14 @@ func (h *WebhookHandler) checkRestrictedAccessAnnotation(req admission.Request) 
 	} else {
 		err = h.Decoder.DecodeRaw(req.Object, obj)
 	}
+	if err != nil {
+		return false, err
+	}
 	annotations := obj.GetAnnotations()
 	return annotations[constants.WorkspaceRestrictedAccessAnnotation] == "true", nil
 }
 
 func checkRestrictedWorkspaceMetadata(oldMeta, newMeta *metav1.ObjectMeta) (allowed bool, msg string) {
-	if oldMeta.Labels[constants.WorkspaceCreatorLabel] != newMeta.Labels[constants.WorkspaceCreatorLabel] {
-		return false, "cannot update controller.devfile.io/creator label"
-	}
 	if oldMeta.Annotations[constants.WorkspaceRestrictedAccessAnnotation] == "true" &&
 		newMeta.Annotations[constants.WorkspaceRestrictedAccessAnnotation] != "true" {
 		return false, "cannot disable restricted-access once it is set"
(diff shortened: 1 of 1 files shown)


webhook/workspace/handler/immutable.go
```diff
@@ -177,7 +187,14 @@ func changePermitted(oldObj, newObj runtime.Object) bool {
...
-	newMeta.SetDeletionTimestamp(oldMeta.GetDeletionTimestamp())
-	return reflect.DeepEqual(oldMeta, newMeta)
+	oldLabels, newLabels := oldMeta.GetLabels(), newMeta.GetLabels()
+	oldAnnotations, newAnnotations := oldMeta.GetAnnotations(), newMeta.GetAnnotations()
+	if oldLabels[constants.WorkspaceCreatorLabel] != newLabels[constants.WorkspaceCreatorLabel] {
+		return false
+	}
+	if oldAnnotations[constants.WorkspaceRestrictedAccessAnnotation] == "true" &&
+		newAnnotations[constants.WorkspaceRestrictedAccessAnnotation] != "true" {
+		return false
```
sleshchenko, line 197: Since we allow to modify some pieces, it makes sense to provide message here, what exactly user tried to modify but it's forbidden, like, when I tried to update creator, I got `object is owned by workspace and cannot be updated` but it could be something like `creator label is managed by devworkspace operator and can't be updated`.

For restricted access it can be: `restricted access should be configure on DevWorkspace level`

webhook/workspace/handler/immutable.go
```diff
@@ -177,7 +187,14 @@ func changePermitted(oldObj, newObj runtime.Object) bool {
 		log.Error(fmt.Errorf("Object %s is not a valid k8s object: does not have metadata", newObj.GetObjectKind()), "Failed to compare objects")
 		return false
 	}
-	newMeta.SetFinalizers(oldMeta.GetFinalizers())
-	newMeta.SetDeletionTimestamp(oldMeta.GetDeletionTimestamp())
-	return reflect.DeepEqual(oldMeta, newMeta)
+	oldLabels, newLabels := oldMeta.GetLabels(), newMeta.GetLabels()
+	oldAnnotations, newAnnotations := oldMeta.GetAnnotations(), newMeta.GetAnnotations()
```
sleshchenko, line 191: I would like to see variables declared nearest their usage.
BTW is there a need to have these vars?

webhook/workspace/handler/immutable.go
```diff
@@ -164,6 +163,17 @@ func (h *WebhookHandler) checkRestrictedAccessAnnotation(req admission.Request)
 	return annotations[constants.WorkspaceRestrictedAccessAnnotation] == "true", nil
 }
 
+func checkRestrictedWorkspaceMetadata(oldMeta, newMeta *metav1.ObjectMeta) (allowed bool, msg string) {
+	if oldMeta.Labels[constants.WorkspaceCreatorLabel] != newMeta.Labels[constants.WorkspaceCreatorLabel] {
```
sleshchenko, line 167: I wonder why we need it?
We already have a creator label logic(a bit better in terms of UX since it can be restored from old obj if it's missing) on higher level.

But maybe restricted access check should be done after creator annotation, how it does not seem to be a case: https://github.com/amisevsk/devworkspace-operator/blob/allow-annotations-webhooks/webhook/workspace/handler/workspace.go#L57

webhook/workspace/handler/immutable.go
```diff
@@ -93,31 +92,31 @@ func (h *WebhookHandler) HandleRestrictedAccessCreate(_ context.Context, req adm
 }
 
 func (h *WebhookHandler) checkRestrictedAccessWorkspaceV1alpha1(oldWksp, newWksp *devworkspacev1alpha1.DevWorkspace, uid string) (allowed bool, msg string) {
-	return h.checkRestrictedAccessWorkspace(oldWksp.Labels[constants.WorkspaceCreatorLabel],
-		oldWksp.Annotations[constants.WorkspaceRestrictedAccessAnnotation],
-		uid,
-		func() bool { return cmp.Equal(oldWksp, newWksp, RestrictedAccessDiffOptions[:]...) })
+	if oldWksp.Annotations[constants.WorkspaceRestrictedAccessAnnotation] != "true" {
+		return true, "workspace does not have restricted access configured"
```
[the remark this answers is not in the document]
amisevsk: We already do this check [here](https://github.com/devfile/devworkspace-operator/blob/015380990dc322974826728a63a2a6aa1fd14c6c/webhook/workspace/handler/workspace.go#L103). This function is checking the restricted-access part.

webhook/workspace/handler/immutable.go
amisevsk: Mostly to avoid the clumsiness of the syntax `oldMeta.GetLabels()[constants.WorkspaceCreatorLabel]`

webhook/workspace/handler/immutable.go
amisevsk: Great idea, done.

webhook/workspace/handler/immutable.go
amisevsk: :+1:

Removed the check here and reordered the checks (check creator first)

amisevsk: Actually, looking at it a second more: the reason we do the restricted-access check first is because of
```go
	newCreator, found := newWksp.Labels[constants.WorkspaceCreatorLabel]
	if !found {
		newWksp.Labels[constants.WorkspaceCreatorLabel] = oldCreator
		return h.returnPatched(req, newWksp)
	}
```
I've dropped the commit that changed the order.

webhook/workspace/handler/immutable.go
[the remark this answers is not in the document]
sleshchenko: It was my first comment, I later wanted to remove, since I figured out the stuff )

webhook/workspace/handler/immutable.go
```diff
@@ -160,24 +150,42 @@ func (h *WebhookHandler) checkRestrictedAccessAnnotation(req admission.Request)
...
 	}
 	newMeta, ok := newCopy.(metav1.Object)
 	if !ok {
-		log.Error(fmt.Errorf("Object %s is not a valid k8s object: does not have metadata", newObj.GetObjectKind()), "Failed to compare objects")
-		return false
+		log.Error(fmt.Errorf("object %s is not a valid k8s object: does not have metadata", newObj.GetObjectKind()), "Failed to compare objects")
+		return false, "Internal error"
+	}
+	oldLabels, newLabels := oldMeta.GetLabels(), newMeta.GetLabels()
+	if oldLabels[constants.WorkspaceCreatorLabel] != newLabels[constants.WorkspaceCreatorLabel] {
```
sleshchenko, line 182: I assume here we lose a feature that we have - restore workspace creator label, if it's missing.
Just because it's called before the code that do the feature.
But I haven't tested yet.

amisevsk: I thought the same, but our code is too complicated for such simple conclusions ;) 

`changePermitted` is only called for other workspace-involved objects, not workspaces themselves. The apply creator label feature applies only to DevWorkspaces themselves, as it's assumed that the DWO will propagate that label to all workspace-related resources it creates.

JPinkney (approved):
Re-tested again and everything was working (I had tested before I just forgot to hit approve :facepalm:)

sleshchenko:
After merge please cherry-pick changes into v1alphax branch

amisevsk:
> Prow tests artifacts does not seem to be useful at this point.
> Only PR's Author can figure out failure when they take a look on OS cluster until it's alive.

I've never successfully done it :)

PR #331, opened 2021-03-17
Check that cert-manager is installed when installing on k8s

amisevsk:
### What does this PR do?
Fail `make install` when cert_manager is not installed.

Also removes the `|| true` safety rail in make install, since it's no longer required (originally it was added to avoid changing deployment files when we fail to apply changes)

I can't seem to find it now

### Is it tested? How?
1. Start a minikube cluster without installing cert manager (or run `kubectl delete -f https://github.com/jetstack/cert-manager/releases/download/v1.0.4/cert-manager.yaml` on an existing cluster)
2. Run `make install` -> fails with message `Cert-manager is required for deploying on Kubernetes. See 'make install_cert_manager'`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,11 +174,11 @@ debug: _print_vars _gen_configuration_env _bump_kubeconfig _login_with_devworksp
 		dlv debug --listen=:2345 --headless=true --api-version=2 ./main.go --
 
 ### install: Install controller in the configured Kubernetes cluster in ~/.kube/config
-install: _print_vars _init_devworkspace_crds _create_namespace generate_deployment
+install: _check_cert_manager _print_vars _init_devworkspace_crds _create_namespace generate_deployment
 ifeq ($(PLATFORM),kubernetes)
-	$(K8S_CLI) apply -f deploy/current/kubernetes/combined.yaml || true
+	$(K8S_CLI) apply -f deploy/current/kubernetes/combined.yaml
 else
-	$(K8S_CLI) apply -f deploy/current/openshift/combined.yaml || true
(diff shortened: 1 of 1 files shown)


sleshchenko (approved):

benoitf (approved):
```
$ make install                                                                                                                                                                                                                                                                 bash: operator-sdk: command not found
Cert-manager is required for deploying on Kubernetes. See 'make install_cert_manager'
make: *** [_check_cert_manager] Error 1
```


after running `make install_cert_manager` and running `make install` the check is detecting it's there.

BTW it seems now minikube is using `jettech/kube-webhook-certgen`

```
$ minikube addons enable ingress                                                                                                                                                                                                                                            
    ▪ Using image us.gcr.io/k8s-artifacts-prod/ingress-nginx/controller:v0.40.2
    ▪ Using image jettech/kube-webhook-certgen:v1.2.2
    ▪ Using image jettech/kube-webhook-certgen:v1.3.0
```

so should it be enough as well ? (and not use cert-manager on a vanilla minikube)

sleshchenko:
> so should it be enough as well ? (and not use cert-manager on a vanilla minikube)

We would need to adapt our code(or deployment) to it I think, but should we rely on one or another method for cert generation?
Is it only minikube addon, or it's supposed to be part of a vanilla minikube?
It worths a dedicated issue and investigating more, when it's a good time for it.

benoitf:
@sleshchenko filed https://github.com/devfile/devworkspace-operator/issues/332 to be discussed

it's with the ingress addon that you have when using minikube (no 3rd party addon, it's in the available list)

PR #335, opened 2021-03-18
fix hanging terminal

tinakurian:
### What does this PR do?
This PR forces the webhook-server to restart by updating the webhook-server deployment with an arbitrary annotation that contains the time at which the pod was restarted.

https://issues.redhat.com/browse/WTO-77

### Is it tested? How?
Reproduced the issue by installing, uninstalling and re-installing the web-terminal. The following are the screenshots of each stage and the terminal starting: 

Shows the pod restarting
image (3)

Pod has restarted
image (4)

Timestamp in webhook deployment
image (5)

Terminal starting after re-install
image (6)

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -60,6 +60,9 @@ const (
 	// Stopped by Annotation holds the information as to why a workspace has stopped
 	StoppedByAnnotation = "controller.devfile.io/stopped-by"
 
+	// WebhookRestartAtAnnotation holds the the time (unixnano) of when the webhook server was forced to restart by controller
+	WebhookRestartedAtAnnotation = "controller.devfile.io/restarted-at"
+
 	// ControllerServiceAccountNameEnvVar stores the name of the serviceaccount used in the controller.
 	ControllerServiceAccountNameEnvVar = "CONTROLLER_SERVICE_ACCOUNT_NAME"
 )
--- a/pkg/webhook/deployment.go
+++ b/pkg/webhook/deployment.go
@@ -14,12 +14,10 @@ package webhook
 
 import (
 	"context"
-
+	"github.com/devfile/devworkspace-operator/internal/images"
+	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/webhook/server"
 
-	"github.com/devfile/devworkspace-operator/internal/images"
-
-	"github.com/devfile/devworkspace-operator/pkg/config"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -89,9 +87,10 @@ func getSpecDeployment(namespace string) (*appsv1.Deployment, error) {
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      server.WebhookServerDeploymentName,
-					Namespace: namespace,
-					Labels:    server.WebhookServerAppLabels(),
+					Name:        server.WebhookServerDeploymentName,
+					Namespace:   namespace,
+					Labels:      server.WebhookServerAppLabels(),
+					Annotations: server.WebhookServerAppAnnotations(),
 				},
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{
--- a/webhook/server/server.go
+++ b/webhook/server/server.go
@@ -14,8 +14,12 @@ import (
 	"errors"
 	"io/ioutil"
 	"os"
+	"strconv"
+	"time"
 
 	"github.com/devfile/devworkspace-operator/internal/cluster"
+	"github.com/devfile/devworkspace-operator/pkg/config"
+
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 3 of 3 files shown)


cmd/manager/main.go
```diff
@@ -37,8 +36,14 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
```
sleshchenko, line 43: I'm not sure I understand why we need Azure, GCP, OIDC...

cmd/manager/main.go
```diff
@@ -137,6 +142,42 @@ func main() {
...
+		log.Error(err, "Unable to get webhook deployment")
+		os.Exit(1)
+	}
+
+	//Add restart timestamp
+	if webhookDeployment.Spec.Template.Annotations == nil {
+		webhookDeployment.Spec.Template.Annotations = map[string]string{}
+	}
+	now := time.Now()
+	webhookDeployment.Spec.Template.Annotations[constants.RestartAtAnnotation] = strconv.FormatInt(now.UnixNano(), 10)
```
sleshchenko, line 170: I think it can be done in simpler way, if we just do it in which already update webhook server on each https://github.com/devfile/devworkspace-operator/blob/main/pkg/webhook/deployment.go

Since we have rolling update for webhook server, I think we should not care that webhook server will be restarted on each controller restart.

sleshchenko (commented):
The idea LGTM, I just think it can be done simpler.

Alternative solution, which handles SA changes even without controller restart: we could make webhook server listen to Controller SA, and refresh UID when it's recreated. But it requires more changes, but is not necessary. I believe the proposed solution is even better than enough :+1:

cmd/manager/main.go
tinakurian: Ohhhhh! okay, that makes it easier, there is already the client and the deployment in deployment.go. I've updated the PR to add the annotation in deployment.go. Thanks @sleshchenko

cmd/manager/main.go
tinakurian: Hmmm, i'm not too sure, i was organizing the imports and thought i would move the comment with the associated import. However, since i'm no longer touching this file i've undone my changes there - hope that is okay.

pkg/webhook/deployment.go
```diff
@@ -47,6 +49,17 @@ func CreateWebhookServerDeployment(
...
+
+		//Add restart timestamp which will update the webhook server
+		//deployment to force restart. This is done so that the
+		//serviceaccount userid is updated to use the latest and the
+		//web-terminal does not hang.
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = map[string]string{}
+		}
+		now := time.Now()
+		deployment.Spec.Template.Annotations[constants.RestartAtAnnotation] = strconv.FormatInt(now.UnixNano(), 10)
```
sleshchenko, line 61: what if we move it into getSpecDeployment?

sleshchenko (approved):
Just one more small proposal

cmd/manager/main.go
sleshchenko: I missed it was already there, my fault :man_facepalming:

pkg/webhook/deployment.go
sleshchenko: Ah, I see. Then it will be assigned to newly created resource... 

I don't have a strong opinion.
A bit better to make code a bit simpler

tinakurian: Ohhh okay, i see, it works that way as well and it is cleaner, sorry, I wasn't sure if that would work so i put it above the update. I've changed to code to add it to the getSpecDeployment func though! Thanks for the suggestion @sleshchenko!

cmd/manager/main.go
tinakurian: :smile: no worries! sorry for the confusion!

pkg/config/constants.go
```diff
@@ -60,6 +60,9 @@ const (
 	// Stopped by Annotation holds the information as to why a workspace has stopped
 	StoppedByAnnotation = "controller.devfile.io/stopped-by"
 
+	// RestartAt Annotation holds the the time (unixnano) of when the webhook pod was restarted
+	RestartAtAnnotation = "controller.devfile.io/restart-at"
```
amisevsk, line 64: Maybe it makes sense to use `restarted-at`, since this annotation will always refer to a time in the past

amisevsk (approved):
LGTM, one minor suggestion

pkg/config/constants.go
tinakurian: Good idea! Changed to past tense!

sleshchenko: Kubectl uses `kubectl.kubernetes.io/restartedAt`

tinakurian: Ohhhh okay, maybe it makes sense to change it to that then

JPinkney (approved):
LGTM

pkg/config/constants.go
sleshchenko: Sorry, as usually I was not clear enough.

I wanted to show that kubectl uses restarted (not restart)

But I did not catch that there is still difference with camel-case and - as word divider.

So, I don't have a strong opinion if one of both is better, I just care that we're consistent on controller side.
Which does not seem to be true even now:
- `workspace_name`
- `discoverable-server`
But it's out of the current PR scope.

We may change them when do workspace -> devworkspace renaming.

sleshchenko (approved):

pkg/config/constants.go
tinakurian: Ohhh okay, got it :-) makes sense, thanks @sleshchenko

PR #336, opened 2021-03-18
Remove Component CRD and related code

amisevsk:
### What does this PR do?
Remove the Components CRD. This subresource is unused since https://github.com/devfile/devworkspace-operator/pull/237 but CRDs were left for compatibility. 

Follow up to https://github.com/devfile/devworkspace-operator/pull/237

### Is it tested? How?
Test as normal; should be no functional changes as https://github.com/devfile/devworkspace-operator/pull/237 removed the Components controller.


sleshchenko (approved):
LGTM

We need to sync with deploy team when we change our deployments, ideally we can provide fixes for che operator and chectl

sleshchenko:
It must be merged before the release

JPinkney (approved):
LGTM

PR #340, opened 2021-03-19
fix hanging terminal

tinakurian:
### What does this PR do?
This PR forces the webhook-server to restart by updating the webhook-server deployment with an arbitrary annotation that contains the time at which the pod was restarted.

https://issues.redhat.com/browse/WTO-77

### Is it tested? How?
Tested via PR on alphax: https://github.com/devfile/devworkspace-operator/pull/335

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -79,6 +79,9 @@ const (
 	// this annotation will be cleared
 	WorkspaceStopReasonAnnotation = "controller.devfile.io/stopped-by"
 
+	// WebhookRestartedAtAnnotation holds the the time (unixnano) of when the webhook server was forced to restart by controller
+	WebhookRestartedAtAnnotation = "controller.devfile.io/restarted-at"
+
 	// PVCCleanupPodMemoryLimit is the memory limit used for PVC clean up pods
 	PVCCleanupPodMemoryLimit = "100Mi"
 
--- a/webhook/server/server.go
+++ b/webhook/server/server.go
@@ -14,7 +14,10 @@ import (
 	"errors"
 	"io/ioutil"
 	"os"
+	"strconv"
+	"time"
 
+	"github.com/devfile/devworkspace-operator/pkg/constants"
 	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
 
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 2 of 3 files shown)


pkg/constants/constants.go
```diff
@@ -79,6 +79,9 @@ const (
 	// this annotation will be cleared
 	WorkspaceStopReasonAnnotation = "controller.devfile.io/stopped-by"
 
+	// WebhookRestartedAtAnnotation holds the the time (unixnano) of when the webhook pod was restarted
+	WebhookRestartedAtAnnotation = "controller.devfile.io/restart-at"
```
sleshchenko, line 83: as discussed in another PR
```suggestion
	WebhookRestartedAtAnnotation = "controller.devfile.io/restarted-at"
```

pkg/constants/constants.go
```diff
@@ -79,6 +79,9 @@ const (
 	// this annotation will be cleared
 	WorkspaceStopReasonAnnotation = "controller.devfile.io/stopped-by"
 
+	// WebhookRestartedAtAnnotation holds the the time (unixnano) of when the webhook pod was restarted
```
sleshchenko, line 82: We can't tell for sure when pod is restarted, it's deployment work.
```suggestion
	// WebhookRestartedAtAnnotation holds the time (unixnano) of when the webhook server was forced to restart by controller
```

webhook/server/server.go
```diff
@@ -54,6 +57,17 @@ var WebhookServerAppLabels = func() map[string]string {
 	}
 }
 
+var WebhookServerAppAnnotations = func() map[string]string {
+	//Add restart timestamp which will update the webhook server
+	//deployment to force restart. This is done so that the
+	//serviceaccount userid is updated to use the latest and the
```
sleshchenko, line 63: I think for SA uid is `Unique identifier` but not user id.
```suggestion
	//serviceaccount uid is updated to use the latest and the
```

sleshchenko (approved):

pkg/constants/constants.go
tinakurian: fixed!

webhook/server/server.go
tinakurian: fixed!

pkg/constants/constants.go
tinakurian: fixed!

amisevsk (approved):
:+1:

JPinkney (approved):
LGTM

PR #343, opened 2021-03-22
Fixups to async storage

amisevsk:
### What does this PR do?
This PR addresses some non-single-line fixes from review on https://github.com/devfile/devworkspace-operator/pull/296:
- https://github.com/devfile/devworkspace-operator/pull/296#discussion_r596653750
- https://github.com/devfile/devworkspace-operator/pull/296#discussion_r596681476
- https://github.com/devfile/devworkspace-operator/pull/296#discussion_r596722914

It does not address one open question from review:
- https://github.com/devfile/devworkspace-operator/pull/296#discussion_r596714302
as I'm not sure it makes sense to go in this direction.

Let me know if there's any comments I missed here and I'll add them to the PR.

Follow up to https://github.com/devfile/devworkspace-operator/pull/296

### Is it tested? How?
Should functionally be no changes. For testing async workspaces, there is the process from the original PR:
> This PR adds a yaml file: `samples/with-k8s-ref/async-theia-next.yaml`, which is the usual theia-next sample with async storage enabled. This can be used to test basic use cases (create workspace, change a file, stop workspace, restart workspace -- note it takes a second for the files to repopulate after startup). 
> 
> To test cleanup, it's necessary to ssh into either a dedicated pod or the async storage server and check the contents of `/async-storage`. A flow that does this is
> ```bash
> kubectl apply -f samples/with-k8s-ref/async-theia-next.yaml
> # wait started; change file; stop workspace; wait till workspace pods terminated
> kubectl exec -it async-storage-xxxxx -- /bin/sh
> # ls -al /async-storage, look at files
> yq -y '.metadata.name = "alt-theia-next"' samples/with-k8s-ref/async-theia-next.yaml | kubectl apply -f -
> # note: this image won't be able to sync due to issue described above
> kubectl patch dw alt-theia-next --type merge -p '{"spec": {"started": false}}'
> kubectl delete dw theia # Delete initial workspace
> kubectl exec -it async-storage-xxxxx -- ls -al /async-storage # should be empty
> kubectl delete dw alt-theia-next
> # should be no async-storage deployment left on cluster after all workspaces deleted.
> ```

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -177,9 +177,17 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{}, nil
 	}
 	workspace.Spec.Template = *flattenedWorkspace
+
+	storageProvisioner, err := storage.GetProvisioner(workspace)
+	if err != nil {
+		reqLogger.Info("DevWorkspace start failed")
+		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
+		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning storage: %s", err)
+		return reconcile.Result{}, nil
+	}
 	// Set finalizer on DevWorkspace if necessary
 	// Note: we need to check the flattened workspace to see if a finalizer is needed, as plugins could require storage
-	if isFinalizerNecessary(workspace) {
+	if isFinalizerNecessary(workspace, storageProvisioner) {
 		coputil.AddFinalizer(clusterWorkspace, storageCleanupFinalizer)
 		if err := r.Update(ctx, clusterWorkspace); err != nil {
 			return reconcile.Result{}, err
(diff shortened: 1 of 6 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -177,9 +177,17 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{}, nil
 	}
 	workspace.Spec.Template = *flattenedWorkspace
+
+	storageProvisioner, err := storage.GetProvisioner(workspace)
```
sleshchenko, line 181: I see the issue we may have with an ability to change storage type.
As far as I understand, user may have devworkspace with any valid storage, devworkspace will get finalizers, but then after user sets invalid storage type, they'll get workspace which will not be possible to delete, until storage type is not fixed.

If you think it will work in that way, please create an issue or let me know, I'll do.

P.S. It's not related to these changes in any case, just observation when we may take care of to improve stability.

controllers/workspace/provision/object.go
```diff
@@ -87,13 +84,12 @@ func SyncObject(object runtime.Object, client client.Client, reqLogger logr.Logg
 		diffOpt = cmp.Options{}
 	}
 	if !cmp.Equal(object, found, diffOpt) {
-		reqLogger.Info("    => Updating "+objType.String(), "namespace", objMeta.GetNamespace(), "name", objMeta.GetName())
+		reqLogger.Info("Updating "+objType.String(), "namespace", objMeta.GetNamespace(), "name", objMeta.GetName())
```
sleshchenko, line 87: it may be the stuff we eventually would like to propagate as events, at least when `debug` mode is enabled.
Don't thanks me for commenting the stuff unrelated to the PR.

sleshchenko (approved):

controllers/workspace/devworkspace_controller.go
amisevsk: Yeah, we need to improve webhooks to forbid making changes to storage type. Hosted Che had this problem, IIRC, where you could set a persistent workspace to ephemeral and the storage wouldn't necessarily be cleaned up. 

I think the safe bet is to block all changes to storage -- what if async storage eventually decides to use a different PVC than common? If we eventually implement per-workspace storage, we'll also run into this problem.

JPinkney (approved):
The changes LGTM

Issue #337, opened 2021-03-18
Support setting condition status to 'false'


amisevsk:
## Description

Currently, DevWorkspace conditions get set to one of two states: `true` (meaning the condition is satisfied) and `unknown` (meaning the condition is unobserved). We should also support setting conditions to `false`, to explicitly note that the condition is not satisfied. This would also allow propagating information more easily (use the message of the first false condition in the INFO field).

## Additional Info
Inspired by https://github.com/devfile/devworkspace-operator/pull/296#discussion_r596697393

PR #345, opened 2021-03-23
Improve condition handling to better represent workspace state

amisevsk:
### What does this PR do?
Improves how we handle conditions to allow multiple "waiting" messages to be defined, and explicitly set conditions to be false when we know they're not true (e.g. if we're waiting for routing, it makes sense to set the RoutingReady condition to false).

Since we already have a condition for `PullSecretsReady`, I figured I'd add a few more for big steps in workspace startup. Semantically, the only changed usage is `WorkspaceReady` is now true when the workspace is in the running phase (whereas previously we still had to wait for the editor).

Closes https://github.com/devfile/devworkspace-operator/issues/337

### Is it tested? How?
1. Deploy operator as normal
2. `kubectl apply -f samples/with-k8s-ref/theia-next.yaml && kubectl get dw -w`
3. Observe status conditions progressing as workspace starts up.

### DIFF
--- a/controllers/workspace/condition.go
+++ b/controllers/workspace/condition.go
@@ -1,0 +1,65 @@
+//
+// Copyright (c) 2019-2021 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package controllers
+
+import (
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	PullSecretsReady     dw.WorkspaceConditionType = "PullSecretsReady"
+	DevWorkspaceResolved dw.WorkspaceConditionType = "DevWorkspaceResolved"
+	StorageReady         dw.WorkspaceConditionType = "StorageReady"
+	DeploymentReady      dw.WorkspaceConditionType = "DeploymentReady"
+)
+
+var conditionOrder = []dw.WorkspaceConditionType{
+	DevWorkspaceResolved,
+	StorageReady,
+	dw.WorkspaceRoutingReady,
+	dw.WorkspaceServiceAccountReady,
+	PullSecretsReady,
+	DeploymentReady,
+	dw.WorkspaceReady,
+}
+
+// workspaceConditions is a description of last-observed workspace conditions.
+type workspaceConditions struct {
+	conditions map[dw.WorkspaceConditionType]dw.WorkspaceCondition
+}
+
+func (c *workspaceConditions) setConditionTrue(conditionType dw.WorkspaceConditionType, msg string) {
+	c.conditions[conditionType] = dw.WorkspaceCondition{
+		Status:  corev1.ConditionTrue,
+		Message: msg,
+	}
+}
+
+func (c *workspaceConditions) setConditionFalse(conditionType dw.WorkspaceConditionType, msg string) {
+	c.conditions[conditionType] = dw.WorkspaceCondition{
+		Status:  corev1.ConditionFalse,
+		Message: msg,
+	}
+}
+
+// getFirstFalse checks current conditions in a set order (defined by conditionOrder) and returns the first
+// condition with a 'false' status. Returns nil if there is no currently observed false condition
+func (c *workspaceConditions) getFirstFalse() *dw.WorkspaceCondition {
+	for _, cond := range conditionOrder {
+		if condition, present := c.conditions[cond]; present && condition.Status == corev1.ConditionFalse {
+			return &condition
+		}
+	}
+	return nil
+}
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -47,14 +47,6 @@ import (
 	devworkspace "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 )
 
-type currentStatus struct {
-	// Map of condition types that are true for the current workspace. Key is valid condition, value is optional
-	// message to be filled into condition's 'Message' field.
-	Conditions map[devworkspace.WorkspaceConditionType]string
-	// Current workspace phase
-	Phase devworkspace.WorkspacePhase
-}
-
 // DevWorkspaceReconciler reconciles a DevWorkspace object
 type DevWorkspaceReconciler struct {
 	client.Client
@@ -132,10 +124,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	}
 
 	// Prepare handling workspace status and condition
-	reconcileStatus := currentStatus{
-		Conditions: map[devworkspace.WorkspaceConditionType]string{},
-		Phase:      devworkspace.WorkspaceStatusStarting,
-	}
+	reconcileStatus := initCurrentStatus()
 	clusterWorkspace := workspace.DeepCopy()
 	timingInfo := map[string]string{}
 	timing.SetTime(timingInfo, timing.WorkspaceStarted)
@@ -171,19 +160,13 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return r.failWorkspace(workspace, fmt.Sprintf("Error processing devfile: %s", err), reqLogger, &reconcileStatus)
 	}
 	workspace.Spec.Template = *flattenedWorkspace
+	reconcileStatus.setConditionTrue(DevWorkspaceResolved, "")
 
 	storageProvisioner, err := storage.GetProvisioner(workspace)
 	if err != nil {
-		reqLogger.Info("DevWorkspace start failed")
-		// Clean up cluster deployment
-		scaleErr := provision.ScaleDeploymentToZero(workspace, r.Client)
-		if scaleErr != nil {
-			return reconcile.Result{}, scaleErr
-		}
-		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
-		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning storage: %s", err)
-		return reconcile.Result{}, nil
+		return r.failWorkspace(workspace, fmt.Sprintf("Error provisioning storage: %s", err), reqLogger, &reconcileStatus)
 	}
+
 	// Set finalizer on DevWorkspace if necessary
 	// Note: we need to check the flattened workspace to see if a finalizer is needed, as plugins could require storage
 	if isFinalizerNecessary(workspace, storageProvisioner) {
@@ -203,6 +186,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		switch storageErr := err.(type) {
 		case *storage.NotReadyError:
 			reqLogger.Info(storageErr.Message)
+			reconcileStatus.setConditionFalse(StorageReady, fmt.Sprintf("Provisioning storage: %s", storageErr.Message))
 			return reconcile.Result{Requeue: true, RequeueAfter: storageErr.RequeueAfter}, nil
 		case *storage.ProvisioningError:
 			return r.failWorkspace(workspace, fmt.Sprintf("Error provisioning storage: %s", storageErr), reqLogger, &reconcileStatus)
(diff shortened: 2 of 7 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -170,13 +157,11 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	}
 	flattenedWorkspace, err := flatten.ResolveDevWorkspace(workspace.Spec.Template, flattenHelpers)
 	if err != nil {
-		reqLogger.Info("DevWorkspace start failed")
-		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
-		// TODO: Handle error more elegantly
-		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error processing devfile: %s", err)
-		return reconcile.Result{}, nil
+		return r.failWorkspace(fmt.Sprintf("Error processing devfile: %s", err), reqLogger, &reconcileStatus)
```
sleshchenko, line 160: Does it make any sense if we duplicate message to condition where it happened?
```suggestion
		reconcileStatus.setConditionFalse(ResolvedDevWorkspace, fmt.Sprintf("Error processing devfile: %s", err))
		return r.failWorkspace(fmt.Sprintf("Error processing devfile: %s", err), reqLogger, &reconcileStatus)
```

sleshchenko: or simpler way, if failWorkspace would supply the condition.
```suggestion
		return r.failWorkspace(ResolvedDevWorkspace, fmt.Sprintf("Error processing devfile: %s", err), reqLogger, &reconcileStatus)
```

controllers/workspace/devworkspace_controller.go
```diff
@@ -188,38 +173,29 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
...
-		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning storage: %s", err)
-		return reconcile.Result{}, nil
+		return r.failWorkspace(fmt.Sprintf("Error provisioning storage: %s", err), reqLogger, &reconcileStatus)
 	}
 	err = storageProvisioner.ProvisionStorage(devfilePodAdditions, workspace, clusterAPI)
 	if err != nil {
 		switch storageErr := err.(type) {
 		case *storage.NotReadyError:
 			reqLogger.Info(storageErr.Message)
+			reconcileStatus.setConditionFalse(StorageProvisioned, "Provisioning storage")
```
sleshchenko, line 188: Does it makes sense to propagate storageErr.Message here?

controllers/workspace/devworkspace_controller.go
```diff
@@ -361,11 +334,20 @@ func (r *DevWorkspaceReconciler) stopWorkspace(workspace *devworkspace.DevWorksp
...
 	if workspaceDeployment.Status.Replicas == 0 {
 		logger.Info("Workspace stopped")
-		status.Phase = devworkspace.WorkspaceStatusStopped
+		status.phase = devworkspace.WorkspaceStatusStopped
 	}
 	return r.updateWorkspaceStatus(workspace, logger, status, reconcile.Result{}, nil)
 }
 
+// failWorkspace marks a workspace as failed by setting relevant fields in the status struct. These changes are not
+// propagated to the cluster, and must be synced via updateWorkspaceStatus()
```
sleshchenko, line 343: if it's not propagated to the cluster, what is the reason to return `reconcile.Result` here?

controllers/workspace/devworkspace_controller.go
```diff
@@ -263,10 +237,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 			reqLogger.Info(provisionErr.Message)
 			return reconcile.Result{Requeue: true, RequeueAfter: provisionErr.RequeueAfter}, nil
 		case *metadata.ProvisioningError:
-			reqLogger.Info(fmt.Sprintf("DevWorkspace start failed: %s", provisionErr))
-			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
-			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning metadata configmap: %s", provisionErr)
-			return reconcile.Result{}, nil
+			return r.failWorkspace(fmt.Sprintf("Error provisioning metadata configmap: %s", provisionErr), reqLogger, &reconcileStatus)
```
sleshchenko, line 240: Docs says updateWorkspaceStatus must be called https://github.com/devfile/devworkspace-operator/pull/345/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R343, or it's when you need to sync immediately but going to continue loop logic?

controllers/workspace/condition.go
```diff
@@ -0,0 +1,65 @@
...
+
+package controllers
+
+import (
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	PullSecretsReadyCondition dw.WorkspaceConditionType = "PullSecretsReady"
```
sleshchenko, line 21: If we try to align conditions names, it should be
```suggestion
	PullSecretsReady dw.WorkspaceConditionType = "PullSecretsReady"
```

controllers/workspace/condition.go
```diff
@@ -0,0 +1,65 @@
...
+	ResolvedDevWorkspace      dw.WorkspaceConditionType = "ResolvedWorkspace"
```
sleshchenko, line 22:
```suggestion
	DevWorkspaceResolved      dw.WorkspaceConditionType = "WorkspaceResolved"
```

controllers/workspace/condition.go
```diff
@@ -0,0 +1,65 @@
...
+	StorageProvisioned        dw.WorkspaceConditionType = "StorageReady"
```
sleshchenko, line 23: StorageProvisioned or Storage ready?
I like the second a bit more
```suggestion
	StorageReady             dw.WorkspaceConditionType = "StorageReady"
```

sleshchenko (commented):
In general LGTM
I'll do deeper review with testing later

controllers/workspace/devworkspace_controller.go
amisevsk: So that you can call `return failWorkspace()` instead of `return reconcile.Result{}, failWorkspace()` in the reconcile loop.

controllers/workspace/devworkspace_controller.go
amisevsk: I thought about this a bit more and I'm not sure it makes total sense:
- If we want to move setting the condition into failWorkspace, it makes failWorkspace depend on a condition, which we don't have for e.g. when `validateCreatorLabel()` fails. We could add a condition there as well (`WorkspaceInitialized` perhaps), but I also would prefer to set conditions when they're useful (`WorkspaceInitialized` will be true in 99% of cases, and when it's not the `FailedStart` condition will contain the same info).
- Further, failure points in the reconcile loop aren't directly tied to conditions. For example, we could fail parsing resourceLimits, which happens after we set `resolveDevWorkspace` but before `storageProvisioned`. I'd fear it would require us to add too many conditions to represent rare edge cases.

controllers/workspace/devworkspace_controller.go
amisevsk: I suppose the doc is unclear. I mean to say that `failWorkspace` doesn't do an update on its own. However, from this call site, the return value will be intercepted by the deferred function in the main reconcile loop, which will call `updateWorkspaceStatus`. I'm not sure the best way to explain this in a short doc though.

controllers/workspace/devworkspace_controller.go
sleshchenko: OK, then changes must be synced with updateWorkspaceStatus only in when client needs these changes by flushed before `return reconcile.Result{}`

Then maybe doc should be
```
// These changes are not synced to cluster immediately.
// The cluster will be reflected with the made changes after reconcile loop finished.
// Or if needed, it can be flushed via updateWorkspaceStatus()
```
WDYT?

controllers/workspace/devworkspace_controller.go
sleshchenko: Left a proposal for documentation

controllers/workspace/devworkspace_controller.go
amisevsk: Makes sense, will do.

controllers/workspace/devworkspace_controller.go
sleshchenko: The issue I had in mind, if phase throws a common error, it may help to understand on which condition is happened, like ResolveDevWorkspace has error `could not unmarshal devfile from response: ${err}` where error is like, is not a valid yaml.
^ but that error can be improved as well, to contain the related component failure, or something. Maybe it's already done on some step of throwing error up, but I assume we just propagate err, without provisioning details.

I agree that in some cases it may be a bit redundant, like `failed to resolve devworkspace: parent has invalid URL`

About cases, where error happens independent from phase, the phase may be optional.

So, I agree that my suggestion still has some question, but I agree to proceed without it.

sleshchenko: Well, the user-facing messages should be good enough, since you solve the issue with face identifying by appending prefix to error message.

controllers/workspace/devworkspace_controller.go
```diff
@@ -232,16 +208,14 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	routingStatus := provision.SyncRoutingToCluster(workspace, clusterAPI)
 	if !routingStatus.Continue {
 		if routingStatus.FailStartup {
-			reqLogger.Info("DevWorkspace start failed")
-			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
 			// TODO: Propagate failure reason from devWorkspaceRouting
-			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = "Failed to install network objects required for devworkspace"
-			return reconcile.Result{}, routingStatus.Err
+			return r.failWorkspace("Failed to install network objects required for devworkspace", reqLogger, &reconcileStatus)
```
sleshchenko, line 212: Maybe it's a good time to fix TODO above?

controllers/workspace/devworkspace_controller.go
```diff
@@ -288,44 +259,46 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
...
 	// Step six: Create deployment and wait for it to be ready
 	timing.SetTime(timingInfo, timing.DeploymentCreated)
 	deploymentStatus := provision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI)
 	if !deploymentStatus.Continue {
 		if deploymentStatus.FailStartup {
-			reqLogger.Info("Workspace start failed")
-			reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
-			reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Devworkspace spec is invalid: %s", deploymentStatus.Err)
-			return reconcile.Result{}, deploymentStatus.Err
+			return r.failWorkspace(fmt.Sprintf("Devworkspace spec is invalid: %s", deploymentStatus.Err), reqLogger, &reconcileStatus)
```
sleshchenko, line 281:
```suggestion
			return r.failWorkspace(fmt.Sprintf("DevWorkspace spec is invalid: %s", deploymentStatus.Err), reqLogger, &reconcileStatus)
```

sleshchenko (commented):
I'm not sure if it's an issue on my end, but conditions seems to have issues with setting in True. I tried to debug but everything looks good
Screenshot_20210329_092526
but it does not help for the cluster state
Screenshot_20210329_091114

controllers/workspace/devworkspace_controller.go
amisevsk: I think it's for a separate PR -- the underlying issue is that devworkspacerouting objects don't get a failure message, they just get marked as failed. The fix is to add a message field to the dwr spec and update the subcontroller to set it appropriately.

amisevsk:
The issue with false conditions is
```
Error updating workspace status: Operation cannot be fulfilled on devworkspaces.workspace.devfile.io "theia": the object has been modified; please apply your changes to the latest version and try again
```
it seems to be happening pretty continuously now.

I'm not sure it's caused by this PR since we do see it in main (though perhaps less often), but I'll look into it.

amisevsk:
Well that was an [embarrasing](https://github.com/devfile/devworkspace-operator/pull/345/commits/46e5c03039b2d9eb14e8b9a72c2bffd30523f0b3#diff-65fcf1aa32aa49850ef197edbe1ff894195f860d3ac58908dc8801757db95616L69-R70) way to spend a few hours :smile: 

Good catch @sleshchenko

controllers/workspace/devworkspace_controller.go
sleshchenko: Makes sense, there is a dedicated issue for that https://github.com/eclipse/che/issues/17455

sleshchenko (approved):
Such a great UX improvement <3

Tested an this time works as expected. Even tried to uncomment waiting for PVC to be bound and now user gets understanding why it hangs
Screenshot_20210331_113334


P.S. Don't forget to adapt your PR after Tina's changes.

Issue #333, opened 2021-03-18
DevWorkspace should have a dedicated phase for terminating


sleshchenko:
Currently, when dw is terminating(pvc files are cleaning up, or owning objects are removing), it got the latest state (STARTING, RUNNING, FAILED, STOPPED). It would be great if we have a dedicated phase to expose that workspace is checked to be removed and just waits some cleaning up to be done before.

PR #355, opened 2021-03-25
Add terminating phase for workspaces waiting on finalizers

amisevsk:
### What does this PR do?
Adds `Terminating` phase to DevWorkspaces, to show deleted workspaces that are waiting on a finalizer:
```
theia   workspacea79dbc5374564e5f   Terminating   Cleaning up resources for deletion
```

This PR adds the new Phase to code in this repo. Once we settle on naming, it should be moved into `devfile/api` with the other phases.

Closes https://github.com/devfile/devworkspace-operator/issues/333

### Is it tested? How?
1. `kubectl apply -f samples/with-k8s-ref/theia-next.yaml`
2. `kubectl get dw -w`
3. (new window) `kubectl delete dw theia`
4. The watch from step 2 should show workspace in `Terminating` phase.

### DIFF
--- a/controllers/workspace/finalize.go
+++ b/controllers/workspace/finalize.go
@@ -30,6 +30,9 @@ import (
 
 const (
 	storageCleanupFinalizer = "storage.controller.devfile.io"
+	// devworkspacePhaseTerminating represents a DevWorkspace that has been deleted but is waiting on a finalizer.
+	// TODO: Should be moved to devfile/api side.
+	devworkspacePhaseTerminating devworkspace.WorkspacePhase = "Terminating"
 )
 
 func (r *DevWorkspaceReconciler) finalize(ctx context.Context, log logr.Logger, workspace *devworkspace.DevWorkspace) (reconcile.Result, error) {
@@ -37,6 +40,7 @@ func (r *DevWorkspaceReconciler) finalize(ctx context.Context, log logr.Logger, 
 		return reconcile.Result{}, nil
 	}
 	workspace.Status.Message = "Cleaning up resources for deletion"
+	workspace.Status.Phase = devworkspacePhaseTerminating
 	err := r.Client.Status().Update(ctx, workspace)
 	if err != nil && !k8sErrors.IsConflict(err) {
 		return reconcile.Result{}, err


JPinkney (approved):
Makes sense to me, I also tried looking up synonyms for terminating but I didn't find any better options

controllers/workspace/finalize.go
```diff
@@ -30,13 +30,17 @@ import (
 
 const (
 	storageCleanupFinalizer = "storage.controller.devfile.io"
+	// devworkspacePhaseTerminating represents a DevWorkspace that has been deleted but is waiting on a finalizer.
+	// TODO: Should be moved to devfile/api side.
+	devworkspacePhaseTerminating devworkspace.WorkspacePhase = "Terminating"
```
sleshchenko, line 35: I wrongly assumed that our DevWorkspace can wait dependent objects to be removed with empty finalizers.

But it's not true. According to https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/
```
The object's metadata.finalizers contains the value "foregroundDeletion".
```

So, if we want DevWorkspace to wait dependent objects, technically it's finalizing.
From that perspective, `"Finalizing"` phase a bit win, but just a bit since:
- `Terminating` can be confused with `Stopping`, but if you know K8s API and Namespaces are terminating before removing, it's pretty clear;
- `Finalizing` can be confused with anything, like finalizing after stop, finalizing after start. If you know about K8s garbage collector, you probably will guess to check DevWorkspace CR finalizers, to check the hanging cause, in case it's third-party finalizer.

I'm 50\50 between `Terminating` and `Finalizing`

Alternative from tmp-che-devworkspace:
`REMOVAL_WAIT`, `FINALIZING_DELETE`, `PendingDeletion`(DWO uses one when there are still workspaces on the cluster)

sleshchenko (approved):

controllers/workspace/finalize.go
sleshchenko: from https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
> When a Pod is being deleted, it is shown as Terminating by some kubectl commands. This Terminating status is not one of the Pod phases.

sleshchenko: for namespace it's 

> Terminating the namespace is being deleted, and can not be used for new objects

which may be for devworkspace:

> Terminating the devworkspace is being deleted, cleaning up the dependent resources, can't not be started anymore

PR #359, opened 2021-03-29
Add Che happy-path selenium test as PR check

SkorikSergey:
### What does this PR do?
Start happy-path selenium test in devworkspace repo. 
It uses built  **devworkspace-operator** image from branch as `--dev-workspace-controller-image` "chectl server:deploy" option.

This PR will not change `openshift_e2e.sh` script and uses it temporary for starting happy-path test on already existed `openshift-ci` job.


JPinkney:
I haven't looked much at the changes yet but isn't it a bit strange that we're adding che specific tests to the devworkspace-operator which is technically supposed to vendor neutral and eventually added to the cncf?

amisevsk (commented):
I agree with Josh. It would be great to have happy-path/e2e tests in this repo, but depending on Che is something we're trying to avoid.

SkorikSergey:
> I agree with Josh. It would be great to have happy-path/e2e tests in this repo, but depending on Che is something we're trying to avoid.

@l0rd What do you think about it? Happy-path test expects to start on Eclipse Che.

l0rd:
I am not shocked if we verify that downstream projects don't get broken by a PR. The downstream projects today are Che and WTO. 

On the other hand that should not block the merge of a PR, in particular considered the fragility of Che happy path tests.

flacatus:
LGTM

.ci/oci-devworkspace-happy-path.sh
```diff
@@ -0,0 +1,170 @@
...
+
+deployChe() {
+  cat > /tmp/che-cr-patch.yaml <<EOL
+spec:
+  devWorkspace:
+    enable: true
+    controllerImage: '${DEVWORKSPACE_OPERATOR}'
+  server:
+    customCheProperties:
+      CHE_FACTORY_DEFAULT__PLUGINS: ""
```
sleshchenko, line 82: why do we need it?

.ci/oci-devworkspace-happy-path.sh
```diff
@@ -0,0 +1,170 @@
...
+    echo "======== oc get events ========"
+    oc get events -n ${DEVWORKSPACE_CONTROLLER_NAMESPACE}| tee get_events.log
+
+  mkdir -p ${ARTIFACTS_DIR}
+  /tmp/chectl/bin/chectl server:logs --chenamespace=${NAMESPACE} --directory=${ARTIFACTS_DIR} --telemetry=off
+}
+
+installChectl() {
+  curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && \
+  chmod +x ./kubectl
```
sleshchenko, line 62: why do we need kubectl?

.ci/oci-devworkspace-happy-path.sh
```diff
@@ -0,0 +1,170 @@
...
+export HAPPY_PATH_DEVFILE='https://gist.githubusercontent.com/l0rd/71a04dd0d8c8e921b16ba2690f7d5a47/raw/d520086e148c359b18c229328824dfefcf85e5ef/spring-petclinic-devfile-v2.0.0.yaml'
+
+# Pod created by openshift ci don't have user. Using this envs should avoid errors with git user.
+export GIT_COMMITTER_NAME="CI BOT"
+export GIT_COMMITTER_EMAIL="ci_bot@notused.com"
+
+function getDevWorkspaceOperatorLogs() {
+    mkdir -p ${ARTIFACTS_DIR}/devworkspace-operator
+    cd ${ARTIFACTS_DIR}/devworkspace-operator
+    for POD in $(oc get pods -o name -n ${DEVWORKSPACE_CONTROLLER_NAMESPACE}); do
```
sleshchenko, line 45: I would love not to duplicate the same we have in existing script.
instead I would move getPodLogs with namespace as argument into `.ci/common.sh` and do source here.

.ci/oci-devworkspace-happy-path.sh
SkorikSergey: It is patch configuration from this issue - https://github.com/eclipse/che/issues/19347.

sleshchenko: I believe it can be dropped at this point.

.ci/oci-devworkspace-happy-path.sh
SkorikSergey: You are right. `oc` client is enough here.

.ci/oci-devworkspace-happy-path.sh
SkorikSergey:
```
spec:
  devWorkspace:
    enable: true
```

.ci/happy-path-che.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - name: test-run-results
+          mountPath: /tmp/e2e/report/
+      resources:
+        requests:
+          memory: "3Gi"
+          cpu: "2"
+        limits:
+          memory: "4Gi"
+          cpu: "2"
+# Download results
```
sleshchenko, line 51:
```suggestion
    # Download results
```

.ci/oci-devworkspace-happy-path.sh
```diff
@@ -0,0 +1,170 @@
...
+    controllerImage: '${DEVWORKSPACE_OPERATOR}'
+  server:
+    customCheProperties:
+      CHE_FACTORY_DEFAULT__PLUGINS: ""
+      CHE_WORKSPACE_DEVFILE_DEFAULT__EDITOR_PLUGINS: ""
+  auth:
+    updateAdminPassword: false
+EOL
+
+  cat /tmp/che-cr-patch.yaml
```
sleshchenko, line 88: it's just for debugging, right?

SkorikSergey: It mostly to check `DEVWORKSPACE_OPERATOR` env has built devworkspace-operator image path.

sleshchenko:
```suggestion
  echo "INFO: Using the following Che Cluster CR"
  cat /tmp/che-cr-patch.yaml
  echo "----------------------------------"
```

.ci/resources/pod-che-happy-path.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - name: test-run-results
+          mountPath: /tmp/e2e/report/
+      resources:
+        requests:
+          memory: "3Gi"
+          cpu: "2"
+        limits:
+          memory: "4Gi"
+          cpu: "2"
+    # Download results
```
sleshchenko, line 51:
```suggestion
    # Download results
```

sleshchenko (approved):
LGTM, I hope I don't break anything with my refactor stuff :crossed_fingers:

sleshchenko:
@amisevsk please take a look as well

sleshchenko:
Happy path test passed, merging. If any comments will appear - they will be addressed in a separate PR.

PR #360, opened 2021-03-29
Add volume into custom web terminal sample

sleshchenko:
### What does this PR do?
Add volume into custom web terminal sample.

N/A

### Is it tested? How?
Made sure that the sample work fine with the latest tooling image after https://github.com/redhat-developer/web-terminal-tooling/pull/26 is merged.

It did not work for me with default latest WTO, so a dedicated issue is created https://issues.redhat.com/browse/WTO-89

### DIFF
--- a/samples/web-terminal-custom-tooling.yaml
+++ b/samples/web-terminal-custom-tooling.yaml
@@ -24,3 +24,8 @@ spec:
           env:
             - name: PS1
               value: \[\e[34m\]>\[\e[m\]\[\e[33m\]>\[\e[m\]
+          volumeMounts:
+            - name: home-storage
+              path: "/home/user"
+      - volume:
+          name: home-storage
--- a/internal-registry/redhat-developer/web-terminal-dev/latest/devworkspacetemplate.yaml
+++ b/internal-registry/redhat-developer/web-terminal-dev/latest/devworkspacetemplate.yaml
@@ -10,7 +10,7 @@ spec:
         command: ["/go/bin/che-machine-exec",
                   "--authenticated-user-id", "$(DEVWORKSPACE_CREATOR)",
                   "--idle-timeout", "$(DEVWORKSPACE_IDLE_TIMEOUT)",
-                  "--pod-selector", "controller.devfile.io/workspace_id=$(DEVWORKSPACE_ID)",
+                  "--pod-selector", "controller.devfile.io/devworkspace_id=$(DEVWORKSPACE_ID)",
                   "--use-bearer-token"]
         endpoints:
           - name: web-terminal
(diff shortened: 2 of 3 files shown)


samples/web-terminal-custom-tooling.yaml
```diff
@@ -24,3 +24,8 @@ spec:
           env:
             - name: PS1
               value: \[\e[34m\]>\[\e[m\]\[\e[33m\]>\[\e[m\]
+          volumeMounts:
+            - path: "/home/"
```
amisevsk, line 28: Shouldn't we mount to `/home/user`?

amisevsk (commented):
This may cause a few other issues. Looking at the contents of the web terminal tooling container, without the entrypoint, I see
```
$ ls -al $HOME
total 20
drwxrwxr-x. 2 root root 4096 Feb 25 15:39 .
drwxr-xr-x. 1 root root 4096 Feb 25 15:38 ..
-rw-rw-r--. 1 root root   77 Feb 25 15:39 .bashrc
-rw-rw----. 1 root root  533 Feb 25 15:39 .viminfo
```
where `.bashrc` contains some completions. Mounting a volume into this path will wipe out these files.

We should update the web-terminal-tooling image to accomodate a mounted `$HOME` dir (e.g. by putting completions `/tmp/home` and copying over to the PVC on start.

I'm not sure why, but with the current PR, the `/home/user` directory is being created by root and not the running user :thinking:

amisevsk:
Created https://issues.redhat.com/browse/WTO-82 to track updates needed for the tooling image.

amisevsk:
PR https://github.com/redhat-developer/web-terminal-tooling/pull/26 is merged so this should be good to go as soon as the change is propagated into the image.

sleshchenko:
@amisevsk I tried the sample with the latest manually built tooling (quay.io/sleshche/web-terminal-tooling:latest) and it does not seem to work as expected:
Screenshot_20210414_125828
```sh
> kc logs workspace2ca8278ee0e3489f-7b6f75765d-dj2tn -c tooling 
cp: cannot create regular file '/home/user/.bashrc': Permission denied

```

samples/web-terminal-custom-tooling.yaml
sleshchenko: Changed to /home/user.

but actually, it's strange that I got permissions denied error when I try /home, somebody creates non-writable /home/user.

amisevsk (approved):
LGTM

sleshchenko:
> /test all

does nothing :confused: 
I tried to patch triggers but they are generated in not the best way :man_shrugging:

amisevsk:
>> /test all

> does nothing
> I tried to patch triggers but they are generated in not the best way

The bot [*LIES*](https://github.com/devfile/devworkspace-operator/pull/381#issuecomment-819275259):

> Use `/test all` to run the following jobs:
> * `pull-ci-devfile-devworkspace-operator-main-v7-images`

PR #361, opened 2021-03-29
Enable github action to build v1alphax image

sleshchenko:
### What does this PR do?
It's back-porting for https://github.com/devfile/devworkspace-operator/pull/329
So, the main purpose of this PR is introduce the latest web terminal plugin.

In addition it enables building v1alphax container image, which previously we built manually after merge.

back-ports https://github.com/devfile/devworkspace-operator/pull/329

### Is it tested? How?
Not yet. but I'm going to test that samples/web-terminal.yaml works as expected.

### DIFF
--- a/.github/workflows/dockerimage-next.yml
+++ b/.github/workflows/dockerimage-next.yml
@@ -10,7 +10,7 @@ name: Next Dockerimage
 
 on:
   push:
-    branches: [ master ]
+    branches: [ v1alphax ]
 
 jobs:
 
@@ -27,5 +27,5 @@ jobs:
         registry: quay.io
         repository: devfile/devworkspace-controller
         dockerfile: ./build/Dockerfile
-        tags: next
+        tags: v1alphax
         tag_with_sha: true
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ SHELL := bash
 
 OPERATOR_SDK_VERSION = v0.17.0
 NAMESPACE ?= devworkspace-controller
-IMG ?= quay.io/devfile/devworkspace-controller:next
+IMG ?= quay.io/devfile/devworkspace-controller:v1alphax
 TOOL ?= oc
 ROUTING_SUFFIX ?= 192.168.99.100.nip.io
 PULL_POLICY ?= Always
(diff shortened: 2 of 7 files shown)


.github/workflows/dockerimage-next.yml
```diff
@@ -27,5 +27,5 @@ jobs:
         registry: quay.io
         repository: devfile/devworkspace-controller
         dockerfile: ./build/Dockerfile
-        tags: next
+        tags: v1alphax
```
JPinkney, line 30: Since this branch represents the web terminal operator should we just publish to https://quay.io/repository/wto/web-terminal-operator instead?

sleshchenko: You reminded me the stuff I planed but forgot - default IMG value in Makefile.
Technically we can, but how good it will be if we'll use WTO image as default value for DWO v1alphax.

I like more the way we have now, when main branch of WTO depends on DWO:alpha https://github.com/redhat-developer/web-terminal-operator/blob/main/manifests/web-terminal.clusterserviceversion.yaml#L38
While quay.io/repository/wto/web-terminal-operator is supposed to host stuff built on brew.

JPinkney: My whole idea with https://issues.redhat.com/projects/WTO/issues/WTO-78 was to make it so that we no longer had to do anything manually after we finish a release (including pushing images built from brew). If we continue with using the images from brew we will always have manual steps after a release. Though, when we create a release I can probably just have a github action on web-terminal-operator that tags the current version of quay.io/devfile/devworkspace-controller:v1alphax and then push that to quay.io/wto/devworkspace-controller so it shouldn't be an issue

amisevsk (approved):
LGTM so long as @JPinkney's concerns are addressed.

.github/workflows/dockerimage-next.yml
sleshchenko: Well, for me WTO stuff must be built in brew, otherwise it's not WTO.
It's like if we build Che x branch and push it as CRW.

So, I'm proceeding with the current flow, which does not change anything and automates image build, and separately we can think more about how we can avoid manual post release activities

JPinkney: If we do it that way though the quay.io/wto bundle will only ever refer to the images on the redhat registry and never the ones in quay.io/wto. The only way to refer to the quay.io/wto images would to build another bundle in a github action, but if we do that then it's also not built in brew and also not technically wto

Issue #338, opened 2021-03-18
Clean up workspace resources when a workspace fails


amisevsk:
## Description

Currently, when a workspace fails to start, we simply mark it as failed and show a message to the user. However, any resources already created for the failed workspace stay on the cluster, which could potentially cause problems (e.g. a deployment that mounts a RWO PVC). When a workspace fails, we should at least delete or scale the deployment for the workspace to zero.

PR #362, opened 2021-03-29
cleaning up workspace on failure

tinakurian:
### What does this PR do?
Cleans up cluster deployments by scaling them to 0. 

https://github.com/devfile/devworkspace-operator/issues/338

### Is it tested? How?
1. Applied a sample yaml (Theia)
2. Wait for devworkspace to be running
3. Add invalid storage-type label

The result: 
image (7)

image (8)

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -171,6 +171,11 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	flattenedWorkspace, err := flatten.ResolveDevWorkspace(&workspace.Spec.Template, flattenHelpers)
 	if err != nil {
 		reqLogger.Info("DevWorkspace start failed")
+		// Clean up cluster deployment
+		err := provision.ScaleDeploymentToZero(workspace, r.Client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
 		// TODO: Handle error more elegantly
 		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error processing devfile: %s", err)
@@ -181,6 +186,11 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	storageProvisioner, err := storage.GetProvisioner(workspace)
 	if err != nil {
 		reqLogger.Info("DevWorkspace start failed")
+		// Clean up cluster deployment
+		err := provision.ScaleDeploymentToZero(workspace, r.Client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
 		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error provisioning storage: %s", err)
 		return reconcile.Result{}, nil
@@ -197,6 +207,11 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	devfilePodAdditions, err := containerlib.GetKubeContainersFromDevfile(&workspace.Spec.Template)
 	if err != nil {
 		reqLogger.Info("DevWorkspace start failed")
+		// Clean up cluster deployment
+		err := provision.ScaleDeploymentToZero(workspace, r.Client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
 		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error processing devfile: %s", err)
 		return reconcile.Result{}, nil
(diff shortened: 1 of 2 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -330,6 +337,15 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	return reconcile.Result{}, nil
 }
 
+func cleanupWorkspace(ctx context.Context, workspace *devworkspace.DevWorkspace, client client.Client, returnError error) (reconcile.Result, error) {
+	// Clean up cluster deployment before starting job to avoid mounting issues for RWO PVCs
+	err := provision.ScaleDeploymentToZero(ctx, workspace, client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, returnError
```
tinakurian, line 346: Not sure i really like this idea of passing in a previous error so we can return it if there was no other error. Is there a better way?

controllers/workspace/provision/deployment.go
```diff
@@ -140,6 +140,31 @@ func DeleteWorkspaceDeployment(ctx context.Context, workspace *devworkspace.DevW
...
+func ScaleDeploymentToZero(ctx context.Context, workspace *devworkspace.DevWorkspace, client runtimeClient.Client) error {
+	defaultRetry := time.Millisecond * 100
+	timeout := time.Second * 60
+	err := wait.Poll(defaultRetry, timeout, func() (done bool, err error) {
+		clusterDeployment, err := getClusterDeployment(common.DeploymentName(workspace.Status.WorkspaceId), workspace.Namespace, client)
+		if err != nil {
+			return false, err
+		}
+		if clusterDeployment == nil {
+			return false, nil
```
amisevsk, line 153: Shouldn't this be `true, nil`? `getClusterDeployment` will only return `nil, nil` if it gets an `isNotFound` error from the cluster, i.e. if the deployment doesn't exist. It looks like this will spin for 60 seconds if a deployment does not exist.

controllers/workspace/provision/deployment.go
```diff
@@ -140,6 +140,31 @@ func DeleteWorkspaceDeployment(ctx context.Context, workspace *devworkspace.DevW
 	return true, nil
 }
 
+// ScaleDeploymentToZero scales the cluster deployment to zero
+func ScaleDeploymentToZero(ctx context.Context, workspace *devworkspace.DevWorkspace, client runtimeClient.Client) error {
+	defaultRetry := time.Millisecond * 100
+	timeout := time.Second * 60
+	err := wait.Poll(defaultRetry, timeout, func() (done bool, err error) {
```
amisevsk, line 147: I'm not sure we can `wait` within the reconcile loop. This risks tying up a controller thread for 60 seconds when most reconciles take less than a second.

We should instead give the scale down one shot, and if it fails, requeue a reconcile to try it again.

controllers/workspace/provision/deployment.go
```diff
@@ -140,6 +140,31 @@ func DeleteWorkspaceDeployment(ctx context.Context, workspace *devworkspace.DevW
...
+		if clusterDeployment == nil {
+			return false, nil
+		}
+
+		replicaCount := int32(0)
+		clusterDeployment.Spec.Replicas = &replicaCount
+
+		err = client.Update(ctx, clusterDeployment)
+		if err != nil {
+			return false, err
```
amisevsk, line 161: We may want to gobble a conflict exception here, if we're going with polling. Otherwise there's no cases where we retry the scale-down (apart from the case above, where no deployment exists)

controllers/workspace/devworkspace_controller.go
tinakurian: fixed!

controllers/workspace/devworkspace_controller.go
```diff
@@ -174,7 +174,8 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		reconcileStatus.Phase = devworkspace.WorkspaceStatusFailed
 		// TODO: Handle error more elegantly
 		reconcileStatus.Conditions[devworkspace.WorkspaceFailedStart] = fmt.Sprintf("Error processing devfile: %s", err)
-		return reconcile.Result{}, nil
+		// Clean up cluster deployment
+		return cleanupWorkspace(ctx, workspace, r.Client, nil)
```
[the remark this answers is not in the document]
tinakurian: Makes sense, fixed!

controllers/workspace/provision/deployment.go
tinakurian: Yeah that makes sense, i've removed the polling and so this should be fixed!

controllers/workspace/provision/deployment.go
tinakurian: fixed!

controllers/workspace/provision/deployment.go
tinakurian: fixed by removing polling :smile:

amisevsk (approved):
LGTM :+1:

JPinkney (approved):
LGTM

sleshchenko (approved):
Be in rush with merging this and make @amisevsk in charge of adapting his PR :-D

controllers/workspace/provision/deployment.go
```diff
@@ -140,6 +138,28 @@ func DeleteWorkspaceDeployment(ctx context.Context, workspace *devworkspace.DevW
...
+	}
+
+	if clusterDeployment == nil {
+		return nil
+	}
+
+	replicaCount := int32(0)
+	clusterDeployment.Spec.Replicas = &replicaCount
+
+	err = client.Update(ctx, clusterDeployment)
```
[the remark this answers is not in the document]
sleshchenko: I should be available
```
// ScaleDeploymentToZero scales the cluster deployment to zero
func ScaleDeploymentToZero(ctx context.Context, workspace *devworkspace.DevWorkspace, client runtimeClient.Client) error {
	patch := []byte(`{"spec":{"replicas": 0}}`)
	err := client.Patch(context.Background(), &appsv1.Deployment{
			Name:      common.DeploymentName(workspace.Status.WorkspaceId),
		},
	}, runtimeClient.RawPatch(types.StrategicMergePatchType, patch))

	if err != nil {
		return err
	}

	return nil
}
```

@amisevsk WDYT about patching here?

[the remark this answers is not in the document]
amisevsk: I've never been able to figure out the semantics of `patch`; sometimes it does what I expect other times it does not :smile:

amisevsk (approved):
Merge away -- I can copy the 
```go
err := provision.ScaleDeploymentToZero(workspace, r.Client)
if err != nil {
}
```
lines into my PR when I rebase.

PR #366, opened 2021-03-31
Fail async workspaces when async storage is not configured

amisevsk:
### What does this PR do?
Fail workspaces early if images required for async storage are not available. 

Graceful failure when `RELATED_IMAGE` for async storage are blank

### Is it tested? How?
1. Modify `manager.yaml` to remove `RELATED_IMAGE_async_storage_server` and/or `RELATED_IMAGE_async_storage_sidecar`
2. Try to start async workspace: `kubectl apply -f samples/with-k8s-ref/async-theia-next.yaml`
3. Workspace should be failed with message `Error provisioning storage: asynchronous storage server image is not configured`


pkg/provision/storage/asyncStorage.go
```diff
@@ -249,3 +256,13 @@ func (*AsyncStorageProvisioner) getAsyncWorkspaceCount(api provision.ClusterAPI)
 	}
 	return started, total, nil
 }
+
+func checkConfigured() error {
+	if images.GetAsyncStorageServerImage() == "" {
+		return fmt.Errorf("asynchronous storage server image is not configured")
```
sleshchenko, line 262: user does not have the ability to fix that issue. Maybe it makes sense to add some reference to admin, like `Contact admin` ?

sleshchenko (approved):
Haven't tested but changes LGTM.
Just one proposal

pkg/provision/storage/asyncStorage.go
amisevsk: Added a note about contacting an admin higher up to avoid a Go warning about punctuation in a string :upside_down_face:

amisevsk:
Requires https://github.com/devfile/devworkspace-operator/pull/367 to test properly

JPinkney (approved):

PR #370, opened 2021-04-01
Return back to v1beta1 CRD API

sleshchenko:
### What does this PR do?
Return back to v1beta1 CRD API.
It's draft since probably we need to workaround:
> Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
+ currently kustomize fails with:
```
Error: no matches for IdId apiextensions.k8s.io_v1_CustomResourceDefinition|~X|devworkspaces.workspace.devfile.io; failed to find unique target for patch apiextensions.k8s.io_v1_CustomResourceDefinition|devworkspaces.workspace.devfile.io
```
Changing https://github.com/devfile/devworkspace-operator/blob/main/deploy/templates/service-ca/crd_webhooks_patch.yaml to v1beta1 does not help.

It's DWO part for https://github.com/eclipse/che/issues/19472

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,8 +224,10 @@ endif
 ### manifests: Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=role webhook paths="./..." \
+			crd:crdVersions=v1beta1 \
 			output:crd:artifacts:config=deploy/templates/crd/bases \
 			output:rbac:artifacts:config=deploy/templates/components/rbac
+	rm -rf ./config
 	patch/patch_crds.sh
 
 ### fmt: Run go fmt against code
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -224,8 +224,10 @@ endif
 ### manifests: Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=role webhook paths="./..." \
+			crd:crdVersions=v1beta1 \
 			output:crd:artifacts:config=deploy/templates/crd/bases \
 			output:rbac:artifacts:config=deploy/templates/components/rbac
+	rm -rf ./config
```
sleshchenko, line 230: controller-gen creates v1 at config folder, I did not figure out how to disable it, so just came to removing this file.

sleshchenko:
on the latest cabal we agreed that instead Che and CRW should drop old openshift versions

Issue #205, opened 2020-12-01
Projects clone should happen before other components are started


l0rd:
## What problem do we want to address?

The ones described [here](https://github.com/eclipse/che/issues/15973)

## How are we going to address it?

- [ ] When a DevWorkspace CR has one or more projects, those should be cloned before any other component is deployed
- [ ] Kubernetes secrets containing Git credentials and SSH key pairs should be used to clone password protected repositories

l0rd:
We may want to use https://github.com/kubernetes/git-sync

sleshchenko:
> We should git clone all projects before a DevWorkspace is started.

Depending on how big the repository[ies] are, it may influences workspace start-up time essentially. Some examples I got locally:

<details>
<summary>Eclipse Che: **1m13s** with 1.76MiB/s at the last moment</summary>

![Screenshot_20201204_144858](https://user-images.githubusercontent.com/5887312/101166055-52ce7100-3640-11eb-96e4-4b9c25cd4a04.png)

</details>

<details>
<summary>Che Plugin Broker: **6s** with 900KiB/s at the last moment</summary>

![Screenshot_20201204_145134](https://user-images.githubusercontent.com/5887312/101166149-7f828880-3640-11eb-9fe6-a844a034ff62.png)

</details>

<details>
<summary>Kubernetes Go Client: **18s** with 1.41MiB/s at the last moment</summary>

![Screenshot_20201204_145052](https://user-images.githubusercontent.com/5887312/101166325-c83a4180-3640-11eb-83b0-ede055fcd6ec.png)

</details>

<details>
<summary>ODO: **1m36s** with 1.48MiB/s at the last moment</summary>

![Screenshot_20201204_145610](https://user-images.githubusercontent.com/5887312/101166387-e011c580-3640-11eb-895b-0501be9e9c7d.png)

</details

sleshchenko:
It seems to be related to 

> Start an editor only workspace and asynchronously update the CR and "recreate" the workspace pod
 
item from https://github.com/devfile/devworkspace-operator/issues/209

amisevsk:
[...]

PR #371, opened 2021-04-01
Clone projects on start

amisevsk:
### What does this PR do?
Implements cloning projects for any DevWorkspace by adding an init-container that parses the projects out of a DevWorkspaceTemplateSpec and clones each one.

There are two parts to this PR:
1. Automatically injecting an init container into the flattened DevWorkspace (via container component, apply command, and prestart event)
2. A standalone binary that reads a DevWorkspace and clones projects

The cloning binary supports most of the features for projects specified by `git` references:
- Multiple projects
- Multiple remotes in one project
- Specific checkoutFrom for branch/tag/revision

PR is a WIP since I'm not sure if we want to keep the separate binary within this repo or create another one for it. A built image of the binary is available at `quay.io/amisevsk/project-clone:dev`, as is used by the updated deployment templates.

Building your own project-clone binary is done by executing
```
docker build -t <image:tag> -f ./project-clone/Dockerfile .
```
from the root of this repo.

Closes https://github.com/devfile/devworkspace-operator/issues/205

### Is it tested? How?
Included is a sample file `samples/plain.yaml` which is a plain container with multiple projects that tests various features of Devfile 2.0 projects. Once running, you'll need to exec into the container and look at the filesystem to ensure everything is in place.

Note that this PR currently does not work with Theia due to https://github.com/eclipse/che/issues/19489

### TODO before removing WIP tag:
- [x] Decide on a quay repo to contain the project-clone image -- quay.io/devfile/project-clone
- [x] Add building project-clone into default CI jobs -- done
- [x] Cover project-clone in PR check (build image) -- done


amisevsk:
Testing on OpenShift works via the DW:
```yaml
kind: DevWorkspace
metadata:
  name: theia
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                image: quay.io/crw_pr/che-theia:1041
            - name: remote-runtime-injector
              container:
                image: quay.io/crw_pr/che-theia-endpoint-runtime-binary:1041
    commands:
      - id: say-hello
        exec:
          component: theia
```

benoitf:
Hello, what is not yet clear to me is when will be executed the logic that will analyze `che-theia-plugins.yaml` and `.vscode/extensions.json` file

if project cloning is happening at init container, it means that it's too late to analyse these files as templates are already there or it means we'll have a workspace restart if we update just after the DevWorkspace templates (or it would mean that dashboard would clone as well the files before ?)

sleshchenko:
Dashboard or any other clients, like chectl, won't do clonning IMHO

> if project cloning is happening at init container, it means that it's too late to analyse these files as templates are already there or it means we'll have a workspace restart if we update just after the DevWorkspace templates (or it would mean that dashboard would clone as well the files before ?)

project clonning and resolving devworkspace are pretty different stuff.

Project clonning is pretty generic for DWO level,
but resolving DW according to file system, is really tight to supported editors and their plugins models.

There are multiple ways to go, for example:
- Che creates DW with just DevWorkspace resolver component.
- Projects are clonned as init container.
- Resolver analyzes file system and update DW and related DWT.
- DevWorkspace is restarted.
- Theia with all plugins are run.

So, it means DW will be restarted, but the first run will be faster.

But not decision is taken at this point. It's just one of the possible ways.

amisevsk:
> if project cloning is happening at init container, it means that it's too late to analyse these files as templates are already there or it means we'll have a workspace restart if we update just after the DevWorkspace templates (or it would mean that dashboard would clone as well the files before ?)

This way of defining plugins will always require a workspace restart unless whatever creates the devworkspace does a clone and figures out the templates needed. It's not something we can handle in the Operator.

project-clone/rhel.Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Build the manager binary
+FROM registry.redhat.io/rhel8/go-toolset:1.13.4-27 as builder
```
sleshchenko, line 2: can't we avoid having two dockerfiles?

project-clone/main.go
```diff
@@ -0,0 +1,278 @@
...
+				log.Printf("Failed to set up remotes for project %s: %s", project.Name, err)
+				os.Exit(1)
+			}
+		} else {
+			log.Printf("Nothing to do for project %s", project.Name)
+		}
+	}
+}
+
+func checkProjectState(project *dw.Project) (needClone, needRemotes bool, err error) {
```
sleshchenko, line 89: main.go is pretty big. I wonder if it makes sense to move lower-level logic to a separate file.

sleshchenko (commented):
Changes LGTM in general, starting testing it

samples/plain.yaml
```diff
@@ -0,0 +1,31 @@
+kind: DevWorkspace
```
sleshchenko, line 1: with the first try this sample, I got 
```
workspace4aba76e40f844658-678777584-6v79c   0/1     Init:OOMKilled   1          29s
```
But then it got better
```
workspace4aba76e40f844658-678777584-6v79c   1/1     Running   0          69s
```
I used RHPDS

samples/plain.yaml
```diff
@@ -0,0 +1,31 @@
...
+    projects:
+      - name: web-nodejs-sample
+        git:
+          remotes:
+            origin: "https://github.com/che-samples/web-nodejs-sample.git"
+      - name: devworkspace-operator
+        git:
+          checkoutFrom:
+            remote: amisevsk
+            revision: clone-projects-on-start
```
sleshchenko, line 18: checkout does not seem to work as expected, I got:
```
> git status

On branch main
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   deploy/deployment/kubernetes/objects/devworkspacetemplates.workspace.devfile.io.CustomResourceDefinition.yaml                                                                                                                           
        deleted:    deploy/deployment/openshift/combined.yaml
        deleted:    deploy/deployment/openshift/objects/devworkspace-controller-configmap.ConfigMap.yaml
        deleted:    deploy/deployment/openshift/objects/devworkspace-controller-edit-workspaces.ClusterRole.yaml
        .....
```

sleshchenko:
checkoutFrom needs fix but clonning and setting up remotes work fine, tested with theia-next and plain samples.

pkg/constants/metadata.go
```diff
@@ -0,0 +1,61 @@
...
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package constants
+
```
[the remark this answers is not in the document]
amisevsk: I'm not sure what you're referring to -- that link just takes me to the top of the PR changes page.

build/rhel.Dockerfile
```diff
@@ -56,6 +56,9 @@ RUN  /usr/local/bin/user_setup
 
 USER ${USER_UID}
 
+COPY build/bin /usr/local/bin
```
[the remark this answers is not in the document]
amisevsk: I'm not sure what you're referring to -- that link just takes me to the `devworkspace_controller.go` file.

pkg/library/projects/clone.go
```diff
@@ -0,0 +1,74 @@
...
+					CpuLimit:      constants.ProjectCloneCPULimit,
+					CpuRequest:    constants.ProjectCloneCPURequest,
+					MountSources:  &boolTrue,
+				},
+			},
+		},
+	}
+}
+
+func getProjectClonerCommand() *dw.Command {
```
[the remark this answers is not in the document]
amisevsk: I'm not sure this is clearer

project-clone/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+var (
+	projectsRoot              string
+	flattenedDevWorkspacePath string
+)
+
+func init() {
+	projectsRoot = os.Getenv(constants.ProjectsRootEnvVar)
+	if projectsRoot == "" {
+		log.Printf("Required environment variable %s is unset", constants.ProjectsRootEnvVar)
```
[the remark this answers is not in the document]
amisevsk: No, we should `os.Exit(1)` since these env vars are required :)

Good catch.

project-clone/main.go
```diff
@@ -0,0 +1,278 @@
...
+		return true, true, nil
+	}
+
+	for projRemote, projRemoteURL := range project.Git.Remotes {
+		gitRemote, err := repo.Remote(projRemote)
+		if err != nil {
+			if err != git.ErrRemoteNotFound {
+				return false, false, fmt.Errorf("error reading remotes: %s", err)
+			}
+			log.Printf("Encountered error trying to read remotes for project %s: %s", project.Name, err)
```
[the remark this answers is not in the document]
amisevsk: Yeah this is done strangely -- I'll fix.

project-clone/main.go
amisevsk: Your wish is my command :D

sleshchenko: Only when you think it makes sense ;)

project-clone/main.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/devfile/devworkspace-operator/pkg/provision/metadata"
+	"github.com/devfile/devworkspace-operator/project-clone/internal"
+)
+
+var (
+	flattenedDevWorkspacePath string
+)
+
+func init() {
+	flattenedDevWorkspacePath = os.Getenv(metadata.FlattenedDevfileMountPathEnvVar)
```
sleshchenko, line 33: it's used only once in main. Any reason to keep it in init?

build/rhel.Dockerfile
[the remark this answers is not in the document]
sleshchenko: it should be better
https://github.com/amisevsk/devworkspace-operator/blob/clone-projects-on-start/build/rhel.Dockerfile#L55

Check that 
```Dockerfile
COPY build/bin /usr/local/bin
RUN  /usr/local/bin/user_setup
```
appears twice

amisevsk:
Commit 623e28214abf7cd08041576e79662ec07a628e5b adds CI functions for testing/pushing the project-clone image.

One thing I haven't done yet (and should probably be a separate PR) is to update the `RELATED_IMAGE` env var for project clone in releases (i.e. in `make-release.sh`) to match the released version. This is also not done for `RELATED_IMAGE_devworkspace_webhook_server` either, though.

project-clone/rhel.Dockerfile
amisevsk: Dockerfiles are merged and the new `project-clone/Dockerfile` is the `rhel` one :)

build/rhel.Dockerfile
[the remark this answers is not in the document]
amisevsk: Ah, no idea how that happened :thinking:

samples/plain.yaml
amisevsk: Should be fixed by bumping the memory limit to `300Mi`, but it's strange that it would require _so_ much memory.

samples/plain.yaml
amisevsk: Should be fixed now.

samples/plain.yaml
sleshchenko: I wonder if bigger projects would require even more memory, which means users need to configure project clone memory limit with devfile attributes or some another way. But I'm OK to postpone it into a dedicated issue https://github.com/devfile/devworkspace-operator/issues/397

pkg/constants/metadata.go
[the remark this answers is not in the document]
sleshchenko: Reference to PR changes is not reliable, don't use it :-D

I meant:
```go
var ControllerAppLabels = func() map[string]string {
	return map[string]string{
		"app.kubernetes.io/name":    "devworkspace-controller",
		"app.kubernetes.io/part-of": "devworkspace-operator",
	}
}
```
https://github.com/amisevsk/devworkspace-operator/blob/clone-projects-on-start/pkg/constants/constants.go#L17
which is still in constants but metadata seems to be a better place.

samples/plain.yaml
```diff
@@ -0,0 +1,31 @@
+kind: DevWorkspace
+apiVersion: workspace.devfile.io/v1alpha2
+metadata:
+  name: plain
```
sleshchenko, line 4: My try to test it on RHPDS was unsuccessful: 
```
2021/04/27 08:39:30 Read DevWorkspace at /devworkspace-metadata/flattened.devworkspace.yaml
2021/04/27 08:39:30 Cloning project web-nodejs-sample to web-nodejs-sample
2021/04/27 08:39:30 failed to clone project web-nodejs-sample: failed to git clone from https://github.com/che-samples/web-nodejs-sample.git: mkdir /projects/web-nodejs-sample: permission denied
```

sleshchenko: Well, seems the reason I face it - I used default namespace.
After I switched to devworkspace-controller(I think any different from default), it works.

sleshchenko (approved):
Tested plain and theia-next samples. Works as expected for me.

samples/plain.yaml
sleshchenko: P.S. BTW My devworkspace did not get failed in such case, please consider adding init container statuses into https://github.com/tinakurian/devworkspace-operator/blob/132acb9016ea6bac99c5416fd736dd2f265ff222/controllers/workspace/provision/deployment.go#L349
```yaml
status:
  containerStatuses:
    ...
  - image: quay.io/eclipse/che-theia:next
    imageID: ""
    lastState: {}
    name: theia-ide
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        reason: PodInitializing
  hostIP: 10.0.189.15
  initContainerStatuses:
  - image: quay.io/eclipse/che-theia-endpoint-runtime-binary:next
    imageID: ""
    lastState: {}
    name: remote-runtime-injector
    ready: false
    restartCount: 0
    state:
      waiting:
        reason: PodInitializing
  phase: Pending
  qosClass: Burstable
  startTime: "2021-04-27T09:37:31Z"

```

tinakurian (approved):

samples/plain.yaml
amisevsk: Done

Issue #353, opened 2021-03-25
DevWorkspace should be failed when its pod has CrashLoopBackOff status


sleshchenko:
Currently, when pod fails, devworkspace will be starting forever.
![Screenshot_20210325_153649](https://user-images.githubusercontent.com/5887312/112481990-418caa00-8d80-11eb-9213-61b998165c1d.png)
we need to better handle that and DevWorkspace should be failed when its pod has CrashLoopBackOff status.
We may limit to restarts to some number to give container a chance to recover, but in any case it would be great if we report CrashLoopBackOff even before we Failed workspace.

PR #375, opened 2021-04-06
DevWorkspace should be failed when its pod has CrashLoopBackOff status

tinakurian:
### What does this PR do?
This PR fetches the pod and checks the the status to see if the pod has the of CrashLoopBackOff status. If so the devworkspace is updated to have the Failed phase.

https://github.com/devfile/devworkspace-operator/issues/353

### Is it tested? How?
Updated web-terminal-custom-tooling.yaml to force CrashLoopBackOff

image

### DIFF
--- a/controllers/workspace/condition.go
+++ b/controllers/workspace/condition.go
@@ -40,17 +40,42 @@ type workspaceConditions struct {
 }
 
 func (c *workspaceConditions) setConditionTrue(conditionType dw.DevWorkspaceConditionType, msg string) {
+	if c.conditions == nil {
+		c.conditions = map[dw.DevWorkspaceConditionType]dw.DevWorkspaceCondition{}
+	}
+
 	c.conditions[conditionType] = dw.DevWorkspaceCondition{
 		Status:  corev1.ConditionTrue,
 		Message: msg,
 	}
 }
 
+func (c *workspaceConditions) setCondition(conditionType dw.DevWorkspaceConditionType, dwCondition dw.DevWorkspaceCondition) {
+	if c.conditions == nil {
+		c.conditions = map[dw.DevWorkspaceConditionType]dw.DevWorkspaceCondition{}
+	}
+
+	c.conditions[conditionType] = dwCondition
+}
+
 func (c *workspaceConditions) setConditionFalse(conditionType dw.DevWorkspaceConditionType, msg string) {
+	if c.conditions == nil {
+		c.conditions = map[dw.DevWorkspaceConditionType]dw.DevWorkspaceCondition{}
+	}
+
 	c.conditions[conditionType] = dw.DevWorkspaceCondition{
 		Status:  corev1.ConditionFalse,
 		Message: msg,
 	}
+}
+
+func getConditionByType(conditions []dw.DevWorkspaceCondition, t dw.DevWorkspaceConditionType) *dw.DevWorkspaceCondition {
+	for _, condition := range conditions {
+		if condition.Type == t {
+			return &condition
+		}
+	}
+	return nil
 }
 
 // getFirstFalse checks current conditions in a set order (defined by conditionOrder) and returns the first
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -36,12 +36,15 @@ import (
 	coputil "github.com/redhat-cop/operator-utils/pkg/util"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 )
@@ -117,6 +120,18 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	// Stop failed workspaces
+	if workspace.Status.Phase == dw.DevWorkspaceStatusFailed && workspace.Spec.Started {
+		patch := []byte(`{"spec":{"started": false}}`)
+		err := r.Client.Patch(context.Background(), workspace, client.RawPatch(types.MergePatchType, patch))
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Requeue reconcile to stop workspace
+		return reconcile.Result{Requeue: true}, nil
+	}
+
 	// Handle stopped workspaces
 	if !workspace.Spec.Started {
 		timing.ClearAnnotations(workspace)
@@ -125,7 +140,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	}
 
 	// Prepare handling workspace status and condition
-	reconcileStatus := initCurrentStatus()
+	reconcileStatus := currentStatus{phase: dw.DevWorkspaceStatusStarting}
 	clusterWorkspace := workspace.DeepCopy()
 	timingInfo := map[string]string{}
 	timing.SetTime(timingInfo, timing.DevWorkspaceStarted)
@@ -282,7 +297,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	deploymentStatus := provision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI)
 	if !deploymentStatus.Continue {
 		if deploymentStatus.FailStartup {
-			return r.failWorkspace(workspace, fmt.Sprintf("DevWorkspace spec is invalid: %s", deploymentStatus.Err), reqLogger, &reconcileStatus)
+			return r.failWorkspace(workspace, deploymentStatus.Info(), reqLogger, &reconcileStatus)
 		}
 		reqLogger.Info("Waiting on deployment to be ready")
 		reconcileStatus.setConditionFalse(DeploymentReady, "Waiting for workspace deployment")
@@ -307,55 +322,68 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 }
 
 func (r *DevWorkspaceReconciler) stopWorkspace(workspace *dw.DevWorkspace, logger logr.Logger) (reconcile.Result, error) {
+	status := currentStatus{phase: dw.DevWorkspaceStatusStopping}
+	if workspace.Status.Phase == dw.DevWorkspaceStatusFailed {
+		status.phase = dw.DevWorkspaceStatusFailed
+		failedCondition := getConditionByType(workspace.Status.Conditions, dw.DevWorkspaceFailedStart)
+		if failedCondition != nil {
+			status.setCondition(dw.DevWorkspaceFailedStart, *failedCondition)
+		}
+	}
+
+	stopped, err := r.doStop(workspace, logger)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if stopped {
+		if status.phase != dw.DevWorkspaceStatusFailed {
+			status.phase = dw.DevWorkspaceStatusStopped
+		}
+	}
+	return r.updateWorkspaceStatus(workspace, logger, &status, reconcile.Result{}, nil)
+}
+
+func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.Logger) (stopped bool, err error) {
 	workspaceDeployment := &appsv1.Deployment{}
 	namespaceName := types.NamespacedName{
 		Name:      common.DeploymentName(workspace.Status.DevWorkspaceId),
 		Namespace: workspace.Namespace,
 	}
-	status := &currentStatus{}
-	err := r.Get(context.TODO(), namespaceName, workspaceDeployment)
+	err = r.Get(context.TODO(), namespaceName, workspaceDeployment)
 	if err != nil {
 		if k8sErrors.IsNotFound(err) {
-			status.phase = dw.DevWorkspaceStatusStopped
-			return r.updateWorkspaceStatus(workspace, logger, status, reconcile.Result{}, nil)
+			return true, nil
 		}
-		return reconcile.Result{}, err
+		return false, err
 	}
 
-	status.phase = dw.DevWorkspaceStatusStopping
 	replicas := workspaceDeployment.Spec.Replicas
 	if replicas == nil || *replicas > 0 {
 		logger.Info("Stopping workspace")
-		patch := client.MergeFrom(workspaceDeployment.DeepCopy())
-		var replicasZero int32 = 0
-		workspaceDeployment.Spec.Replicas = &replicasZero
-		err = r.Patch(context.TODO(), workspaceDeployment, patch)
+		err = provision.ScaleDeploymentToZero(workspace, r.Client)
 		if err != nil && !k8sErrors.IsConflict(err) {
-			return reconcile.Result{}, err
+			return false, err
 		}
-		return r.updateWorkspaceStatus(workspace, logger, status, reconcile.Result{}, nil)
+		return false, nil
 	}
 
 	if workspaceDeployment.Status.Replicas == 0 {
-		logger.Info("Workspace stopped")
-		status.phase = dw.DevWorkspaceStatusStopped
+		return true, nil
 	}
-	return r.updateWorkspaceStatus(workspace, logger, status, reconcile.Result{}, nil)
+	return false, nil
 }
 
 // failWorkspace marks a workspace as failed by setting relevant fields in the status struct.
 // These changes are not synced to cluster immediately, and are intended to be synced to the cluster via a deferred function
 // in the main reconcile loop. If needed, changes can be flushed to the cluster immediately via `updateWorkspaceStatus()`
 func (r *DevWorkspaceReconciler) failWorkspace(workspace *dw.DevWorkspace, msg string, logger logr.Logger, status *currentStatus) (reconcile.Result, error) {
-	logger.Info("Workspace start failed")
-	// Clean up cluster deployment
-	err := provision.ScaleDeploymentToZero(workspace, r.Client)
-	if err != nil {
-		// Return error here without setting phase to failed in order to retry cleaning the deployment
-		return reconcile.Result{}, err
-	}
+	logger.Info("DevWorkspace failed to start: " + msg)
 	status.phase = dw.DevWorkspaceStatusFailed
 	status.setConditionTrue(dw.DevWorkspaceFailedStart, msg)
+	if workspace.Spec.Started {
+		return reconcile.Result{Requeue: true}, nil
+	}
 	return reconcile.Result{}, nil
 }
 
@@ -388,6 +416,32 @@ func getWorkspaceId(instance *dw.DevWorkspace) (string, error) {
 	return "workspace" + strings.Join(strings.Split(uid.String(), "-")[0:3], ""), nil
 }
 
+// Mapping the pod to the devworkspace
+func dwRelatedPodsHandler() handler.EventHandler {
+	podToDW := func(mapObj handler.MapObject) []reconcile.Request {
+		meta := mapObj.Meta
+		labels := meta.GetLabels()
+		if _, ok := labels[constants.DevWorkspaceNameLabel]; !ok {
+			return nil
+		}
+
+		//If the dewworkspace label does not exist, do no reconcile
+		if _, ok := labels[constants.DevWorkspaceIDLabel]; !ok {
+			return nil
+		}
+
+		return []reconcile.Request{
+			{
+				NamespacedName: types.NamespacedName{
+					Name:      labels[constants.DevWorkspaceNameLabel],
+					Namespace: meta.GetNamespace(),
+				},
+			},
+		}
+	}
+	return &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(podToDW)}
+}
+
 func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// TODO: Set up indexing https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup
 	return ctrl.NewControllerManagedBy(mgr).
@@ -399,6 +453,8 @@ func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}).
 		Owns(&batchv1.Job{}).
 		Owns(&controllerv1alpha1.DevWorkspaceRouting{}).
+		Watches(&source.Kind{Type: &corev1.Pod{}}, dwRelatedPodsHandler()).
 		WithEventFilter(predicates).
+		WithEventFilter(podPredicates).
 		Complete(r)
 }
--- a/controllers/workspace/predicates.go
+++ b/controllers/workspace/predicates.go
@@ -14,6 +14,9 @@ package controllers
 
 import (
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
+
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -30,9 +33,11 @@ var predicates = predicate.Funcs{
 		if !ok {
 			return true
 		}
+
 		if newObj.Status.Phase != dw.DevWorkspaceStatusFailed {
 			return true
 		}
+
 		oldObj, ok := ev.ObjectOld.(*dw.DevWorkspace)
 		if !ok {
 			// Should never happen
@@ -42,8 +47,37 @@ var predicates = predicate.Funcs{
 		if newObj.GetDeletionTimestamp() != nil {
 			return true
 		}
+
 		// Trigger a reconcile on failed workspaces if spec is updated.
 		return !equality.Semantic.DeepEqual(oldObj.Spec, newObj.Spec)
 	},
 	GenericFunc: func(_ event.GenericEvent) bool { return true },
+}
+
+var podPredicates = predicate.Funcs{
+	UpdateFunc: func(ev event.UpdateEvent) bool {
+		newObj, ok := ev.ObjectNew.(*corev1.Pod)
+		if !ok {
+			//that's no Pod. Let other predicates decide
+			return true
+		}
+
+		oldObj, ok := ev.ObjectOld.(*corev1.Pod)
+		if !ok {
+			// Should never happen
+			return true
+		}
+
+		// If the devworkspace label does not exist, do no reconcile
+		if _, ok = newObj.Labels[constants.DevWorkspaceNameLabel]; !ok {
+			return false
+		}
+
+		// Trigger a devworkspace reconcile if related pod status is updated.
+		if equality.Semantic.DeepEqual(oldObj.Status, newObj.Status) {
+			return false
+		}
+
+		return true
+	},
 }
--- a/controllers/workspace/provision/deployment.go
+++ b/controllers/workspace/provision/deployment.go
@@ -27,6 +27,7 @@ import (
 	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	appsv1 "k8s.io/api/apps/v1"
@@ -35,8 +36,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	ContainerCrashLoopBackOffReason = "CrashLoopBackOff"
+	ContainerImagePullErr           = "ImagePullBackOff"
+	ContainerCreateErr              = "CreateContainerError"
 )
 
 type DeploymentProvisioningStatus struct {
@@ -61,7 +69,6 @@ func SyncDeploymentToCluster(
 	podAdditions []v1alpha1.PodAdditions,
 	saName string,
 	clusterAPI ClusterAPI) DeploymentProvisioningStatus {
-
 	// [design] we have to pass components and routing pod additions separately because we need mountsources from each
 	// component.
 	specDeployment, err := getSpecDeployment(workspace, podAdditions, saName, clusterAPI.Scheme)
@@ -73,7 +80,6 @@ func SyncDeploymentToCluster(
 			},
 		}
 	}
-
 	clusterDeployment, err := getClusterDeployment(specDeployment.Name, workspace.Namespace, clusterAPI.Client)
 	if err != nil {
 		return DeploymentProvisioningStatus{
@@ -128,7 +134,21 @@ func SyncDeploymentToCluster(
 		}
 	}
 
-	return DeploymentProvisioningStatus{}
+	failureMsg, checkErr := checkFailedPods(workspace, clusterAPI)
+	if checkErr != nil {
+		return DeploymentProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{
+				Err: checkErr,
+			},
+		}
+	}
+
+	return DeploymentProvisioningStatus{
+		ProvisioningStatus: ProvisioningStatus{
+			FailStartup: failureMsg != "",
+			Message:     failureMsg,
+		},
+	}
 }
 
 // DeleteWorkspaceDeployment deletes the deployment for the DevWorkspace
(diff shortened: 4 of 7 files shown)


controllers/workspace/provision/deployment.go
```diff
@@ -107,6 +109,35 @@ func SyncDeploymentToCluster(
...
+			return DeploymentProvisioningStatus{
+				ProvisioningStatus: ProvisioningStatus{Requeue: true},
+			}
+		}
+
+		for _, containerStatuses := range pod.Status.ContainerStatuses {
+			if containerStatuses.State.Waiting != nil && containerStatuses.State.Waiting.Reason == constants.ContainerCrashLoopBackOffReason {
+				return DeploymentProvisioningStatus{
+					ProvisioningStatus: ProvisioningStatus{
+						FailStartup: true,
```
sleshchenko, line 134: let's provide a message with a failure cause here, `some of devworkspace containers crashed`

sleshchenko (approved):
I'll test and do detailed review tomorrow

controllers/workspace/devworkspace_controller.go
```diff
@@ -385,6 +388,21 @@ func getWorkspaceId(instance *devworkspace.DevWorkspace) (string, error) {
...
+// Mapping the pod to the devworkspace
+func mapPodToDW(mapObj handler.MapObject) []reconcile.Request {
+	meta := mapObj.Meta
+	labels := meta.GetLabels()
+	if _, ok := labels[constants.WorkspaceIDLabel]; !ok {
+		return nil
+	}
+	return []reconcile.Request{
+		{NamespacedName: types.NamespacedName{
+			Name:      labels[constants.WorkspaceNameLabel],
```
amisevsk, line 400: We may want to verify that the label is not empty. I don't know what exactly happens when we enqueue a reconcile for an empty name.

Since the predicate should filter pod events when the pod does not have the workspaceID label, maybe we can replace the check above?

```
if _, ok := labels[constants.WorkspaceNameLabel]; !ok {
  return nil
}
```

controllers/workspace/provision/deployment.go
```diff
@@ -107,6 +109,35 @@ func SyncDeploymentToCluster(
 		}
 	}
 
+	// Get pod and update dw status on crashloopbackoff
+	podList, err := getPods(workspace, clusterAPI.Client)
+	if err != nil {
+		return DeploymentProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{
+				Requeue: true,
+				Err:     err,
```
amisevsk, line 118: If `err != nil`, the requeue is unnecessary (returning an error always triggers a requeue).

controllers/workspace/provision/deployment.go
```diff
@@ -107,6 +109,35 @@ func SyncDeploymentToCluster(
...
+			},
+		}
+	}
+
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodPending {
+			return DeploymentProvisioningStatus{
+				ProvisioningStatus: ProvisioningStatus{Requeue: true},
```
amisevsk, line 126: I don't know if we need to requeue here. If a pod is pending, either it will still be pending (status hasn't changed and the requeue has no effect) or the phase will have changed (which would trigger a new reconcile anyways).

controllers/workspace/provision/deployment.go
```diff
@@ -107,6 +109,35 @@ func SyncDeploymentToCluster(
...
+			}
+		}
+
+		for _, containerStatuses := range pod.Status.ContainerStatuses {
+			if containerStatuses.State.Waiting != nil && containerStatuses.State.Waiting.Reason == constants.ContainerCrashLoopBackOffReason {
```
amisevsk, line 131: This is an issue for a separate PR, but it'd be great to eventually check reasons against a list of "failure" states, e.g. `CrashLoopBackOff`, `ImagePullErr`, etc.

controllers/workspace/provision/deployment.go
amisevsk: We can even propagate the specific reason: `Container <containerName> has state CrashLoopBackOff`

amisevsk (commented):
Generally LGTM, I'll test once I'm out of `v1alpha1` land :)

controllers/workspace/devworkspace_controller.go
tinakurian: Yup! Makes sense! I've changed it!

controllers/workspace/provision/deployment.go
tinakurian: Ah yes! Sorry, removed the Requeue! Good catch!

controllers/workspace/provision/deployment.go
tinakurian: Yeah, that's a good point, i've removed it!

controllers/workspace/provision/deployment.go
tinakurian: Good idea! I've updated to add a message!

controllers/workspace/provision/deployment.go
sleshchenko: I faced this issue today. Due `ImagePullErr` my workspace was starting forever.
We should fail workspace in such case or at least propagate it as a message.

sleshchenko (commented):
It does not work yet, devworkspace is changing status forever because failed pod always triggers reconcile loop. Checking what we can do with it.

controllers/workspace/predicates.go
```diff
@@ -47,3 +51,31 @@ var predicates = predicate.Funcs{
...
+			return false
+		}
+
+		// Always reconcile if resource is deleted
+		if newObj.GetDeletionTimestamp() != nil {
+			return true
+		}
+
+		// Trigger a reconcile on failed workspaces if status is updated.
+		return !equality.Semantic.DeepEqual(oldObj.Status, newObj.Status)
```
sleshchenko, line 79: that's going to trigger reconcile for failed DW and as the result DW will be restarted.

I tried to rework it in the following way, but somebody else still trigger reconcile loop:
```go
func podPredicates(c client.Client) predicate.Funcs {
	return predicate.Funcs{
		UpdateFunc: func(ev event.UpdateEvent) bool {
			newObj, ok := ev.ObjectNew.(*corev1.Pod)
			if !ok {
				//that's no Pod. Let other predicates decide
				return true
			}

			oldObj, ok := ev.ObjectOld.(*corev1.Pod)
			if !ok {
				// Should never happen
				return true
			}

			// If the devworkspace label does not exist, do no reconcile
			var dwName string
			if dwName, ok = newObj.Labels[constants.DevWorkspaceNameLabel]; !ok {
				return false
			}

			// Do not trigger a reconcile on if status is not updated
			if equality.Semantic.DeepEqual(oldObj.Status, newObj.Status) {
				return false
			}
			//return false;
			workspace := &dw.DevWorkspace{}
			err := c.Get(context.TODO(), types.NamespacedName{Name: dwName, Namespace: newObj.GetNamespace()}, workspace)
			if err != nil {
				if !k8sErrors.IsNotFound(err) {
					//TODO log error
				}
				//Do not trigger if we failed to evaluate DW state. Otherwise failed workspaces will be restarted
				return false
			}

			noFailed := workspace.Status.Phase != dw.DevWorkspaceStatusFailed
			if noFailed {
				logf.Log.WithName("controller_devworkspace_config").Info("Pod: reconcile")
			} else {
				logf.Log.WithName("controller_devworkspace_config").Info("Pod: Do not reconcile pod for failed dw")
			}
			return noFailed
		},
	}
}
```

```
{"level":"info","ts":1617979194.2063107,"logger":"controller_devworkspace_config","msg":"Pod: Do not reconcile pod for failed dw"}
{"level":"info","ts":1617979194.2237866,"logger":"controllers.DevWorkspace","msg":"Error updating workspace status: Operation cannot be fulfilled on devworkspaces.workspace.devfile.io \"failure-test\": the object has been modified; please apply your changes to the latest version and try again","Request.Namespace":"devworkspace-controller","Request.Name":"failure-test","devworkspace_id":"workspace4531d661a2ff4323"}
{"level":"info","ts":1617979194.2240682,"logger":"controllers.DevWorkspace","msg":"Reconciling Workspace","Request.Namespace":"devworkspace-controller","Request.Name":"failure-test","devworkspace_id":"workspace4531d661a2ff4323"}

```

controllers/workspace/predicates.go
```diff
@@ -47,3 +51,31 @@ var predicates = predicate.Funcs{
...
+		}
+
+		oldObj, ok := ev.ObjectOld.(*corev1.Pod)
+		if !ok {
+			// Should never happen
+			return true
+		}
+
+		// If the dewworkspace label does not exist, do no reconcile
+		if _, ok := newObj.Labels[constants.DevWorkspaceIDLabel]; !ok {
```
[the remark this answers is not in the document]
amisevsk: If there's no devworkspace name label, then we don't enqueue any reconciles and so the functionality is the same. We should be checking workspace ID here since it's the canonical label that identifies resources are related to DevWorkspaces.

controllers/workspace/predicates.go
sleshchenko: That should be updated deployment status, since we scale it to 0. But how then it works in main branch.
We should rethink how we fail workspace start.
We should set started to false and force users to restart after failure, but it's not the best UX from operator perspective.
Or filter all event in the same way as I did pods above

controllers/workspace/predicates.go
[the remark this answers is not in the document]
sleshchenko: True, we could drop it earlier but it does not matter since we must provide both labels.

controllers/workspace/provision/deployment.go
tinakurian: done!

controllers/workspace/devworkspace_controller.go
```diff
@@ -385,6 +388,21 @@ func getWorkspaceId(instance *dw.DevWorkspace) (string, error) {
...
 
+// Mapping the pod to the devworkspace
+func mapPodToDW(mapObj handler.MapObject) []reconcile.Request {
+	meta := mapObj.Meta
+	labels := meta.GetLabels()
+	if _, ok := labels[constants.DevWorkspaceNameLabel]; !ok {
+		return nil
+	}
+	return []reconcile.Request{
+		{NamespacedName: types.NamespacedName{
```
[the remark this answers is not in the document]
tinakurian: agreed, done!

controllers/workspace/devworkspace_controller.go
```diff
@@ -396,6 +414,8 @@ func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}).
 		Owns(&batchv1.Job{}).
 		Owns(&controllerv1alpha1.DevWorkspaceRouting{}).
+		Watches(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(mapPodToDW)}).
```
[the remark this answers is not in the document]
tinakurian: sounds good! updated to use this!

controllers/workspace/predicates.go
```diff
@@ -47,3 +51,31 @@ var predicates = predicate.Funcs{
...
+			return false
+		}
+
+		// Always reconcile if resource is deleted
+		if newObj.GetDeletionTimestamp() != nil {
+			return true
+		}
+
+		// Trigger a reconcile on failed workspaces if status is updated.
```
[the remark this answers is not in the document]
tinakurian: fixed!

controllers/workspace/provision/deployment.go
```diff
@@ -119,6 +121,33 @@ func SyncDeploymentToCluster(
 		}
 	}
 
+	// Get pod and update dw status on crashloopbackoff
+	podList, err := getPods(workspace, clusterAPI.Client)
```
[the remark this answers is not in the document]
tinakurian: :-D thanks Serhii, used the above :-)

controllers/workspace/provision/deployment.go
```diff
@@ -119,6 +121,33 @@ func SyncDeploymentToCluster(
...
+			ProvisioningStatus: ProvisioningStatus{
+				Err: err,
+			},
+		}
+	}
+
+	for _, pod := range podList.Items {
+		if pod.Status.Phase != corev1.PodPending {
+			for _, containerStatus := range pod.Status.ContainerStatuses {
+				if containerStatus.State.Waiting != nil {
```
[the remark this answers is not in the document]
tinakurian: Sounds good! looks better! Used the above!

pkg/constants/constants.go
```diff
@@ -105,4 +105,9 @@ const (
 	// AsyncStorageClassType defines the 'asynchronous' storage policy. An rsync sidecar is added to devworkspaces that uses SSH to connect
 	// to a storage deployment that mounts a common PVC for the namespace.
 	AsyncStorageClassType = "async"
+
+	// ContainerCrashLoopBackOffReason defines the container status waiting reason for CrashLoopBackOff
+	ContainerCrashLoopBackOffReason = "CrashLoopBackOff"
+
+	ContainerImagePullErr = "ImagePullErr"
```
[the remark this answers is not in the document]
tinakurian: fixed! sorry didn't use commit suggestion so i could add it to the deployment.go file instead of keeping them in the constants.go file - hope this is okay

controllers/workspace/provision/deployment.go
```diff
@@ -119,6 +121,33 @@ func SyncDeploymentToCluster(
...
+	if err != nil {
+		return DeploymentProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{
+				Err: err,
+			},
+		}
+	}
+
+	for _, pod := range podList.Items {
+		if pod.Status.Phase != corev1.PodPending {
```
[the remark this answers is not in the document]
tinakurian: makes sense, removed!

pkg/constants/constants.go
```diff
@@ -105,4 +105,9 @@ const (
 	// AsyncStorageClassType defines the 'asynchronous' storage policy. An rsync sidecar is added to devworkspaces that uses SSH to connect
 	// to a storage deployment that mounts a common PVC for the namespace.
 	AsyncStorageClassType = "async"
+
+	// ContainerCrashLoopBackOffReason defines the container status waiting reason for CrashLoopBackOff
+	ContainerCrashLoopBackOffReason = "CrashLoopBackOff"
```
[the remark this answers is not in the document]
tinakurian: agreed, makes sense, i've moved both constants there!

controllers/workspace/provision/deployment.go
```diff
@@ -119,6 +121,33 @@ func SyncDeploymentToCluster(
...
+								Message:     fmt.Sprintf("Container %s has state %s", containerStatus.Name, containerStatus.State.Waiting.Reason),
+							},
+						}
+					}
+				}
+			}
+		}
+	}
+
 	deploymentReady := checkDeploymentStatus(clusterDeployment)
```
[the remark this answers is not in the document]
tinakurian: yup, makes sense, fixed this!

controllers/workspace/predicates.go
```diff
@@ -47,3 +51,31 @@ var predicates = predicate.Funcs{
...
+		}
+
+		// If the dewworkspace label does not exist, do no reconcile
+		if _, ok := newObj.Labels[constants.DevWorkspaceIDLabel]; !ok {
+			return false
+		}
+
+		// Always reconcile if resource is deleted
+		if newObj.GetDeletionTimestamp() != nil {
+			return true
```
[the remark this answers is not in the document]
tinakurian: yeah was just copied, but it doesn't make sense, removed it!

controllers/workspace/predicates.go
tinakurian: @sleshchenko I've updated the predicate to be almost the same, I've removed the part on getting the workspace though. Instead, went with updating the workspace.Spec.Started field and rely on the failWorkspace + stopWorkspace funcs. It seems to work! WDYT?

controllers/workspace/provision/deployment.go
```diff
@@ -167,6 +186,19 @@ func ScaleDeploymentToZero(workspace *dw.DevWorkspace, client runtimeClient.Clie
...
+	patch := []byte(`{"spec":{"started": false}}`)
+	err := client.Patch(context.Background(), &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: workspace.Namespace,
+			Name:      common.DeploymentName(workspace.Status.DevWorkspaceId),
+		},
+	}, runtimeClient.RawPatch(types.StrategicMergePatchType, patch))
+
+	return err
+}
```
amisevsk, line 200: We want to apply this patch to the DevWorkspace, not the deployment -- I think this might be a no-op.

```suggestion
func SetDeploymentSpecStartedToFalse(workspace *dw.DevWorkspace, client runtimeClient.Client) error {
	patch := []byte(`{"spec":{"started": false}}`)
	err := client.Patch(context.Background(), workspace, runtimeClient.RawPatch(types.StrategicMergePatchType, patch))
	return err
}
```

This probably doesn't need a function of its own.

controllers/workspace/predicates.go
sleshchenko: I'm OK with that. I'll take a look changes.

controllers/workspace/devworkspace_controller.go
```diff
@@ -116,6 +119,17 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	// Stop failed workspaces
+	if workspace.Status.Phase == dw.DevWorkspaceStatusFailed && workspace.Spec.Started {
+		err := provision.SetDeploymentSpecStartedToFalse(workspace, r.Client)
```
sleshchenko, line 124:
```suggestion
		err := provision.StopDevWorkspace(workspace, r.Client)
```

controllers/workspace/devworkspace_controller.go
```diff
@@ -279,7 +293,13 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	deploymentStatus := provision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI)
 	if !deploymentStatus.Continue {
 		if deploymentStatus.FailStartup {
-			return r.failWorkspace(workspace, fmt.Sprintf("DevWorkspace spec is invalid: %s", deploymentStatus.Err), reqLogger, &reconcileStatus)
+			var message string
+			if deploymentStatus.Err != nil {
```
sleshchenko, line 297: Why error has a higher priority? Isn't the message supposed to be user-friendly?

controllers/workspace/devworkspace_controller.go
```diff
@@ -309,50 +329,53 @@ func (r *DevWorkspaceReconciler) stopWorkspace(workspace *dw.DevWorkspace, logge
 		Name:      common.DeploymentName(workspace.Status.DevWorkspaceId),
 		Namespace: workspace.Namespace,
 	}
-	status := &currentStatus{}
+	status := initCurrentStatus()
```
sleshchenko, line 332: Well, I think it could simplify the stuff if we start here not from STARTING phase from STOPPING, FAILED, like the following https://github.com/sleshchenko/devworkspace-operator/commit/f7809311360d3ce895d16468600a12f0008284d2

controllers/workspace/provision/deployment.go
sleshchenko: > I think this might be a no-op.

WDYM? Currently, it's invoked only when workspace.Spec.Started.

sleshchenko (commented):
LGTM in general.
I'll take a deeper look and test later, probably tomorrow

controllers/workspace/provision/deployment.go
tinakurian: ohhhh okay done! i've removed the function as well!

controllers/workspace/devworkspace_controller.go
tinakurian: Yeah, sometimes SynDeploymentToCluster doesn't return an error though, so doing deploymentStatus.Err.Error() will fail because Err is nil

controllers/workspace/provision/deployment.go
amisevsk: > WDYM? Currently, it's invoked only when workspace.Spec.Started.

I mean applying `{"spec":{"started": false}}` to a Deployment does nothing since the deployment doesn't have that field:
```
❯ kc patch deploy devworkspace-controller-manager --type merge -p '{"spec": {"started": false}}'                                             
deployment.apps/devworkspace-controller-manager patched (no change)
```

controllers/workspace/devworkspace_controller.go
amisevsk: `initCurrentStatus` mostly is used to make sure the conditions map is initialized, but I agree that `Starting` isn't ideal. However it should always be set to one of `Stopping/Stopped/Failed` in this function so we should never actually *see* `Starting`

controllers/workspace/devworkspace_controller.go
```diff
@@ -116,6 +119,18 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	// Stop failed workspaces
+	if workspace.Status.Phase == dw.DevWorkspaceStatusFailed && workspace.Spec.Started {
+		patch := []byte(`{"spec":{"started": false}}`)
+		err := r.Client.Patch(context.Background(), workspace, client.RawPatch(types.StrategicMergePatchType, patch))
```
amisevsk, line 125: Strategic merge patch is not supported for CRs. 

```suggestion
		err := r.Client.Patch(context.Background(), workspace, client.RawPatch(types.MergePatchType, patch))
```

controllers/workspace/devworkspace_controller.go
tinakurian: fixed by removing function!

controllers/workspace/status.go
```diff
@@ -47,6 +47,31 @@ func initCurrentStatus() currentStatus {
...
+		stoppingStatus.copyConditionFromDevWorkspace(dw.DevWorkspaceFailedStart, dwStatus)
+	} else {
+		stoppingStatus.phase = dw.DevWorkspaceStatusStopping
+	}
+	return stoppingStatus
+}
+
+// changeToStoppedIfNeeded changes status to stopped if needed
+// we keep workspace status failed if it's stopped by workspace startup failure
+func changeToStoppedIfNeeded(c *currentStatus) {
```
amisevsk, line 69: I don't think this function should exist. It's confusing and assumes that the currentStatus is filled when most of the time it is not -- making use of this value would require you to first explicitly set `currentStatus.Phase = Failed`, which is only done in one part of the code. 

Also the naming isn't clear (what makes changing to stopped needed?) without reading the function doc.

controllers/workspace/status.go
```diff
@@ -47,6 +47,31 @@ func initCurrentStatus() currentStatus {
 	}
 }
 
+// initStoppingStatus initialize the status for stopping workspace
+// it may be STOPPING or FAILED in case it's stopped automatically due workspace startup failure
+// in the second case the corresponding condition is propagated
+func initStoppingStatus(dwStatus dw.DevWorkspaceStatus) currentStatus {
```
amisevsk, line 53: This seems like a confusing abstraction for a very specific we need to do. Perhaps a better name would help?

In general, if a function is only going to be called in one place, I don't see much benefit in breaking it out like this.

controllers/workspace/status.go
sleshchenko: Tina sorry me for proposing these changes without thinking enough )
Proposing alternative in https://github.com/devfile/devworkspace-operator/pull/375#discussion_r613487876 thread

controllers/workspace/status.go
sleshchenko: Well, initStartingStatus is called also from one place.
It's also called finalize.go but it has a different use-case because it gets STARTING status and then update phase to Error, so starting does not make sense.

I tried to play with it a bit and @amisevsk WDYT if we:
1. Introduce method `New` that will init conditions.
2. Separate actual workspace stopping logic from managing status, to avoid `changeToStoppedIfNeeded` func as well as doing the same checks in two places

Like the following https://github.com/sleshchenko/devworkspace-operator/commit/caa5a68ea17bd1261b6f67723067a1c6c3d879e0

stop... doStop... may be not the best naming
doStop came to me from Java but in this context `scaleDeploymentDown` may be better at least until we don't do anything apart from scaling deployment to zero on stopping workspaces.

controllers/workspace/devworkspace_controller.go
sleshchenko: let's discuss an alternative proposal in https://github.com/devfile/devworkspace-operator/pull/375#discussion_r614572224

controllers/workspace/devworkspace_controller.go
sleshchenko: With the current codebase it should work since we return error or message.

But I think in case both are returned, we need to do one of the following
1. Mix message and error in the condition message, as we do with ProvisioningError. Actually, what if we change ProvisioningStatus to have `ProvisioningError` instead of Err + Message? Then it will work the same.
2. Or message has higher priority to be provided into status message, error is used only when "user-friendly" is not specified.

controllers/workspace/provision/deployment.go
sleshchenko: It does not matter now but: as far as I saw workspace was patched but the function name was confused. So, it was not no-op.
But with removing this func with don't lose anything )

controllers/workspace/status.go
amisevsk: That makes sense to me -- also I would rework `initCurrentStatus` to just remove it -- all we really need to do is make sure the map isn't `nil`. It's safer to just do that in the functions that write to it, e.g. 
```go
func (c *workspaceConditions) setConditionTrue(conditionType dw.DevWorkspaceConditionType, msg string) {
	if c.conditions == nil {
		c.conditions = map[dw.DevWorkspaceConditionType]dw.DevWorkspaceCondition{}
	}
	c.conditions[conditionType] = dw.DevWorkspaceCondition{
		Status:  corev1.ConditionTrue,
		Message: msg,
	}
}
```

> Well, initStartingStatus is called also from one place.

It used to be called `initCurrentStatus` and was meant to be used in a different way :)

controllers/workspace/condition.go
```diff
@@ -40,19 +40,43 @@ type workspaceConditions struct {
...
+	}
+
 	c.conditions[conditionType] = dw.DevWorkspaceCondition{
 		Status:  corev1.ConditionFalse,
 		Message: msg,
 	}
 }
 
+func (c *workspaceConditions) copyFailedStatus(dwStatus dw.DevWorkspaceStatus) {
+	if dwStatus.Phase == dw.DevWorkspaceStatusFailed {
```
sleshchenko, line 65: Should not this func client check if status is failed before calling it?

sleshchenko: it actually copies FailedStart condition but not failed status. It's a bit confusing.

tinakurian: Yeah, sorry, it was originally called copyConditionFromDevWorkspace - but i changed it to be used specifically for the failed condition (instead of needing to pass in the condition).

controllers/workspace/devworkspace_controller.go
tinakurian: Thanks for pointing this out @amisevsk! I see it's not supported in the docs here: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/

controllers/workspace/provision/deployment.go
```diff
@@ -306,6 +325,38 @@ func getClusterDeployment(name string, namespace string, client runtimeClient.Cl
...
+		constants.DevWorkspaceIDLabel: workspace.Status.DevWorkspaceId,
+	}); err != nil {
+		return nil, err
+	}
+	return pods, nil
+}
+
+// checkFailedPods check if related pods has unrecoverable states: CrashLoopBackOffReason, ImagePullErr
+// Returns optional message with detected unrecoverable state details
+//         error is any happens during check
```
sleshchenko, line 340:
```suggestion
//         error if any happens during check
```

controllers/workspace/devworkspace_controller.go
```diff
@@ -279,7 +294,19 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
...
+				if deploymentStatus.Message != "" {
+					message = message + ": " + deploymentStatus.Message
+				}
+				reqLogger.Info(message)
+			} else {
+				message = deploymentStatus.Message
+				reqLogger.Info(message)
+			}
+
+			return r.failWorkspace(workspace, fmt.Sprintf("DevWorkspace spec is invalid: %s", message), reqLogger, &reconcileStatus)
```
sleshchenko, line 309: I know it's here just from the previous state, but I would say it's not true.
Pod may be failed to start because authentication is missing, or DWO provides the bad init container due DWO config issues or bug, it's not necessary something user may be able to fix with DevWorkspace Spec.
```suggestion
			return r.failWorkspace(workspace, fmt.Sprintf("Startup failed due: %s", message), reqLogger, &reconcileStatus)
```
^ or probably you'll provide even better common wording.

controllers/workspace/devworkspace_controller.go
```diff
@@ -279,7 +294,19 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	deploymentStatus := provision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI)
 	if !deploymentStatus.Continue {
 		if deploymentStatus.FailStartup {
-			return r.failWorkspace(workspace, fmt.Sprintf("DevWorkspace spec is invalid: %s", deploymentStatus.Err), reqLogger, &reconcileStatus)
+			var message string
+			if deploymentStatus.Err != nil {
+				message = deploymentStatus.Err.Error()
+				if deploymentStatus.Message != "" {
+					message = message + ": " + deploymentStatus.Message
```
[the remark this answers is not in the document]
sleshchenko: Actually, that's a pretty big piece of low-level logic to be a part of DW Controller file. I would move it to ProvisioningStatus. Something like https://github.com/sleshchenko/devworkspace-operator/commit/b391c7c4c2144f3d230021948615c0083852c181
WDYT?

controllers/workspace/condition.go
sleshchenko: it still check
```go
if dwStatus.Phase == dw.DevWorkspaceStatusFailed {
```

at the same time as all clients (1) has the same check

sleshchenko: I would refactor it in the following way https://github.com/sleshchenko/devworkspace-operator/commit/4fc74e794649cbb5edb0a34d7738e692e005bf93

WDYT?

controllers/workspace/provision/deployment.go
```diff
@@ -35,10 +36,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	ContainerCrashLoopBackOffReason = "CrashLoopBackOff"
```
sleshchenko, line 45: It works for me when I do `sleep 1` but when I do `exit 1`, it does not work:
```
kind: DevWorkspace
metadata:
  name: web-terminal
  annotations:
    controller.devfile.io/restricted-access: "true"
  labels:
    # it's a label OpenShift console uses a flag to mark terminal's workspaces
    console.openshift.io/terminal: "true"
spec:
  started: true
  routingClass: 'web-terminal'
  template:
    components:
      - plugin:
          name: web-terminal
          id: redhat-developer/web-terminal/latest
      - container:
          memoryLimit: "256Mi"
          name: tooling
          command: ["exit"]
          args: ["1"]
          mountSources: false
          env:
            - name: PS1
              value: \[\e[34m\]>\[\e[m\]\[\e[33m\]>\[\e[m\]
          volumeMounts:
            - name: home-storage
              path: "/home/user"
      - volume:
          name: home-storage
```

```
  containerStatuses:
  - image: quay.io/wto/web-terminal-tooling:latest
    imageID: ""
    lastState: {}
    name: tooling
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        message: |
          container create failed: time="2021-04-20T11:54:11Z" level=error msg="container_linux.go:366: starting container process caused: exec: \"exit\": executable file not found in $PATH"
        reason: CreateContainerError

```

I'm fully OK to move it out of the current PR scope to move forward.

sleshchenko (approved):
I tested and it works fine for me. Good job sorry for bothering with tiny refactoring stuff.

controllers/workspace/condition.go
tinakurian: Looks good to me! I've updated the PR to reflect this

controllers/workspace/devworkspace_controller.go
tinakurian: Okay, makes sense. I've updated this!

controllers/workspace/provision/deployment.go
tinakurian: > I tested and it works fine for me. Good job sorry for bothering with tiny refactoring stuff.

@sleshchenko Thanks for the review! It's not a bother :-) It looks much better. I've incorporated all of the suggestions above.

Ohhh okay, i was using exit 1 this way and it seems to work: 
```
kind: DevWorkspace
metadata:
  name: plain
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: nodejs
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
            - "/bin/bash"
          args:
            - "exit 1"
```

I'll work on figuring out why the sample you provided above doesn't work. I don't mind doing it as part of this PR, it makes sense that it should work for the sample you provided.

sleshchenko: Seems I see the difference: I missed removing one args, so it probably got `exit tail -f /dev/null`.

sleshchenko: BTW it's just one more of the possible errors we can propagate to the user.

tinakurian: fixed! 
image

controllers/workspace/status.go
```diff
@@ -38,13 +38,11 @@ type currentStatus struct {
...
-		workspaceConditions: workspaceConditions{
-			conditions: map[dw.DevWorkspaceConditionType]dw.DevWorkspaceCondition{},
-		},
-		phase: dw.DevWorkspaceStatusStarting,
+func newStatusWithPhase(phase dw.DevWorkspacePhase) currentStatus {
+	status := currentStatus{
+		phase: phase,
 	}
+	return status
 }
```
amisevsk, line 46: Nit: This function could just be replaced with `myStatus := currentStatus{phase: myPhase}` at call sites. Do we need this wrapper over a struct initialization?

amisevsk (approved):

amisevsk:
:tada:

controllers/workspace/status.go
sleshchenko: initially, it also initialized conditions with empty map. Not it makes sense just inline the struct init.

tinakurian: done!

sleshchenko (approved):
LGTM

amisevsk (approved):

Issue #356, opened 2021-03-26
Invalid devfile should lead to failed workspace startup


sleshchenko:
When I tried to run Che DevWorkspace with https://github.com/l0rd/spring-petclinic/blob/master/devfile.yaml:

Workspace was in `STARTING | Processing DevWorkspace stuff` forever.
Only when I checked logs I found the cause:
```
2021-03-26T10:30:45.129Z       ERROR  controller     Reconciler error       {"reconcilerGroup": "workspace.devfile.io", "r
econcilerKind": "DevWorkspace", "controller": "devworkspace", "name": "spring-petclinic", "namespace": "opentlc-mgr-che", "err
or": "container 'maven' references undefined volume 'mavenrepo'"}
```
In such case devworkspace startup must be failed and error should be reported.

PR #381, opened 2021-04-09
Fail devworkspace start if devfile is invalid

tinakurian:
### What does this PR do?
Fails devworkspace when there is an error in rewriteContainerVolumeMounts.

https://github.com/devfile/devworkspace-operator/issues/356

However, we should probably write a utility function that validates a devfile to catch all issues up front.

### Is it tested? How?
Applied yaml and watched to see if the devworkspace would fail
image

### DIFF
--- a/pkg/provision/storage/commonStorage.go
+++ b/pkg/provision/storage/commonStorage.go
@@ -48,8 +48,12 @@ func (p *CommonStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAd
 	}
 
 	if err := p.rewriteContainerVolumeMounts(workspace.Status.DevWorkspaceId, podAdditions, &workspace.Spec.Template); err != nil {
-		return err
+		return &ProvisioningError{
+			Err:     err,
+			Message: "Could not rewrite container volume mounts",
+		}
 	}
+
 	if _, err := syncCommonPVC(workspace.Namespace, clusterAPI); err != nil {
 		return err
 	}


amisevsk (approved):
LGTM :)

sleshchenko (approved):
Changes LGTM
I wonder why the message has duplicated text but if we can easily avoid it

pkg/provision/storage/commonStorage.go
```diff
@@ -48,8 +48,12 @@ func (p *CommonStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAd
 	}
 
 	if err := p.rewriteContainerVolumeMounts(workspace.Status.DevWorkspaceId, podAdditions, &workspace.Spec.Template); err != nil {
-		return err
+		return &ProvisioningError{
+			Err:     err,
```
sleshchenko, line 52: > I wonder why the message has duplicated text but if we can easily avoid it

The answer is in two places:
```go
		case *storage.ProvisioningError:
			return r.failWorkspace(workspace, fmt.Sprintf("Error provisioning storage: %s", storageErr), reqLogger, &reconcileStatus)

//...

// ProvisioningError represents an unrecoverable issue in provisioning storage for a DevWorkspace.
type ProvisioningError struct {
	// Err is the underlying error causing the problem. If nil, it is not included in the output of Error()
	Err error
	// Message is a user-friendly string explaining why the error occurred
	Message string
}

func (e *ProvisioningError) Error() string {
	if e.Err != nil {
		return fmt.Sprintf("%s: %s", e.Message, e.Err)
	}
	return e.Message
}
```

We set in the condition storageErr.Message, but fail workspace with storageErr.

We should not set err here or fail workspace with storageErr.

@amisevsk you're an author of ProvisioningError, what would be your suggestion here?

amisevsk: The intention is that for `ProvisioningError`, `Message` is the important bit and `Err` is additional context, so that you can say something like
```go
return &ProvisioningError{
  Err: // Error from a library call,
  Message: "Failed to do X",
}
```

Message should not include the error, so either we should add context, e.g.
```go
return &ProvisioningError{
  Err:     err,
  Message: "Could not rewrite container volume mounts", // Something better than this of course
}
```

or  we should leave `Err` nil since the message is being consumed:
```go
return &ProvisioningError{
  Message: err.Error()
}
```

amisevsk: The joys of Go error handling mean that you have to know what info is in the error so you don't repeat yourself :upside_down_face:

tinakurian: Sorry! I noticed this but forgot to change it before i created the PR. I've taken @amisevsk example to add context

amisevsk (approved):

sleshchenko (approved):

sleshchenko (approved):

JPinkney (approved):

PR #385, opened 2021-04-14
Routing errors propagation

amisevsk:
### What does this PR do?
Adds a string field `.status.message` to DevWorkspaceRoutings, and uses this value in place of `"preparing networking"` in DevWorkspaces that are waiting for routing to be ready. This means that user-readable failure reasons can be set in DevWorkspaceRoutings and are automatically propagated to the DevWorkspace status.

Also adds printColumns to DevWorkspaceRoutings to improve readability for short output:
```
❯ kc get dwr
NAME                                DEVWORKSPACE ID             PHASE   INFO
routing-workspaceda77c562abd649a1   workspaceda77c562abd649a1   Ready   DevWorkspaceRouting prepared
```

Commit messages should be clear for what each commit adds. Note that I've also added propagating messages from dwr to dw status even when the dwr is not failed, but since a the dwr is ready so quickly in normal usage the message never reaches the DevWorkspace's status, so we never see `Preparing [services|routes|ingresses]`. This might still be useful if e.g. there is an issue applying resources, but might also make tracking progress harder (since the messages may or may not appear)

Closes https://github.com/eclipse/che/issues/17455

### Is it tested? How?
Easiest to test on Kubernetes as otherwise all routingClasses are valid and the default `dwr` reconciler cannot fail:
```bash
{ cat <<EOF | kubectl apply -f -
kind: DevWorkspace
metadata:
  name: theia-next
spec:
  started: true
  routingClass: cluster-tls
  template:
    components:
      - name: theia
        plugin:
EOF
} && kubectl get dw -w
```

Output: 
```
❯ kc apply -f samples/theia-next.yaml && kc get dw -w
devworkspace.workspace.devfile.io/theia-next created
NAME         DEVWORKSPACE ID             PHASE      INFO
theia-next   workspaceba44c7e060774cb3           
theia-next   workspaceba44c7e060774cb3   Starting   Preparing networking
theia-next   workspaceba44c7e060774cb3   Failed     Invalid routingClass for DevWorkspace: routing class cluster-tls only supported on OpenShift
```

### DIFF
--- a/deploy/deployment/openshift/objects/devworkspaceroutings.controller.devfile.io.CustomResourceDefinition.yaml
+++ b/deploy/deployment/openshift/objects/devworkspaceroutings.controller.devfile.io.CustomResourceDefinition.yaml
@@ -19,7 +19,20 @@ spec:
     singular: devworkspacerouting
   scope: Namespaced
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - description: The owner DevWorkspace's unique id
+      jsonPath: .spec.devworkspaceId
+      name: DevWorkspace ID
+      type: string
+    - description: The current phase
+      jsonPath: .status.phase
+      name: Phase
+      type: string
+    - description: Additional info about DevWorkspaceRouting state
+      jsonPath: .status.message
+      name: Info
+      type: string
+    name: v1alpha1
     schema:
       openAPIV3Schema:
         description: DevWorkspaceRouting is the Schema for the devworkspaceroutings
@@ -164,6 +177,10 @@ spec:
                   type: array
                 description: Machine name to exposed endpoint map
                 type: object
+              message:
+                description: Message is a user-readable message explaining the current
+                  phase (e.g. reason for failure)
+                type: string
               phase:
                 description: Routing reconcile phase
                 type: string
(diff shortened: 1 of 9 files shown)


deploy/deployment/openshift/objects/devworkspaceroutings.controller.devfile.io.CustomResourceDefinition.yaml
```diff
@@ -19,7 +19,20 @@ spec:
     singular: devworkspacerouting
   scope: Namespaced
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
```
sleshchenko, line 22: Seems our [gitattributes](https://github.com/devfile/devworkspace-operator/blob/main/.gitattributes) is not up-to-date

sleshchenko (approved):
I haven't tested but changes LGTM

metlos (approved):

deploy/deployment/openshift/objects/devworkspaceroutings.controller.devfile.io.CustomResourceDefinition.yaml
amisevsk: :thinking:

JPinkney (approved):

deploy/deployment/openshift/objects/devworkspaceroutings.controller.devfile.io.CustomResourceDefinition.yaml
amisevsk: Figured it out https://github.com/devfile/devworkspace-operator/pull/389

PR #387, opened 2021-04-15
Fix info in PR template about e2e tests as PR checks

sleshchenko:
### What does this PR do?
Fix info in PR template about e2e tests as PR checks

N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,4 +7,7 @@
 ### Is it tested? How?
 <!-- Please provide instructions here how reviewer can test your changes if applicable -->
 
-<!-- Before PR merging it's required to run e2e tests, to trigger them comment `/test v5-devworkspaces-operator-e2e` -->
+<!-- 
+Before PR merging it's required to run e2e tests, to trigger them comment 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,4 +7,7 @@
 ### Is it tested? How?
 <!-- Please provide instructions here how reviewer can test your changes if applicable -->
 
-<!-- Before PR merging it's required to run e2e tests, to trigger them comment `/test v5-devworkspaces-operator-e2e` -->
+<!-- 
+Before PR merging it's required to run e2e tests, to trigger them comment 
+/test v7-devworkspaces-operator-e2e,v7-devworkspace-happy-path
```
sleshchenko, line 12: I tried to patch trigger to have an ability to use `/test all` but they are autogenerated, probably without an ability to set a custom.

Just leave here to share what I came to, maybe later I'll figure out how to avoid rewriting by generator
```
    # trigger if test list has `all` or test specific trigger
    # like `/test v7-devworkspace-happy-path` or `/test any,v7-devworkspace-happy-path, another
    trigger: (?m)^/test (.*, ?)*((all)|(v7-devworkspace-happy-path))( ?,.*,? ?)*$
```

amisevsk (approved):

Issue #373, opened 2021-04-05
DevWorkspace controller should not care about DWR base host


sleshchenko:
DevWorkspace controller should not care about DWR base host.
Instead it should be on the DWR controller side, similar behavior as with Routes: if DWR is created without base host - it gets assigned by the controller.

Currently, I think that setting is mandatory on DWO side, but if DWO is deployed to be used with Web Terminal + Che, it's not needed at all.

+ Che and Basic routing may want to be exposed with different base hosts.

sleshchenko:
on planning we agreed that DevWorkspaceRouting routingSuffix is currently unused and should be removed.
Later we may reintroduce it via DevWorkspace field or annotation if we'll need an ability to configure routingSuffix per DevWorkspace instance.

sleshchenko:
The prepared PR https://github.com/devfile/devworkspace-operator/pull/391 depends on https://github.com/eclipse/che/issues/19631

sleshchenko:
Fixed by https://github.com/devfile/devworkspace-operator/pull/391

PR #391, opened 2021-04-19
devworkspace controller should not be in charge of routing suffix for DWR CRs

tinakurian:
### What does this PR do?
Removes the routingSuffix. It should be assigned by the DWR controller.

It should be merged in sync with https://github.com/che-incubator/devworkspace-che-operator/pull/45

https://github.com/devfile/devworkspace-operator/issues/373

### Is it tested? How?
Ran unit tests.

### DIFF
--- a/apis/controller/v1alpha1/devworkspacerouting_types.go
+++ b/apis/controller/v1alpha1/devworkspacerouting_types.go
@@ -25,8 +25,6 @@ type DevWorkspaceRoutingSpec struct {
 	DevWorkspaceId string `json:"devworkspaceId"`
 	// Class of the routing: this drives which DevWorkspaceRouting controller will manage this routing
 	RoutingClass DevWorkspaceRoutingClass `json:"routingClass,omitempty"`
-	// Routing suffix for cluster
-	RoutingSuffix string `json:"routingSuffix"`
 	// Machines to endpoints map
 	Endpoints map[string]EndpointList `json:"endpoints"`
 	// Selector that should be used by created services to point to the devworkspace Pod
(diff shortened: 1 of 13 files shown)


controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -115,7 +115,6 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(req ctrl.Request) (ctrl.Result
 		DevWorkspaceId: instance.Spec.DevWorkspaceId,
 		Namespace:      instance.Namespace,
 		PodSelector:    instance.Spec.PodSelector,
-		RoutingSuffix:  instance.Spec.RoutingSuffix,
```
sleshchenko, line 118: You removed assigning DevWorkspaceMetadata.RoutingSuffix but it seems to be still used. I wonder if it should be fully removed from struct and evaluated by routing impl in place where it's needed
Screenshot_20210419_100146

sleshchenko (approved):
I haven't tested but with the latest changes it should work, and it should change nothing for DWCO, but still DWCO probably needs adaptation to remove routingSuffix read from DWR CR

metlos (approved):

apis/controller/v1alpha1/devworkspacerouting_types.go
```diff
@@ -25,8 +25,6 @@ type DevWorkspaceRoutingSpec struct {
 	DevWorkspaceId string `json:"devworkspaceId"`
 	// Class of the routing: this drives which DevWorkspaceRouting controller will manage this routing
 	RoutingClass DevWorkspaceRoutingClass `json:"routingClass,omitempty"`
-	// Routing suffix for cluster
-	RoutingSuffix string `json:"routingSuffix"`
```
amisevsk, line 29: Worth noting that this is a breaking change for v1alpha1 devworkspaceroutings, but I don't think it impacts anything as nothing is deploying devworkspaceroutings yet.

amisevsk (approved):
LGTM (though I didn't test directly)

sleshchenko:
rebased against main branch to run e2e tests on fresh version

Issue #377, opened 2021-04-08
Improve DWO versioning


sleshchenko:
Currently, DWO always reports [1.0.0-next](https://github.com/devfile/devworkspace-operator/blob/main/version/version.go#L16)
it would be extremely useful if we report actual version + git hash

tinakurian:
Thanks @amisevsk :-) forgot to close it!

PR #393, opened 2021-04-22
Improve dwo versioning

tinakurian:
### What does this PR do?
Improves dwo versioning 

https://github.com/devfile/devworkspace-operator/issues/377

### Is it tested? How?
image

image

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@
 SHELL := bash
 .SHELLFLAGS = -ec
 .ONESHELL:
+.DEFAULT_GOAL := help
 
 ifndef VERBOSE
 MAKEFLAGS += --silent
@@ -32,34 +33,11 @@ INTERNAL_TMP_DIR=/tmp/devworkspace-controller
 BUMPED_KUBECONFIG=$(INTERNAL_TMP_DIR)/kubeconfig
 CONTROLLER_ENV_FILE=$(INTERNAL_TMP_DIR)/environment
 
-ifeq (,$(shell which kubectl))
-ifeq (,$(shell which oc))
-$(error oc or kubectl is required to proceed)
-else
-K8S_CLI := oc
-endif
-else
-K8S_CLI := kubectl
-endif
-
-ifeq ($(shell $(K8S_CLI) api-resources --api-group='route.openshift.io'  2>&1 | grep -o routes),routes)
-PLATFORM := openshift
-else
-PLATFORM := kubernetes
-endif
-
-# minikube handling
-ifeq ($(shell $(K8S_CLI) config current-context 2>&1),minikube)
-# check ingress addon is enabled
-ifeq ($(shell minikube addons list -o json | jq -r .ingress.Status), disabled)
-$(error ingress addon should be enabled on top of minikube)
-endif
-export ROUTING_SUFFIX := $(shell minikube ip).nip.io
+ifneq (,$(shell which kubectl 2>/dev/null)$(shell which oc 2>/dev/null))
+include deploy.mk
 endif
 
 # Bootstrapped by Operator-SDK v1.1.0
-# Current Operator version
-VERSION ?= 0.0.1
 OPERATOR_SDK_VERSION = v1.1.0
 # Default bundle image tag
 DWO_BUNDLE_IMG ?= controller-bundle:$(VERSION)
@@ -94,24 +72,8 @@ _print_vars:
 	@echo "    DEFAULT_ROUTING=$(DEFAULT_ROUTING)"
 	@echo "    DEVWORKSPACE_API_VERSION=$(DEVWORKSPACE_API_VERSION)"
 
-_create_namespace:
-	$(K8S_CLI) create namespace $(NAMESPACE) || true
-
-_gen_configuration_env:
-	mkdir -p $(INTERNAL_TMP_DIR)
-	echo "export RELATED_IMAGE_devworkspace_webhook_server=$(DWO_IMG)" > $(CONTROLLER_ENV_FILE)
-ifeq ($(PLATFORM),kubernetes)
-	echo "export WEBHOOK_SECRET_NAME=devworkspace-operator-webhook-cert" >> $(CONTROLLER_ENV_FILE)
-endif
-	cat ./deploy/templates/components/manager/manager.yaml \
-		| yq -r \
-			'.spec.template.spec.containers[]?.env[] | select(.name | startswith("RELATED_IMAGE")) | "export \(.name)=\"$${\(.name):-\(.value)}\""' \
-		>> $(CONTROLLER_ENV_FILE)
-	echo "export MAX_CONCURRENT_RECONCILES=1" >> $(CONTROLLER_ENV_FILE)
-	cat $(CONTROLLER_ENV_FILE)
-
 ##### Rules for dealing with devfile/api
-### update_devworkspace_api: update version of devworkspace crds in go.mod
+### update_devworkspace_api: Updates the version of devworkspace crds in go.mod
 update_devworkspace_api:
 	go mod edit --require github.com/devfile/api/v2@$(DEVWORKSPACE_API_VERSION)
 	go mod download
@@ -120,24 +82,24 @@ update_devworkspace_api:
 _init_devworkspace_crds:
 	./update_devworkspace_crds.sh --init --api-version $(DEVWORKSPACE_API_VERSION)
 
-### update_devworkspace_crds: pull latest devworkspace CRDs to ./devworkspace-crds. Note: pulls master branch
+### update_devworkspace_crds: Pulls the latest devworkspace CRDs to ./devworkspace-crds. Note: pulls master branch
 update_devworkspace_crds:
 	./update_devworkspace_crds.sh --api-version $(DEVWORKSPACE_API_VERSION)
 ###### End rules for dealing with devfile/api
 
-### test: Run tests
+### test: Runs tests
 ENVTEST_ASSETS_DIR = $(shell pwd)/bin/testbin
 test: generate fmt vet manifests
 	mkdir -p $(ENVTEST_ASSETS_DIR)
 	test -f $(ENVTEST_ASSETS_DIR)/setup-envtest.sh || curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.6.3/hack/setup-envtest.sh
 	source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test $(shell go list ./... | grep -v test/e2e) -coverprofile cover.out
 
-### test_e2e: runs e2e test on the cluster set in context. DevWorkspace Operator must be already deployed
+### test_e2e: Runs e2e test on the cluster set in context. DevWorkspace Operator must be already deployed
 test_e2e:
 	CGO_ENABLED=0 go test -v -c -o bin/devworkspace-controller-e2e ./test/e2e/cmd/workspaces_test.go
 	./bin/devworkspace-controller-e2e -ginkgo.failFast
 
-### test_e2e_debug: runs e2e test in debug mode, so it's possible to connect to execution via remote debugger
+### test_e2e_debug: Runs e2e test in debug mode, so it's possible to connect to execution via remote debugger
 test_e2e_debug:
 	dlv test --listen=:2345 --headless=true --api-version=2 ./test/e2e/cmd/workspaces_test.go -- --ginkgo.failFast
 
@@ -184,54 +146,22 @@ else
 	$(K8S_CLI) apply -f deploy/current/openshift/combined.yaml
 endif
 
-### generate_deployment: Generate files used for deployment from kustomize templates, using environment variables
+### generate_deployment: Generates the files used for deployment from kustomize templates, using environment variables
 generate_deployment:
 	deploy/generate-deployment.sh
 
-### generate_default_deployment: Generate files used for deployment from kustomize templates with default values
+### generate_default_deployment: Generates the files used for deployment from kustomize templates with default values
 generate_default_deployment:
 	deploy/generate-deployment.sh --use-defaults
 
-### install_plugin_templates: Deploy sample plugin templates to namespace devworkspace-plugins:
-install_plugin_templates: _print_vars
-	$(K8S_CLI) create namespace devworkspace-plugins || true
-	$(K8S_CLI) apply -f samples/plugins -n devworkspace-plugins
-
-### restart: Restart devworkspace-controller deployment
-restart:
-	$(K8S_CLI) rollout restart -n $(NAMESPACE) deployment/devworkspace-controller-manager
-
-### restart_webhook: Restart devworkspace-controller webhook deployment
-restart_webhook:
-	$(K8S_CLI) rollout restart -n $(NAMESPACE) deployment/devworkspace-webhook-server
-
-### uninstall: Remove controller resources from the cluster
-uninstall: generate_deployment
-# It's safer to delete all workspaces before deleting the controller; otherwise we could
-# leave workspaces in a hanging state if we add finalizers.
-	$(K8S_CLI) delete devworkspaces.workspace.devfile.io --all-namespaces --all --wait || true
-	$(K8S_CLI) delete devworkspacetemplates.workspace.devfile.io --all-namespaces --all || true
-	$(K8S_CLI) delete devworkspaceroutings.controller.devfile.io --all-namespaces --all --wait || true
-
-ifeq ($(PLATFORM),kubernetes)
-	$(K8S_CLI) delete --ignore-not-found -f deploy/current/kubernetes/combined.yaml || true
-else
-	$(K8S_CLI) delete --ignore-not-found -f deploy/current/openshift/combined.yaml || true
-endif
-
-	$(K8S_CLI) delete all -l "app.kubernetes.io/part-of=devworkspace-operator" --all-namespaces
-	$(K8S_CLI) delete mutatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io --ignore-not-found
-	$(K8S_CLI) delete validatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io --ignore-not-found
-	$(K8S_CLI) delete namespace $(NAMESPACE) --ignore-not-found
-
-### manifests: Generate manifests e.g. CRD, RBAC etc.
+### manifests: Generates the manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=role webhook paths="./..." \
 			output:crd:artifacts:config=deploy/templates/crd/bases \
 			output:rbac:artifacts:config=deploy/templates/components/rbac
 	patch/patch_crds.sh
 
-### fmt: Run go fmt against code
+### fmt: Runs go fmt against code
 fmt:
 ifneq ($(shell command -v goimports 2> /dev/null),)
 	find . -name '*.go' -exec goimports -w {} \;
@@ -241,7 +171,7 @@ else
 	go fmt -x ./...
 endif
 
-### fmt_license: ensure license header is set on all files
+### fmt_license: Ensures the license header is set on all files
 fmt_license:
 ifneq ($(shell command -v addlicense 2> /dev/null),)
 	@echo 'addlicense -v -f license_header.txt **/*.go'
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -25,6 +25,26 @@ export DEFAULT_ROUTING ?= basic
 export KUBECONFIG ?= ${HOME}/.kube/config
 export DEVWORKSPACE_API_VERSION ?= 1f335562c475972132851c68227dd36558317bb3
 
+# By default the project should be build under GOPATH/src/github.com/<orgname>/<reponame>
```
sleshchenko, line 28: That's not true for `go mod`
we should hardcode it or read from `https://github.com/devfile/devworkspace-operator/blob/main/go.mod#L1`

Makefile
```diff
@@ -25,6 +25,26 @@ export DEFAULT_ROUTING ?= basic
...
+GO_PACKAGE_ORG_NAME ?= $(shell basename $$(dirname $$PWD))
+GO_PACKAGE_REPO_NAME ?= $(shell basename $$PWD)
+GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
+
+# Set GIT_COMMIT_ID
+GIT_COMMIT_ID := $(shell git rev-parse HEAD)
+ifneq ($(shell git status --porcelain --untracked-files=no),)
+       GIT_COMMIT_ID := $(GIT_COMMIT_ID)-dirty
```
sleshchenko, line 36: Something, like PROW CI, don't use Makefile to build image.
What if we move this logic to Dockerfile?

Makefile
```diff
@@ -58,7 +78,7 @@ endif
 
 # Bootstrapped by Operator-SDK v1.1.0
 # Current Operator version
-VERSION ?= 0.0.1
+DWO_VERSION ?= 0.0.1
```
sleshchenko, line 81: I'm not sure if we need it in Makefile

Makefile
tinakurian: Done, reading from go mod instead!

Makefile
tinakurian: @sleshchenko that makes sense - does this look better now?

k8s.mk
```diff
@@ -0,0 +1,116 @@
...
+	echo "export WEBHOOK_SECRET_NAME=devworkspace-operator-webhook-cert" >> $(RELATED_IMAGES_FILE)
+endif
+	cat ./deploy/templates/components/manager/manager.yaml \
+		| yq -r \
+			'.spec.template.spec.containers[]?.env[] | select(.name | startswith("RELATED_IMAGE")) | "export \(.name)=\"$${\(.name):-\(.value)}\""' \
+		>> $(RELATED_IMAGES_FILE)
+	cat $(RELATED_IMAGES_FILE)
+
+### install: Install controller in the configured Kubernetes cluster in ~/.kube/config
+install: _check_cert_manager _print_vars _init_devworkspace_crds _create_namespace generate_deployment
```
sleshchenko, line 43: it's not only for K8s but also for OpenShift. Maybe then better name file `deploy.mk`?

Makefile
```diff
@@ -25,40 +25,29 @@ export DEFAULT_ROUTING ?= basic
...
-endif
-
-# minikube handling
-ifeq ($(shell $(K8S_CLI) config current-context 2>&1),minikube)
-# check ingress addon is enabled
-ifeq ($(shell minikube addons list -o json | jq -r .ingress.Status), disabled)
-$(error ingress addon should be enabled on top of minikube)
-endif
-export ROUTING_SUFFIX := $(shell minikube ip).nip.io
+ifneq (,$(shell which kubectl)$(shell which oc))
```
sleshchenko, line 44: I'm not sure I fully got what you try to achieve with it.

Makefile
sleshchenko: I still think we should have the only one place to devfile semver for DWO and currently it's https://github.com/devfile/devworkspace-operator/blob/main/version/version.go#L16

Makefile
```diff
@@ -25,40 +25,29 @@ export DEFAULT_ROUTING ?= basic
 export KUBECONFIG ?= ${HOME}/.kube/config
 export DEVWORKSPACE_API_VERSION ?= 1f335562c475972132851c68227dd36558317bb3
 
+# Set GO_PACKAGE_PATH to what is found in go.mod
+GO_PACKAGE_PATH ?= $(shell head -n1 go.mod | cut -d " " -f 2)
```
sleshchenko, line 29: can we move it closer to targets where it's used `compile_...` or even more to internal target, like `_eval_build_info`

Makefile
sleshchenko: now it should work and looks better.

Initially, I meant that we evaluate that variables during Dockerfile build, without Makefile at all. But I don't have any strong objection against reusing Makefile in Dockerfile, as far as it works fine for Dockerfile.rhel as well

build/rhel.Dockerfile
```diff
@@ -28,18 +28,8 @@ RUN go mod download
...
-  -gcflags all=-trimpath=/ \
-  -asmflags all=-trimpath=/ \
-  main.go
-RUN export ARCH="$(uname -m)" && if [[ ${ARCH} == "x86_64" ]]; then export ARCH="amd64"; elif [[ ${ARCH} == "aarch64" ]]; then export ARCH="arm64"; fi && \
-  CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} GO111MODULE=on go build \
-  -o _output/bin/webhook-server \
-  -gcflags all=-trimpath=/ \
-  -asmflags all=-trimpath=/ \
-  webhook/main.go
+RUN make devworkspace-controller
```
sleshchenko, line 31: It leads to non-critical but still warnings:

```
which: no kubectl in (/opt/rh/go-toolset-1.13/root/usr/bin:/opt/rh/go-toolset-1.13/root/usr/sbin:/opt/rh/go-toolset-1.13/root/usr/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
which: no oc in (/opt/rh/go-toolset-1.13/root/usr/bin:/opt/rh/go-toolset-1.13/root/usr/sbin:/opt/rh/go-toolset-1.13/root/usr/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
which: no controller-gen in (/opt/rh/go-toolset-1.13/root/usr/bin:/opt/rh/go-toolset-1.13/root/usr/sbin:/opt/rh/go-toolset-1.13/root/usr/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
```

Makefile
tinakurian: This was to avoid errors around kubetclt/oc not being present. in particular this: 

```
ifeq (,$(shell which kubectl))
ifeq (,$(shell which oc))
$(error oc or kubectl is required to proceed)
```

deploy.mk should only be included when oc and/or kubectl is present

Makefile
tinakurian: sorry i didn't realize there was a comment that said #internal params :woman_facepalming:  would it be okay if i moved it where the rest of the internal params are maybe?

Makefile
tinakurian: Sorry, missed this comment earlier. Okay, I've removed it!

k8s.mk
tinakurian: Makes sense - I've renamed it to deploy.mk!

k8s.mk
```diff
@@ -0,0 +1,116 @@
...
+_login_with_devworkspace_sa:
+	$(eval SA_TOKEN := $(shell $(K8S_CLI) get secrets -o=json -n $(NAMESPACE) | jq -r '[.items[] | select (.type == "kubernetes.io/service-account-token" and .metadata.annotations."kubernetes.io/service-account.name" == "$(DEVWORKSPACE_CTRL_SA)")][0].data.token' | base64 --decode ))
+	echo "Logging as controller's SA in $(NAMESPACE)"
+	oc login --token=$(SA_TOKEN) --kubeconfig=$(BUMPED_KUBECONFIG)
+
+### install_cert_manager: install Cert Mananger v1.0.4 on the cluster
+install_cert_manager:
+	${K8S_CLI} apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v1.0.4/cert-manager.yaml
+
+_operator_sdk:
```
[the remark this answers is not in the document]
tinakurian: Doesn't look like it - good catch! I've removed it!

build/rhel.Dockerfile
tinakurian: Okay, should be fixed now!

Makefile
amisevsk: I know this one is outdated, but I think it makes sense to keep all the top-level vars in one place; we've gotten confused by putting this stuff in a rule before. Variable declarations like this are processed before any rules.

Makefile
amisevsk: The issue is that other rules will fail for a more cryptic reason if `kubectl/oc` aren't installed, e.g.
```
make run
Current env vars:
    DWO_IMG=quay.io/devfile/devworkspace-controller:next
    PULL_POLICY=Always
    ROUTING_SUFFIX=192.168.99.100.nip.io
    DEFAULT_ROUTING=basic
    DEVWORKSPACE_API_VERSION=1f335562c475972132851c68227dd36558317bb3
Versions:
    GIT_COMMIT_ID=819de03bd85fa9c6ea67596f575e5a0bc36b7eb3
    BUILD_TIME=2021-04-27T19:05:33Z
make: *** No rule to make target '_gen_configuration_env', needed by 'run'.  Stop.
```

Makefile
```diff
@@ -30,35 +30,16 @@ DEVWORKSPACE_CTRL_SA=devworkspace-controller-serviceaccount
...
-# minikube handling
-ifeq ($(shell $(K8S_CLI) config current-context 2>&1),minikube)
-# check ingress addon is enabled
-ifeq ($(shell minikube addons list -o json | jq -r .ingress.Status), disabled)
-$(error ingress addon should be enabled on top of minikube)
-endif
-export ROUTING_SUFFIX := $(shell minikube ip).nip.io
+ifneq (,$(shell which kubectl 2>/dev/null)$(shell which oc 2>/dev/null))
+include deploy.mk
 endif
```
amisevsk, line 40: The check for `oc/kubectl` is necessary as most rules require it in some way or another. This should either always include `deploy.mk` and let it fail if kubectl/oc are not installed, or catch this case with an else:
```makefile
ifneq (,$(shell which kubectl 2>/dev/null)$(shell which oc 2>/dev/null))
include deploy.mk
else
$(error either oc or kubectl must be installed)
endif
```

Makefile
tinakurian: @amisevsk Yeah, that makes sense.. except since we are calling these new targets from the docker file it doesn't have oc/kubectl so we will always see this error if we always include it... hmm.. maybe there is another way around this though?

Makefile
tinakurian: Okay, will keep it at the top. I've added it with the other top-level vars

Makefile
```diff
@@ -305,35 +229,38 @@ else
...
-	@echo 'WARN: Please use the recommended operator-sdk if you face any issue.'
-endif
-
-_check_cert_manager:
-ifeq ($(PLATFORM),kubernetes)
-	if ! ${K8S_CLI} api-versions | grep -q '^cert-manager.io/v1$$' ; then \
-		echo "Cert-manager is required for deploying on Kubernetes. See 'make install_cert_manager'" ;\
-		exit 1 ;\
+.PHONY: compile-devworkspace-controller
+compile-devworkspace-controller:
```
amisevsk, line 234: Could you add some doc lines according to the hacked-together `help:` rule we have?
```suggestion
.PHONY: compile-devworkspace-controller
### compile-devworkspace-controller: compile the devworkspace-controller binary
compile-devworkspace-controller:
```

Makefile
```diff
@@ -305,35 +229,38 @@ else
...
+	export ARCH="$(shell uname -m)";
+	if [[ "$(ARCH)" == "x86_64" ]]; then
+		export ARCH="amd64";
+	elif [[ "$(ARCH)" == "aarch64" ]]; then
+		export ARCH="arm64";
 	fi
```
amisevsk, line 240: The whole rule should take place on one line or Florent will complain :D (unless these rules are only used within docker builds with a modern version of `make` installed). Something like
```suggestion
	export ARCH="$(shell uname -m)"; \
	if [[ "$(ARCH)" == "x86_64" ]]; then \
		export ARCH="amd64"; \
	elif [[ "$(ARCH)" == "aarch64" ]]; then \
		export ARCH="arm64"; \
	fi ;\
```

This is because macOS by default uses `make` from 15 years ago before `.ONESHELL` was added.

Makefile
amisevsk: Ohh I see the problem, in the dockerfile we don't have `oc` and `kubectl` installed. Hmmm -- the only way I see around this that keeps the user-friendly(-ish) error message is something like a hidden target
```makefile
_check_oc_kubectl:
	if ! command -v kubectl &> /dev/null && ! command -v oc &> /dev/null; then \
		echo "Either kubectl or oc must be installed for this target" ;\
		exit 1 ;\
	fi
```
that all targets that use `oc`/`kubectl` depend on

Makefile
```diff
@@ -30,35 +30,16 @@ DEVWORKSPACE_CTRL_SA=devworkspace-controller-serviceaccount
...
-
-# minikube handling
-ifeq ($(shell $(K8S_CLI) config current-context 2>&1),minikube)
-# check ingress addon is enabled
-ifeq ($(shell minikube addons list -o json | jq -r .ingress.Status), disabled)
-$(error ingress addon should be enabled on top of minikube)
-endif
-export ROUTING_SUFFIX := $(shell minikube ip).nip.io
+ifneq (,$(shell which kubectl 2>/dev/null)$(shell which oc 2>/dev/null))
+include deploy.mk
```
amisevsk, line 39: Splitting the file in two breaks our `make help` functionality :(

Line 269 in `Makefile` should be updated to
```bash
	@sed -n 's/^### /    /p' $< deploy.mk | awk 'BEGIN { FS=":" } { printf "%-30s -%s\n", $$1, $$2 }'
```

amisevsk (commented):
Generally LGTM. I like splitting the makefile but it does make a few things more complicated. I'm also concerned that there's a cross-referencing for rules (e.g. `deploy.mk` uses `_print_vars` which is in `Makefile`) but I also don't think that's a huge deal.

Makefile
tinakurian: @amisevsk yeahh.. i was thinking to update it to be like this: https://github.com/kwk/llvm-ci/blob/ffa5ce5245/infra/help.mk#L10-L35 - I've tested it and it seems to work! WDYT?

Makefile
tinakurian: @amisevsk I've moved the other oc/kubectl targets to the deploy.mk - sorry for the confusion, should have been more clear on the split :woman_facepalming:

Makefile
tinakurian: :smile: Sorry, didn't realize that - I've updated it!

Makefile
tinakurian: I've moved the other related targets to the deploy.mk - sorry for not being more clear :woman_facepalming:

Makefile
tinakurian: Actually for now, maybe lets just go with adding deploy.mk to the help target. We can think about updating the help if we need to later on! I've updated to add deploy.mk!

Makefile
tinakurian: Yes! Sorry! I've added this!

Makefile
```diff
@@ -305,41 +203,46 @@ else
...
-endif
+### compile-devworkspace-controller: Compiles the devworkspace-controller binary
+.PHONY: compile-devworkspace-controller
+compile-devworkspace-controller:
+	export ARCH="$(shell uname -m)";  \
+	if [[ "$(ARCH)" == "x86_64" ]]; then  \
+		export ARCH="amd64";  \
+	elif [[ "$(ARCH)" == "aarch64" ]]; then  \
+		export ARCH="arm64";  \
+	fi  \
```
amisevsk, line 215: Missing a semicolon here (and in `compile-webhook-server`):
```suggestion
	fi  ;\
```

```
Step 8/21 : RUN make compile-devworkspace-controller
 ---> Running in cc6ffbdb3a97
bash: -c: line 6: syntax error near unexpected token `CGO_ENABLED=0'
make: *** [Makefile:210: compile-devworkspace-controller] Error 2
The command '/bin/sh -c make compile-devworkspace-controller' returned a non-zero code: 2
make: *** [Makefile:178: docker-build] Error 2
```

Makefile
```diff
@@ -305,41 +203,46 @@ else
...
+	-X $(GO_PACKAGE_PATH)/version.BuildTime=$(BUILD_TIME)" \
+	webhook/main.go
 
 .PHONY: help
-### help: print this message
+### help: Prints this message
 help: Makefile
 	@echo 'Available rules:'
-	@sed -n 's/^### /    /p' $< | awk 'BEGIN { FS=":" } { printf "%-30s -%s\n", $$1, $$2 }'
+	@sed -n 's/^### /    /p' $< deploy.mk | awk 'BEGIN { FS=":" } { printf "%-30s -%s\n", $$1, $$2 }'
```
amisevsk, line 245: I like your suggestion of using `$(MAKEFILE_LIST)` here -- that way the help stays accurate if `deploy.mk` is not included.
```suggestion
	@sed -n 's/^### /    /p' $(MAKEFILE_LIST) | awk 'BEGIN { FS=":" } { printf "%-30s -%s\n", $$1, $$2 }'
```

Makefile
```diff
@@ -117,118 +86,47 @@ update_devworkspace_api:
...
 
-### test_e2e_debug: runs e2e test in debug mode, so it's possible to connect to execution via remote debugger
+### test_e2e_debug: Runs e2e test in debug mode, so it's possible to connect to execution via remote debugger
 test_e2e_debug:
 	dlv test --listen=:2345 --headless=true --api-version=2 ./test/e2e/cmd/workspaces_test.go -- --ginkgo.failFast
 
-### manager: Build manager binary
+### manager: Builds the manager binary
 manager: generate fmt vet
 	go build -o bin/manager main.go
```
amisevsk, line 112: While I'm nitpicking, we should just remove this target from the file completely. It's vestigial at this point.

amisevsk (approved):
A couple of minor fixups but apart from that LGTM!

Thanks for fixing the wording on help messages, too :)

Makefile
tinakurian: yikes! good catch, added it :-)

Makefile
tinakurian: Okay :-) done!

Makefile
tinakurian: Okay, sounds good, I've removed it!

sleshchenko (approved):
LGTM

Makefile
```diff
@@ -30,35 +30,16 @@ DEVWORKSPACE_CTRL_SA=devworkspace-controller-serviceaccount
 INTERNAL_TMP_DIR=/tmp/devworkspace-controller
 BUMPED_KUBECONFIG=$(INTERNAL_TMP_DIR)/kubeconfig
 RELATED_IMAGES_FILE=$(INTERNAL_TMP_DIR)/environment
+GO_PACKAGE_PATH ?= $(shell head -n1 go.mod | cut -d " " -f 2)
+BUILD_TIME = $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+REF = $(shell cat .git/HEAD | sed 's|ref: ||')
+GIT_COMMIT_ID := $(shell cat .git/${REF})
 
-ifeq (,$(shell which kubectl))
```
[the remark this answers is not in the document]
tinakurian: Heh, sorry -  this was because i changed the order by splitting up our old Makefile. I added a default so that the help target is what gets called when you type `make` now.

Makefile
```diff
@@ -30,35 +30,16 @@ DEVWORKSPACE_CTRL_SA=devworkspace-controller-serviceaccount
 INTERNAL_TMP_DIR=/tmp/devworkspace-controller
 BUMPED_KUBECONFIG=$(INTERNAL_TMP_DIR)/kubeconfig
 RELATED_IMAGES_FILE=$(INTERNAL_TMP_DIR)/environment
+GO_PACKAGE_PATH ?= $(shell head -n1 go.mod | cut -d " " -f 2)
+BUILD_TIME = $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
```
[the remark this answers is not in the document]
tinakurian: sure, i've moved them into the two targets!

sleshchenko (approved):

PR #394, opened 2021-04-22
Configure Dependabot for Dockerfiles

mkuznyetsov:

### What does this PR do?
Configure dependabot for Dockerfiles, so it will propose updates to the new versions of base images

https://github.com/eclipse/che/issues/18831

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,4 +6,4 @@ updates:
     directory: "/build"
     # Check for updates once a week
     schedule:
-      interval: "weekly"
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for Docker
+  - package-ecosystem: "docker"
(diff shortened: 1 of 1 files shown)


nickboldt (approved):
Let's merge this and see what happens. If nothing happens after a few days, we could try a shorter interval to get faster feedback than `weekly`. Are crontab options available, to do @hourly? for example?

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for Docker
+  - package-ecosystem: "docker"
+    # Look for a `Dockerfile` in the `build` directory
+    directory: "/build"
+    # Check for updates once a week
+    schedule:
+      interval: "weekly"
```
sleshchenko, line 9:
```suggestion
      interval: "weekly"

```

sleshchenko: and yeah, maybe let's start with daily and if it will spam us too much, we'll be able to return back to weekly

sleshchenko (approved):

nickboldt:
merged and c'picked to 0.3.x

PR #395, opened 2021-04-23
Simplify condition handling for DevWorkspaces

amisevsk:
### What does this PR do?
Rework how conditions are synced for simplicity and avoid updating `LastTransitionTime` unless the condition's status actually has transitioned. 

When testing DWO in `crc` with many DevWorkspaces running at once, I found that repeatedly updating the status caused every DevWorkspace to reconcile in a loop:

running-devworkspaces

(Effect is seen in the last 10 minutes of these metrics0

### Is it tested? How?
Test that conditions are set to `true|false|unknown` as expected.

### DIFF
--- a/controllers/workspace/status.go
+++ b/controllers/workspace/status.go
@@ -19,7 +19,6 @@ import (
 	"net/http"
 	"net/url"
 	"sort"
-	"strings"
 
 	"github.com/go-logr/logr"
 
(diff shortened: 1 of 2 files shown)


controllers/workspace/status.go
```diff
@@ -109,6 +81,51 @@ func (r *DevWorkspaceReconciler) updateWorkspaceStatus(workspace *dw.DevWorkspac
...
+			LastTransitionTime: currTransitionTime,
+			Type:               condType,
+			Status:             cond.Status,
+			Message:            cond.Message,
+		})
+	}
+
+	// Sort conditions to avoid unnecessary updates
+	sort.SliceStable(workspaceStatus.Conditions, func(i, j int) bool {
+		return strings.Compare(string(workspaceStatus.Conditions[i].Type), string(workspaceStatus.Conditions[j].Type)) > 0
```
sleshchenko, line 125: Would it be a good UX improvement if you order conditions in the order they are expected to appear?

sleshchenko (approved):

JPinkney (approved):

controllers/workspace/status.go
amisevsk: :+1: Done

PR #396, opened 2021-04-23
Enable concurrent reconciles

amisevsk:
### What does this PR do?
Enable concurrent reconciles by setting an env var. Default value is 1 since I can't really tell if increasing concurrence helps.

### Is it tested? How?
Should be no difference.

### DIFF
--- a/deploy/templates/components/manager/manager.yaml
+++ b/deploy/templates/components/manager/manager.yaml
@@ -58,6 +58,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.serviceAccountName
+            - name: MAX_CONCURRENT_RECONCILES
+              value: "1"
             - name: RELATED_IMAGE_plugin_redhat_developer_web_terminal_4_5_0
               value: "quay.io/eclipse/che-machine-exec:nightly"
             - name: RELATED_IMAGE_web_terminal_tooling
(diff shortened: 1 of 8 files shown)


deploy/templates/components/manager/manager.yaml
```diff
@@ -58,6 +58,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.serviceAccountName
+            - name: MAX_CONCURRENT_RECONCILES
+              value: "1"
```
sleshchenko, line 62: Does it make sense to increase it to a few?
If it does not influence memory consumption significantly, I don't see a reason why not to set like 3,5 here.

sleshchenko (approved):

JPinkney (approved):

deploy/templates/components/manager/manager.yaml
amisevsk: From what I can tell, memory consumption is not meaningfully impacted, but at the same time increasing concurrency should be done at the same time as increasing resource limits (CPU in particular). From brief testing using workspace annotations for start time, increasing concurrency to 3, 5, or 10 had no real effect in startup time when starting 5 workspaces simultaneously (because doing two things at half speed is effectively the same as doing one thing at normal speed). 

Until we can accurately test this better, I opted to leave it at the default value to avoid future issues. I suspect that my locally-running clusters are already CPU-starved so it's hard to test higher CPU limits accurately. 

For example, here's some metrics from `crc` where we have either `MAX_CONCURRENT_RECONCILES=1` or `MAX_CONCURRENT_RECONCILES=5`, with the current CPU limits. I'm starting a simple workspace every two seconds and deleting each workspace after 30 seconds (so a consistent ~15 starting workspaces on the cluster):

concurrence-1-v-5

(Note the spike is due to other load on the system I think). With 5 concurrence, I'm seeing way fewer reconciles in the 25-50ms range, so our throughput may not actually be increasing with concurrency.

PR #402, opened 2021-04-28
Set the project clone RELATED_IMAGE env var on releases

amisevsk:
### What does this PR do?
Set the value of `RELATED_IMAGE_project_clone` used by the DevWorkspace operator to match the released version of DWO. I had mixed up how DWO_IMG is used in `generate-deployment.sh` so it's actually easier to do this than I thought.

Follow up on PR review for https://github.com/devfile/devworkspace-operator/pull/371

### Is it tested? How?
`./deploy/generate-deployment.sh --use-defaults --project-clone-image test:test` should modify templates as expected.

### DIFF
--- a/make-release.sh
+++ b/make-release.sh
@@ -120,7 +120,10 @@ docker build -t "${PROJECT_CLONE_QUAY_REPO}" -f ./project-clone/Dockerfile .
 docker push "${PROJECT_CLONE_QUAY_REPO}"
 
 set -x
-bash -x ./deploy/generate-deployment.sh --use-defaults --default-image ${QUAY_REPO}
+bash -x ./deploy/generate-deployment.sh \
+  --use-defaults \
+  --default-image ${QUAY_REPO} \
+  --project-clone-image ${PROJECT_CLONE_QUAY_REPO}
 
 # tag the release if the version/version.go file has changed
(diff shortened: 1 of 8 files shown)


make-release.sh
```diff
@@ -120,7 +120,10 @@ docker build -t "${PROJECT_CLONE_QUAY_REPO}" -f ./project-clone/Dockerfile .
 docker push "${PROJECT_CLONE_QUAY_REPO}"
 
 set -x
-bash -x ./deploy/generate-deployment.sh --use-defaults --default-image ${QUAY_REPO}
+bash -x ./deploy/generate-deployment.sh \
+  --use-defaults \
+  --default-image ${QUAY_REPO} \
+  --project-clone-image ${PROJECT_CLONE_QUAY_REPO}
```
sleshchenko, line 126: Well, it's out of the current PR scope, I must miss it in the previous PRs, and I think it's just sync with `QUAY_REPO` but it's not actually a good name. It's not just repo but image
```suggestion
  --project-clone-image ${PROJECT_CLONE_QUAY_IMG}
```

sleshchenko: Please consider changing QUAY_REPO -> DWO_IMG or QUAY_DWO_IMG

sleshchenko (approved):

Issue #384, opened 2021-04-14
Provide a mechanism to automount configmap/secret into devworkspaces


sleshchenko:
Che Server provides an ability to annotate secrets in a special way which will lead it to be mounted into each/specified devworkspace/container.

DevWorkspace Operator should provide the same ability, it's going to be used for:
- user may store maven settings or other stuff there;
- Che can provide custom CA or different settings, like registries URL, dashboard URL, ...;

P.S. we may need also provide additional selectors to match only Che or WTO devworkspaces.

sleshchenko:
fixed by https://github.com/devfile/devworkspace-operator/pull/417

PR #417, opened 2021-05-05
Automount configmaps/secrets into devworkspaces

tinakurian:
### What does this PR do?
Automounts configmaps/secrets to **all** devworkspace if the automount label is present. 

https://github.com/devfile/devworkspace-operator/issues/384

### Is it tested? How?
1. make install
2. 
```
cat << EOF | oc apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: file-configmap
  labels:
#  annotations:
#    controller.devfile.io/mount-as: file # default behavior
type: Opaque
data:
  important-cfg: "test"
EOF
```
3.
```
cat << EOF | oc apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: file-configmap-custom-path
  labels:
  annotations:
#    controller.devfile.io/mount-as: file # default behavior
    controller.devfile.io/mount-path: /etc/config/custom-mount-path/here
type: Opaque
data:
  important-cfg: "test"
EOF
```
4.
```
cat << EOF | oc apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: env-var-secret
  labels:
  annotations:
    controller.devfile.io/mount-as: env
stringData:
  secret_username: admin
  secret_password: t0p-Secret
EOF
```
4. oc apply -f samples/web-terminal-custom-tooling.yaml
5. kubectl exec -it <pod> -c tooling -- bash
6. ls /etc/config
7. env | grep secret_

### DIFF
--- a/controllers/workspace/provision/deployment.go
+++ b/controllers/workspace/provision/deployment.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
@@ -70,9 +71,34 @@ func SyncDeploymentToCluster(
 	podAdditions []v1alpha1.PodAdditions,
 	saName string,
 	clusterAPI ClusterAPI) DeploymentProvisioningStatus {
+
+	cmPodAdditions, configmapEnvFromSourceAdditions, err := getDevWorkspaceConfigmaps(workspace.Namespace, clusterAPI.Client)
+	if err != nil {
+		return DeploymentProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{Err: err},
+		}
+	}
+	podAdditions = append(podAdditions, *cmPodAdditions)
+
+	sPodAdditions, secretEnvFromSourceAdditions, err := getDevWorkspaceSecrets(workspace.Namespace, clusterAPI.Client)
+	if err != nil {
+		return DeploymentProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{Err: err},
+		}
+	}
+	podAdditions = append(podAdditions, *sPodAdditions)
+
+	var envFromSourceAdditions []corev1.EnvFromSource
+	if configmapEnvFromSourceAdditions != nil {
+		envFromSourceAdditions = append(envFromSourceAdditions, configmapEnvFromSourceAdditions...)
+	}
+	if secretEnvFromSourceAdditions != nil {
+		envFromSourceAdditions = append(envFromSourceAdditions, secretEnvFromSourceAdditions...)
+	}
+
 	// [design] we have to pass components and routing pod additions separately because we need mountsources from each
 	// component.
-	specDeployment, err := getSpecDeployment(workspace, podAdditions, saName, clusterAPI.Scheme)
+	specDeployment, err := getSpecDeployment(workspace, podAdditions, envFromSourceAdditions, saName, clusterAPI.Scheme)
 	if err != nil {
 		return DeploymentProvisioningStatus{
 			ProvisioningStatus: ProvisioningStatus{
@@ -209,6 +235,7 @@ func checkDeploymentStatus(deployment *appsv1.Deployment) (ready bool) {
 func getSpecDeployment(
 	workspace *dw.DevWorkspace,
 	podAdditionsList []v1alpha1.PodAdditions,
+	envFromSourceAdditions []corev1.EnvFromSource,
 	saName string,
 	scheme *runtime.Scheme) (*appsv1.Deployment, error) {
 	replicas := int32(1)
@@ -220,14 +247,17 @@ func getSpecDeployment(
 	}
 
 	creator := workspace.Labels[constants.DevWorkspaceCreatorLabel]
-	commonEnv := env.CommonEnvironmentVariables(workspace.Name, workspace.Status.DevWorkspaceId, workspace.Namespace, creator)
+	var envVars []corev1.EnvVar
+	envVars = append(envVars, env.CommonEnvironmentVariables(workspace.Name, workspace.Status.DevWorkspaceId, workspace.Namespace, creator)...)
 	for idx := range podAdditions.Containers {
-		podAdditions.Containers[idx].Env = append(podAdditions.Containers[idx].Env, commonEnv...)
+		podAdditions.Containers[idx].Env = append(podAdditions.Containers[idx].Env, envVars...)
 		podAdditions.Containers[idx].VolumeMounts = append(podAdditions.Containers[idx].VolumeMounts, podAdditions.VolumeMounts...)
+		podAdditions.Containers[idx].EnvFrom = append(podAdditions.Containers[idx].EnvFrom, envFromSourceAdditions...)
 	}
 	for idx := range podAdditions.InitContainers {
-		podAdditions.InitContainers[idx].Env = append(podAdditions.InitContainers[idx].Env, commonEnv...)
+		podAdditions.InitContainers[idx].Env = append(podAdditions.InitContainers[idx].Env, envVars...)
 		podAdditions.InitContainers[idx].VolumeMounts = append(podAdditions.InitContainers[idx].VolumeMounts, podAdditions.VolumeMounts...)
+		podAdditions.InitContainers[idx].EnvFrom = append(podAdditions.InitContainers[idx].EnvFrom, envFromSourceAdditions...)
 	}
 
 	deployment := &appsv1.Deployment{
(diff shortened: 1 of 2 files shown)


controllers/workspace/provision/deployment.go
```diff
@@ -70,9 +70,24 @@ func SyncDeploymentToCluster(
...
+	if err != nil {
+		return DeploymentProvisioningStatus{
+			ProvisioningStatus: ProvisioningStatus{Err: err},
+		}
+	}
+
 	// [design] we have to pass components and routing pod additions separately because we need mountsources from each
 	// component.
-	specDeployment, err := getSpecDeployment(workspace, podAdditions, saName, clusterAPI.Scheme)
+	specDeployment, err := getSpecDeployment(workspace, configmaps, secrets, podAdditions, saName, clusterAPI.Scheme)
```
amisevsk, line 90: An alternate approach I would suggest here is to reuse `volumes` and volumeMounts` in `podAdditionsList`. This function (`getSpecDeployment`) will already automatically add all volumes/volumeMounts to the workspace deployment/containers, so we're kind of duplicating work right now.

In other words, we could have `getDevWorkspaceSecrets` and `getDevWorkspaceConfigMaps` return `*v1alpha1.podAdditions`, and then just append those to the `podAdditions` we have here. Most of the code required is already present, so this would mainly be reorganizing things.

controllers/workspace/provision/deployment.go
```diff
@@ -433,6 +511,41 @@ func getWorkspaceSubpathVolumeMount(workspaceId string) corev1.VolumeMount {
...
+		},
+	}
+	return workspaceVolumeMount
+}
+
+func getVolumeMounts(segment, name string) corev1.VolumeMount {
+	workspaceVolumeMount := corev1.VolumeMount{
+		Name:      name,
+		ReadOnly:  true,
+		MountPath: "/etc" + segment + name,
```
amisevsk, line 544:
```suggestion
		MountPath: path.Join("/etc/", segment, name),
```

amisevsk (commented):
A couple structural comments

pkg/constants/metadata.go
```diff
@@ -23,6 +23,9 @@ const (
 	// DevWorkspaceNameLabel is label key to store workspace name
 	DevWorkspaceNameLabel = "controller.devfile.io/devworkspace_name"
 
+	// DevWorkspaceAutomountLabel is label key to store if a configmap or secret she should be automounted to the devworkspace
+	DevWorkspaceAutomountLabel = "controller.devfile.io/automount-to-workspace"
```
benoitf, line 27: hello, I was wondering if we need the `auto` prefix because I don't see why it can't be automatically mounted if I only specify "controller.devfile.io/mount-to-workspace" ( auto = the job of the controller anyway)

for example in devfiles I have `mountSources : true|false`

controllers/workspace/provision/deployment.go
tinakurian: Got it! Finally. Much cleaner! Thanks @amisevsk

pkg/constants/metadata.go
tinakurian: @benoitf makes sense, I've changed the label to the following: 
```
DevWorkspaceMountLabel = "controller.devfile.io/mount-to-workspace"
```

controllers/workspace/provision/deployment.go
```diff
@@ -433,6 +481,41 @@ func getWorkspaceSubpathVolumeMount(workspaceId string) corev1.VolumeMount {
...
+		Name: name,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: name,
+				},
+			},
+		},
+	}
+	return workspaceVolumeMount
```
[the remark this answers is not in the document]
tinakurian: Sounds good, i've added it to `getVolumeWithSecret` as well! Thanks @amisevsk :smile:

pkg/constants/metadata.go
```diff
@@ -14,15 +14,18 @@ package constants
...
 	DevWorkspaceIDLabel = "controller.devfile.io/devworkspace_id"
 
 	// DevWorkspaceCreatorLabel is the label key for storing the UID of the user who created the workspace
 	DevWorkspaceCreatorLabel = "controller.devfile.io/creator"
 
-	// DevWorkspaceNameLabel is label key to store workspace name
+	// DevWorkspaceNameLabel is the label key to store workspace name
 	DevWorkspaceNameLabel = "controller.devfile.io/devworkspace_name"
 
+	// DevWorkspaceMountLabel is the label key to store if a configmap or secret she should be mounted to the devworkspace
```
sleshchenko, line 26:
```suggestion
	// DevWorkspaceMountLabel is the label key to store if a configmap or secret should be mounted to the devworkspace
```

pkg/constants/metadata.go
```diff
@@ -14,15 +14,18 @@ package constants
...
+	DevWorkspaceMountLabel = "controller.devfile.io/mount-to-workspace"
```
sleshchenko, line 27: we would be a bit more consistent with devworkspace_id, devworkspace_name then
```suggestion
	DevWorkspaceMountLabel = "controller.devfile.io/mount-to-devworkspace"
```

controllers/workspace/provision/deployment.go
```diff
@@ -433,6 +481,45 @@ func getWorkspaceSubpathVolumeMount(workspaceId string) corev1.VolumeMount {
...
+}
+
+func getVolumeMounts(segment, name string) corev1.VolumeMount {
+	workspaceVolumeMount := corev1.VolumeMount{
+		Name:      name,
+		ReadOnly:  true,
+		MountPath: path.Join("/etc/", segment, name),
```
sleshchenko, line 518: What Che currently has:
- an ability to customize is it should be mounted as env var or file;
- an ability to customize where the file should be mounted;
- an ability to specify container name where it should be injected into;

sleshchenko (approved):
In general LGTM, just a few stuff that Che may also need.
But I'm OK with handling it out of the current PR

pkg/constants/metadata.go
tinakurian: :+1: thanks for catching that!

controllers/workspace/provision/deployment.go
sleshchenko: and yeah, sorry for missing it before, the items above I got from Che Documentation https://www.eclipse.org/che/docs/che-7/end-user-guide/mounting-a-secret-as-a-file-or-an-environment-variable-into-a-workspace-container/

sleshchenko: @tinakurian thanks for catching that item I referenced to

> an ability to specify container name where it should be injected into;

is actually deprecated. And instead users are able to define
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: git-credentials-secret
  labels:
    app.kubernetes.io/part-of: che.eclipse.org
    app.kubernetes.io/component: workspace-secret
  annotations:
    che.eclipse.org/automount-workspace-secret: false
```
and then mount such disabled by default secrets into the particular component with [automountWorkspaceSecrets](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/resources/schema/1.0.0/devfile.json#L277).

Let's move this feature to a dedicated issue where we agreed on the desired model and workflow for devfile v2 and devworkspaces.

amisevsk (approved):
PR in its current state looks good to me -- I'm :+1: to merge as-is and add additional features (mount as env or mount as volume) separately, or add to this PR if that's easier.

pkg/constants/metadata.go
```diff
@@ -14,15 +14,24 @@ package constants
...
+	// DevWorkspaceNameLabel is the label key to store workspace name
 	DevWorkspaceNameLabel = "controller.devfile.io/devworkspace_name"
 
+	// DevWorkspaceMountLabel is the label key to store if a configmap or secret should be mounted to the devworkspace
+	DevWorkspaceMountLabel = "controller.devfile.io/mount-to-devworkspace"
+
+	// DevWorkspaceMountPathAnnotation is the label key to store the mount path for the secret or configmap
+	DevWorkspaceMountPathAnnotation = "controller.devfile.io/mount-path"
+
+	// DevWorkspaceMountAsAnnotation is the label key to store if the configmap or secret should be mounted as a file or environment variables
```
sleshchenko, line 32: please consider adding expected possible values. I guess it's `env`/`file`

controllers/workspace/provision/deployment.go
```diff
@@ -453,3 +574,11 @@ func checkContainerStatusForFailure(containerStatus *corev1.ContainerStatus) (ok
 	}
 	return true
 }
+
+func getEnvVar(key, value string) corev1.EnvVar {
+	envKey := DevWorkspaceEnvPrefix + "_" + (strings.ToUpper(strings.ReplaceAll(strings.ReplaceAll(key, ".", "_"), "-", "_")))
```
sleshchenko, line 579: Does Che Server appends their prefix as well?

If we replace . and -, should not we take care of other not allowed special characters, !@#$%... ?

sleshchenko (commented):
LGTM
I'm going to test it tomorrow
It seems to be a good time that we start documenting devWorkspaces feature, but it's out of the current PR scope.

controllers/workspace/provision/deployment.go
```diff
@@ -337,6 +365,60 @@ func getPods(workspace *dw.DevWorkspace, client runtimeClient.Client) (*corev1.P
...
+	if err := client.List(context.TODO(), configmaps, k8sclient.InNamespace(namespace), k8sclient.MatchingLabels{
+		constants.DevWorkspaceMountLabel: "true",
+	}); err != nil {
+		return nil, nil, err
+	}
+	podAdditions := &v1alpha1.PodAdditions{}
+	additionalEnvVars := []corev1.EnvVar{}
+	for _, configmap := range configmaps.Items {
+		mountAs := configmap.Annotations[constants.DevWorkspaceMountAsAnnotation]
+		if mountAs == "env" {
```
amisevsk, line 379: I would assume this would be done via [`envFrom`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables). If someone is creating a configmap and specifically marking it as "mount this as environment variables", this is the k8s analaogue I'd choose.

sleshchenko: Do you mean linking to the configmap through K8s mechanism(instead of propagating it manually),
Not changing the expected value for this annotation?

pkg/constants/metadata.go
tinakurian: done!

amisevsk (approved):
Tested again on OpenShift (yesterday) and both mount-as: env and the default work for me. 

I did run into an issue on `crc` but I think that's a me problem rather than a problem with this code. That issue also reproduced on the main branch, also.

tinakurian:
> Tested again on OpenShift (yesterday) and both mount-as: env and the default work for me.
> 
> I did run into an issue on `crc` but I think that's a me problem rather than a problem with this code. That issue also reproduced on the main branch, also.

Thanks @amisevsk I realized the issue i was seeing happens if i just use `make install` + `make uninstall` a few times. But running `make docker install restart` seems to resolve it. Thanks for testing it for me! Greatly appreciated!

pkg/constants/metadata.go
```diff
@@ -14,15 +14,28 @@ package constants
...
 	DevWorkspaceCreatorLabel = "controller.devfile.io/creator"
 
-	// DevWorkspaceNameLabel is label key to store workspace name
+	// DevWorkspaceNameLabel is the label key to store workspace name
 	DevWorkspaceNameLabel = "controller.devfile.io/devworkspace_name"
 
+	// DevWorkspaceMountLabel is the label key to store if a configmap or secret should be mounted to the devworkspace
+	DevWorkspaceMountLabel = "controller.devfile.io/mount-to-devworkspace"
+
+	// DevWorkspaceMountPathAnnotation is the label key to store the mount path for the secret or configmap
```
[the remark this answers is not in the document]
sleshchenko: It makes sense to specify default mount paths here.

sleshchenko (approved):
LGTM I left just a few minor comments
Tested and works as expected. I hope you don't mind that I updated testing instructiond

tinakurian:
> LGTM I left just a few minor comments
> Tested and works as expected. I hope you don't mind that I updated testing instructiond

Not at all! Thanks for taking the time to update it @sleshchenko!

controllers/workspace/provision/deployment.go
sleshchenko: I've created  Add an ability to select secret/config maps to mount for particular devworkspace/component https://github.com/devfile/devworkspace-operator/issues/435, so resolving this conversation.

PR #420, opened 2021-05-06
Grab all env vars in manager.yaml when using make run/debug

amisevsk:
### What does this PR do?
To avoid having to maintain a list of environment variables the operator expects to support running locally, set all environment variables used by the controller instead of just RELATED_IMAGE ones

`make run` and `make debug` are broken

### Is it tested? How?
`make run`

`make debug`


tinakurian (approved):
LGTM!

JPinkney (approved):

PR #421, opened 2021-05-07
Rework version info to work without a `.git` directory

amisevsk:
### What does this PR do?
Rework Git commit SHA-1 resolving to 
1. Check for `.git/HEAD` and `.git/refs`, and use a value from there if possible
2. Otherwise, use value in file `.commit_hash` in root directory of repo
3. Otherwise, use `unknown`

Also, splits buildinfo resolving into a separate file, and move included makefiles into `build/make`

Builds fail in mid/downstream repos since `.git` directory is not present in root directory (instead, it's one level up)

### Is it tested? How?
Should work the same as before for this repo.


build/make/version.mk
```diff
@@ -0,0 +1,26 @@
...
+export GO_PACKAGE_PATH := $(shell head -n1 go.mod | cut -d " " -f 2)
+export BUILD_TIME := $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+
+# Architecture we're building on
+ifeq (x86_64,$(shell uname -m))
+export ARCH := "amd64"
+else ifeq (aarch64,$(shell uname -m))
+export ARCH := "arm64"
+else
+export ARCH := "unknown"
```
nickboldt, line 25: what about other arches like s390x and ppc64le, which are the same as their `uname -m` values? why fall back to unknown when you can fall back to `$(shell uname -m)` ?

build/make/version.mk
```diff
@@ -0,0 +1,26 @@
...
+else
+export GIT_COMMIT_ID := "unknown"
+endif
+
+# Additional build info
+export GO_PACKAGE_PATH := $(shell head -n1 go.mod | cut -d " " -f 2)
+export BUILD_TIME := $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+
+# Architecture we're building on
+ifeq (x86_64,$(shell uname -m))
```
nickboldt, line 20: insert after line 19:
```
export ARCH := $(shell uname -m)
```
(or whatever the make syntax is)

build/make/version.mk
amisevsk: Ah, didn't realize how that worked. I'll fix it.

nickboldt (approved):
+1

Issue #399, opened 2021-04-28
Adapt to ideUrl -> mainUrl renaming


sleshchenko:
DWO, devWorkspace client and dashboard should be adapted to ideUrl -> mainUrl renaming

JPinkney:
devworkspace-client PR: https://github.com/che-incubator/devworkspace-client/pull/22
dashboard PR: https://github.com/eclipse-che/che-dashboard/pull/232
devworkspace operator PR: https://github.com/devfile/devworkspace-operator/pull/423

sleshchenko:
All components now should be adapted to ideUrl -> mainUrl renaming.
But there is one item missing: make endpoints define `type: main` instead of `type: ide`. See https://github.com/devfile/devworkspace-operator/pull/423/files#r630758675
I assume this changes needs to be done on DWO and plugin registry side.

sleshchenko:
- [ ] Plugin registry PR https://github.com/eclipse-che/che-plugin-registry/pull/963

sleshchenko:
All expected changes are moved to use main endpoint type.

PR #423, opened 2021-05-11
Rename ideUrl to mainUrl

tinakurian:
### What does this PR do?
renames ideurl to mainurl

https://github.com/devfile/devworkspace-operator/issues/399

### Is it tested? How?
unit tests


controllers/workspace/status.go
```diff
@@ -155,7 +155,7 @@ func checkServerStatus(workspace *dw.DevWorkspace) (ok bool, err error) {
 	return ok, nil
 }
 
-func getIdeUrl(exposedEndpoints map[string]v1alpha1.ExposedEndpointList) string {
+func getMainUrl(exposedEndpoints map[string]v1alpha1.ExposedEndpointList) string {
 	for _, endpoints := range exposedEndpoints {
 		for _, endpoint := range endpoints {
 			if endpoint.Attributes.GetString(string(v1alpha1.TypeEndpointAttribute), nil) == "ide" {
```
sleshchenko, line 161: I believe we should change it accordingly.
@amisevsk Do you recall what we agreed on? Attribute `main: true`, or `type: main`?
Also, I'm OK with handling it in a separate PR since it depends on plugin registry in addition

amisevsk: No memory of the consensus, if we even reached one :D

sleshchenko:
As far as I see Theia does not use ideURL from workspace status yet. We should be fine with merging it

sleshchenko (approved):

amisevsk:
> As far as I see Theia does not use ideURL from workspace status yet. We should be fine with merging it

Haven't checked, but can the dashboard handle this change?

JPinkney:
Yeah, I made the changes for the dashboard side

Issue #413, opened 2021-05-03
Cannot start failed workspaces


amisevsk:
### Description
The controller will stop (i.e. set `.spec.started=false`) any workspace that has `.status.phase=failed`. Since starting a workspace does not clear the phase/conditions, it is impossible to start a workspace once it has entered the Failed state since it just hits the same condition and is stopped again.

We can work around this in multiple ways:
- We can use the `Failed` phase only to mark a running workspace that still has to be stopped, and use the `Stopped` phase once the stop has been completed (failure reason can still be shown in the INFO field).
- We can introduce a "Failing" state or similar, that is used to mark a workspace that has failed and should be stopped
- We can clear a DevWorkspace's `.status.phase` and `.status.conditions` when `.spec.started` is changed from false to true in a webhook

benoitf:
Also how can we grab logs of the pod (or one container of the pod) when it's stopping the workspace that has a failed status ?

amisevsk:
@benoitf That's another issue I encountered when I ran into this one (and probably deserves its own issue). My plan was to support an annotation or attribute on DevWorkspaces to direct the controller to not scale deployments to zero on failure (similar to how we have a "debug start" option in Che itself) -- WDYT?

Created https://github.com/devfile/devworkspace-operator/issues/418 for follow-up

benoitf:
@amisevsk yes I agree that it's a separate issue, thanks for the follow-up issue (and annotation's idea to control if scale to zero or not)

PR #424, opened 2021-05-11
Fix starting a workspace once it enters the failed state

amisevsk:
### What does this PR do?
Add a new `DevWorkspacePhase`, `"Failing"` to represent workspaces that have hit a fail condition but haven't yet been stopped. When the controller encounters a DevWorkspace with `phase: Failing` and `.spec.started: true`, it will stop that workspace and update the phase to `Failed`.

Fixes https://github.com/devfile/devworkspace-operator/issues/413

### Is it tested? How?
```yaml
kind: DevWorkspace
metadata:
  name: plain
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "exit"
           - "1"
EOF
kubectl get dw -w
```

1. The workspace should enter phases `Failing` and `Failed` with message `Container web-terminal has state RunContainerError`
2. If you restart the workspace (`kubectl patch dw plain --type merge -p '{"spec": {"started": true}}'`), it should enter the starting phase and continue until it fails a second time.

### DIFF
--- a/controllers/workspace/status.go
+++ b/controllers/workspace/status.go
@@ -31,6 +31,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// devworkspacePhaseTerminating represents a DevWorkspace that has been deleted but is waiting on a finalizer.
+	// TODO: Should be moved to devfile/api side.
+	devworkspacePhaseTerminating dw.DevWorkspacePhase = "Terminating"
+
+	// devworkspacePhaseFailing represents a DevWorkspace that has encountered an unrecoverable error and is in
+	// the process of stopping.
+	devworkspacePhaseFailing dw.DevWorkspacePhase = "Failing"
+)
+
 type currentStatus struct {
 	workspaceConditions
 	// Current workspace phase
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -124,7 +124,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 	}
 
 	// Stop failed workspaces
-	if workspace.Status.Phase == dw.DevWorkspaceStatusFailed && workspace.Spec.Started {
+	if workspace.Status.Phase == devworkspacePhaseFailing && workspace.Spec.Started {
 		patch := []byte(`{"spec":{"started": false}}`)
 		err := r.Client.Patch(context.Background(), workspace, client.RawPatch(types.MergePatchType, patch))
 		if err != nil {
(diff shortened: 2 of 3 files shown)


controllers/workspace/status.go
```diff
@@ -31,6 +31,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// devworkspacePhaseTerminating represents a DevWorkspace that has been deleted but is waiting on a finalizer.
+	// TODO: Should be moved to devfile/api side.
+	devworkspacePhaseTerminating dw.DevWorkspacePhase = "Terminating"
+
+	// devworkspacePhaseFailing represents a DevWorkspace that has encountered an unrecoverable error and will
+	// terminate
```
sleshchenko, line 40: this terminate at first look may be confusing since it's similar to `Terminating` about. Maybe this could help
```suggestion
	// terminate start up
```

controllers/workspace/status.go
```diff
@@ -31,6 +31,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// devworkspacePhaseTerminating represents a DevWorkspace that has been deleted but is waiting on a finalizer.
+	// TODO: Should be moved to devfile/api side.
+	devworkspacePhaseTerminating dw.DevWorkspacePhase = "Terminating"
+
+	// devworkspacePhaseFailing represents a DevWorkspace that has encountered an unrecoverable error and will
```
sleshchenko, line 39:
```suggestion
	// devworkspacePhaseFailing represents a DevWorkspace that has encountered an unrecoverable start-up error and will
```
Are we going to use it for anything else apart from startup failing?

sleshchenko (commented):
Changes LGTM. I'll test later

tinakurian (approved):

JPinkney (approved):

controllers/workspace/status.go
amisevsk: This is worth considering. I'd err on the side of generality, since it could potentially happen after a workspace is `Running` -- the logic applies based on whether or not `spec.started` is true. 

One scenario I could imagine us detecting in the future would be something like repeated OOMKilled in a workspace pod. Such a workspace would also require the `Failing` phase.

controllers/workspace/status.go
amisevsk: I reworded the whole section to avoid use of "terminate" since it does seem confusing with `Terminating` above. I changed it to

```
	// devworkspacePhaseFailing represents a DevWorkspace that has encountered an unrecoverable error and is in
	// the process of stopping.
```

sleshchenko (approved):
Tested with custom terminal image and works as expected, I was able to restart failed workspace
Screenshot_20210524_102445

Issue #418, opened 2021-05-06
Add annotation/attribute to disable scale-to-zero on DevWorkspace failure


amisevsk:
### Description

When a DevWorkspace enters a failed state, the deployment for it is scaled to zero. This means that any logs that could be used to debug a `CrashLoopBackoff` are unavailable. 

It would be great to have an annotation or attribute we can apply to DevWorkspaces to disable the scale-to-zero behavior and allow users to view logs when required.

sleshchenko:
:+1: And dashboard can expose this feature to a user with `Open In Verbose Mode`

PR #425, opened 2021-05-11
Add annotation for debugging workspace startup

amisevsk:
### What does this PR do?
Adds support for an annotation (`controller.devfile.io/debug-start: true`) on devworkspaces that disables scale-to-zero when a workspace fails. This leaves the workspace deployment/pod on the cluster, allowing for debugging issues with startup.

#### Notes
- **Annotation vs Attribute**: I opted to use annotations for this purpose since it's a DWO-specific feature and shouldn't be included in devfiles by default.
- **Annotation name**: Instead of something more specific like `disable-scale-to-zero` I named the annotation `controller.devfile.io/debug-start`. This annotation could be used for enabling additional debug information in the future.

Closes https://github.com/devfile/devworkspace-operator/issues/418

### Is it tested? How?
```yaml
kind: DevWorkspace
metadata:
  name: plain
  annotations:
    controller.devfile.io/debug-start: "true"
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "exit"
           - "1"
EOF
kubectl get dw -w
```
- Deployment and pod should be left on cluster once workspace fails

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -125,6 +125,12 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 
 	// Stop failed workspaces
 	if workspace.Status.Phase == devworkspacePhaseFailing && workspace.Spec.Started {
+		// If debug annotation is present, leave the deployment in place to let users
+		// view logs.
+		if workspace.Annotations[constants.DevWorkspaceDebugStartAnnotation] == "true" {
+			return reconcile.Result{}, nil
+		}
+
 		patch := []byte(`{"spec":{"started": false}}`)
 		err := r.Client.Patch(context.Background(), workspace, client.RawPatch(types.MergePatchType, patch))
 		if err != nil {
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -47,6 +47,10 @@ const (
 	// this annotation will be cleared
 	DevWorkspaceStopReasonAnnotation = "controller.devfile.io/stopped-by"
 
+	// DevWorkspaceDebugStartAnnotation enables debugging workspace startup if set to "true". If a workspace with this annotation
+	// fails to start (i.e. enters the "Failed" phase), its deployment will not be scaled down in order to allow viewing logs, etc.
+	DevWorkspaceDebugStartAnnotation = "controller.devfile.io/debug-start"
+
 	// WebhookRestartedAtAnnotation holds the the time (unixnano) of when the webhook server was forced to restart by controller
 	WebhookRestartedAtAnnotation = "controller.devfile.io/restarted-at"
 


pkg/constants/metadata.go
```diff
@@ -32,6 +32,10 @@ const (
 	// this annotation will be cleared
 	DevWorkspaceStopReasonAnnotation = "controller.devfile.io/stopped-by"
 
+	// DevWorkspaceDebugStartAnnotation enables debugging workspace startup if set to "true". If a workspace with this annotation
+	// fails to start (i.e. enters the "Failed" phase), its deployment will not be scaled down in order to allow viewing logs, etc.
+	DevWorkspaceDebugStartAnnotation = "controller.devfile.io/debug-start"
```
sleshchenko, line 37: Seems it's a good time(or too late) to agree on a convention on annotations/labels names we use.
We have - and _, like `stopped-by` `debug-start` but `endpoint_name`, `devworkspace_name`, `devworkspace_id`.
BTW we need to continue with one format even if leave another existing untouched.

controllers/workspace/devworkspace_controller.go
```diff
@@ -125,6 +125,12 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 
 	// Stop failed workspaces
 	if workspace.Status.Phase == dw.DevWorkspaceStatusFailed && workspace.Spec.Started {
+		// If debug annotation is present, leave the deployment in place to let users
```
sleshchenko, line 128: leaving the deployment in place is the result, but what actually it does - do not initialize stopping of failed/[failing](https://github.com/amisevsk/devworkspace-operator/blob/start-failed-workspace/controllers/workspace/devworkspace_controller.go#L127) workspace.

Do I understand correctly, that with https://github.com/devfile/devworkspace-operator/pull/424/files users are supposed to manually stop workspace before starting it again if debug mode is activated?

amisevsk: That would be the requirement, yes -- it leaves failed workspaces with `spec.started=true` so that you can debug issues (basically it treats workspaces with the annotation similar to how they were treated before https://github.com/devfile/devworkspace-operator/pull/362).

The other way to implement this would be to leave the deployment in place for stopped workspaces when the debug annotation is present, which I think is more confusing.

This PR will need to be reworked slightly once #424 is merged.

pkg/constants/metadata.go
sleshchenko: +1 to use - for a new annotations and properties.

sleshchenko (commented):
LGTM but I haven't tested since it's going to be reworked after PR with failing state is merged

amisevsk:
Currently, the way the debug annotation works is:

1. Apply debug annotation and start failing workspace
2. Workspace enters `failing` state
3. Debug annotation blocks stopping and entering `failed` state, leaving deployment on cluster.
4. Debug issues with deployment by checking logs, etc.
5. Once problem is figured out, remove debug annotation from DevWorkspace and workspace enters `failed` state, deployment is scaled to zero.
6. Workspace can be restarted normally.

I think this is an okay flow. Any suggestions for improvement? I considered trying to allow workspaces to still enter the `failed` state, but thinking about it more that may be more confusing and also make the flow above harder to use.

sleshchenko (approved):
LGTM

Recently, I already had a case when I could suggest QE to use to investigate failure

JPinkney (approved):

amisevsk:
> Recently, I already had a case when I could suggest QE to use to investigate failure

I've had to temporarily cherry-pick this commit into multiple PR branches to figure out why my workspace wasn't starting :D

PR #427, opened 2021-05-13
Upgrade the project to use go 1.14

JPinkney:

### What does this PR do?
This PR updates the project to use go 1.14. The main reason for this change is for:
```
By default, if the go version in go.mod is 1.14 or higher and a vendor directory is present, the go command acts as if -mod=vendor were used. Otherwise, the go command acts as if -mod=readonly were used.
```
from https://golang.org/ref/mod#build-commands making productization a bit simpler

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -9,7 +9,7 @@
 #   Red Hat, Inc. - initial API and implementation
 #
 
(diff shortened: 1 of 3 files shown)


sleshchenko (approved):

nickboldt:
cherrypicked to 0.4.x where I actually need this change :)

PR #433, opened 2021-05-21
Rework generate-deployment.sh to be simpler

amisevsk:
### What does this PR do?
Rework how generate-deployment.sh works to be simpler:
* Restrict envsubst to only replace predefined env vars
* Use envsubst on kustomize-built yaml to avoid needing to copy templates to .bak files

This PR is a cherry-pick of https://github.com/devfile/devworkspace-operator/pull/405/commits/14fa459e1192fd80976597d68f1f95d90af96b35 from #405 

N/A

### Is it tested? How?
Running `make generate_default_deployment` does not change `deploy/deployment`

### DIFF
--- a/deploy/generate-deployment.sh
+++ b/deploy/generate-deployment.sh
@@ -24,6 +24,9 @@
 
 set -e
 
+# List of environment variables that will be replaced by envsubst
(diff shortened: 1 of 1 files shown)


amisevsk:
This PR also removes the frustration of `.bak` files being left over if `generate_default_deployment.sh` is cancelled.

sleshchenko (approved):

Issue #392, opened 2021-04-21
Move storage type setting to devfile attributes


sleshchenko:
Currently storage type is configured via DevWorkspace CR label/annotation (can't tell for sure without checking code).
After https://github.com/devfile/api/pull/352#event-4621440702 is merged, it's needed to move it into devfile attributes, to make it possible to configure storage type via devfile in the repository root:

PR #437, opened 2021-05-25
Move storage type setting to devfile attributes

tinakurian:
### What does this PR do?
Moves the storage type setting from using lables to devfile attributes. 

https://github.com/devfile/devworkspace-operator/issues/392

### Is it tested? How?
tested to see if the storage type gets set properly when specifying it through devfile attributes using samples/with-k8s-ref/async-theia-next.yaml

### DIFF
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -39,13 +39,13 @@ const (
 	// The full annotation name is supposed to be "<routingClass>.routing.controller.devfile.io/<anything>"
 	RoutingAnnotationInfix = ".routing.controller.devfile.io/"
 
-	// DevWorkspaceStorageTypeLabel defines the strategy used for provisioning storage for the workspace.
+	// DevWorkspaceStorageTypeAtrr defines the strategy used for provisioning storage for the workspace.
 	// If empty, the common PVC strategy is used.
 	// Supported options:
 	// - "common": Create one PVC per namespace, and store data for all workspaces in that namespace in that PVC
 	// - "async" : Create one PVC per namespace, and create a remote server that syncs data from workspaces to the PVC.
 	//             All volumeMounts used for devworkspaces are emptyDir
-	DevWorkspaceStorageTypeLabel = "controller.devfile.io/storage-type"
+	DevWorkspaceStorageTypeAtrr = "controller.devfile.io/storage-type"
 
 	// WorkspaceEndpointNameAnnotation is the annotation key for storing an endpoint's name from the devfile representation
 	DevWorkspaceEndpointNameAnnotation = "controller.devfile.io/endpoint_name"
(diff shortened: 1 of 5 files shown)


pkg/provision/storage/asyncStorage.go
```diff
@@ -246,7 +246,8 @@ func (*AsyncStorageProvisioner) getAsyncWorkspaceCount(api provision.ClusterAPI)
 		return 0, 0, err
 	}
 	for _, workspace := range workspaces.Items {
-		if workspace.Labels[constants.DevWorkspaceStorageTypeLabel] == constants.AsyncStorageClassType {
+		storageClass := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeLabel, nil)
```
sleshchenko, line 249:
```suggestion
		storageClass := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAtrr, nil)
```
its not label anymore

tinakurian: sorry, changed it to be `DevWorkspaceStorageTypeAtrr`

pkg/constants/metadata.go
```diff
@@ -39,13 +39,13 @@ const (
...
 
-	// DevWorkspaceStorageTypeLabel defines the strategy used for provisioning storage for the workspace.
+	// DevWorkspaceStorageTypeAtrr defines the strategy used for provisioning storage for the workspace.
 	// If empty, the common PVC strategy is used.
 	// Supported options:
 	// - "common": Create one PVC per namespace, and store data for all workspaces in that namespace in that PVC
 	// - "async" : Create one PVC per namespace, and create a remote server that syncs data from workspaces to the PVC.
 	//             All volumeMounts used for devworkspaces are emptyDir
-	DevWorkspaceStorageTypeLabel = "controller.devfile.io/storage-type"
+	DevWorkspaceStorageTypeAtrr = "controller.devfile.io/storage-type"
```
sleshchenko, line 48: I wonder if storage impl is common enough that we can move it to devfile/api side, where each implementation may support or does not the configured implementation.
But I'm fine with moving it out of the current PR scope.

sleshchenko (approved):

pkg/constants/metadata.go
amisevsk: I think moving it would mean that if we want to add another storage type for our use-case, we'd have to first do it on the devfile/api side. I think I prefer it scoped to the operator (as it's a feature primarily for development) -- I think odo treats volumes differently than we do.

JPinkney (approved):

sleshchenko:
OpenShift CI is broken ATM
@tinakurian I can merge after you confirm locally e2e tests work fine

PR #452, opened 2021-06-11
feat: use ubi8 minimal image as go builder base image

mkuznyetsov:

### What does this PR do?
Use ubi 8 image as base image for building go in all Dockerfiles across the project 

https://github.com/eclipse/che/issues/19762

### Is it tested? How?

only with manual `docker build` command for all affected Dockerfiles


amisevsk (approved):

build/Dockerfile
```diff
@@ -9,8 +9,11 @@
 #   Red Hat, Inc. - initial API and implementation
 #
 
-FROM quay.io/bitnami/golang:1.14 as builder
-RUN go env GOPROXY
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
+FROM registry.access.redhat.com/ubi8-minimal:8.4 as builder
```
amisevsk, line 13: Why not use the same ubi image further down, rather than using `8.4` in the builder and `8.3` in the final image?

nickboldt: should be the same, so updated both to 8.4-200.1622548483

build/rhel.Dockerfile
```diff
@@ -9,10 +9,10 @@
...
 #
 
-# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/devtools/go-toolset-rhel7
-FROM registry.access.redhat.com/devtools/go-toolset-rhel7:1.14.12-4.1615820747  as builder
-ENV PATH=/opt/rh/go-toolset-1.14/root/usr/bin:${PATH} \
-    GOPATH=/go/
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
+FROM registry.access.redhat.com/ubi8-minimal:8.4 as builder
+RUN microdnf install -y make golang shadow-utils && \
+    go version
```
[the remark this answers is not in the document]
nickboldt: yes! please trash it upstream and we can simplify downstream too, so that upstream file Dockerfile is transformed into midstream (brew friendly) Dockerfile via https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-devworkspace-controller/build/scripts/sync.sh#L91-L92

sleshchenko (approved):

benoitf:
Could we use ubi8-micro or just scratch ?

build/Dockerfile
```diff
@@ -9,8 +9,11 @@
 #   Red Hat, Inc. - initial API and implementation
 #
 
-FROM quay.io/bitnami/golang:1.14 as builder
-RUN go env GOPROXY
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
+FROM registry.access.redhat.com/ubi8-minimal:8.4-200.1622548483 as builder
```
benoitf, line 13: we should use https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1 (use the image instead of installing it)

nickboldt:
Once merged, if we cherrypick this to 7.32.x / 0.6.x branch, we need to remember to update https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-devworkspace-controller/build/scripts/sync.sh#L91-L92 to use Dockerfile instead of rhel.Dockerfile.

nickboldt (approved):

nickboldt:
> Could we use ubi8-micro or just scratch ?

That's up to @amisevsk and the team to decide, but I'm good with it from the prod side. You can remove the `rpm -qa` line too since there's nothing being installed or updated in a micro image.

sleshchenko:
> Could we use ubi8-micro or just scratch ?

I don't have a strong opinion. The minimal working is OK for me.

nickboldt:
> I don't have a strong opinion. The minimal working is OK for me.

micro: 14M, with no package management at all (basically it's a scratch image + sleep binary to use with KIP)
minimal: 40M, including microdnf

I have no strong opinion either. :D

amisevsk:
No preference either. I lean towards slightly bigger images that might have bash included, since it's nice to be able to exec in. These things should be pulled less than weekly, so image size isn't a huge concern IMO.

benoitf:
There is Bash in micro
https://catalog.redhat.com/software/containers/ubi8-micro/601a84aadd19c7786c47c8ea?container-tabs=packages

amisevsk:
> micro: 14M, with no package management at all (basically it's a scratch image + sleep binary to use with KIP)

Does this mean running `microdnf  upgrade` is unnecessary, or does it mean we have to make sure the FROM image is continually updated?

benoitf:
there is no package manager in micro ( no yum, dnf or microdnf) so just need to update tags

nickboldt:
> there is no package manager in micro ( no yum, dnf or microdnf) so just need to update tags

correct, we will use https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/product/updateBaseImages.sh as per https://github.com/eclipse/che/issues/18831 (though that work got stalled a bit when we spent time exploring dependabot, which doesn't work for all our use cases).

nickboldt:
Merging this for now so we can move forward on this issue and on the downstream updates needed to adapt to it. 

If a future PR appears to move to micro, I'll approve that one at that time, but it's (arguably) beyond the scope of the original issue, https://github.com/eclipse/che/issues/19762

build/Dockerfile
```diff
@@ -9,15 +9,18 @@
 #   Red Hat, Inc. - initial API and implementation
 #
 
-FROM quay.io/bitnami/golang:1.14 as builder
-RUN go env GOPROXY
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8/go-toolset
+FROM registry.access.redhat.com/ubi8/go-toolset:1.15.7-11 as builder
+ENV GOPATH=/go/
```
nickboldt, line 14: do we need to mkdir /go so that golang can write here? 

Seeing this error in the PR check:

```
go: writing stat cache: mkdir /go: permission denied
```

nickboldt:
Looks like a problem happened in the test:

```
go: writing stat cache: mkdir /go: permission denied
```

https://github.com/devfile/devworkspace-operator/pull/452/files#diff-5aecd73bbcb9b7410938673269f0d4e676dd18509277b0ff2f2787c9d3e22604R14

nickboldt:
weird, must have been a previous PR check that failed, since this PR currently shows:

image

Issue #454, opened 2021-06-11
Investigate project-clone CPU and memory requirements


amisevsk:
## Description
PR https://github.com/devfile/devworkspace-operator/pull/453 increases the memory/CPU limits for the project clone image to avoid 

1. Slow clones due to limited CPU
2. OOMKilled due to limited memory.

In testing memory/CPU usage, I've noticed that the project-clone binary consumes a significant amount of memory on larger repos, e.g. for openshift/console:
```
	Command being timed: "/usr/local/bin/project-clone"
	User time (seconds): 21.38
	System time (seconds): 2.27
	Percent of CPU this job got: 96%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:24.47
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 686456
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 7418
	Voluntary context switches: 49597
	Involuntary context switches: 568
	Exit status: 0
```

Running just `git clone` locally requires roughly 1/5th the memory. 

Is this acceptable, or should project-clone be reworked to delegate to a commandline `git` binary? A max RSS of 680MB means there's a chance that even larger projects could fail to clone under a `1Gi` memory limit.

Testing memory usage/timing can be done directly using the image `quay.io/amisevsk/project-clone:debug`, which wraps the clone binary in `/usr/bin/time -v` (check the logs once the container is terminated).

Issue #416, opened 2021-05-04
Add support for zip-type projects to project-clone init container


amisevsk:
### Description
Currently, the project-clone init container only supports Git-type projects. This should be extended to also support "Zip"-type projects to fully support the devfile/api spec.

PR #460, opened 2021-06-16
feat: add support for zip projects in project-clone init container

amisevsk:
### What does this PR do?
* Add support for zip-type projects in devfiles. Zip files are downloaded and extracted as-is.
* Delegate git clone step to standard `git` binary rather than relying on the go-git library

Using git instead of go-git for cloning projects results in the project-clone image being ~180MB larger (due to requiring git). However, from some brief testing, it also reduces the memory requirement of project clone by around 5x for large projects.

Closes https://github.com/devfile/devworkspace-operator/issues/416
Closes https://github.com/devfile/devworkspace-operator/issues/454

### Is it tested? How?
Changes are built and pushed to `quay.io/amisevsk/project-clone:zip-support` and can be tested with the devfile below
```yaml
kind: DevWorkspace
metadata:
  name: projects-testing
spec:
  started: true
  template:
    projects:
      - name: dwo
        zip:
          location: https://github.com/devfile/devworkspace-operator/archive/refs/heads/0.6.x.zip
        clonePath: test-project-zip
      - name: dwo-2
        git:
          checkoutFrom:
            revision: 0.6.x
          remotes:
            origin: https://github.com/devfile/devworkspace-operator.git
        clonePath: "test-project-git"

    components:
      - name: tools
        container:
          image: quay.io/amisevsk/project-clone:debug
          command:
            - "tail"
            - "-f"
            - "/dev/null"
EOF
```

Alternatively, the project clone image can be tested directly:
```bash
TMPDIR=$(mktemp -d)
cat <<EOF > "$TMPDIR/dw.yaml"
projects:
  - name: dwo
    zip:
      location: https://github.com/devfile/devworkspace-operator/archive/refs/heads/0.6.x.zip
    clonePath: test-project-zip
  - name: dwo-2
    git:
      checkoutFrom:
        revision: 0.6.x
      remotes:
        origin: https://github.com/devfile/devworkspace-operator.git
    clonePath: "test-project-git"
EOF

podman run \
  -e DEVWORKSPACE_FLATTENED_DEVFILE=/tmp/dw.yaml \
  -e PROJECTS_ROOT=/projects \
  --mount type=bind,src=$TMPDIR/dw.yaml,dst=/tmp/dw.yaml,z \
  --mount type=tmpfs,tmpfs-size=500M,dst=/projects/ \
  quay.io/amisevsk/project-clone:zip-support -- /bin/bash -c "/usr/local/bin/project-clone; sleep 10d"
```
(once clone is completed, run `podman exec -it <the container> /bin/bash` and check `/projects` manually)

### DIFF
--- a/project-clone/internal/devfile.go
+++ b/project-clone/internal/devfile.go
@@ -24,6 +24,7 @@ import (
 	"github.com/devfile/devworkspace-operator/pkg/provision/metadata"
 )
 
+// GetClonePath gets the correct clonePath for a project, given the semantics in devfile/api
 func GetClonePath(project *dw.Project) string {
 	if project.ClonePath != "" {
 		return project.ClonePath
@@ -31,6 +32,8 @@ func GetClonePath(project *dw.Project) string {
 	return project.Name
 }
 
+// ReadFlattenedDevWorkspace reads the flattened DevWorkspaceTemplateSpec from disk. The location of the flattened
+// yaml is determined from the DevWorkspace Operator-provisioned environment variable.
 func ReadFlattenedDevWorkspace() (*dw.DevWorkspaceTemplateSpec, error) {
 	flattenedDevWorkspacePath := os.Getenv(metadata.FlattenedDevfileMountPathEnvVar)
 	if flattenedDevWorkspacePath == "" {
--- a/project-clone/internal/devfile.go
+++ b/project-clone/internal/devfile.go
@@ -42,7 +42,7 @@ func ReadFlattenedDevWorkspace() (*dw.DevWorkspaceTemplateSpec, error) {
 
 	fileBytes, err := ioutil.ReadFile(flattenedDevWorkspacePath)
 	if err != nil {
-		return nil, fmt.Errorf("error reading current DevWorkspace YAML: %s", err)
+		return nil, fmt.Errorf("error reading YAML file: %s", err)
 	}
 
 	dwts := &dw.DevWorkspaceTemplateSpec{}
--- a/project-clone/internal/devfile.go
+++ b/project-clone/internal/devfile.go
@@ -12,11 +12,41 @@
 
 package internal
 
-import dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"sigs.k8s.io/yaml"
+
+	"github.com/devfile/devworkspace-operator/pkg/provision/metadata"
+)
 
 func GetClonePath(project *dw.Project) string {
 	if project.ClonePath != "" {
 		return project.ClonePath
 	}
 	return project.Name
+}
+
+func ReadFlattenedDevWorkspace() (*dw.DevWorkspaceTemplateSpec, error) {
+	flattenedDevWorkspacePath := os.Getenv(metadata.FlattenedDevfileMountPathEnvVar)
+	if flattenedDevWorkspacePath == "" {
+		return nil, fmt.Errorf("required environment variable %s is unset", metadata.FlattenedDevfileMountPathEnvVar)
+	}
+
+	fileBytes, err := ioutil.ReadFile(flattenedDevWorkspacePath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading current DevWorkspace YAML: %s", err)
+	}
+
+	dwts := &dw.DevWorkspaceTemplateSpec{}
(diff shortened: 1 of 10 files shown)


benoitf:
hello, do you think we could use https://github.com/kubernetes/git-sync instead as the utility to clone projects ?

amisevsk:
git-sync may be overkill or targeting a different use-case. We basically only ever want to sync the git repo once -- otherwise we risk overwriting user changes on start.

JPinkney (approved):

project-clone/internal/zip/setup.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+
+	return nil
+}
+
+// closeSafe is a wrapper on io.Closer.Close() that just prints an error if encountered.
+//
+// Adapted from the Che plugin broker:
+// https://github.com/eclipse/che-plugin-broker/blob/27e7c6953c92633cbe7e8ce746a16ca10d240ea2/utils/ioutil.go#L326
```
sleshchenko, line 162: how these references from sources could help us in future? Maybe just drop them?

project-clone/internal/devfile.go
```diff
@@ -12,11 +12,44 @@
...
+	}
+
+	fileBytes, err := ioutil.ReadFile(flattenedDevWorkspacePath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading current DevWorkspace YAML: %s", err)
+	}
+
+	dwts := &dw.DevWorkspaceTemplateSpec{}
+	if err := yaml.Unmarshal(fileBytes, dwts); err != nil {
+		return nil, fmt.Errorf("error unmarshalling DevWorkspace YAML: %s", err)
```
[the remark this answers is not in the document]
amisevsk: I think using `DevWorkspaceSpecTemplate` and `Devfile` are more confusing here, though. This is a user-facing message at the end of the day -- they created a DevWorkspace, and `DevWorkspaceTemplateSpec` is a term only in the Go structs. Devfile could work here but it's just as inaccurate as `DevWorkspace`.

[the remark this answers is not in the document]
sleshchenko: you won

project-clone/internal/zip/setup.go
```diff
@@ -0,0 +1,168 @@
...
+
+	zipFilePath := path.Join(tmpDir, fmt.Sprintf("%s.zip", clonePath))
+	log.Printf("Downloading project archive from %s", url)
+	err := downloadZip(url, zipFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to download archive: %s", err)
+	}
+
+	log.Printf("Extracting project archive to %s", zipFilePath)
+	err = unzip(zipFilePath, projectPath)
```
sleshchenko, line 57: I wonder if we should not drop the top level folder.
Like
```
kind: DevWorkspace
metadata:
  name: projects-testing
spec:
  started: true
  template:
    projects:
      - name: dwo
        zip:
          location: https://github.com/devfile/devworkspace-operator/archive/refs/heads/main.zip
        clonePath: test-project-zip
      - name: dwo-2
        git:
          checkoutFrom:
            revision: 0.6.x
          remotes:
            origin: https://github.com/devfile/devworkspace-operator.git
        clonePath: "test-project-git"

    components:
      - name: tools
        container:
          image: quay.io/amisevsk/project-clone:debug
          command:
            - "tail"
            - "-f"
            - "/dev/null"
EOF
```
^ the project is the same but git will be in `test-project-git` when zip in `test-project-zip/devworkspace-operator`.
I'm not sure about the current Theia behavior
Screenshot_20210622_160606
Screenshot_20210622_161354

amisevsk: I'll work on it :+1:

I was mostly going for compatibility with how Theia currently does it, but there's no real reason for that :)

project-clone/internal/zip/setup.go
amisevsk: It feels strange to copy-paste code from another project without referencing its source, even if I helped write both functions :)

project-clone/main.go
```diff
@@ -13,85 +13,36 @@
...
-				os.Exit(1)
-			}
-			if err := internal.SetupRemotes(repo, &project); err != nil {
-				log.Printf("Failed to set up remotes for project %s: %s", project.Name, err)
-				os.Exit(1)
-			}
-		} else {
-			log.Printf("Project '%s' is already cloned and has all remotes configured", project.Name)
+		if err != nil {
+			log.Printf("Encountered error while setting up project %s: %s", project.Name, err)
```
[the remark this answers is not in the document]
amisevsk: :+1:

sleshchenko (approved):
tested and it works fine

PR #461, opened 2021-06-17
Update operator description in CSV

amisevsk:
### What does this PR do?
Updates the operator description and begins an uninstall doc.

### Is it tested? How?

N/A

### See also
https://github.com/redhat-developer/web-terminal-operator/pull/84

### DIFF
--- a/doc/uninstall.md
+++ b/doc/uninstall.md
@@ -1,0 +1,40 @@
+# Uninstalling the DevWorkspace Operator
+The DevWorkspace Operator makes use of finalizers on DevWorkspace resources and webhooks to restrict access to development resources. As these are not automatically removed when the Operator is uninstalled, they must manually be removed to fully clean the operator from the cluster.
+
+**Note:** The DevWorkspace Operator utilizes validating webhooks to restict `pods/exec` access to workspace resources. As Kubernetes provides no way to restrict which pods these webhooks apply to, improper removal of the webhook server could result in exec access to all pods on the cluster being blocked.
+
+1. Ensure that all DevWorkspace Custom Resources are removed along with their related k8s objects, like deployments.
+
+	```
+	kubectl delete devworkspaces.workspace.devfile.io --all-namespaces --all --wait
+	kubectl delete devworkspaceroutings.controller.devfile.io --all-namespaces --all --wait
+	```
+	Note: This step must be done first, as otherwise the resources above may have finalizers that block automatic cleanup.
+
+2. Uninstall the Operator
+
+3. Remove the custom resource definitions installed by the operator
+
+	```
+	kubectl delete customresourcedefinitions.apiextensions.k8s.io devworkspaceroutings.controller.devfile.io
+	kubectl delete customresourcedefinitions.apiextensions.k8s.io devworkspaces.workspace.devfile.io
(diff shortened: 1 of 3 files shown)


doc/uninstall.md
```diff
@@ -0,0 +1,44 @@
...
+	kubectl delete serviceaccounts devworkspace-webhook-server -n openshift-operators
+	kubectl delete configmap devworkspace-controller -n openshift-operators
+	kubectl delete clusterrole devworkspace-webhook-server
+	kubectl delete clusterrolebinding devworkspace-webhook-server
+	```
+
+6. Remove mutating/validating webhook configurations.
+
+	```
+	kubectl delete mutatingwebhookconfigurations controller.devfile.io
```
sleshchenko, line 42: maybe it's better to do it earlier to unblock pod/exec ASAP. It can be done as soon as CRs are removed and operator is uninstalled.

sleshchenko (approved):
LGTM
DO you plan to fix TODOs before merging?

doc/uninstall.md
```diff
@@ -0,0 +1,44 @@
+# Uninstalling the DevWorkspace Operator
+The DevWorkspace Operator makes use of finalizers on DevWorkspace resources and webhooks to restrict access to development resources. As these are not automatically removed when the Operator is uninstalled, they must manually be removed to fully clean the operator from the cluster.
+
+**Note:** The DevWorkspace Operator utilizes validating webhooks to restict `pods/exec` access to workspace resources. As Kubernetes provides no way to restrict which pods these webhooks apply to, improper removal of the webhook server could result in exec access to all pods on the cluster being blocked.
+
+1. Ensure that all DevWorkspace Custom Resources are removed along with their related k8s objects, like deployments.
+
+	```
+	kubectl delete devworkspaces.workspace.devfile.io --all-namespaces --all --wait
+	kubectl delete devworkspaceroutings.controller.devfile.io --all-namespaces --all --wait
```
sleshchenko, line 10: Seems devworkspace templates are missing. It's not a requirement to remove them since they don't have finalizers yet but still what is the point to leave them on the cluster?

JPinkney (approved):

doc/uninstall.md
amisevsk: Will do. I think the reason we chose the order here is because webhooks may be re-created if the webhooks server restarts.

doc/uninstall.md
amisevsk: Good catch! I forgot we excluded them from WTO so they wouldn't be included in my copy-paste :)

doc/uninstall.md
sleshchenko: that's true. So we should remove it after webhook server is removed.

amisevsk:
> Do you plan to fix TODOs before merging?

The remaining TODOs are for links to documentation and devfile samples. We need to fill them before release, but I don't think we have good options for this yet. Ideally linking to something like https://doc.crds.dev/github.com/devfile/api/workspace.devfile.io/DevWorkspace/v1alpha2@v2.0.0 would be great, but I don't know if we have a stable/official link to put here.

amisevsk:
Merging this PR to get it off the backlog, opened issue https://github.com/devfile/devworkspace-operator/issues/473 to track remaining work.

Issue #462, opened 2021-06-18
failure to start a devWorkspace with low memoryLimit


benoitf:
Using the following devWorkspace, workspace is not starting

```
NAMESPACE   NAME      DEVWORKSPACE ID             PHASE      INFO
default     failure   workspace622225a967194e5b   Starting   Waiting for workspace deployment
```

devWorkspace to use:
```yaml
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: failure
spec:
  started: true
  template:
    components:
      - name: tools
        container:
          image: quay.io/eclipse/che-quarkus:next
          memoryLimit: 1G
      - name: ubi-minimal
        container:
          image: registry.access.redhat.com/ubi8/ubi-minimal
          command: ['tail']
          args: ['-f', '/dev/null']
          memoryLimit: 32M
```

Error is shown in devworkspace-controller pod:
```
spec.template.spec.containers[1].resources.requests: Invalid value: \"64M\": must be less than or equal to memory 
```

```
{"level":"info","ts":1624016878.2157364,"logger":"controllers.DevWorkspace","msg":"Waiting on deployment to be ready","Request.Namespace":"default","Request.Name":"failure","devworkspace_id":"workspace622225a967194e5b"}
[...]

PR #464, opened 2021-06-18
fix: Verify that memory and CPU limits are greater than requests

amisevsk:
### What does this PR do?
* Fail workspace and set error message when memory/CPU limit is less than request
* Set memory request to match limit if default request is used and limit is less (e.g. user sets `32Mi` as a limit and no request, use `32Mi` as the request instead of `64Mi`).

Fixes https://github.com/devfile/devworkspace-operator/issues/462

### Is it tested? How?
Added test cases.


benoitf:
I think it would be interesting to report the error to the user as well if there is a deployment issue ( today it's mangled)

JPinkney (approved):

sleshchenko:
> I think it would be interesting to report the error to the user as well if there is a deployment issue ( today it's mangled)

DevWorkspace Operator should report deployment issues https://github.com/devfile/devworkspace-operator/issues/465

pkg/library/container/testdata/fixes-memory-limit-less-than-default-request.yaml
```diff
@@ -0,0 +1,24 @@
+name: "Gives useful error when memory limit is less than default request"
```
sleshchenko, line 1: Sorry for being late to the party, and that's fully OK you merged without my approval.

It seems to be copy/paste mistake. Is fixed in https://github.com/devfile/devworkspace-operator/pull/467

amisevsk: Good catch, I should've waited :smile: 

https://github.com/devfile/devworkspace-operator/pull/469

Issue #403, opened 2021-04-28
Should Kubernetes references in plugin/parent resolve across the entire cluster?


amisevsk:
### Description

Currently, when using a name/namespace pair to reference a DWT in a plugin or parent, the DevWorkspace controller will resolve that reference regardless of the namespace. This means that one could read DevWorkspaceTemplates in another namespace, even if they do not have permissions to read that namespace, provided they know the name/namespace pair.

Should we restrict plugin/parent references to a single namespace? If we do decide to restrict in this way, what's the point of having a namespace field at all?

The current approach allows for easier management of parents/plugins (you can just keep them all in one namespace). Removing this capability would mean that every namespace would need any parents/plugin DWTs the DevWorkspace intends to use. On the other hand, allowing DWTs to come from any namespace is a (minor) privilege escalation.

amisevsk:
Note: restricting `namespace` for plugins would mean the proposed solution for having WTO depend on DWO would no longer work (since WTO would be deploying DWTs to a specific namespace for use by all Web Terminals)

sleshchenko:
> Note: restricting namespace for plugins would mean the proposed solution for having WTO depend on DWO would no longer work (since WTO would be deploying DWTs to a specific namespace for use by all Web Terminals)

It can be configured. With OpenShift CSVs will get allowed the current namespace + openshift-operators.

sleshchenko:
As we agreed on DevWorkspace Cabal, it's not good to make DWT read-only available for all by default, and actually does not have permissions to create private one.
At first step we could allow openshift-operators by default to expose WTO's DWT.
Later we can introduce:
1. CRD that could define DWT in namespace as exposed.
2. Simpler - annotations on DWT with rules in which namespaces it should be available. By default - only the current one,

PR #466, opened 2021-06-22
Restrict reading DWTs in other namespaces

amisevsk:
### What does this PR do?
Removes the ability for DevWorkspaces to read dwts across the cluster, defaulting to only being able to read dwts from the current namespace. Adds an annotation (on dwts) to allow them to be readable in either all namespaces or a specific set of namespaces.

Closes https://github.com/devfile/devworkspace-operator/issues/403

### Is it tested? How?
Test cases added. 

### PR Checklist

### DIFF
--- a/pkg/library/flatten/flatten.go
+++ b/pkg/library/flatten/flatten.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"net/url"
 	"path"
+	"strings"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/api/v2/pkg/utils/overriding"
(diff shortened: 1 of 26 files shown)


pkg/library/flatten/flatten.go
```diff
@@ -272,3 +287,30 @@ func resolveElementByURI(
 	}
 	return dwt, nil
 }
+
+// canImportDW returns true if a DevWorkspace in dwNamespace is allowed to reference the provided DevWorkspaceTemplate
+// DevWorkspaces can by default only read DevWorkspaceTemplates in their own namespace, unless the DevWorkspaceTemplate
+// has the
```
JPinkney, line 293: Looks like the comment wasn't finished :smile:

pkg/library/flatten/flatten.go
```diff
@@ -29,12 +30,21 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// DWTSupportedNamespacesAnnotation defines additional namespaces from which a DevWorkspace can import a DevWorkspaceTemplate.
+	// Options are:
+	// - '*': allow importing by all DevWorkspaces on the cluster
+	// - 'namespaceA,namespaceB,namespaceC': Allow importing by DevWorkspaces in list of specific namespaces
```
sleshchenko, line 37: I don't think that anybody care about such a case, but for clarity, even when you specify `namespaceA`, the current namespace should still work as well.

pkg/library/flatten/flatten.go
```diff
@@ -29,12 +30,21 @@ import (
...
+	// If the annotation does not exist or is empty, only DevWorkspaces in the same namespace as the template can reference it.
+	DWTSupportedNamespacesAnnotation = "controller.devfile.io/supported-namespaces"
```
sleshchenko, line 39: suported-namespace is pretty common and it's not clear which level of support is meant to be here.
Maybe `controller.devfile.io/allow-from-namespaces`, `controller.devfile.io/accessible-from-namespaces` would be a bit clear but I'm not sure, I don't like them much as well.

> DWTSupportedNamespacesAnnotation defines additional namespaces from which a DevWorkspace can import a DevWorkspaceTemplate.

`controller.devfile.io/allow-import-from` or `controller.devfile.io/allow-references-from`?

sleshchenko: @amisevsk @JPinkney I just flushed with all I managed to come to, feel free to share your opinion or propose your variants. I'm actually OK with any )

sleshchenko (commented):
in general LGTM, I'll take a deeper look tomorrow and will test it

pkg/library/flatten/flatten.go
amisevsk: Woopsie :D

pkg/library/flatten/flatten.go
amisevsk: I went with `controller.devfile.io/allow-import-from`

pkg/library/flatten/flatten.go
amisevsk: Clarified. Alternatively, we could change the annotation to `allow-all-namespaces: ['true'|'false']` if that's simpler.

sleshchenko (commented):
changes LGTM

JPinkney (approved):

sleshchenko (approved):
Tested and it works as expected

PR #467, opened 2021-06-24
fix: fixes test cases title & restructure files

sleshchenko:
### What does this PR do?
this PR restructure test cases files in the way that it starts with component it tests, like
- resource-... (invalid-memory-limit/cpu-limit-less-than-request/...)
- volume-... (mounts-project-directly/use-mount-sources)
it helps easier to catch which cases are covered for a particular component.

Also, it fixes a few copied/pasted test-cases titles.

### PR Checklist


JPinkney:
instead of doing something like `pkg/library/container/testdata/components-ignores-non-container-components.yaml` does it make sense to put specific test data under their own folder like: `pkg/library/container/testdata/components/ignores-non-container-components.yaml`

amisevsk (approved):

JPinkney (approved):

PR #472, opened 2021-06-30
Don't remove openshift-operators namespace on make uninstall

amisevsk:
### What does this PR do?
Make it so that `make uninstall` doesn't remove `openshift-operators` accidentally (a problem I keep running into when trying to uninstall the operator :wink:)

### PR Checklist

### DIFF
--- a/build/make/deploy.mk
+++ b/build/make/deploy.mk
@@ -76,7 +76,9 @@ endif
 	$(K8S_CLI) delete all -l "app.kubernetes.io/part-of=devworkspace-operator" --all-namespaces
 	$(K8S_CLI) delete mutatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io --ignore-not-found
 	$(K8S_CLI) delete validatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io --ignore-not-found
+ifneq ($(NAMESPACE),openshift-operators)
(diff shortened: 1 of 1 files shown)


JPinkney (approved):

build/make/deploy.mk
```diff
@@ -76,7 +76,9 @@ endif
 	$(K8S_CLI) delete all -l "app.kubernetes.io/part-of=devworkspace-operator" --all-namespaces
 	$(K8S_CLI) delete mutatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io --ignore-not-found
 	$(K8S_CLI) delete validatingwebhookconfigurations.admissionregistration.k8s.io controller.devfile.io --ignore-not-found
+ifneq ($(NAMESPACE),openshift-operators)
```
sleshchenko, line 79: I wonder if openshift-operators should become our default.

sleshchenko (approved):

build/make/deploy.mk
amisevsk: I think openshift-operators should be reserved for the OLM case. The operator doesn't actually care where it's deployed, and it's a lot easier to reason about if everything is in a dedicated namespace. 

We can potentially look to `cert-manager` as an example. If you install via applying yaml files (as we do on k8s) it gets installed to the `cert-manager` namespace, but if you install via OLM the default is `openshift-operators` since it's an all-namespaces operator.

Issue #428, opened 2021-05-18
DWO version is not bumped after release


sleshchenko:
DWO version is not bumped after release.
We still have [v0.3.0](https://github.com/devfile/devworkspace-operator/blob/main/version/version.go#L17) when [v0.5.0](https://github.com/devfile/devworkspace-operator/tags) is expected

sleshchenko:
According to https://github.com/devfile/devworkspace-operator/blob/main/version/version.go it's fixed

Issue #415, opened 2021-05-04
use tag image in Makefile when tagging the release


benoitf:
If I checkout `v0.2.3` [[1]](https://github.com/devfile/devworkspace-operator/blob/v0.2.3/Makefile#L21) or `v0.3.2` [[2]](https://github.com/devfile/devworkspace-operator/blob/v0.3.2/Makefile#L21) tag of devworkspace operator, the Makefile still reference `:next` tag so `make install` command doesn't work out of the box

I think release process [[3]](https://github.com/devfile/devworkspace-operator/commit/af1b6bd88986cce2a7983e069070d4570994a378) should also update Makefile when cutting a release

sleshchenko:
it should be fixed already, at least 0.7.2 seems to have the right values https://github.com/devfile/devworkspace-operator/blob/v0.7.2/Makefile

PR #480, opened 2021-07-08
fix: Add several fixes to the release process

mkuznyetsov:

### What does this PR do?
- Fix bumping version in bugfix/main branches
- Change image tags in Makefile to released ones when tagging the release

https://github.com/devfile/devworkspace-operator/issues/428
https://github.com/devfile/devworkspace-operator/issues/415

### PR Checklist

### DIFF
--- a/make-release.sh
+++ b/make-release.sh
@@ -33,7 +33,7 @@ bump_version () {
 
   echo "Updating project version to ${NEXT_VERSION}"
   # change version/version.go file
-  sed -i version/version.go -r -e 's#(Version = ")(v?[0-9.]+)(")#\1'"${VERSION}"'\3#g'
+  sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+)(\")#\1${NEXT_VERSION}\3#g"
   if [[ ! -z $(git status -s) ]]; then # dirty
     git add version/version.go
     COMMIT_MSG="[release] Bump to ${NEXT_VERSION} in ${BUMP_BRANCH}"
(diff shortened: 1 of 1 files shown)


make-release.sh
```diff
@@ -109,8 +109,10 @@ fi
 set -e
 
 # change version/version.go file
-sed -i version/version.go -r -e 's#(Version = ")(v?[0-9.]+)(")#\1'"${VERSION}"'\3#g'
-
+sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+)(\")#\1${VERSION}\3#g"
+# change image tag in Makefile
+sed -i Makefile -r -e "s#export DWO_IMG \?= quay.io/devfile/devworkspace-controller:.*#export DWO_IMG \?= quay.io/devfile/devworkspace-controller:${VERSION}#g"
```
sleshchenko, line 114: that works fine but I would say it's double work. Instead maybe move it below and do
```suggestion
sed -i Makefile -r -e "s#export DWO_IMG \?= .*#export DWO_IMG \?= ${DWO_QUAY_IMG}#g"
```
and the same for PROJECT_CLONE

sleshchenko (commented):
LGTM and seems to work as far as I tested but I have one suggestion left

make-release.sh
mkuznyetsov: ok, thanks, I made an even more generic sed, to just match the image name anywhere in Makefile

sleshchenko (approved):
Seems to work as expected after I executed changed commands

PR #486, opened 2021-07-09
Fix run/debug locally

amisevsk:
## What does this PR do?
PR https://github.com/devfile/devworkspace-operator/pull/474 moved the conversion webhooks to the controller deployment, which requires the controller deployment to read mounted secrets to get the correct TLS cert.

For `make run` and `make debug` to work locally, we need to grab this secret and store its keys as files on the local disk as if it was a mounted secret (stored in `/tmp/k8s-webhook-server/serving-certs/`).

Note that since conversion webhooks are now in the main deployment, conversion won't work when running locally.

Make run/debug broken

### Is it tested? How?
`make run`

`make debug`

### PR Checklist


sleshchenko (approved):
I haven't tested but LGTM

JPinkney (approved):

Issue #477, opened 2021-07-02
Add an ability to configure common PVC size on namespace level


sleshchenko:
Add an ability to configure common PVC size on namespace level.
Like on devsandbox, the default Che used value is 15Gi, but devworkspace init PVC with hard-coded [1 Gi](https://github.com/devfile/devworkspace-operator/blob/9e0ffdda400f256c575127b6bda5e3e3e7557645/pkg/constants/constants.go#L36).
We need a way to define PVC size on downstream side (WTO, Che, CRW), I assume we can do it with configmap in the devworkspace namespace.

PR #487, opened 2021-07-10
Add support for per-namespace configuration and add setting for workspace PVC request size

amisevsk:
### What does this PR do?
* Add mechanism for defining namespace-level configuration for DevWorkspaces via configmap
    * A configmap is marked as relevant with the label `controller.devfile.io/namespaced-config`
* Add capability to define PVC size via configmap above.

TODOs: (either here or separate issues)
* Workspace PVCs are not updated once they are created, so the setting above only applies when the claim-devworkspace PVC does not exist. https://github.com/devfile/devworkspace-operator/issues/488
    * It's not straightforward to enable this, as *only* the PVC storage `request` can be changed after creation. The SyncObjects method won't work if it uses `Update()` and switching to `Patch()` fails for reasons I don't fully understand yet (likely due to dealing with runtime.Object)
* The configuration setting currently applies to the PVC's storage `request` and no limit is supplied. Should we allow the configmap to define both request and limit?

Closes https://github.com/devfile/devworkspace-operator/issues/477

### Is it tested? How?
Create a configmap to set defaults for a specific namespace:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-nsconfig
  labels:
    controller.devfile.io/namespaced-config: "true"
data:
  commonPVCSize: 5Gi
EOF
```
and then create a DevWorkspace in that namespace.

If the namespace already has a claim-devworkspace, it will not be updated and so should be deleted before testing.

### PR Checklist

### DIFF
--- a/pkg/provision/config/config.go
+++ b/pkg/provision/config/config.go
@@ -1,0 +1,71 @@
+//
+// Copyright (c) 2019-2021 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/devfile/devworkspace-operator/controllers/workspace/provision"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
+)
+
+const (
+	commonPVCSizeKey = "commonPVCSize"
+)
+
+type NamespacedConfig struct {
+	CommonPVCSize string
+}
+
+// ReadNamespacedConfig reads the per-namespace DevWorkspace configmap and returns it as a struct. If there are
+// no valid configmaps in the specified namespace, returns (nil, nil). If there are multiple configmaps with the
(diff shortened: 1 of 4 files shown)


sleshchenko (commented):
Changes LGTM
will test soon

pkg/provision/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/devfile/devworkspace-operator/controllers/workspace/provision"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
+)
+
+const (
+	commonPVCSizeKey = "commonPVCSize"
```
sleshchenko, line 28: well, during testing I found more config we may want to configure, but I'm not sure.
Like pvc name. For CRW it's [claim-che-workspace](https://github.com/eclipse-che/che-server/blob/main/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L385), not sure if CRW expects Che Workspace and DevWorkspaces uses different PVCs.

Also, see che.infra.kubernetes.pvc.storage_class_name.

it's out of the current PR scope for sure, but WDYT?

sleshchenko (approved):
Tested and it works fine

pkg/provision/config/config.go
amisevsk: My idea was that the config struct could be extended to include additional information as needed (storage class makes sense). 

For common PVC name, I think it's a limited use case that would be complex to support -- e.g. what happens if it is updated after `claim-che-workspace` has been created? It seems like it would be an open door for leaking storage after workspaces are deleted, etc.

PR #500, opened 2021-07-21
Add DevWorkspace-specific metrics

amisevsk:
### What does this PR do?
Adds DevWorkspace-specific metrics to the metrics server, tracking
1. Number of workspace starts
2. Number of successfully started workspaces (i.e. workspaces entering 'Running' phase)
3. Number of devworkspaces that enter the Failed state
4. A histogram of workspace startup time, bucketed in 10-second intervals

Currently, the metrics come with a `source` label. The value of this label is derived from the value of the `controller.devfile.io/devworkspace-source` label on the DevWorkspace (optional). If set, this allows metrics to be partitioned based on the tool creating the DevWorkspace (e.g. Web Terminal Operator, Eclipse Che, etc.)

Metrics are updated on phase changes in the DevWorkspace (i.e. going from phase 'Started' -> 'Running') to avoid double counting DevWorkspaces.

Additional nice-to-haves for https://github.com/devfile/devworkspace-operator/issues/241

### Is it tested? How?
Changes can be tested in same way as https://github.com/devfile/devworkspace-operator/pull/405 (note metrics won't appear until they're updated at least once)

A simpler way to test these changes directly is by running the controller locally:
1. `make run` the controller in one window
2. `kubectl apply -f samples/theia-next.yaml`
3. curl -s http://localhost:8080/metrics | grep '^dev'

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"github.com/devfile/devworkspace-operator/controllers/workspace/metrics"
 	"github.com/devfile/devworkspace-operator/controllers/workspace/provision"
 	"github.com/devfile/devworkspace-operator/pkg/common"
 	"github.com/devfile/devworkspace-operator/pkg/config"
@@ -120,6 +121,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		workspace.Status.Phase = dw.DevWorkspaceStatusStarting
 		workspace.Status.Message = "Initializing DevWorkspace"
 		err = r.Status().Update(ctx, workspace)
+		metrics.WorkspaceStarted(workspace, reqLogger)
 		return reconcile.Result{Requeue: true}, err
 	}
 
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -120,18 +120,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 			return reconcile.Result{}, err
 		}
 		workspace.Status.DevWorkspaceId = workspaceId
-		workspace.Status.Phase = dw.DevWorkspaceStatusStarting
-		workspace.Status.Message = "Initializing DevWorkspace"
-		workspace.Status.Conditions = []dw.DevWorkspaceCondition{
-			{
-				Type:               conditions.Started,
-				Status:             corev1.ConditionTrue,
-				LastTransitionTime: metav1.Time{Time: clock.Now()},
-				Message:            "DevWorkspace is starting",
-			},
-		}
 		err = r.Status().Update(ctx, workspace)
-		metrics.WorkspaceStarted(workspace, reqLogger)
 		return reconcile.Result{Requeue: true}, err
 	}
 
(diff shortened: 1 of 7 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -120,6 +121,7 @@ func (r *DevWorkspaceReconciler) Reconcile(req ctrl.Request) (reconcileResult ct
 		workspace.Status.Phase = dw.DevWorkspaceStatusStarting
 		workspace.Status.Message = "Initializing DevWorkspace"
 		err = r.Status().Update(ctx, workspace)
+		metrics.WorkspaceStarted(workspace, reqLogger)
```
sleshchenko, line 134: Should not we check error before incrementing Started?

sleshchenko: Won't it increment Started when creating devworkspace with Started: false.

I'm afraid I don't understand why we even set Starting phase here, I would say we just should update status with DevWorkspace, and STOPPED.

controllers/workspace/status.go
```diff
@@ -62,6 +63,7 @@ var healthHttpClient = &http.Client{
 // Parameters for result and error are returned unmodified, unless error is nil and another error is encountered while
 // updating the status.
 func (r *DevWorkspaceReconciler) updateWorkspaceStatus(workspace *dw.DevWorkspace, logger logr.Logger, status *currentStatus, reconcileResult reconcile.Result, reconcileError error) (reconcile.Result, error) {
+	updateMetricsForPhase(workspace, status.phase, logger)
```
sleshchenko, line 67: I assume metrics will be more precise if we move after check that status update error is nil

controllers/workspace/metrics/update.go
```diff
@@ -34,7 +44,27 @@ func incrementMetricForWorkspace(metric *prometheus.CounterVec, wksp *dw.DevWork
...
 	if err != nil {
-		log.Info("Failed to increment metric: %s", err)
+		log.Error(err, "Failed to increment metric")
 	}
 	ctr.Inc()
 }
+
+func incrementStartTimeBucketForWorkspace(wksp *dw.DevWorkspace, log logr.Logger) {
+	sourceLabel := wksp.Labels[workspaceSourceLabel]
+	hist, err := workspaceStartupTimesHist.GetMetricWith(map[string]string{metricSourceLabel: sourceLabel})
```
sleshchenko, line 54: So, we would need to create PR against OpenShift Console, and probably backport it to OS 4.8, so also update existing CR is Console if label is not detected.

sleshchenko (approved):

controllers/workspace/devworkspace_controller.go
amisevsk: You're right, I don't know what I was thinking doing it this way. I'll fix this before merging, but I got side-tracked with debugging https://github.com/devfile/devworkspace-operator/pull/527 today.

controllers/workspace/status.go
amisevsk: Move to only update metrics if the corresponding status update succeeds.

controllers/workspace/devworkspace_controller.go
amisevsk: Should be fixed, though there's still a bit of strangeness in having to set the Started condition ASAP and then also having to do
```
	reconcileStatus := currentStatus{phase: dw.DevWorkspaceStatusStarting}
	reconcileStatus.setConditionTrue(conditions.Started, "DevWorkspace is starting")
```
to ensure the condition isn't wiped out later.

sleshchenko (approved):

controllers/workspace/metrics/update.go
amisevsk: Done (finally): https://github.com/openshift/console/pull/9752

Issue #508, opened 2021-07-23
Add support for devfile 2.0 variables


amisevsk:
### Description

The devfile 2.0 spec includes global variables, which define key-value pairs, e.g.
```
variables:
  my-var: "some value"
```
These variables can be referenced in non-key string fields within the devfile using `{{my-var}}` to template different versions of a devfile by setting only the variable field.

We should add support for this in DWO.

See issue https://github.com/devfile/api/issues/239 and PR https://github.com/devfile/api/pull/352 for details.

sleshchenko:
is implemented with https://github.com/devfile/devworkspace-operator/pull/509

sleshchenko:
duplicates https://github.com/devfile/devworkspace-operator/issues/408

PR #509, opened 2021-07-23
Add support for devfile 2.0 global variables

amisevsk:
### What does this PR do?
Add support for [devfile 2.0 variables](https://docs.devfile.io/devfile/2.1.0/user-guide/api-reference.html). In most string fields in a DevWorkspace, DWO will attempt to replace the string `{{ some-variable }}` with the corresponding variable from `.spec.template.variables`.

If a replacement cannot be found, a new condition `DevWorkspaceWarning` is set on the DevWorkspace with a message like
> `Error processing variable replacements: invalid variables in component test-container: 'test-variable'"`

and the info message will be updated to include some warning text, e.g.
```
NAME         DEVWORKSPACE ID             PHASE      INFO
theia-next   workspace89e5bf5e04b34b6b   Starting   [warnings present] Waiting for workspace deployment
```

https://github.com/devfile/devworkspace-operator/issues/508

### Is it tested? How?
Apply the DevWorkspace below:
```yaml
kind: DevWorkspace
metadata:
  name: theia-next
spec:
  started: true
  template:
    attributes:
      controller.devfile.io/storage-type: "ephemeral"
    variables:
      test-variable: "testing_value"
    components:
      - name: test-container
        container:
          image: quay.io/libpod/busybox:latest
          command: ["tail"]
          args: ["-f", "/dev/null"]
          env:
          - name: TEST_ENV
            value: "{{test-variable}}"
      - name: theia
        plugin:
EOF
```
The value in the env var should be replaced once the pod is created.

To test warnings, delete the variable from the devworkspace

### PR Checklist

### DIFF
--- a/pkg/library/flatten/flatten.go
+++ b/pkg/library/flatten/flatten.go
@@ -21,6 +21,7 @@ import (
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/api/v2/pkg/utils/overriding"
+	"github.com/devfile/api/v2/pkg/validation/variables"
 	"github.com/devfile/devworkspace-operator/pkg/library/annotate"
 	registry "github.com/devfile/devworkspace-operator/pkg/library/flatten/internal_registry"
 	"github.com/devfile/devworkspace-operator/pkg/library/flatten/network"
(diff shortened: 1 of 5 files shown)


JPinkney (approved):
Just tested and it LGTM

pkg/library/flatten/flatten.go
```diff
@@ -315,3 +322,20 @@ func canImportDWT(dwNamespace string, dwt *dw.DevWorkspaceTemplate) bool {
...
+	for commandName, warnings := range warn.Commands {
+		msg = append(msg, fmt.Sprintf("invalid variables in component %s: '%s'", commandName, strings.Join(warnings, "', '")))
+	}
+	for projectName, warnings := range warn.Projects {
+		msg = append(msg, fmt.Sprintf("invalid variables in project %s: '%s'", projectName, strings.Join(warnings, "', '")))
+	}
+	for starterProjectName, warnings := range warn.StarterProjects {
+		msg = append(msg, fmt.Sprintf("invalid variables in starter project %s: '%s'", starterProjectName, strings.Join(warnings, "', '")))
+	}
+	return fmt.Sprintf("Error processing variable replacements: %s", strings.Join(msg, "; "))
```
sleshchenko, line 340: when I tested I noticed that probably there are even more quotas than we would like to have
```yaml
  conditions:
  - lastTransitionTime: "2021-07-28T12:57:33Z"
    message: 'Error processing variable replacements: invalid variables in component
      vsix-installer: ''test-variable''; invalid variables in component tools: ''test-variable'''
```
Ah, I see. It's ' is used for multilines, and `''` is escaped `'`

sleshchenko (approved):
I just tested and works as expected. Please create an issue for dashboard to take into account DevWorkspaceWarning.

pkg/library/flatten/flatten.go
amisevsk: The joys of YAML :D 

For a while I wanted to use `%q` instead of `%s` (so that it would `print "var"` instead of `print var`), but that ends up requiring escapes all the time.

amisevsk:
Created dashboard issue: https://github.com/eclipse/che/issues/20220

PR #510, opened 2021-07-23
Fix make run/debug locally in k8s

amisevsk:
### What does this PR do?
On Kubernetes, the makefile was referencing the wrong secret for the
webhooks server, which resulted in it failing to start locally.

### Is it tested? How?
`make run` or `make debug` on Kubernetes

### PR Checklist

### DIFF
--- a/build/make/deploy.mk
+++ b/build/make/deploy.mk
@@ -43,8 +43,13 @@ _gen_configuration_env:
 
 _store_tls_cert:
 	mkdir -p /tmp/k8s-webhook-server/serving-certs/
+ifeq ($(PLATFORM),kubernetes)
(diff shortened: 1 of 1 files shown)


JPinkney (approved):

sleshchenko (approved):
Would not it easier to align secrets names on k8s and OS?

amisevsk:
> Would not it easier to align secrets names on k8s and OS?

I don't think so (but could be convinced) -- on k8s, we use one cert to serve both the controller and webhook server, and on OpenShift we use separate ones.

k8s: 
- `devworkspace-operator-webhook-cert`

OpenShift:
- `devworkspace-webhooks-tls` (controller)
- `devworkspace-webhookserver-tls` (webhooks server)

Issue #357, opened 2021-03-26
DevWorkspace MVP (minimal viable product)


l0rd:
We would like to submit DevWorkspace as CNCF incubator project. Here we want to track tasks to accomplish before pushing the submit button:

**Mandatory**

- [x] [Projects clone should happen before other components are started](https://github.com/devfile/devworkspace-operator/issues/205)
- [x] Replace (or add) Apache-2.0 license
- [ ] Register a domain for the project landing page
- [ ] Setup the website content at https://github.com/devworkspace/devworkspace.github.io
- [x] Review and fix devfile v2 documentation
- [x] Remove Web Terminal Operator-specific code from DevWorkspace Operator and develop a plan to continue supporting WTO.

**Nice to have**

- [x] Publish DWO on the Operator Hub
- [x] Release devfile spec v2.1
- [ ] Record a demo of `<kubectl apply -f dw.yaml>` | `<kubectl exec>`  `<code (with vim)>` | `<build>` | `<run>`

amisevsk:
Added a note about web-terminal-specific functionality that should probably be removed before going forward.

amisevsk:
Closing as all CNCF-related stuff is done as far as I know (devfile/api is in CNCF, DWO will not be)

l0rd:
For the records we decided to move forward with the Devfile rather than the DevWorskspace Operator to be a CNCF project. That's because there was a cross-team, cross-company interest in the Devfile but not in the DevWorkspace. At the time I created the [DevWorkspaec GitHub org](https://github.com/devworkspace) to host a DevWorkspace website but then we abandoned that plan. Now the DevWorkspace Operator is one subproject of the CNCF sandbox project Devfile.

PR #512, opened 2021-07-23
chore: Remove dependabot for dockerfiles

mkuznyetsov:

### What does this PR do?
Remove Dependabot for dockerfiles, which will be replaced by `update-base-images.yml` workflow in eclipse-che/che-release project (to manage all updates from one place)
https://github.com/eclipse/che/issues/18831

### PR Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,0 @@
-version: 2
-updates:
-  # Enable version updates for Docker
-  - package-ecosystem: "docker"
-    # Look for a `Dockerfile` in the `build` directory
-    directory: "/build"
-    # Check for updates once a week
-    schedule:
-      interval: "daily"


nickboldt (approved):
+1, replaced by https://github.com/eclipse-che/che-release/blob/main/.github/workflows/update-base-images.yml

amisevsk (commented):
Not sure how this interacts with https://github.com/devfile/devworkspace-operator/issues/357, etc. It's a little strange to have a part of the devfile org infrastructure housed in a different project from a different organization.

nickboldt:
> strange to have a part of the devfile org infrastructure housed in a different project from a different organization

Ah, but eclipse-che org projects has dependencies on che-incubator projects, and redhat-developer (CRW) has dependencies on both. Why does the repo location matter? Would you rather have a copy of the CRW updateBaseImages.sh script in this repo, which can be its own lifecycle/fork ?

nickboldt:
Can we merge this? Replacement script is already running -- see https://github.com/devfile/devworkspace-operator/commit/8fb82395e25b9fad3205d028fbd68a8252fbe1d6#diff-5aecd73bbcb9b7410938673269f0d4e676dd18509277b0ff2f2787c9d3e22604

sleshchenko:
> Can we merge this? The replacement script is already running

@nickboldt could you point who invoke that script?
I think since DWO has a separate release cadence, not Che or CRW strongly connected, DWO should perform it on its side.... And what is the drawback from relying on dependabot?

nickboldt:
1. It's run via this script once a week https://github.com/eclipse-che/che-release/actions/workflows/update-base-images.yml 
https://github.com/eclipse-che/che-release/blob/main/.github/workflows/update-base-images.yml#L29

Once Che and CRW are simply CONSUMING DWO, rather than releasing / rebuilding it, then yes, you could run this weekly check from another entrypoint. 

But honestly, I'm not sure I see the problem w/ a bot process checking once a week to see if a new release of UBI8 images are available, and sending you a PR if a change it proposed. That's exactly how dependabot works (when it works) and it's also running from a 3rd party code base outside your organization.

2. Dependabot doesn't work for UBI base images - see proof here https://github.com/eclipse/che/issues/18831#issuecomment-839761482 We tried to use Dependabot, but switched to my script after proving it didn't work as expected.

sleshchenko:
> Once Che and CRW are simply CONSUMING DWO, rather than releasing / rebuilding it, then yes, you could run this weekly check from another entrypoint.

it should be the next step I believe, but we can proceed with merging this PR as well.

sleshchenko (approved):

nickboldt:
Removal of dependabot shouldn't require e2e tests... but hey. If there's no manual override, I guess we wait for pointless rebuilds to tell you that things in the build are in fact unchanged by this removal. :)

nickboldt:
Just for reference/completeness... if base image update process is to be done separate from the weekly Che release action, we'll want to remove this line:

https://github.com/eclipse-che/che-release/blob/main/.github/workflows/update-base-images.yml#L29

... and to move that functionality into another repo/org/whatever.

PR #527, opened 2021-07-28
Avoid unnecessary condition updates when status is 'Unknown'

amisevsk:
### What does this PR do?
Only update conditions with `Unknown` status when the existing condition is not `Unknown`. Basically, this avoids updating the last transition time on every reconcile loop if there are persistently 'unknown' conditions.

This is a subtle issue since a sufficiently quick re-reconcile will result in no changes to the DevWorkspace (we update last transition time but it's within the accuracy of the timer), resulting in no additional reconciles. However, if the workqueue duration grows significantly, this can result in continually requeuing reconciles where all we're changing is `LastTransitionTime`.

### Is it tested? How?
The best way I've found to reproduce the issue:
1. Create 1000 stopped DevWorkspaces
2. Quickly update all of them to flood the reconcile queue

With these changes, the controller should work through the queue without enqueuing additional requests.

### PR Checklist

### DIFF
--- a/controllers/workspace/status.go
+++ b/controllers/workspace/status.go
@@ -92,9 +92,11 @@ func syncConditions(workspaceStatus *dw.DevWorkspaceStatus, currentStatus *curre
 		currCondition, ok := currentStatus.conditions[workspaceCondition.Type]
 		if !ok {
 			// Didn't observe this condition this time; set status to unknown
-			workspaceStatus.Conditions[idx].LastTransitionTime = currTransitionTime
-			workspaceStatus.Conditions[idx].Status = corev1.ConditionUnknown
-			workspaceStatus.Conditions[idx].Message = ""
+			if workspaceCondition.Status != corev1.ConditionUnknown {
+				workspaceStatus.Conditions[idx].LastTransitionTime = currTransitionTime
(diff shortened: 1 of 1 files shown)


JPinkney (approved):

controllers/workspace/status.go
```diff
@@ -92,9 +92,11 @@ func syncConditions(workspaceStatus *dw.DevWorkspaceStatus, currentStatus *curre
 		currCondition, ok := currentStatus.conditions[workspaceCondition.Type]
 		if !ok {
 			// Didn't observe this condition this time; set status to unknown
```
sleshchenko, line 94: is this up to date?

sleshchenko (approved):
I haven't tested but LGTM

controllers/workspace/status.go
amisevsk: Yeah, the idea is that if a workspace had the `Ready` condition set to `true`, but this reconcile ended before we confirmed the workspace was still running, we want to set this condition to `Unknown` for now (since it was not observed). 

The difference in this PR is to not update that `Unknown` conditions `LastTransitionTime` if it's not actually transitioning.

Another option is deleting `Unknown` conditions from the list, which would do a similar thing, but does lose _some_ information.

Issue #520, opened 2021-07-26
Document "upstream" release


sleshchenko:
We need to document what is "upstream" release and how it's connected with the product.

Like each 6 weeks product release is going to happen

Questions:
1. does upstream needs tags? How often they should be made? On time bases or after product is released and we know the actual state we used for it?
2. does controller/webhookserver/project-clone should have tagged containers?
3. do we do next bits with github workflow and quay.io that allows to install the next DWO with custom catalog source>

sleshchenko:
https://github.com/devfile/devworkspace-operator/pull/528 is merged and handles creating x branches and tags.
Some pieces are still missing for upstream release:
- an ability to release with revision;
- set up github workflow to do the stuff;
- ... ?

PR #528, opened 2021-07-29
Adapt make-release to a new DWO release cycle

sleshchenko:
### What does this PR do?
Adapt make-release to a new DWO release cycle.

It's not well tested yet but the idea is to do releases like the following:
1. trigger prerelease, which will:
  - create a new .x branch like `v0.8.x`
  - bump version in main branch to `v0.9.0-dev`
2. do downstreaming from v0.8.x branch
3. after DWO is released and pushed to prod we proceed the next step.
4. Trigger release, which will:
  - patch resources with the right tagged values for deployment templates and Makefile;
    :question: need to decide if the right tagged values should be from redhat catalog or quay.io;
  - create a tag
  - bump version in .x branch for next bugfix release, like `v0.8.1` if it will happen;

it's for https://github.com/devfile/devworkspace-operator/issues/520

### Is it tested? How?
I've tested `./make-release.sh --prerelease --version v0.9.0 --tmp-dir`, the result is https://github.com/devfile/devworkspace-operator/pull/533

Also, I've tested it with
```
export DWO_REPO=git@github.com:sleshchenko/devworkspace-operator
export DWO_QUAY_REPO=quay.io/sleshche/devworkspace-controller
export PROJECT_CLONE_QUAY_REPO=quay.io/sleshche/project-clone
./make-release.sh --prerelease --version v0.9.0 --tmp-dir
./make-release.sh --release --version v0.9.0 --tmp-dir
./make-release.sh --release --version v0.9.1 --tmp-dir
```
So, you can check 
https://github.com/sleshchenko/devworkspace-operator/tree/main
https://github.com/sleshchenko/devworkspace-operator/tree/0.9.x
https://github.com/sleshchenko/devworkspace-operator/releases/tag/v0.9.0
https://github.com/sleshchenko/devworkspace-operator/releases/tag/v0.9.1

https://quay.io/sleshche/devworkspace-controller
https://quay.io/sleshche/project-clone

### PR Checklist

### DIFF
--- a/make-release.sh
+++ b/make-release.sh
@@ -33,7 +33,7 @@ Arguments:
   --version     : version to release, v0.8.0
   --prerelease  : flag to perform prerelease. Is performed from main branch
   --release     : flag to perform release. Is performed from v0.x.y branch
-  --revision    : GIT_SHA of the target brach to be used for releasing
+  --revision    : [is not implemented yet] GIT_SHA of the target brach to be used for releasing
 
 Development arguments to debug:
   --dry-run     : do not push changes locally
@@ -59,16 +59,17 @@ parse_args() {
       '--tmp-dir') TMP=$(mktemp -d); shift 0;;
       '--release') DO_RELEASE=true; shift 0;;
       '--prerelease') DO_PRERELEASE='true'; shift 0;;
+      '--revision') echo "[ERROR] --revision is not implemented yet"; exit 1;;
       '--dry-run') DRY_RUN='dryrun'; shift 0;;
       '--verbose') VERBOSE='true'; shift 0;;
       '--help') usage; exit 0;;
-      *) echo "Unknown parameter is used: $1."; usage; exit 1;;
+      *) echo "[ERROR] Unknown parameter is used: $1."; usage; exit 1;;
     esac
     shift 1
   done
 
   if [ -z "${VERSION}" ]; then
-    echo "Required parameter --version is missing."
+    echo "[ERROR] Required parameter --version is missing."
     usage
     exit 1
   fi
@@ -86,12 +87,17 @@ bump_version () {
   git checkout "${BUMP_BRANCH}"
 
   # change version/version.go file
-  sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+).*(\")#\1v${NEXT_VERSION}+rc\3#g"
+  VERSION_GO="v${NEXT_VERSION}"
+  if [[ "$BUMP_BRANCH" == "$MAIN_BRANCH" ]]; then
+    VERSION_GO="$VERSION_GO+dev"
+  fi
+  sed -i version/version.go -e "s#Version = \".*\"#Version = \"${VERSION_GO}\"#g"
   sed -i deploy/templates/components/csv/clusterserviceversion.yaml -r -e "s#(name: devworkspace-operator.)(v[0-9.]+)#\1v${NEXT_VERSION}#g"
-  sed -i deploy/templates/components/csv/clusterserviceversion.yaml -r -e "s#(version: )([0-9.]+)#\1${NEXT_VERSION}#g" 
+  sed -i deploy/templates/components/csv/clusterserviceversion.yaml -r -e "s#(version: )([0-9.]+)#\1${NEXT_VERSION}#g"
+  git add -A
   if [[ ! -z $(git status -s) ]]; then # dirty
     COMMIT_MSG="[release] Bump to ${NEXT_VERSION} in ${BUMP_BRANCH}"
-    git commit -asm "${COMMIT_MSG}"
+    git commit -m "${COMMIT_MSG}" --signoff
   fi
   git pull origin "${BUMP_BRANCH}"
 
@@ -118,58 +124,47 @@ bump_version () {
 }
 
 prerelease() {
+  if [ "${VERSION##*.}" != "0" ]; then
+    echo "[ERROR] Flag --prerelease should not be used for bugfix versions"
+    exit 1
+  fi
+
   echo "[INFO] Starting prerelease procedure"
   # derive bugfix branch from version
-  BRANCH=${VERSION#v}
-  BRANCH=${BRANCH%.*}.x
+  X_BRANCH=${VERSION#v}
+  X_BRANCH=${X_BRANCH%.*}.x
 
-  echo "[INFO] Creating ${BRANCH} from ${MAIN_BRANCH}"
+  echo "[INFO] Creating ${X_BRANCH} from ${MAIN_BRANCH}"
   git checkout ${MAIN_BRANCH}
-  git branch "${BRANCH}" || git checkout "${BRANCH}"
-  sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+-rc)(\")#\1${VERSION}\3#g"
+  git checkout -b "${X_BRANCH}"
+  sed -i version/version.go -e "s#Version = \".*\"#Version = \"${VERSION}\"#g"
   if [[ ! -z $(git status -s) ]]; then # dirty
-    COMMIT_MSG="[prerelease] Remove rc from version"
-    git commit -asm "${COMMIT_MSG}"
+    git add -A
+    COMMIT_MSG="[prerelease] Remove dev from version"
+    git commit -m "${COMMIT_MSG}" --signoff
   fi
-  $DRY_RUN git push origin "${BRANCH}"
-  git fetch origin "${BRANCH}:${BRANCH}" || true
-  git checkout "${BRANCH}"
+  $DRY_RUN git push origin "${X_BRANCH}"
 
-  # change VERSION file + commit change into ${BASEBRANCH} branch
-  # bump the y digit, if it is a major release
-  [[ $BRANCH =~ ^([0-9]+)\.([0-9]+)\.x ]] && BASE=${BASH_REMATCH[1]}; NEXT=${BASH_REMATCH[2]}; (( NEXT=NEXT+1 )) # for BRANCH=0.1.x, get BASE=0, NEXT=2
-  NEXT_VERSION_Y="${BASE}.${NEXT}.0"
-  bump_version "${NEXT_VERSION_Y}" "${MAIN_BRANCH}"
-  echo "Prerelease is done"
+  # bump version in MAIN_BRANCH to next dev version
+  [[ $X_BRANCH =~ ^([0-9]+)\.([0-9]+)\.x ]] \
+    && BASE=${BASH_REMATCH[1]}; \
+    NEXT=${BASH_REMATCH[2]}; \
+    (( NEXT=NEXT+1 )) # for X_BRANCH=0.1.x, get BASE=0, NEXT=2
+
+  NEXT_DEV_VERSION="${BASE}.${NEXT}.0"
+  bump_version "${NEXT_DEV_VERSION}" "${MAIN_BRANCH}"
+  echo "[INFO] Prerelease is done"
 }
 
 release() {
   echo "[INFO] Starting Release procedure"
   # derive bugfix branch from version
-  BRANCH=${VERSION#v}
-  BRANCH=${BRANCH%.*}.x
+  X_BRANCH=${VERSION#v}
+  X_BRANCH=${X_BRANCH%.*}.x
 
-  BASEBRANCH="${BRANCH}"
+  git fetch origin "${X_BRANCH}:${X_BRANCH}" || true
+  git checkout "${X_BRANCH}"
 
-  git remote show origin
-
-  # get sources from ${BASEBRANCH} branch
-  git fetch origin "${BASEBRANCH}":"${BASEBRANCH}" || true
-  git checkout "${BASEBRANCH}"
-
-  # create new branch off ${BASEBRANCH} (or check out latest commits if branch already exists), then push to origin
-  if [[ "${BASEBRANCH}" != "${BRANCH}" ]]; then
-    git branch "${BRANCH}" || git checkout "${BRANCH}"
-    $DRY_RUN git push origin "${BRANCH}"
-    git fetch origin "${BRANCH}:${BRANCH}" || true
-    git checkout "${BRANCH}"
-  else
-    git fetch origin "${BRANCH}:${BRANCH}" || true
-    git checkout "${BRANCH}"
-  fi
-
-  # change version/version.go file
-  sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+)(\")#\1${VERSION}\3#g"
   # change image tag in Makefile
   DWO_QUAY_IMG="${DWO_QUAY_REPO}:${VERSION}"
   sed -i Makefile -r -e "s#quay.io/devfile/devworkspace-controller:[0-9a-zA-Z._-]+#${DWO_QUAY_IMG}#g"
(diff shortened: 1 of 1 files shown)


make-release.sh
```diff
@@ -15,14 +15,57 @@ REPO=git@github.com:devfile/devworkspace-operator
...
-done
+usage () {
+cat << EOF
+This scripts handles upstream releasing stuff.
 
+Prerelease is supposed to happen first, that will create .x
+branch that will be used as a based for downstream DWO release.
+
+After DWO release candidate bits are built, tested and pushed to prod
+Release is supposed to happen, which will tag the DWO dowstream release
```
amisevsk, line 26:
```suggestion
Release is supposed to happen, which will tag the DWO downstream release
```

make-release.sh
```diff
@@ -15,14 +15,57 @@ REPO=git@github.com:devfile/devworkspace-operator
...
+
+Arguments:
+  --version     : version to release, v0.8.0
+  --prerelease  : flag to perform prerelease. Is performed from main branch
+  --release     : flag to perform release. Is performed from v0.x.y branch
+  --revision    : GIT_SHA of the target brach to be used for releasing
+
+Development arguments to debug:
+  --dry-run     : do not push changes locally
+  --tmp-dir     : perform operation in repo clonned into temporary folder
```
amisevsk, line 37:
```suggestion
  --tmp-dir     : perform operation in repo cloned into temporary folder 
```

make-release.sh
```diff
@@ -15,14 +15,57 @@ REPO=git@github.com:devfile/devworkspace-operator
...
+
+Development arguments to debug:
+  --dry-run     : do not push changes locally
+  --tmp-dir     : perform operation in repo clonned into temporary folder 
+
+Examples:
+$0 --prerelease --version v0.1.0
+$0 --release --version v0.1.0 --revision ${GIT_SHA}
+
+It's supposed to be triggered with GitHub Actions on the repo.
```
amisevsk, line 43:
```suggestion
This script is intended to be triggered by GitHub Actions on the repo.
```

make-release.sh
```diff
@@ -15,14 +15,57 @@ REPO=git@github.com:devfile/devworkspace-operator
...
+}
+
+parse_args() {
+  while [[ "$#" -gt 0 ]]; do
+    case $1 in
+      '-v'|'--version') VERSION="$2"; shift 1;;
+      '--tmp-dir') TMP=$(mktemp -d); shift 0;;
+      '--release') DO_RELEASE=true; shift 0;;
+      '--prerelease') DO_PRERELEASE=true; shift 0;;
+      '--dry-run') DRY_RUN='true'; shift 0;;
```
amisevsk, line 54: For consistency

```suggestion
      '--prerelease') DO_PRERELEASE='true'; shift 0;;
      '--dry-run') DRY_RUN='true'; shift 0;;
```

make-release.sh
```diff
@@ -15,14 +15,57 @@ REPO=git@github.com:devfile/devworkspace-operator
...
+      '--help') usage; exit 0;;
+      *) echo "Unknown parameter is used $1."; usage; exit 1;;
```
amisevsk, line 56:
```suggestion
      *) echo "Unknown parameter is used: $1."; usage; exit 1;;
```

make-release.sh
```diff
@@ -15,14 +15,57 @@ REPO=git@github.com:devfile/devworkspace-operator
...
+    esac
+    shift 1
+  done
+
+  if [[ ! ${VERSION} ]]; then
```
amisevsk, line 61: Should this be `if [ -z ${VERSION} ];`?

make-release.sh
```diff
@@ -33,55 +76,114 @@ bump_version () {
...
   if [[ ! -z $(git status -s) ]]; then # dirty
-    git add version/version.go
     COMMIT_MSG="[release] Bump to ${NEXT_VERSION} in ${BUMP_BRANCH}"
     git commit -asm "${COMMIT_MSG}"
   fi
   git pull origin "${BUMP_BRANCH}"
 
   set +e
-  PUSH_TRY="$(git push origin "${BUMP_BRANCH}")"
+  [[ -v "$DRY_RUN" ]] && PUSH_TRY="$(git push origin "${BUMP_BRANCH}")" || PUSH_TRY="protected branch hook declined"
```
amisevsk, line 89: I don't know if `-v` is the right option here; also it seems to have been added in bash `4.2` (news to me) so we'll probably run into issues when someone with a mac tries to run it :D

make-release.sh
```diff
@@ -33,55 +76,114 @@ bump_version () {
...
     PR_BRANCH=pr-${BUMP_BRANCH}-to-${NEXT_VERSION}
     # create pull request for the main branch branch, as branch is restricted
     git branch "${PR_BRANCH}"
     git checkout "${PR_BRANCH}"
     git pull origin "${PR_BRANCH}"
-    git push origin "${PR_BRANCH}"
+    [[ -v "$DRY_RUN" ]] && git push origin "${PR_BRANCH}" || echo "Dry run: push ${PR_BRANCH}"
     lastCommitComment="$(git log -1 --pretty=%B)"
-    hub pull-request -f -m "${lastCommitComment}" -b "${BUMP_BRANCH}" -h "${PR_BRANCH}"
+    [[ -v "$DRY_RUN" ]] && hub pull-request -f -m "${lastCommitComment}" -b "${BUMP_BRANCH}" -h "${PR_BRANCH}" || echo "Dry run: create PR"
```
amisevsk, line 99: Outside of the scope of this PR, but should we move from [`hub`](https://github.com/github/hub) to [`gh`](https://github.com/cli/cli)?

make-release.sh
```diff
@@ -33,55 +76,114 @@ bump_version () {
...
+  BRANCH=${VERSION#v}
+  BRANCH=${BRANCH%.*}.x
 
-if [[ ! ${VERSION} ]]; then
-  usage
-  exit 1
-fi
+  git branch "${BRANCH}" || git checkout "${BRANCH}"
+  [[ -v "$DRY_RUN" ]] && git push origin "${BRANCH}" || echo "Dry run: push ${BRANCH}"
+  git fetch origin "${BRANCH}:${BRANCH}" || true
```
amisevsk, line 112: Wondering how this will work for bugfix releases -- if e.g. I have `0.8.0` released locally and `0.8.x` branch existing, but `0.8.1` has been released in upstream, this looks like it will fetch changes but not merge them to the local branch.

make-release.sh
```diff
@@ -33,55 +76,114 @@ bump_version () {
...
+  git checkout "${BRANCH}"
 
+  # change VERSION file + commit change into ${BASEBRANCH} branch
+  # bump the y digit, if it is a major release
+  [[ $BRANCH =~ ^([0-9]+)\.([0-9]+)\.x ]] && BASE=${BASH_REMATCH[1]}; NEXT=${BASH_REMATCH[2]}; (( NEXT=NEXT+1 )) # for BRANCH=0.1.x, get BASE=0, NEXT=2
+  NEXT_VERSION_Y="${BASE}.${NEXT}.0"
+  bump_version "${NEXT_VERSION_Y}" "${MAIN_BRANCH}"
+
```
amisevsk, line 120:
```suggestion
```

make-release.sh
```diff
@@ -33,55 +76,114 @@ bump_version () {
...
-  exit 1
-fi
+  git branch "${BRANCH}" || git checkout "${BRANCH}"
+  [[ -v "$DRY_RUN" ]] && git push origin "${BRANCH}" || echo "Dry run: push ${BRANCH}"
+  git fetch origin "${BRANCH}:${BRANCH}" || true
+  git checkout "${BRANCH}"
 
+  # change VERSION file + commit change into ${BASEBRANCH} branch
+  # bump the y digit, if it is a major release
+  [[ $BRANCH =~ ^([0-9]+)\.([0-9]+)\.x ]] && BASE=${BASH_REMATCH[1]}; NEXT=${BASH_REMATCH[2]}; (( NEXT=NEXT+1 )) # for BRANCH=0.1.x, get BASE=0, NEXT=2
```
amisevsk, line 117: Lines like this make me wish this script was written in python or something :smile:

make-release.sh
sleshchenko: ¯\_(ツ)_/¯
I just moved this code but it's nice catch.
I'll double check

make-release.sh
sleshchenko: Now that's multiline and it's easier to read

make-release.sh
sleshchenko: changes to -z

make-release.sh
sleshchenko: good catch. I'll replace it with -z.

sleshchenko: Well, shell check suggests -n instead https://github.com/koalaman/shellcheck/wiki/SC2236.
but now I rewrote it in the following way:
```bash
  [[ -z "$DRY_RUN" ]] && echo "dry run command" \
    || echo "real command"
```

sleshchenko: then shellcheck suggest not to use such construction for our purposes https://github.com/koalaman/shellcheck/wiki/SC2015 )

make-release.sh
sleshchenko: prerelease should not happen for bugfix releases, since it's purpose is just to create .x branch from main and bump version in main branch

amisevsk: Makes sense :+1:

JPinkney (approved):

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
-usage ()
-{
-  echo "Usage: $0 --version [VERSION TO RELEASE]"
-  echo "Example: $0 --version v0.1.0"; echo
-}
+prerelease() {
+  echo "[INFO] Starting prerelease procedure"
+  # derive bugfix branch from version
+  BRANCH=${VERSION#v}
+  BRANCH=${BRANCH%.*}.x
```
amisevsk, line 124: Should we include a check along the lines of
```bash
if [ "${VERSION##*.}" != "0" ]; then 
  echo "Flag --prerelease should not be used for bugfix versions"
  exit 1
fi
```

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
-if [[ ${VERSION} == *".0" ]]; then
-  BASEBRANCH="${MAIN_BRANCH}"
-else
   BASEBRANCH="${BRANCH}"
-fi
 
-# work in tmp dir
-if [[ $TMP ]] && [[ -d $TMP ]]; then
-  pushd "$TMP" > /dev/null || exit 1
+  git remote show origin
```
amisevsk, line 154: Is this a debugging message?

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
   # get sources from ${BASEBRANCH} branch
-  echo "Check out ${REPO} to ${TMP}/${REPO##*/}"
-  git clone "${REPO}" -q
-  cd "${REPO##*/}" || exit 1
-fi
+  git fetch origin "${BASEBRANCH}":"${BASEBRANCH}" || true
+  git checkout "${BASEBRANCH}"
+
+  # create new branch off ${BASEBRANCH} (or check out latest commits if branch already exists), then push to origin
+  if [[ "${BASEBRANCH}" != "${BRANCH}" ]]; then
```
amisevsk, line 161: Can we ever hit this branch? We set `BASEBRANCH="${BRANCH}"` on line 154 above and there's no writes to either variable in between.

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
+
   # get sources from ${BASEBRANCH} branch
-  echo "Check out ${REPO} to ${TMP}/${REPO##*/}"
-  git clone "${REPO}" -q
-  cd "${REPO##*/}" || exit 1
-fi
+  git fetch origin "${BASEBRANCH}":"${BASEBRANCH}" || true
+  git checkout "${BASEBRANCH}"
+
+  # create new branch off ${BASEBRANCH} (or check out latest commits if branch already exists), then push to origin
```
amisevsk, line 160: This use case is covered by `--prerelease`, isn't it?

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
+    git fetch origin "${BRANCH}:${BRANCH}" || true
+    git checkout "${BRANCH}"
+  else
+    git fetch origin "${BRANCH}:${BRANCH}" || true
+    git checkout "${BRANCH}"
+  fi
 
-git remote show origin
+  # change version/version.go file
+  sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+)(\")#\1${VERSION}\3#g"
```
amisevsk, line 172: This should be doable without grouping:
```suggestion
  sed -i version/version.go -e "s#Version = \".*\"#Version = \"${VERSION}\"#g"
```
The current regex will also fail to match the version if it is formatted e.g. `v0.1.0+dev` since it only looks for numbers and dots.

We could also avoid a bit of repetition by doing a line match + replace:
```
sed version/version.go -e "/Version =/ s|\".*\"|\"${VERSION}\"|"
```

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
+  # change image tag in Makefile
+  DWO_QUAY_IMG="${DWO_QUAY_REPO}:${VERSION}"
+  sed -i Makefile -r -e "s#quay.io/devfile/devworkspace-controller:[0-9a-zA-Z._-]+#${DWO_QUAY_IMG}#g"
+  docker build -t "${DWO_QUAY_IMG}" -f ./build/Dockerfile .
+  $DRY_RUN docker push "${DWO_QUAY_IMG}"
+
+  PROJECT_CLONE_QUAY_IMG="${PROJECT_CLONE_QUAY_REPO}:${VERSION}"
+  sed -i Makefile -r -e "s#quay.io/devfile/project-clone:.*[0-9a-zA-Z._-]#${PROJECT_CLONE_QUAY_IMG}#g"
```
amisevsk, line 180:
```suggestion
  sed -i Makefile -r -e "s#quay.io/devfile/project-clone:[0-9a-zA-Z._-]+#${PROJECT_CLONE_QUAY_IMG}#g"
```

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
+    git checkout "${BRANCH}"
+  else
+    git fetch origin "${BRANCH}:${BRANCH}" || true
+    git checkout "${BRANCH}"
+  fi
 
-git remote show origin
+  # change version/version.go file
+  sed -i version/version.go -r -e "s#(Version = \")(v?[0-9.]+)(\")#\1${VERSION}\3#g"
+  # change image tag in Makefile
```
amisevsk, line 173: Is it necessary to change image tags in the `Makefile`? I don't think make is used anywhere in the release script and it's already written to support setting alternate images via env vars.

make-release.sh
```diff
@@ -11,147 +11,221 @@
...
+  bash -x ./deploy/generate-deployment.sh \
+    --use-defaults \
+    --default-image "${DWO_QUAY_IMG}" \
+    --project-clone-image "${PROJECT_CLONE_QUAY_IMG}"
+
+  # tag the release if the version/version.go file has changed
+  if [[ ! -z $(git status -s) ]]; then # dirty
+    COMMIT_MSG="[release] Release ${VERSION}"
+    git add version/version.go
+    git commit -asm "${COMMIT_MSG}"
```
amisevsk, line 193: `git add version/version.go` is redundant from the `-a` flag in `git commit`, however using the `-a` flag has slightly different semantics from `git add .` as untracked files are ignored.

```suggestion
    git add .
    git commit -sm "${COMMIT_MSG}"
```

make-release.sh
sleshchenko: it's not necessary but it may be handy or not very but here is a request https://github.com/devfile/devworkspace-operator/issues/415

make-release.sh
sleshchenko: Blame me for moving stuff without paying enough attention.
I suppose so, will remove it until we don't see need in it.

make-release.sh
sleshchenko: :man_facepalming: <- this is the reaction to my reworking the script with low quality. Revisiting all pieces

make-release.sh
sleshchenko: removed

make-release.sh
sleshchenko: should be fixed

make-release.sh
sleshchenko: removed groups for sed in version.go

make-release.sh
sleshchenko: now we do `git add -A` and git commit -m "" to do stuff more explicit + not to lose some changes we make but not commit.

sleshchenko:
@amisevsk thanks for such a great review. I believe current state should be much better and simpler, but there still may be places to improve

sleshchenko:
merging this to make further review easier. But please do review and let me know if anything is still unaddressed or needs to be improved.

Issue #530, opened 2021-07-29
Add support for defining environment variables for all containers in a component


amisevsk:
### Description

Since https://github.com/devfile/api/issues/149 will not be included in the Devfile 2.0 spec, we should consider supporting similar functionality via a controller-specific attribute on components. Some Che components (e.g. telemetry for the publicly hosted deployment) require the feature to function correctly, and reimplementing it as suggested in https://github.com/devfile/api/issues/239 is difficult.

#### 1. Implement workspaceEnv as global variables
This would require
1. A plugin to define a devfile 2.0 global variable, e.g.
    ```yaml
    template:
      variables:
        MY_ENV: "desired_value"
    ```
2. All plugins that expect to use the variable to have an environment variable defined
    ```yaml
    env:
      - name: MY_ENV
        value: "{{MY_ENV}}"
    ```

However, the dependent plugins from step 2 cannot assume that the variable will be provided. If the plugin from step 1 is not included in a DevWorkspace, the DevWorkspace will issue a warning that there's no replacement for `{{MY_ENV}}`.

#### 2. Implement workspaceEnv as an attribute
A plugin could define an attribute
```yaml
attributes:
  workspaceEnv:
    name: MY_ENV
    value: "desired_value"
```

This would signal the controller to add the env var `MY_ENV=desired_value` to all containers in the DevWorkspace. However, this would be DevWorkspace controller-specific behaviour and completely outside the Devfile 2.0 spec as it's currently written.

PR #529, opened 2021-07-29
Support workspace env

amisevsk:
### What does this PR do?
Adds support for workspaceEnv as in devfile 1.0 via the attribute `workspaceEnv` on components. Environment variables can be added to all containers in a DevWorkspace by specifying e.g.

```yaml
components:
  - name: my-component
    attributes:
      workspaceEnv: 
        TEST_ENV_1: TEST_VAL_1
        TEST_ENV_2: TEST_VAL_2
```

### What issues does this PR fix or reference?
Closes https://github.com/devfile/devworkspace-operator/issues/530

### Is it tested? How?
Test cases are added. Can be tested directly using the DevWorkspace 
```yaml
cat <<EOF | kubectl apply -f -
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: theia-next
spec:
  started: true
  template:
    attributes:
      controller.devfile.io/storage-type: "ephemeral"
    components:
      - name: tooling
        attributes:
          workspaceEnv: 
            TEST_ENV_1: TEST_VAL_1
            TEST_ENV_2: TEST_VAL_2
        container:
          image: quay.io/wto/web-terminal-tooling
          args: ["tail", "-f", "/dev/null"]
      - name: theia
        plugin:
          uri: https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml

```

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v7-devworkspaces-operator-e2e, v7-devworkspace-happy-path` to trigger)
    - [ ] `v7-devworkspaces-operator-e2e`: DevWorkspace e2e test
    - [ ] `v7-devworkspace-happy-path`: DevWorkspace e2e test

PR #538, opened 2021-07-30
Support global environment variables defined by a component-level attribute

amisevsk:
This PR is a slightly cleaned up iteration of https://github.com/devfile/devworkspace-operator/pull/529

### What does this PR do?
Add support for the devfile 1.0 feature `workspaceEnv` (env vars supplied by a component that get added to all containers in the workspace). This is defined using an attribute on *components* named `workspaceEnv` formatted as `[]EnvVar`, i.e.:
```yaml
attributes:
  workspaceEnv: 
    - name: MY_ENV_1
      value: value
    - name: MY_ENV_2
      value: value
```

Note: it's necessary to define the attribute on components for now, because merging plugin components with top-level attributes is unsuitable to some use cases:

1. If the base DevWorkspace defines a workspaceEnv attribute, any plugin that also defines a top-level workspaceEnv attribute will fail to merge during flattening
2. If two plugins define workspaceEnv attributes at the top level, the second plugin will overwrite the value from the first, i.e. it's only possible to define workspaceEnv in one place in this way.

An alternative approach that would allow using top-level attributes would be some sort of namespacing mechanism, e.g. using `workspace-env/my-component` as the attribute, but this would be harder to process and write.

Closes https://github.com/devfile/devworkspace-operator/issues/530

### Is it tested? How?
Test cases are included; can be tested directly using the DevWorkspace 
```yaml
kind: DevWorkspace
metadata:
  name: theia-next
spec:
  started: true
  template:
    components:
      - name: tooling
        attributes:
          workspaceEnv: 
            - name: TEST_ENV_1
              value: TEST_VAL_1
            - name: TEST_ENV_2
              value: TEST_VAL_2
        container:
          image: quay.io/wto/web-terminal-tooling
      - name: theia
        plugin:
```
and checking that all containers in the resulting pod have the environment variables listed.

### PR Checklist

### DIFF
--- a/pkg/library/flatten/helper.go
+++ b/pkg/library/flatten/helper.go
@@ -17,6 +17,7 @@ import (
 	"reflect"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
 )
 
 // resolutionContextTree is a recursive structure representing information about the devworkspace that is
@@ -63,4 +64,21 @@ func formatImportCycle(end *resolutionContextTree) string {
 		cycle = fmt.Sprintf("%s -> %s", end.componentName, cycle)
 	}
 	return cycle
+}
+
+// getSourceForComponent returns the 'original' name for a component in a flattened DevWorkspace. Given a component, it
+// returns the name of the plugin component that imported it if the component came via a plugin, and the actual
+// component name otherwise.
+//
+// The purpose of this function is mainly to enable providing better messages to end-users, as a component name may
+// not match the name of the plugin in the original DevWorkspace.
+func getSourceForComponent(component dw.Component) string {
+	if component.Attributes.Exists(constants.PluginSourceAttribute) {
+		var err error
+		componentName := component.Attributes.GetString(constants.PluginSourceAttribute, &err)
+		if err == nil {
+			return componentName
+		}
+	}
+	return component.Name
 }
--- a/pkg/library/flatten/testdata/workspace-env/error_workspaceEnv-formatted-wrong.yaml
+++ b/pkg/library/flatten/testdata/workspace-env/error_workspaceEnv-formatted-wrong.yaml
@@ -1,0 +1,29 @@
+name: "Returns an error when workspaceEnv cannot be read as []EnvVar"
+
+input:
+  devworkspace:
+    components:
+      - name: test-plugin
+        plugin:
+          uri: "https://my-plugin.io/test"
+  devfileResources:
+    "https://my-plugin.io/test":
+      schemaVersion: 2.0.0
+      metadata:
+        name: "plugin-a"
+      components:
+        - name: plugin-a
+          attributes:
+            # Note this fails to deserialize because it's just a yaml string, not a list.
+            # JSON can be used here if you remove the pipe
+            workspaceEnv: |
+              [{"name": "TEST_ENV_1",
+               "value": "TEST_VAL_1"},
+              {"name": "TEST_ENV_2",
+               "value": "TEST_VAL_2"}]
+          container:
+            name: test-container
+            image: test-image
+
+output:
+  errRegexp: "failed to read attribute workspaceEnv on component test-plugin"
--- a/pkg/library/flatten/workspaceEnv.go
+++ b/pkg/library/flatten/workspaceEnv.go
@@ -1,0 +1,83 @@
+//
+// Copyright (c) 2019-2021 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package flatten
+
+import (
+	"fmt"
+
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+)
+
+const (
+	// WorkspaceEnvAttribute is an attribute that specifies a set of environment variables provided by a component
+	// that should be added to all workspace containers. The structure of the attribute value should be a list of
+	// Devfile 2.0 EnvVar, e.g.
+	//
+	//   attributes:
+	//     workspaceEnv:
+	//       - name: ENV_1
+	//         value: VAL_1
+	//       - name: ENV_2
+	//         value: VAL_2
+	WorkspaceEnvAttribute = "workspaceEnv"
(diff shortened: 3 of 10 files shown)


pkg/library/flatten/helper.go
```diff
@@ -64,3 +65,14 @@ func formatImportCycle(end *resolutionContextTree) string {
 	}
 	return cycle
 }
+
+func getOriginalNameForComponent(component dw.Component) string {
```
sleshchenko, line 69: Well, we don't modify components names, it returns exactly what is written in body: SourceComponent. Maybe then rename it to `getSourceForComponent`

pkg/library/flatten/workspaceEnv.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"fmt"
+
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+)
+
+const (
+	// WorkspaceEnvAttribute is an attribute that specifies a set of environment variables provided by a component
+	// that should be added to all workspace containers. The structure of the attribute value should be a map of strings
```
sleshchenko, line 23: > The structure of the attribute value should be a map of strings
> 	// to strings, e.g.

it's not true, right? Because it's an array of EnvVar objects

pkg/library/flatten/workspaceEnv.go
```diff
@@ -0,0 +1,83 @@
...
+
+	for _, component := range flattenedDW.Components {
+		if !component.Attributes.Exists(WorkspaceEnvAttribute) {
+			continue
+		}
+
+		var componentEnv []dw.EnvVar
+		err := component.Attributes.GetInto(WorkspaceEnvAttribute, &componentEnv)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read attribute %s on component %s: %w", WorkspaceEnvAttribute, getOriginalNameForComponent(component), err)
```
sleshchenko, line 65: validate the format of know attribute can be +1 things we can validate with validate webhooks.

pkg/library/flatten/testdata/workspace-env/no-error-on-duplicate-workspace-env.yaml
```diff
@@ -0,0 +1,75 @@
+name: "Adds workspace environment variables from plugin"
```
sleshchenko, line 1: please allign test files names, now some are with _, some only -, some with CamelCase for workspaceEnv, some has workspace-env.

pkg/library/flatten/testdata/workspace-env/adds-workspace-env-written-as-json.yaml
```diff
@@ -0,0 +1,84 @@
+name: "Adds JSON workspace environment variables from plugin "
```
sleshchenko, line 1: File name and case name does not seem to be really the same.

pkg/library/flatten/testdata/workspace-env/adds-workspace-env-to-all-containers.yaml
```diff
@@ -0,0 +1,84 @@
+name: "Adds workspace environment variables from plugin"
```
sleshchenko, line 1: This suite case appear the second time

pkg/library/flatten/testdata/workspace-env/error_workspaceEnv-formatted-wrong.yaml
```diff
@@ -0,0 +1,29 @@
...
+              [{"name": "TEST_ENV_1",
+               "value": "TEST_VAL_1"},
+              {"name": "TEST_ENV_2",
+               "value": "TEST_VAL_2"}]
+          container:
+            name: test-container
+            image: test-image
+
+output:
+  errRegexp: "failed to read attribute workspaceEnv on component test-plugin"
```
sleshchenko, line 29: I know, it's not something that DWO controls but maybe it makes sense to include error as well, to understand better user-facing message. Currently it's going to be

```
failed to read attribute workspaceEnv on component test-plugin: json: cannot unmarshal string into Go value of type []v1alpha2.EnvVar
```

sleshchenko: Or even `failed to process workspaceEnv: failed to read attribute workspaceEnv on component test-plugin: json: cannot unmarshal string into Go value of type []v1alpha2.EnvVar
`

sleshchenko (approved):
LGTM

Just a few minor comments. Good job!

pkg/library/flatten/helper.go
amisevsk: We don't modify component names per se, but we delete plugin components and add e.g. container components:
```yaml
components:
  - name: my-plugin
    plugin: [...]
```
becomes
```yaml
components:
  - name: plugin-container
    container: [...]
```
But I see your point -- I'll update.

pkg/library/flatten/workspaceEnv.go
amisevsk: Yup, forgot to update the comment as a switched back and forth between the structures :)

JPinkney (approved):
Other then Serhii's comments it looks good to me

pkg/library/flatten/testdata/workspace-env/error_workspaceEnv-formatted-wrong.yaml
amisevsk: I'm not strongly against testing the wording of errors, but I would want a better way to check. While the chance is small, this could ostensibly depending on Go version if we're looking for exact wording.

I'm also of the opinion that testing exact error messages is basically just writing the error message twice, but that's a more philosophical topic :)

pkg/library/flatten/testdata/workspace-env/no-error-on-duplicate-workspace-env.yaml
amisevsk: Fixed for consistency -- note the format I've used thus far for `_` vs `-` is that dashes replace spaces in the sentence describing the test, and `error_` is the prefix that specifies that a test case is expected to return an error (it's mostly so that an alphabetical sort puts all error testcases together). If I've missed a place in that framework, I'll fix it.

pkg/library/flatten/testdata/workspace-env/adds-workspace-env-written-as-json.yaml
amisevsk: Copy-paste error :)

sleshchenko (approved):

sleshchenko:
@amisevsk feel free to merge with squash or rebase to get rid of fixups and rerun tests.

Issue #516, opened 2021-07-26
Update K8s API to newest


sleshchenko:
It's a pre-req for https://github.com/devfile/devworkspace-operator/pull/504
https://github.com/devfile/api should be updated too.

amisevsk:
@sleshchenko I assigned you as well since you have https://github.com/devfile/api/pull/565 open. Let me know if you'd like me to adopt it

PR #540, opened 2021-08-04
Update controller runtime to v0.9.5 and k8s dependencies to v0.21.3

amisevsk:
### What does this PR do?
Update controller-runtime to `v0.9.5` and associated dependencies to match (mainly, k8s deps to `v0.21.3`)

Commits walk through required changes step-by-step -- we're going through 3 versions of breaking changes so there's a lot of scattered updates needed.

Fixes https://github.com/devfile/devworkspace-operator/issues/516
Fixes https://github.com/devfile/devworkspace-operator/issues/543

### Is it tested? How?
Should test controller thoroughly; I've done basic testing and found no issues.

### PR Checklist


amisevsk:
One note about the updated controller-runtime dependency: the latest controller-runtime releases suggest updating to go 1.16, which we currently cannot do since no 1.16 UBI8 images appear to be available at this time. This isn't a huge problem, but some internal packages in controller-runtime make use of the [`io` package](https://golang.org/doc/go1.16#ioutil), which causes `go mod tidy` to fail on `go <1.15.10`. See issue https://github.com/golang/go/issues/44557 for details.

I've tried to sync everything to match go 1.15.13, which is used in our current UBI image, but we should update to 1.16 once that image becomes available.

sleshchenko (approved):
LGTM

sleshchenko:
rebased against main to fix PR checks

JPinkney (approved):
LGTM

amisevsk:
Great news: turns out prow CI is using some older version of Go and so checks will always fail due to https://github.com/devfile/devworkspace-operator/pull/540#issuecomment-892747218

amisevsk:
Added to PR:

* Update devfile/api dependency to match k8s v0.21.3 change (https://github.com/devfile/api/pull/565)
* Update webhooks config to use `v1` webhooks instead of `v1beta1` (and also `networking.k8s.io/v1` Ingresses)

metlos (approved):

metlos:
Actually, I would prefer if there were 2 PRs, if possible - one for ingress v1 and the other for the controller runtime upgrade. Upgrading the controller runtime in che-operator (which depends on DWO) is going to be a much harder task and we should not block progress on v1beta1 removal there by requiring a simulatenous upgrade of controller runtime.

sleshchenko:
Controller runtime is upgraded in https://github.com/devfile/devworkspace-operator/pull/567
K8s API and dropping usage of beta APIs are part of https://github.com/devfile/devworkspace-operator/pull/556

PR #547, opened 2021-08-10
Add sample grafana dashboard and short README doc

amisevsk:
### What does this PR do?
Adds sample Grafana dashboard that consumes controller metrics to `doc/grafana`. The large json blob basically comes from exporting from within Grafana running locally -- hopefully it's easily compatible/importable elsewhere, but I haven't tested. It's also worth testing how reproducible exports are (if you export the dashboard after making a change, how much does the json file change?)

### Is it tested? How?
README.md doc added which explains how to use this locally.

### PR Checklist

### DIFF
--- a/doc/grafana/README.md
+++ b/doc/grafana/README.md
@@ -1,0 +1,78 @@
+# Grafana dashboards for the DevWorkspace Operator
+
+This directory contains a sample Grafana dashboard that processes some of the prometheus metrics exposed by the DevWorkspace Operator. Once Prometheus and Grafana are configured with the DevWorkspace Operator's metrics server as a source, the dashboard can be imported from [json](grafana-dashbaord.json).
+
+## Testing DevWorkspace Operator metrics locally
+
+To quickly test metrics locally (without installing Prometheus Grafana in the cluster), it's possible to deploy prometheus and grafana as containers locally.
+
+1. Create a clusterrolebinding for a service account to allow the metrics-reader role (this example uses the controller's service account)
+    ```bash
+    NAMESPACE=devworkspace-controller # Namespace where controller is installed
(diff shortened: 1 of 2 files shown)


doc/grafana/README.md
```diff
@@ -0,0 +1,78 @@
...
+    ```bash
+    TOKEN=$(kubectl get secrets -o=json -n ${NAMESPACE} | jq -r '[.items[] |
+      select (
+        .type == "kubernetes.io/service-account-token" and
+        .metadata.annotations."kubernetes.io/service-account.name" == "devworkspace-controller-serviceaccount")][0].data.token' \
+      | base64 --decode)
+    ```
+3. In another terminal, use `kubectl port-forward` to expose the controller's service locally
+    ```bash
+    kubectl port-forward service/devworkspace-controller-manager-service 8443:8443 &;
```
sleshchenko, line 26:
```
error: Service devworkspace-controller-manager-service does not have a service port 8443
[1]  + 30425 exit 1     kubectl port-forward service/devworkspace-controller-manager-service 8443:8443
```

```suggestion
    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &;
```

sleshchenko (approved):
LGTM, just one fix is needed to point prometheus to the right service.

Some minor stuff is not very clear on dashboard, like what is 1.25 (actually 0 Successes, 1 Failures for that period) I may just don't get something important point.
Screenshot_20210811_144701

But let's merge it and then contribute together to make it ideal :)

doc/grafana/README.md
amisevsk: Right, forgot we had to change that -- thanks!

JPinkney (approved):
Other then the small issue Serhii found it LGTM

Issue #465, opened 2021-06-22
DevWorkspace Operator should report deployment issues


sleshchenko:
Currently, devworkspace operator does not track deployment issues, like invalid memory limit and request (note this particular case we handle on DWO case after https://github.com/devfile/devworkspace-operator/pull/464) and some more cases when Deployment can not create pod.
So, this issue is about making DWO track deployment events and detect unrecoverable ones, I guess it can be similar to what Che Server has.

Issue #484, opened 2021-07-08
Happy path tests don't use PR deployment templates


amisevsk:
### Description

The happy path test uses `chectl` to deploy the Che operator with DevWorkspaces support enabled. The only configuration applied to the CheCluster CR is:
```
+ cat /tmp/che-cr-patch.yaml
spec:
  devWorkspace:
    enable: true
    controllerImage: 'registry.build02.ci.openshift.org/ci-op-3z668v78/pipeline@sha256:322dc913cd574dbe5c2069f36452cac18c5fcaaaec1653173bfcfd5157fa0f2a'
```
where the controllerImage is the result of the CI build. 

This means any PR that needs to update deployment templates will fail the happy path test, as the test will deploy an updated image with stale templates.

PR #549, opened 2021-08-10
Check devworkspace pod events for "Unrecoverable" events

amisevsk:
### What does this PR do?
Checks events related to the DevWorkspace pod for what we would consider "unrecoverable" events (list stolen from [Che](https://github.com/eclipse-che/che-server/blob/72bfc85436022bc87aad109829411ef7f9b48663/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L358) :stuck_out_tongue:)

This requires adding an index to the manager, to allow us to list events with the appropriate fieldSelector.

Also adds `get/list/watch` Events RBAC to the controller.

Closes https://github.com/devfile/devworkspace-operator/issues/465

### Is it tested? How?
It's hard to reproduce these events manually; the easiest way I found was to patch `commonStorage.go` to read
```go
// Line 115
	podAdditions.Volumes = append(podAdditions.Volumes, corev1.Volume{
		Name: commonPVCName,
		VolumeSource: corev1.VolumeSource{
			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
				ClaimName: commonPVCName+"fake",
			},
		},
	})
```
and trying to start a non-ephemeral DevWorkspace.

### PR Checklist

### DIFF
--- a/controllers/workspace/provision/deployment.go
+++ b/controllers/workspace/provision/deployment.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/fields"
+
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/controllers/workspace/env"
 	maputils "github.com/devfile/devworkspace-operator/internal/map"
@@ -41,11 +43,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-var ContainerFailureStateReasons = []string{
+var containerFailureStateReasons = []string{
 	"CrashLoopBackOff",
 	"ImagePullBackOff",
 	"CreateContainerError",
 	"RunContainerError",
+}
+
+var unrecoverablePodEventReasons = []string{
+	"FailedMount",
+	"FailedScheduling",
+	"MountVolume.SetUp failed",
+	"FailedCreate",
+	"ReplicaSetCreateError",
 }
 
 type DeploymentProvisioningStatus struct {
@@ -163,7 +173,7 @@ func SyncDeploymentToCluster(
 		}
 	}
 
-	failureMsg, checkErr := checkFailedPods(workspace, clusterAPI)
+	failureMsg, checkErr := checkPodsState(workspace, clusterAPI)
 	if checkErr != nil {
 		return DeploymentProvisioningStatus{
 			ProvisioningStatus: ProvisioningStatus{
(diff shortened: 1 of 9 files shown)


controllers/workspace/provision/deployment.go
```diff
@@ -41,13 +43,21 @@ import (
...
-var ContainerFailureStateReasons = []string{
+var containerFailureStateReasons = []string{
 	"CrashLoopBackOff",
 	"ImagePullBackOff",
 	"CreateContainerError",
 	"RunContainerError",
 }
 
+var unrecoverablePodEventReasons = []string{
+	"FailedMount",
```
sleshchenko, line 54: On some infrastructure, FailedMount or FailedScheduling can be an usual situation when you just need to be more patient and then on Che we suggest users to drop them from unrecoverable events list.
We should have it configurable in some way or let users annotate particular workspace, maybe get it from namespaced configmap settings if present.
WDYT?

sleshchenko (approved):

controllers/workspace/provision/deployment.go
amisevsk: I feel like an annotation system might complicate things too much -- we should just do what we think is best (whether that's check or don't check).

One issue I've seen in Che before is that there can be a bit of a [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem) -- if I've got 10 workspaces waiting for the common PVC to be mountable, it can be very confusing because they'll randomly start up and have to be stopped until it gets to the one I want to actually use. 

One option that just occurred to me from looking at the Event struct a bit more -- what if we have a threshold for "number of times the event has occurred"? We could fail workspace start if e.g. we've seen the `FailedMount` even three times.

controllers/workspace/provision/deployment.go
```diff
@@ -390,6 +405,31 @@ func checkFailedPods(workspace *dw.DevWorkspace,
...
+}
+
+func checkPodEvents(pod *corev1.Pod, clusterAPI ClusterAPI) (msg string, err error) {
+	evs := &corev1.EventList{}
+	selector, err := fields.ParseSelector(fmt.Sprintf("involvedObject.name=%s", pod.Name))
+	if err != nil {
+		return "", fmt.Errorf("failed to parse field selector: %s", err)
+	}
+	if err := clusterAPI.Client.List(clusterAPI.Ctx, evs, k8sclient.InNamespace(pod.Namespace), k8sclient.MatchingFieldsSelector{Selector: selector}); err != nil {
+		return "", fmt.Errorf("failed ot list events in namespace %s: %w", pod.Namespace, err)
```
JPinkney, line 422:
```suggestion
		return "", fmt.Errorf("failed to list events in namespace %s: %w", pod.Namespace, err)
```

JPinkney (approved):

controllers/workspace/provision/deployment.go
amisevsk: Created https://github.com/devfile/devworkspace-operator/issues/550 to follow up

amisevsk:
Happy path tests fail due to https://github.com/devfile/devworkspace-operator/issues/484 (Events not in RBAC used in the test itself).

I'm merging this one now, with the note that we should follow up in https://github.com/devfile/devworkspace-operator/issues/550

Issue #479, opened 2021-07-08
Implement validation webhooks which would validate devfile consistency


sleshchenko:
Implement validation webhooks which would validate devfile consistency, like referencing non-existing remote for project, or volume, before devworkspace creation.

sleshchenko:
implemented in https://github.com/devfile/devworkspace-operator/pull/555

PR #555, opened 2021-08-20
Validate incoming devfiles

JPinkney:
### What does this PR do?
This PR introduces validation for incoming devfiles. It does a couple things:
1. Introduces the ability to easily debug the webhooks using telepresence
2. Introduces validation for devfiles based off the https://github.com/devfile/api/tree/main/pkg/validation
3. Fixes a couple samples so that they pass validation

Things currently being validated can be found here: https://github.com/devfile/api/blob/main/pkg/validation/validation-rule.md

https://github.com/devfile/devworkspace-operator/issues/479

### PR Checklist


deploy/scripts/generate-cert.sh
```diff
@@ -0,0 +1,33 @@
...
+
+chmod 0700 "$key_dir"
+cd "$key_dir"
+
+# Generate the CA cert and private key
+openssl req -nodes -new -x509 -keyout ca.key -out ca.crt -days 1024 -subj "/CN=Admission Workspace Controller Webhook"
+# Generate the private key for the webhook server
+openssl genrsa -out tls.key 2048
+# Generate a Certificate Signing Request (CSR) for the private key, and sign it with the private key of the CA.
+openssl req -new -key tls.key -subj "/CN=devworkspace-webhookserver.devworkspace-controller.svc" \
```
sleshchenko, line 32: can't we just grab generated for in-cluster webserver? By openshift or certmanager?

sleshchenko (commented):
I haven't tested yet but changes LGTM

sleshchenko (approved):
~~Tested on minikube v1.22.0, OpenShift 4.8 and 4.9 and everything seems to work fine.~~ haha wrong tab

deploy/scripts/generate-cert.sh
sleshchenko: if we can't use one the real webhook server uses, I would move it somewhere in `/local/scripts` folder instead of deploy, since it's not part of K8s/OS deployment.

webhook/workspace/handler/validate.go
```diff
@@ -0,0 +1,105 @@
...
+	components := workspace.Components
+	events := workspace.Events
+	projects := workspace.Projects
+	starterProjects := workspace.StarterProjects
+
+	var devfileErrors []string
+
+	// validate commands
+	if commands != nil && components != nil {
+		cmdErrors := devfilevalidation.ValidateCommands(commands, components)
```
sleshchenko, line 62: well, we must think carefully about what and when we should validate... The corner case I have in mind. We verify and create DevWorkspace which is valid because it used volume from the plugin...

Then DWT is updated and volume is dropped.
DevWorkspace can't be started/stopped/updated because of DWT.

So, the state we can proceed with:
- drop all validation from webhooks except ones that can't use stuff from DWT, like projects;
- do whole validation on try to start a workspace;

Think more carefully about what we can validate in webhooks... Probably we should not do flattening here at all.

JPinkney: I've switched to only validating projects, starterProjects, and events since none of those validations rely on DWT's and don't need flattening. I also find out that if you change `_store_tls_cert` to grab the `devworkspace-webhookserver-tls` secret on openshift then you don't need the additional scripts for generating the serving certs locally

sleshchenko:
e2e test fails with the following error

> Error from server (InternalError): error when creating "samples/web-terminal.yaml": Internal error occurred: failed calling webhook "mutate.devworkspace-controller.svc": Post "https://devworkspace-webhookserver.devworkspace-controller.svc:443/mutate?timeout=10s": no endpoints available for service "devworkspace-webhookserver"

sleshchenko (approved):
LGTM

PR #560, opened 2021-08-25
[release] Bump to 0.10.0 in main

sleshchenko:
Signed-off-by: Sergii Leshchenko <sleshche@redhat.com>


JPinkney (approved):

deploy/templates/components/csv/clusterserviceversion.yaml
```diff
@@ -97,4 +97,4 @@ spec:
     url: https://devfile.io
   # uncomment the below when devworkspace-operator.v0.8.0 is availble
   # replaces: devworkspace-operator.v0.8.0
-  version: 0.9.0
+  version: 0.10.0
```
JPinkney, line 100: I think we can uncomment the replaces now, we will also have to do that in the 0.9.x branch as well

sleshchenko: well, 0.9.0 is not available yet... Should `replaces: devworkspace-operator.v0.8.0` be there for time being?

JPinkney: Yeah, that makes sense

Issue #506, opened 2021-07-23
Figure out how to support using personal access tokens in project clone


amisevsk:
### Description

The project-clone init container supports automounting SSH keys to interact with private git repos. We also need to support setting a password, e.g. using [`personal access tokens`](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token).

sleshchenko:
Should be fixed by https://github.com/devfile/devworkspace-operator/pull/564

PR #564, opened 2021-08-27
feat: Automount git credentials if they exist

JPinkney:
### What does this PR do?
This PR enables the ability to automount git credentials if they exist. The way it works is fundamentally the same as [Che](https://www.eclipse.org/che/docs/che-7/end-user-guide/mounting-a-secret-as-a-file-or-an-environment-variable-into-a-workspace-container/#mounting-a-git-credential-store-into-a-workspace-container_che).
1. Users provide a secret containing their credentials with a label of `"controller.devfile.io/git-credential": "true"`
2. When a devworkspace is created the devworkspace controller looks for all secrets with that label, combines them into a new secret (so that you can have multiple credentials referenced in one credential file. See https://github.com/eclipse/che/issues/18721) and then mounts both that new combined secret and a system wide gitconfig that points to the secret so that user credentials are already picked up when cloning
3. Devworkspace now has access to those credentials and can use them for cloning inside of project clone

https://github.com/devfile/devworkspace-operator/issues/506

### Is it tested? How?

1. Create a secret on the cluster containing:
```
kind: Secret
apiVersion: v1
metadata:
  name: git-credentials-secret
  annotations:
    controller.devfile.io/mount-path: /home/theia/.git-credentials/
  labels:
    controller.devfile.io/git-credential: 'true'
data:
  credentials: >- ${your base64 credentials here}
type: Opaque
```
${your base64 credentials here} should be base64 encoded in the format of "https://{USERNAME}:{PERSONAL_ACCESS_TOKEN}@github.com". E.g. `echo -n "https://{USERNAME}:{PERSONAL_ACCESS_TOKEN}@github.com" | base64` 
2. Create a private repo on github
3. Create a devworkspace that uses your private repo in the same namespace as your secret
4. See that project cloned correctly

### PR Checklist

### DIFF
--- a/pkg/provision/workspace/automount/git-credentials.go
+++ b/pkg/provision/workspace/automount/git-credentials.go
@@ -44,9 +44,10 @@ const credentialTemplate = "[credential]\n\thelper = store --file %s\n"
 func getDevWorkspaceGitConfig(devworkspace *dw.DevWorkspace, client k8sclient.Client, scheme *runtime.Scheme) (*v1alpha1.PodAdditions, error) {
 	namespace := devworkspace.GetNamespace()
 	secrets := &corev1.SecretList{}
-	if err := client.List(context.TODO(), secrets, k8sclient.InNamespace(namespace), k8sclient.MatchingLabels{
+	err := client.List(context.TODO(), secrets, k8sclient.InNamespace(namespace), k8sclient.MatchingLabels{
 		constants.DevWorkspaceGitCredentialLabel: "true",
-	}); err != nil {
+	})
+	if err != nil {
 		return nil, err
 	}
 	var credentials []string
@@ -78,7 +79,6 @@ func getDevWorkspaceGitConfig(devworkspace *dw.DevWorkspace, client k8sclient.Cl
 		}
 		podAdditions.Volumes = append(podAdditions.Volumes, secretAdditions.Volumes...)
 		podAdditions.VolumeMounts = append(podAdditions.VolumeMounts, secretAdditions.VolumeMounts...)
-
 	}
 	return podAdditions, nil
 }
--- a/pkg/provision/workspace/automount/git-credentials.go
+++ b/pkg/provision/workspace/automount/git-credentials.go
@@ -18,7 +18,6 @@ import (
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
-	"github.com/prometheus/common/log"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


sleshchenko (commented):
I haven't done good review yet but from the first look, from the testing instructions:
```
    controller.devfile.io/mount-path: /home/theia/.git-credentials/
```
I would expect project clone to be using it but not theia.

JPinkney:
Technically the mount path can be anywhere and both project clone and theia will pick it up automatically. The only reason the mount path was:
```
 controller.devfile.io/mount-path: /home/theia/.git-credentials/
```
was so that I could test that the project clone was working and the credentials were also successfully mounted to theia to make sure that I could clone a private repo that wasn't in the devfile

sleshchenko (approved):
I haven't had a chance to test it but changes LGTM now

pkg/provision/workspace/automount/git-credentials.go
```diff
@@ -0,0 +1,253 @@
...
+	for _, secret := range secrets.Items {
+		credentials = append(credentials, string(secret.Data[gitCredentialsName]))
+		if val, ok := secret.Annotations[constants.DevWorkspaceMountPathAnnotation]; ok {
+			mountpath = val
+		}
+	}
+
+	podAdditions := &v1alpha1.PodAdditions{}
+	if len(credentials) > 0 {
+		gitCredsName := devworkspace.Status.DevWorkspaceId + "-" + gitConfigName
```
amisevsk, line 64: I'm concerned that will effectively take a fixed number of secrets and result in a variable number of secrets based on the number of workspaces -- this could cause issues if a quota is in place. 

Since all workspaces in a namespace should be fine sharing the combined credentials secret, we should be okay defining a fixed name (that's unlikely to collide with others, e.g. `devworkspace-merged-git-credentials`, if that fits in the max length) and using that.

Created https://github.com/devfile/devworkspace-operator/issues/580 for follow-up

(Sorry to bump an already-merged PR, just got back from PTO)

pkg/provision/workspace/automount/git-credentials.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, err
+	}
+
+	// Create the volume for the configmap
+	podAdditions.Volumes = append(podAdditions.Volumes, GetAutoMountVolumeWithConfigMap(configMapName))
+
+	// Create the gitconfig volume mount and set it's location as /etc/gitconfig so it's automatically picked up by git
+	gitConfigMapVolumeMount := GetAutoMountConfigMapVolumeMount(gitConfigLocation, configMapName)
+	gitConfigMapVolumeMount.SubPath = gitConfigName
+	gitConfigMapVolumeMount.ReadOnly = false
```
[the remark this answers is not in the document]
JPinkney: Not technically no, I just followed what Che is currently doing

pkg/provision/workspace/automount/git-credentials.go
sleshchenko: > effectively take a fixed number of secrets

I just wonder where this requirement is really coming from.
Like I understand limiting deployment/pods, they are heavy...
But configmap is like 1Mb at maximum right? Then we somebody should limit you to use only 10,20... )
But good point. We should save objects if it's possible and does not hurt anyone.

PR #576, opened 2021-09-06
Fix location of fake DWO CSV

sleshchenko:
### What does this PR do?
It has the following fixes and improvements for happy path test:
- Fix location of fake DWO CSV
- Fixed and Improved logs fetching;
- Adapt happy path test to be runnable locally against any OpenShift Cluster with Che and DWO deployed.

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspaces-operator-e2e, v8-devworkspace-happy-path` to trigger)
    - [ ] `v8-devworkspaces-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-devworkspace-happy-path`: DevWorkspace e2e test

### DIFF
--- a/.ci/che-happy-path.sh
+++ b/.ci/che-happy-path.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+#
+# Copyright (c) 2012-2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+set -e
+
+export ARTIFACT_DIR=${ARTIFACT_DIR:-"/tmp/che-happy-path"}
+mkdir -p "${ARTIFACT_DIR}"
+
+SCRIPT_DIR=$(dirname $(readlink -f "$0"))
+
+export CHE_NAMESPACE="eclipse-che"
+export HAPPY_PATH_POD_NAME=happy-path-che
+export HAPPY_PATH_TEST_PROJECT='https://github.com/che-samples/java-spring-petclinic/tree/devfilev2'
+
+# Create cluster-admin user inside of openshift cluster and login
+function provisionOpenShiftOAuthUser() {
+  if oc login -u che-user -p user --insecure-skip-tls-verify=false; then
+    echo "Che User already exists. Using it"
+    return 0
+  fi
+  echo "Che User does not exist. Setting up htpassw oauth for it."
+  SCRIPT_DIR=$(dirname $(readlink -f "$0"))
+  oc delete secret che-htpass-secret -n openshift-config || true
+  oc create secret generic che-htpass-secret --from-file=htpasswd="$SCRIPT_DIR/resources/users.htpasswd" -n openshift-config
+
+# The idea was to patch oauth not to break existing ones but our selenium test is not adapted to it yet.
+# So, just override ATM
+#  oc patch oauth/cluster --type=json \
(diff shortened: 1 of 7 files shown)


sleshchenko:
DevWorkspace is not deployed correctly
Screenshot_20210907_104142

sleshchenko:
now happy path test fails to deploy DWO
>   message: 'MountVolume.SetUp failed for volume "webhook-tls-certs" : secret "devworkspace-webhooks-tls" not found'

sleshchenko:
Finally:
```
INFO[2021-09-07T12:52:31Z] Step che-happy-path-che-happy-path succeeded after 17m20s. 
```

.ci/che-happy-path.sh
```diff
@@ -0,0 +1,106 @@
...
+export HAPPY_PATH_TEST_PROJECT='https://github.com/che-samples/java-spring-petclinic/tree/devfilev2'
+
+# Create cluster-admin user inside of openshift cluster and login
+function provisionOpenShiftOAuthUser() {
+  if oc login -u che-user -p user --insecure-skip-tls-verify=false; then
+    echo "Che User already exists. Using it"
+    return 0
+  fi
+  echo "Che User does not exist. Setting up htpassw oauth for it."
+  SCRIPT_DIR=$(dirname $(readlink -f "$0"))
```
amisevsk, line 29: Duplicates line 16 above

.ci/che-happy-path.sh
```diff
@@ -0,0 +1,106 @@
...
+  oc delete secret che-htpass-secret -n openshift-config || true
+  oc create secret generic che-htpass-secret --from-file=htpasswd="$SCRIPT_DIR/resources/users.htpasswd" -n openshift-config
+
+# The idea was to patch oauth not to break existing ones but our selenium test is not adapted to it yet.
+# So, just override ATM
+#  oc patch oauth/cluster --type=json \
+#    -p '[{"op": "add", "path": "/spec/identityProviders/0", "value": {"name":"che-htpasswd","mappingMethod":"claim","type":"HTPasswd","htpasswd":{"fileData":{"name":"che-htpass-secret"}}}}]'
+  oc apply -f "$SCRIPT_DIR/resources/htpasswdProvider.yaml"
```
amisevsk, line 37: Should we store the original and restore it once the run is complete? I don't know if it's easy to do.

.ci/che-happy-path.sh
```diff
@@ -0,0 +1,106 @@
...
+startHappyPathTest
+
+echo "1"
+# wait for the test to finish
+oc logs -n ${CHE_NAMESPACE} ${HAPPY_PATH_POD_NAME} -c happy-path-test -f
+echo "2"
+# just to sleep
+sleep 3
+
+echo "3"
```
amisevsk, line 95: Can we log something more useful than "1", "2", etc.?

.ci/resources/users.htpasswd
```diff
@@ -1 +1 @@
-user:{SHA}Et6pb+wgWTVmq3VpLJlJWWgzrck=
+che-user:{SHA}Et6pb+wgWTVmq3VpLJlJWWgzrck=
```
amisevsk, line 1: Is it possible to include a comment here showing what the password is? I think .htpasswd files support `#` for line comments.

amisevsk (approved):
LGTM with a few minor comments

.ci/che-happy-path.sh
sleshchenko: we may but instead I would like just Che provide its own and use it without cleaning up.
Or, even since kubeadmin now works great in Che, we can avoid creating it at all :partying_face: 
But we'll handle it separately

.ci/che-happy-path.sh
sleshchenko: LOL
It should not be there, I just tried to find exact place where it hangs. I'll revert.

sleshchenko:
I'm merging it just to get working happy path test in main branch. And will create a separate PR for fixes.

Issue #581, opened 2021-09-07
DevWorkspace operator should better reported found differences on reconcilation


sleshchenko:
Sometimes it happens that some object (deployment, serviceaccount, service, ...) is endless updating because difference is found. And in most cases it caused by some new or missed default values in desired configuration.

We need to consider adding the same as Che Operator does - report what is actually found differences, always or when running in non-production mode
https://github.com/eclipse-che/che-operator/blob/826475928cd6a3f0be14579663862358d473ea11/pkg/deploy/sync.go#L220

PR #583, opened 2021-09-07
feat: Enable config to log diff on objects when updating

amisevsk:
### What does this PR do?
Use the experimental features flag to enable additional logging when updating objects on the cluster. If experimental features are enabled and a spec object is found to be different from the cluster one, log a
diff between the two.

Fixes https://github.com/devfile/devworkspace-operator/issues/581

### Is it tested? How?
1. Start DWO with experimental features enabled
2. Create a devworkspace 
3. Edit e.g. the deployment/service/devworkspacerouting and check logs

Note that since the controller uses json-structured logging, newlines and tabs are escaped in log messages. An easy way to work around this is to pipe to `sed -e 's|\\n|\n|g; s|\\t|    |g'` -- though, note if using `make run` it's also necessary to pipe stderr to stdout, i.e. `make run 2>&1 | sed -e 's|\\n|\n|g; s|\\t|    |g'`. Another option is to copy paste a `msg` and feed it to `printf`.

### PR Checklist

### DIFF
--- a/controllers/controller/devworkspacerouting/sync_routes.go
+++ b/controllers/controller/devworkspacerouting/sync_routes.go
@@ -55,6 +55,10 @@ func (r *DevWorkspaceRoutingReconciler) syncRoutes(routing *controllerv1alpha1.D
 		if contains, idx := listContainsRouteByName(specRoute, clusterRoutes); contains {
 			clusterRoute := clusterRoutes[idx]
 			if !cmp.Equal(specRoute, clusterRoute, routeDiffOpts) {
+				r.Log.Info("Updating route")
+				if r.DebugLogging {
+					r.Log.Info(fmt.Sprintf("Diff: %s", cmp.Diff(specRoute, clusterRoute, routeDiffOpts)))
+				}
 				// Update route's spec
 				clusterRoute.Spec = specRoute.Spec
 				err := r.Update(context.TODO(), &clusterRoute)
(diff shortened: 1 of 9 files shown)


controllers/controller/devworkspacerouting/sync_routes.go
```diff
@@ -55,6 +55,10 @@ func (r *DevWorkspaceRoutingReconciler) syncRoutes(routing *controllerv1alpha1.D
 		if contains, idx := listContainsRouteByName(specRoute, clusterRoutes); contains {
 			clusterRoute := clusterRoutes[idx]
 			if !cmp.Equal(specRoute, clusterRoute, routeDiffOpts) {
+				r.Log.Info("Updating route")
```
sleshchenko, line 58: Route name is not in log context only workspace id and probably name/namespace.
So, maybe adding object name makes sense as well.

sleshchenko (approved):
LGTM
Just assume that the object name may be useful as well in logs.

JPinkney (approved):

controllers/controller/devworkspacerouting/sync_routes.go
amisevsk: Added

pkg/provision/workspace/deployment.go
```diff
@@ -121,7 +121,7 @@ func SyncDeploymentToCluster(
 	}
 
 	if clusterDeployment == nil {
-		clusterAPI.Logger.Info("Creating deployment...")
+		clusterAPI.Logger.Info("Creating deployment")
```
sleshchenko, line 124: deployment name can be useful here as well, but since ATM we have only one and one deployment, that's OK

sleshchenko (approved):

PR #584, opened 2021-09-08
Fix: improve watching for changes in ancillary resources

amisevsk:
### What does this PR do?
1. Mark the controller as `Own`ing configmaps, secrets, and serviceaccounts, in order to ensure reconciles are triggered when these objects are modified
2. Restrict DWO configmap syncs to only events involving that configmap (instead of all configmaps)

1. Editing e.g. the `workspaceID-metadata` configmap doesn't trigger a reconcile (state is restored on next devworkspace reconcile)
2. Probably harmless but frequent syncs with the `devworkspace-controller-configmap` (check by putting a `Printf` in `updateConfigMap`) -- previously, any event relating to any configmap would trigger a reconcile 

### Is it tested? How?
Start a devworkspace and try to edit the `metadata` configmap (or an owned secret/the sa)

### PR Checklist

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -114,7 +114,7 @@ func (wc *ControllerConfig) GetWorkspaceControllerSA() (string, error) {
 	return saName, nil
 }
 
-func updateConfigMap(client client.Client, obj client.Object) {
+func syncConfigmapFromCluster(client client.Client, obj client.Object) {
 	if obj.GetNamespace() != ConfigMapReference.Namespace ||
 		obj.GetName() != ConfigMapReference.Name {
 		return
@@ -185,7 +185,7 @@ func WatchControllerConfig(mgr manager.Manager) error {
 		return err
 	}
 
-	updateConfigMap(nonCachedClient, configMap)
+	syncConfigmapFromCluster(nonCachedClient, configMap)
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -242,11 +242,15 @@ func fillOpenShiftRouteSuffixIfNecessary(nonCachedClient client.Client, configMa
 func ConfigMapPredicates(mgr manager.Manager) predicate.Predicate {
 	return predicate.Funcs{
 		UpdateFunc: func(evt event.UpdateEvent) bool {
-			updateConfigMap(mgr.GetClient(), evt.ObjectNew)
+			if evt.ObjectNew.GetName() == "devworkspace-controller-configmap" {
```
sleshchenko, line 245: please, let's use it from constant

sleshchenko (approved):
Looks meaningful

pkg/config/config.go
```diff
@@ -242,11 +242,15 @@ func fillOpenShiftRouteSuffixIfNecessary(nonCachedClient client.Client, configMa
...
+				updateConfigMap(mgr.GetClient(), evt.ObjectNew)
```
sleshchenko, line 246: well, let's rename updateConfigMap because it's confusing. Like `loadConfigFromCM`

JPinkney (approved):

pkg/config/config.go
amisevsk: Good point -- we're not really watching anything there anymore.

pkg/config/config.go
amisevsk: Yeah, accidentally left the "lets see if this works" implementation here. Updated to match name/namespace from `ConfigMapReference` (which can be overridden via env var)

JPinkney (approved):

sleshchenko (approved):

Issue #562, opened 2021-08-25
None of make target works if there is no active cluster in kubeconfig


sleshchenko:
it was not obvious and I spent sometime to figure out why prerelease script hangs on make generate_olm, it turns out that I did oc logout previously. So, we should make `Makefile` working like `help`, `docker`, `fmt` targets even is there is no active k8s/OS cluster.

PR #587, opened 2021-09-08
Workaround issue where makefile can't run commands after oc logout

amisevsk:
### What does this PR do?
Closes `stdin` for all top-level kubectl/oc commands (i.e. ones that are run before any rules). 

Executing `oc logout` drops user info from kubeconfig. This does not pose a problem for `oc` as it understands what it has done, but kubectl will prompt for login info when executing commands. The order in which we evaluate `$K8S_CLI` makes it so that kubectl is used if available.

This causes any make rule to hang if the kubectl is installed, as deploy.mk is imported and contains a check for `kubectl api-resources`. Since this call will try to read a password from stdin, the call will never return and so the comparison will hang silently.

This PR also adds a warning if the current context does not appear to be authenticated, to hopefully make it clearer what's going wrong if you e.g. `make install`

Fixes https://github.com/devfile/devworkspace-operator/issues/562

### Is it tested? How?
1. Login to OpenShift cluster
2. Run `oc logout`
3. Try to run make rules (e.g. `make fmt`)
4. For good measure, make rules that require cluster interaction (e.g. deploy) should fail 

### PR Checklist

### DIFF
--- a/build/make/deploy.mk
+++ b/build/make/deploy.mk
@@ -11,12 +11,14 @@ K8S_CLI := kubectl <&-
 # Check if logged in and warn if not
 ifeq ($(findstring Please enter Username,$(shell kubectl auth can-i '*' '*' <&- 2>&1)),Please enter Username)
 $(info Warning: current context is not authenticated with a cluster)
+LOGGED_IN="false"
 endif
 else
 K8S_CLI := oc <&-
 # Check if logged in and warn if not
 ifeq ($(findstring Forbidden,$(shell oc whoami 2>&1)),Forbidden)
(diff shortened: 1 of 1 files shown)


sleshchenko (approved):
Thanks makefile/shell master! I assume I would kill week figuring out that :-D

Tested the following
```bash
$ make install
Warning: current context is not authenticated with a cluster
Current env vars:
....
Please enter Username: error: read /dev/stdin: invalid argument
build/make/deploy.mk:71: recipe for target 'install' failed
make: *** [install] Error 1
```
```bash
$ make fmt
Warning: current context is not authenticated with a cluster
```

PR #589, opened 2021-09-10
docs: Adding instructinon for switching to the right namespace

ibuziuk:
### What does this PR do?
docs: Adding instruction for switching to the right project

### PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ To see all rules supported by the makefile, run `make help`
 ### Run controller locally
 ```bash
 make install
+oc project devworkspace-controller
 oc patch deployment/devworkspace-controller-manager --patch "{\"spec\":{\"replicas\":0}}"
 make run
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +107,7 @@ To see all rules supported by the makefile, run `make help`
 ### Run controller locally
 ```bash
 make install
+oc project devworkspace-controller
```
sleshchenko, line 110: It would be better to make like the following:
```
export NAMESPACE="devworkspace-controller"
make install
oc patch deployment/devworkspace-controller-manager --patch "{\"spec\":{\"replicas\":0}}" -n $NAMESPACE
make run
```
otherwise it may happen that DWO is deployed in one namespace, but run try to patch in another.

sleshchenko (approved):
Thanks for providing a fix. Left one comment and from our end ideally it should be a part of `make run`, not to bother people with copying/pasting commands.

@amisevsk @JPinkney WDYT?

amisevsk (approved):

JPinkney (approved):

PR #592, opened 2021-09-10
doc: fixing a typo in the grafana port-forwarding instructions

ibuziuk:
### What does this PR do?
fixing a typo in the grafana port-forwarding instructions

N/A

### PR Checklist

### DIFF
--- a/doc/grafana/README.md
+++ b/doc/grafana/README.md
@@ -23,7 +23,7 @@ To quickly test metrics locally (without installing Prometheus Grafana in the cl
     ```
 3. In another terminal, use `kubectl port-forward` to expose the controller's service locally
     ```bash
-    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &&
+    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &
     kubectl port-forward service/devworkspace-webhookserver 9443:9443
     ```
     We can check that the token works and metrics are accessible:
--- a/doc/grafana/README.md
+++ b/doc/grafana/README.md
@@ -23,7 +23,7 @@ To quickly test metrics locally (without installing Prometheus Grafana in the cl
     ```
 3. In another terminal, use `kubectl port-forward` to expose the controller's service locally
     ```bash
-    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &;
+    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &&
     kubectl port-forward service/devworkspace-webhookserver 9443:9443
     ```
     We can check that the token works and metrics are accessible:


doc/grafana/README.md
```diff
@@ -23,7 +23,7 @@ To quickly test metrics locally (without installing Prometheus Grafana in the cl
     ```
 3. In another terminal, use `kubectl port-forward` to expose the controller's service locally
     ```bash
-    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &;
+    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &&
```
sleshchenko, line 26: well, I'm not sure if it's typo.
`command1 && command2`: in my understanding command2 won't be launched until command1 ends, and it's the case when command1 is expected to run until it's interrupted.
But I let @amisevsk provide the truth here )

ibuziuk:
```
[ibuziuk@fedora ~]$  kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &;
bash: syntax error near unexpected token `;'

```

sleshchenko: @ibuziuk which shell do you use?

it's how it works for me on zsh.
Screenshot_20210910_164717
Screenshot_20210910_164705
Note there is difference ^

ibuziuk: yeah, hmm 
I'm using vanilla bash 
```
[ibuziuk@fedora ~]$ echo "$SHELL" 
/bin/bash
[ibuziuk@fedora ~]$ ps -p $$
  PID TTY          TIME CMD
25736 pts/2    00:00:00 bash
```

amisevsk: I get the feeling this is a quirk in bash that was smoothed over in `zsh` without me noticing. Serhii is correct that we don't want to use `&&` since `kubectl port-forward` does not terminate, so I think a plain `&` is the right option (we want to run the first command in the background and then run the second.

Some background: the semicolon is there from muscle memory (e.g. `sleep 2; sleep 1` to run two commands in a one-liner) -- the port forward command was one line originally and I split it for documentation without removing the `;`. I originally had both port-forward commands together, separated by a semicolon, then realized I needed the first to run in the background so I added the `&`.

In bash and zsh,
```bash
(sleep 2; echo hi) & (sleep 1; echo bye)
```
works as expected, whereas
```bash
(sleep 2; echo hi) &; (sleep 1; echo bye)
```
only works in zsh.

sleshchenko:
```suggestion
    kubectl port-forward service/devworkspace-controller-metrics 8443:8443 &
```

ibuziuk:
Thanks for review
PR has been updated

amisevsk (approved):

sleshchenko (approved):
Thanks for contribution!

Issue #586, opened 2021-09-08
Volumes from DevWorkspace and DevWorkspaceTemplates should be merged


sleshchenko:
The context:
1. In Che flow, user's devfile.yaml describe only their application related component, and everything related to tooling and IDE come from plugins;
2. DevWorkspace Operator powers that flow with DevWorkspaces(aka DW) (for devfile) and DevWorkspaceTemplates(aka DWT)(for plugins), and then use flattening devfile logic from devfile spec;
3. Flattening devfile logic requires for each devfile item(DW, DWT) (unless it's parent logic which we don't consider) provision objects(component, commands, projects) with unique names.

So, let's consider two cases in details:
1. [Upstream] User has java-maven application that has devfile with volume for maven repository, and java plugin that then just mounts that repo in its container;
2. [Downstream] Java plugin is designed in the common way to work for both of maven and gradle, so it mounts maven and gradle repos volumes, which requires from users to declare both volumes in their devfiles, but it does not make much sense for application to use both build tools at the same time.

So, it's proposed to allow plugin define their volumes and to make it working add DevWorkspace specific logic for processing DevWorkspaceTemplates volumes, which will work in the following way:
1. Process each DevWorkspaceTemplate volumes:
  1.1. Check if it's in DevWorkspace(user's devfile):
     + it's there - use one from DevWorkspace;
     - it's **not** there - use one from DevWorkspace Template;

Downstream issue https://issues.redhat.com/browse/CRW-2253

PR #593, opened 2021-09-10
Merge duplicate volume components from parent/plugins

amisevsk:
### What does this PR do?
Adds merging functionality to combine volumes that are duplicated in plugins and parents. This is done by removing duplicate volume components from parents/plugins if they are defined earlier in the flattened devworkspace, in the order devworkspace <- parents <- plugins (i.e. if a volume is defined in the devworkspace, duplicates in the parent and plugins are removed). If a volume is defined in multiple plugins, the first processed instance is kept and the rest are dropped.

Duplicated volumes are merged so that the larger size is used, and so that a persistent volume overrides an ephemeral one. This means that if a plugin defines e.g. a `.m2` volume with size `1Gi`, it will overwrite a main-devfile volume with size `512Gi`.  This also means defining volume sizes in plugins can modify main-devfile volumes, which may be undesirable. Generally, a good practice is probably to define plugin volumes as ephemeral unless otherwise necessary, and let the main devworkspace make them persistent if needed.

Closes https://github.com/devfile/devworkspace-operator/issues/586

### Is it tested? How?
Test cases are included in PR.

### PR Checklist

### DIFF
--- a/pkg/library/flatten/flatten.go
+++ b/pkg/library/flatten/flatten.go
@@ -22,14 +22,14 @@ import (
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/api/v2/pkg/utils/overriding"
 	"github.com/devfile/api/v2/pkg/validation/variables"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	"github.com/devfile/devworkspace-operator/pkg/library/annotate"
 	registry "github.com/devfile/devworkspace-operator/pkg/library/flatten/internal_registry"
 	"github.com/devfile/devworkspace-operator/pkg/library/flatten/network"
 	"github.com/devfile/devworkspace-operator/pkg/library/flatten/web_terminal"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/resource"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -123,16 +123,13 @@ func recursiveResolve(workspace *dw.DevWorkspaceTemplateSpec, tooling ResolverTo
 		}
 	}
 
-	if err := mergeVolumeComponents(resolvedContent, resolvedParent, pluginSpecContents...); err != nil {
-		return nil, fmt.Errorf("failed to merge DevWorkspace volumes: %w", err)
-	}
-	resolvedContent, err := overriding.MergeDevWorkspaceTemplateSpec(resolvedContent, resolvedParent, pluginSpecContents...)
+	mergedSpecContent, err := mergeDevWorkspaceElements(resolvedContent, resolvedParent, pluginSpecContents...)
 	if err != nil {
-		return nil, fmt.Errorf("failed to merge DevWorkspace parents/plugins: %w", err)
+		return nil, err
 	}
 
 	return &dw.DevWorkspaceTemplateSpec{
-		DevWorkspaceTemplateSpecContent: *resolvedContent,
+		DevWorkspaceTemplateSpecContent: *mergedSpecContent,
 	}, nil
 }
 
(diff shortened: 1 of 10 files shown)


pkg/library/flatten/flatten.go
```diff
@@ -127,6 +123,9 @@ func recursiveResolve(workspace *dw.DevWorkspaceTemplateSpec, tooling ResolverTo
 		}
 	}
 
+	if err := mergeVolumeComponents(resolvedContent, resolvedParent, pluginSpecContents...); err != nil {
```
sleshchenko, line 126: Maybe it makes sense to move merging logic to a separate file?

here we just call abstract merge, and inside merge volumes and merging with devfile/api.

sleshchenko (commented):
Merging rules looks meaningful to me. So, not approved just since it's WIP

pkg/library/flatten/flatten.go
amisevsk: Good idea, I've moved this logic into `merge.go` with the more-abstract `mergeDevWorkspaceElements`

pkg/library/flatten/flatten.go
```diff
@@ -127,6 +123,9 @@ func recursiveResolve(workspace *dw.DevWorkspaceTemplateSpec, tooling ResolverTo
 		}
 	}
 
+	if err := mergeDevWorkspaceElements(resolvedContent, resolvedParent, pluginSpecContents...); err != nil {
+		return nil, fmt.Errorf("failed to merge DevWorkspace volumes: %w", err)
+	}
 	resolvedContent, err := overriding.MergeDevWorkspaceTemplateSpec(resolvedContent, resolvedParent, pluginSpecContents...)
```
sleshchenko, line 129: maybe let's move it there into pkg/library/flatten/merge.go as well?

sleshchenko (approved):
I haven't tested but tests suggests it works )

Please create an issue for plugins team to define volumes on Devfile level of plugins.

pkg/library/flatten/flatten.go
amisevsk: :+1: Good point. Squashed this change into the latest commit, PTAL

pkg/library/flatten/flatten.go
```diff
@@ -127,13 +123,13 @@ func recursiveResolve(workspace *dw.DevWorkspaceTemplateSpec, tooling ResolverTo
 		}
 	}
 
-	resolvedContent, err := overriding.MergeDevWorkspaceTemplateSpec(resolvedContent, resolvedParent, pluginSpecContents...)
+	mergedSpecContent, err := mergeDevWorkspaceElements(resolvedContent, resolvedParent, pluginSpecContents...)
 	if err != nil {
-		return nil, fmt.Errorf("failed to merge DevWorkspace parents/plugins: %w", err)
+		return nil, fmt.Errorf("failed to merge DevWorkspace volumes: %w", err)
```
sleshchenko, line 128:
```suggestion
		return nil, fmt.Errorf("failed to resolve DevWorkspace: %w", err)
```

sleshchenko: the details are already in 

```go
	if err := mergeVolumeComponents(main, parent, plugins...); err != nil {
		return nil, fmt.Errorf("failed to merge DevWorkspace volumes: %w", err)
	}
	mergedDevWorkspaceTemplateSpec, err := overriding.MergeDevWorkspaceTemplateSpec(main, parent, plugins...)
	if err != nil {
		return nil, fmt.Errorf("failed to merge DevWorkspace parents/plugins: %w", err)
	}
	return mergedDevWorkspaceTemplateSpec, nil
```

amisevsk: Formatting errors continues to be annoying :sweat_smile:

sleshchenko: yeah, it will never ends.

sleshchenko:
from e2e tests

> 2021/09/14 19:56:32 Now current status of developer workspace is: Failed. Message: Detected unrecoverable event FailedMount: MountVolume.SetUp failed for volume "devworkspace-serving-cert-workspace3ff4357331c7441e-service" : secret "workspace3ff4357331c7441e-service" not found

¯\_(ツ)_/¯

sleshchenko:
Ran tests against RHPDS OpenShift 4.8 and they works.
Let's give it a second try and if no luck - probably `FailedMount` is not something we should use as unrecoverable event.

sleshchenko:
e2e path seems to pass on https://github.com/devfile/devworkspace-operator/pull/596.

So, trying one more time

amisevsk:
I have _no_ idea why e2e is consistently failing here.

amisevsk:
Lets see if the mountVolumes issue is transient I guess (still don't understand why only this PR is affected, and I also can't reproduce locally.)

/test v8-devworkspace-operator-e2e

amisevsk:
The last failure had a different message:
>  2021/09/16 17:05:42 Now current status of developer workspace is: Failed. Message: Detected unrecoverable event FailedMount: MountVolume.SetUp failed for volume "devworkspace-serving-cert-workspacedaf72da98cff4016-service" : failed to sync secret cache: timed out waiting for the condition

sleshchenko:
Tests passed! :mage_man: :tada:

PR #597, opened 2021-09-16
feat: Adding the 'reason' label to the 'devworkspace_fail_total' counter Prometheus metricter Prometheus metric

ibuziuk:
### What does this PR do?
- Adding the 'reason' label to the 'devworkspace_fail_total' counter Prometheus metric
- Removing 'metricSourceLabel' / 'metricsRoutingClassLabel' to improve the cardinality 

Related to https://github.com/eclipse/che/issues/20261

### Is it tested? How?
Locally https://github.com/devfile/devworkspace-operator/tree/main/doc/grafana#testing-devworkspace-operator-metrics-locally

### PR Checklist


controllers/workspace/metrics/metrics.go
```diff
@@ -53,8 +65,7 @@ var (
 			Help:      "Number of failed DevWorkspaces",
 		},
 		[]string{
-			metricSourceLabel,
```
amisevsk, line 56: It might still be useful to have `source` here (to e.g. distinguish between web terminal and Che) but I'm okay removing it for now.

amisevsk (approved):
LGTM

controllers/workspace/metrics/metrics.go
ibuziuk: added back :+1:

JPinkney (approved):

controllers/workspace/metrics/metrics.go
```diff
@@ -21,14 +21,26 @@ const (
...
 	metricSourceLabel        = "source"
 	metricsRoutingClassLabel = "routingclass"
+	metricsReasonLabel       = "reason"
+
+	reasonInfrastructureFailure = "infrastructure_failure"
+	reasonUnknown               = "unknown"
 )
 
+var infrastructureFailures = []string{
+	"CrashLoopBackOff",
```
sleshchenko, line 31: Well, it's not a necessary infrastructure error. I would say in most cases it's caused by Che bug or a user's mistake.

controllers/workspace/metrics/metrics.go
```diff
@@ -21,14 +21,26 @@ const (
...
+	"ImagePullBackOff",
```
sleshchenko, line 32: The same is here as above like typically we face this issue when CRW bits are broken and plugin registry references non-existing plugin tag. Or also user specifies non-existing image.

So, I think we should better describe what is infrastructure error (like something user can do nothing by configuring devworkspace), and what exactly we expect from each of failures.

ibuziuk: @l0rd WDYT? how should we handle this situation? I have raised similar concerns in https://github.com/eclipse/che/issues/20261#issuecomment-916152162

sleshchenko: > I have raised similar concerns in eclipse/che#20261 (comment)

yeah, sorry for losing that request and did not provide the answer from my end.

ibuziuk: no worries, the easiest solution would be to have a reason per failed event e.g. `reason=image_pull_back_off`, but I do not think we can afford this from the cardinality perspective

l0rd: ImagePullBackOff is a request problem. Even if that's a registry misconfiguration that's still a request problem.

ibuziuk: so, you mean it should be under `bad_request` reason, right? 
in this case should we put`"CrashLoopBackOff", "ImagePullBackOff"  under `bad_request`, and `CreateContainerError", "RunContainerError", "ErrImagePull" under `infrastructure_failure`?

controllers/workspace/metrics/update.go
```diff
@@ -59,6 +61,32 @@ func incrementMetricForWorkspace(metric *prometheus.CounterVec, wksp *dw.DevWork
...
+func incrementMetricForWorkspaceFailure(metric *prometheus.CounterVec, wksp *dw.DevWorkspace, log logr.Logger) {
+	sourceLabel := wksp.Labels[workspaceSourceLabel]
+	if sourceLabel == "" {
+		sourceLabel = "unknown"
+	}
+	reason := reasonUnknown
+	for _, failure := range infrastructureFailures {
+		if strings.Contains(wksp.Status.Message, failure) {
+			reason = reasonInfrastructureFailure
+			break
```
sleshchenko, line 73: Just matter of taste, so just proposal without strong requirement to apply:
```go
func incrementMetricForWorkspaceFailure(metric *prometheus.CounterVec, wksp *dw.DevWorkspace, log logr.Logger) {
	sourceLabel := wksp.Labels[workspaceSourceLabel]
	if sourceLabel == "" {
		sourceLabel = "unknown"
	}
	reason := determinateFailureReason(wksp)

	ctr, err := metric.GetMetricWith(map[string]string{metricSourceLabel: sourceLabel, metricsReasonLabel: reason})
	if err != nil {
		log.Error(err, "Failed to increment metric")
	}
	ctr.Inc()
}

// determinateFailureReason scans devworkspace .status.message and try to match
// it to the know category. Unknown is return when no one is matched
func determinateFailureReason(wksp *dw.DevWorkspace) string {
	for _, failure := range infrastructureFailures {
		if strings.Contains(wksp.Status.Message, failure) {
			return reasonInfrastructureFailure
		}
	}
	for _, failure := range badRequestFailures {
		if strings.Contains(wksp.Status.Message, failure) {
			return reasonBadRequest
		}
	}
	return reasonUnknown
}
```

ibuziuk:
@sleshchenko thanks for the meticulous review. Suggestions have been applied

Issue #550, opened 2021-08-11
Make unrecoverable events in controller configurable


amisevsk:
### Description

As of PR https://github.com/devfile/devworkspace-operator/pull/549, there is a list of events that cause the controller to mark DevWorkspaces as failed. However, some of these events [may be temporary](https://github.com/devfile/devworkspace-operator/pull/549#discussion_r686757052) states that can be safely ignored (e.g. a pod fails to schedule for a brief period before succeeding).

We should consider making the list of unrecoverable events customizable in order to avoid unnecessary workspace failures. 

Note however that whether these events should be ignored depends on *how* they occur (i.e. if scheduling fails for ~3-5 seconds it's fine, but if it's longer workspaces could hang for a long time, in which case it makes sense to fail the DevWorkspace).

Issue #191, opened 2020-11-02
Remove configmap from operator


amisevsk:
### Is your enhancement related to a problem? Please describe.
Currently, the devworkspace operator relies on a configmap to define a few options (default routingClass, etc.). However, operators as managed by OLM don't come with configmaps, and instead we create a configmap on cluster during startup as a way of manually configuring the operator after the fact. 

### Describe the solution you'd like
We should remove configmap functionality as its generally not used for operators. Instead configuration should be defined in a standard way, e.g. by [OLM descriptors](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#7-number)

### Additional context
This issue was created for a [TODO](https://github.com/devfile/devworkspace-operator/pull/187/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R314) added in PR #187

davidfestal:
> We should remove configmap functionality as its generally not used for operators. Instead configuration should be defined in a standard way, e.g. by OLM descriptors

I'm not sure I understand your point here. Aren't OLM descriptors used to specify the UI for presenting / changing Custom Resources in the OperatorHub console. I don't see how you plan to use them to set controller settings...

amisevsk:
> Aren't OLM descriptors used to specify the UI for presenting / changing Custom Resources in the OperatorHub console. I don't see how you plan to use them to set controller settings...

Maybe I misinterpreted the purpose of OLM descriptors, but even so, configuring an operator via a configmap does not seem to be a standard flow (happy to be wrong, though).

Issue #577, opened 2021-09-07
Documentation for DevWorkpace Operator specific features


sleshchenko:
DevWorkspace provides a few features on the top of devfile:
- mounting secrets/config maps into devworkspace;
- configuring image pull secrets;
- configuring git https tokens;
- configuring devworkspace on namespace level;
- annotate the devworkspace to debug startup;

Features list is growing on the way we're moving forward integration with Che, and the only place where these features are described - go code.
It's needed to start DevWorkspace Operator Documentation to describe such features, which will be the source of truth for Che and other dependent projects.

From docs team:
>  a documentation built using Antora and following modular docs guidelines, like what we have for devfile or che-docs would be perfect and much appreciated in the eventuality of a future downstream reuse.
> that a new docs project would be best to start using the toolchain we  already have in place for Che and devfile docs

JPinkney:
Initial devworkspace configuration documentation branch is: https://github.com/devfile/devworkspace-operator-docs/tree/configuration-documentation

PR #598, opened 2021-09-16
Move operator configuration into a custom resource and enable ignoring unrecoverable events

amisevsk:
### What does this PR do?
* Move Operator configuration into a custom resource (`DevWorkspaceOperatorConfiguration`, shortname `dwoc`), hopefully making configuration easier
  * Default config is stored internally and overridden by the CR, so deleting entries/the CR restores default values. This means that the operator works fine (on OpenShift, at least) without any additional resources.
  * On OpenShift, clusterRoutingSuffix is detected automatically, must be supplied in Kubernetes
  * Since config is a CR, it cannot be part of `combined.yaml` anymore (rejected as unrecognized). I've moved creation of a config from env vars to a separate step in `make install`
* Add support for ignoring unrecoverable events

Currently, the config CR's name is hard-coded to `devworkspace-operator-config`. It might be worth changing that since it could be unclear (you have to create a CR with this *specific* name)

One potential task to complete before merging is to implement a sort of migration process that converts an existing configmap into a config CR, to allow seamless migration from previous installs.

Hopefully the commit history is legible; the change is fairly wide-ranging due to how frequently config is used in DWO.

Closes https://github.com/devfile/devworkspace-operator/issues/550
Closes https://github.com/devfile/devworkspace-operator/issues/191
Related to https://github.com/devfile/devworkspace-operator/issues/577

### Is it tested? How?
1. Deploy DWO
2. Try to start DevWorkspace that will encounter unrecoverable event
    ```yaml
    kind: DevWorkspace
    metadata:
      name: theia-next
    spec:
      started: true
      template:
        components:
          - name: theia
            plugin:
              components:
                - name: theia-ide
                  container:
                    memoryRequest: 100Gi
                    memoryLimit: 100Gi
    EOF
    ```
3. Workspace fails to start; `kubectl delete dw theia-next
4. Apply a new configuration to ignore `FailedScheduling` events:
    ```yaml
    metadata:
    config:
      workspace:
        ignoredUnrecoverableEvents:
          - FailedScheduling
    EOF
    ```
5. Re-apply workspace from step 2 -- startup should hang but not fail.

Bonus:
dwo-config-explain

### PR Checklist


amisevsk:
I'm curious is we can reproduce the e2e test failure from https://github.com/devfile/devworkspace-operator/pull/593 in this PR

sleshchenko:
[go tests](https://github.com/devfile/devworkspace-operator/runs/3635873666) fail with 
```
vet: pkg/library/flatten/flatten_test.go:237:11: SetupControllerCfg not declared by package testutil
make: *** [Makefile:171: vet] Error 2
Error: Process completed with exit code 2.
```

pkg/config/sync.go
```diff
@@ -104,6 +115,37 @@ func updatePublicConfig() {
...
+		},
+		Spec: routeV1.RouteSpec{
+			To: routeV1.RouteTargetReference{
+				Kind: "Service",
+				Name: "devworkspace-controller-test-route",
+			},
+		},
+	}
+
+	err := client.Create(context.TODO(), testRoute)
```
sleshchenko, line 140: on DevSanbox stage I faced the following failure:
Screenshot_20210917_150634
Screenshot_20210917_150649
So, I had to remove it manually.

:man_facepalming: for myself not to capture whole error message but it was about: `can't create test route since it already exists`. Not sure why terminal entered into that state but maybe we need first to check if it exists, and use if it does.

sleshchenko: I left that comment and in the next commit I see it's fixed. @amisevsk this kudo for you :+1: for addressing feedback before it's even flushed :-D
So, please just resolve this thread )

pkg/config/defaults.go
```diff
@@ -0,0 +1,27 @@
...
+//
+
+package config
+
+import "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+
+// DefaultConfig represents the default configuration for the DevWorkspace Operator.
+var DefaultConfig = &v1alpha1.OperatorConfiguration{
+	Routing: &v1alpha1.RoutingConfig{
+		DefaultRoutingClass: "basic",
```
sleshchenko, line 20:
```suggestion
		DefaultRoutingClass: "basic",
		ClusterHostSuffix: "", // is auto discovered when running on OpenShift. Must be defined by CR on Kubernetes.
```
if you think it makes sense

pkg/config/sync.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func restoreDefaultConfig() {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+	internalConfig = DefaultConfig.DeepCopy()
+	updatePublicConfig()
+}
+
+func updatePublicConfig() {
```
sleshchenko, line 117: Maybe it makes sense to report in logs when changes are detected and Config is reloaded? Probably also print the result how it's merged with default

pkg/config/sync.go
```diff
@@ -0,0 +1,207 @@
...
+	Routing = internalConfig.Routing.DeepCopy()
+	Workspace = internalConfig.Workspace.DeepCopy()
+}
+
+func fillOpenShiftRouteSuffix(client crclient.Client) error {
+	if !infrastructure.IsOpenShift() {
```
sleshchenko, line 123: This check is done twice:
```go
...
   if internalConfig.Routing.ClusterHostSuffix == "" && infrastructure.IsOpenShift() {
		if err := fillOpenShiftRouteSuffix(client); err != nil {
			return err
		}
		updatePublicConfig()
	}
...

func fillOpenShiftRouteSuffix(client crclient.Client) error {
	if !infrastructure.IsOpenShift() {	
```
This suggests something is wrong.

Maybe it makes sense to refactor it in the following way:
```go
    if internalConfig.Routing.ClusterHostSuffix == "" && infrastructure.IsOpenShift() {
        routeSuffix, err := discoverRouteSuffix(client);
		err != nil {
			return err
		}
		internalConfig.Routing.ClusterHostSuffix = clusterHostSuffix
		// Set routing suffix in default config as well to ensure value is persisted across config changes
		DefaultConfig.Routing.ClusterHostSuffix = clusterHostSuffix
		updatePublicConfig()
	}
```

sleshchenko (commented):
Good job, I haven't tested yet but flushing code review result.

pkg/config/cmd_terminal.go
```diff
@@ -1,69 +0,0 @@
...
-)
-
-const (
-	// property name for value with yaml for default dockerimage component
-	// that should be provisioned if devfile DOES have redhat-developer/web-terminal plugin
-	// and DOES NOT have any dockerimage component
-	defaultTerminalDockerimageProperty = "devworkspace.default_dockerimage.redhat-developer.web-terminal"
-)
-
-func (wc *ControllerConfig) GetDefaultTerminalDockerimage() (*dw.Component, error) {
```
sleshchenko, line 32: I think e2e test needs to be adapted after these changes:
```
------------------------------
• Failure [1.527 seconds]
[Create OpenShift Web Terminal Workspace]
/home/sleshche/projects/devworkspace-operator/test/e2e/pkg/tests/devworkspaces_tests.go:25
  Check that pod creator can execute a command in the container [It]
  /home/sleshche/projects/devworkspace-operator/test/e2e/pkg/tests/devworkspaces_tests.go:56

  Cannot execute command in the devworkspace container. Error: `exit status 1`. Exec output: `Error from server (BadRequest): container dev is not valid for pod workspaceda33612faf9843c6-5c775dc868-26vs7         
```

Currently, it references plugin from internal registry
https://github.com/devfile/devworkspace-operator/blob/main/samples/web-terminal.yaml#L16

I think the way we should rework it - make it not terminal specific test but restricted access. Then we can pretty any container component and `controller.devfile.io/restricted-access: "true"` annotation

main.go
```diff
@@ -183,21 +183,11 @@ func main() {
 }
 
 func setupControllerConfig(mgr ctrl.Manager) error {
-	operatorNamespace, err := infrastructure.GetOperatorNamespace()
-	if err == nil {
-		config.ConfigMapReference.Namespace = operatorNamespace
-	} else {
-		config.ConfigMapReference.Namespace = os.Getenv(infrastructure.WatchNamespaceEnvVar)
-	}
-	err = config.WatchControllerConfig(mgr)
```
sleshchenko, line 192: Che Operator needs to be adapted to these changes not to break Che + DevWorkspace integration on Kubernetes: 

https://github.com/eclipse-che/che-operator/blob/main/pkg/deploy/dev-workspace/dev_workspace.go#L332

please note that Operator CR must be already available when Che Operator does this logic

deploy/default-config.yaml
```diff
@@ -0,0 +1,10 @@
+kind: DevWorkspaceOperatorConfig
+apiVersion: controller.devfile.io/v1alpha1
+metadata:
+  name: devworkspace-operator-config
```
sleshchenko, line 4:
```suggestion
  name: devworkspace-operator-config
  namespace: ${NAMESPACE}
```

sleshchenko: Also, I got my routing suffix wrongly assigned
Screenshot_20210920_144601

deploy/default-config.yaml
```diff
@@ -0,0 +1,10 @@
...
+config:
+  routing:
+    clusterHostSuffix: ${ROUTING_SUFFIX}
```
sleshchenko, line 7: This actually shows another issue we have: it would be better to report: `Waiting for main endpoint be ready`, health check fails:
Screenshot_20210920_145448

it makes sense to create a separate issue for that. I'll do

sleshchenko: Probably the background for one more issue:
After clusterSuffix is cleaned up, it's not propagated to existing workspaces, even after restart. Removing DevWorkspace Routing helps but it's not desired behaviour I think.

PROJECT
```diff
@@ -1,30 +1,36 @@
...
--
-  controller: false
-  domain: devfile.io
+- domain: devfile.io
   group: workspace
   kind: DevWorkspaceTemplate
   path: github.com/devfile/devworkspace-operator/apis/workspace/v1alpha1
   version: v1alpha2
+- api:
+    crdVersion: v1beta1
```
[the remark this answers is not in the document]
amisevsk: Huh, weird. I assume it's because I used Operator SDK `v1.8.0` to bootstrap the new CRD. Removed.

pkg/config/sync.go
amisevsk: I actually had this for debugging and removed it before opening the PR :D 

I'll re-add

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+type WorkspaceConfig struct {
+	// ImagePullPolicy defines the imagePullPolicy used for containers in a DevWorkspace
+	// For additional information, see Kubernetes documentation for imagePullPolicy. If
+	// not specified, the default value of "Always" is used.
+	// +kubebuilder:validation:Enum=IfNotPresent;Always;Never
+	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
+	// PVCName defines the name used for the persistent volume claim created
+	// to support workspace storage when the 'common' storage class is used.
+	// If not specified, the default value of claim-devworkspace is used.
```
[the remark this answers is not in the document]
amisevsk: :+1: done

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// +kubebuilder:validation:MaxLength=63
+	PVCName string `json:"pvcName,omitempty"`
+	// IdleTimeout determines how long a workspace should sit idle before being
+	// automatically scaled down. Proper functionality of this configuration property
+	// requires support in the workspace being started. If not specified, the default
+	// value of "15m" is used.
+	IdleTimeout string `json:"idleTimeout,omitempty"`
+	// StorageClassName defines and optional storageClass to use for persistent
+	// volume claims created to support DevWorkspaces
+	StorageClassName *string `json:"storageClassName,omitempty"`
```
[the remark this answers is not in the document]
amisevsk: Moved! :+1:

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// IgnoredUnrecoverableEvents defines a list of Kubernetes event names that should
+	// be ignored when deciding to fail a DevWorkspace startup. This option should be used
+	// if a transient cluster issue is triggering false-positives (for example, if
+	// the cluster occasionally encounters FailedScheduling events). Events listed
+	// here will not trigger DevWorkspace failures.
```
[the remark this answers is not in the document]
amisevsk: I would worry that it's another piece of documentation that has to be maintained (e.g. would we have updated this when we removed the mountvolume setup error?)

Maybe it makes sense to have examples, e.g.
```
For example, to disable failing DevWorkspace startup when a pod cannot be scheduled on the cluster, add `FailedScheduling` to this list
```
or something?

pkg/config/env.go
```diff
@@ -68,3 +69,11 @@ func GetResourceQuantityFromEnvVar(env string) (*resource.Quantity, error) {
 	}
 	return &quantity, nil
 }
+
+func GetWorkspaceControllerSA() (string, error) {
+	saName := os.Getenv(constants.ControllerServiceAccountNameEnvVar)
+	if saName == "" {
+		return "", fmt.Errorf("could not get service account name")
```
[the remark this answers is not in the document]
amisevsk: Fixed (to match the other ones, message is `env var %s is unset`)

controllers/controller/devworkspacerouting/solvers/basic_solver.go
```diff
@@ -56,19 +56,19 @@ func (s *BasicSolver) Finalize(*controllerv1alpha1.DevWorkspaceRouting) error {
 func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.DevWorkspaceRouting, workspaceMeta DevWorkspaceMetadata) (RoutingObjects, error) {
 	routingObjects := RoutingObjects{}
 
-	routingSuffix := config.ControllerCfg.GetProperty(config.RoutingSuffix)
-	if routingSuffix == nil {
-		return routingObjects, errors.New(config.RoutingSuffix + " must be set for basic routing")
+	routingSuffix := config.Routing.ClusterHostSuffix
+	if routingSuffix == "" {
+		return routingObjects, errors.New("config setting clusterHostSuffix must be set for basic routing")
```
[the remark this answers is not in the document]
amisevsk: Applied

pkg/config/sync.go
amisevsk: There were a good number of edge cases in OpenShift handling, hopefully it's stable now.

pkg/config/defaults.go
amisevsk: Makes a lot of sense, thanks!

pkg/config/sync.go
amisevsk: Added. Currently, we log non-default values only (otherwise it can be very long). 

One question: should we be logging clusterHostSuffix? Someone using DWO might want to suppress the URL of their cluster in logs, especially if sharing publicly in e.g. a bug report.

pkg/config/sync.go
amisevsk: Reworked as suggested. I'm only checking `IsOpenShift()` once now, and `discoverRouteSuffix` returns `"", nil` when on Kubernetes.

deploy/default-config.yaml
amisevsk: Ahh yes, previously, on OpenShift we would always overwrite `clusterHostSuffix` from a route, even if the configmap contained a different value. With the current changes, we only create the test route when `clusterHostSuffix` is unset (which doesn't happen with the makefile, since it creates a config).

I'll think about how best to handle this -- should we just assume OpenShift users want to use the default fill for `route.spec.host`?

pkg/config/cmd_terminal.go
amisevsk: Here's hoping I fixed it -- I don't have OpenShift handy to test changes :crossed_fingers:

deploy/default-config.yaml
amisevsk: I've updated the Makefile to not set a default `ROUTING_SUFFIX` anymore. On OpenShift, this should result in the config getting an empty value, which would be auto-filled by the controller. On Kubernetes, we auto-detect it in the makefile for minikube and print a warning if it's empty.

deploy/default-config.yaml
amisevsk: Hmm I'm not sure how to reproduce the above issue (deployment is running but devworkspace reports `waiting for workspace deployment`). When I start a DevWorkspace locally, I see the `Waiting for editor to start` message (sometimes, only briefly) -- I assume if startup is really hung on the health check, that message should show up.

amisevsk: > After clusterSuffix is cleaned up, it's not propagated to existing workspaces, even after restart. Removing DevWorkspace Routing helps but it's not desired behavior I think.

This is because the DevWorkspaceRouting CR doesn't store `clusterRouteSuffix`, so reconciling the DevWorkspace won't trigger any reconciles to the DevWorkspaceRouting. Anything triggering a reconcile for the routing object will cause the change to be reflected (e.g. modifying annotations/labels on routes) but otherwise there are no events for the routing controller to respond to.

We could work around this by propagating the `.spec.started` field to `DevWorkspaceRouting`; this would at least enable stopping + restarting the workspace to propagate the config change.

amisevsk: Created https://github.com/devfile/devworkspace-operator/issues/602

main.go
amisevsk: Opened PR https://github.com/eclipse-che/che-operator/pull/1081 (note no config CRD is necessary to run DWO anymore, and `clusterHostSuffix` is only required for the basic routing solver, which is outside the Che Operator use-case).

sleshchenko:
PR check is failing https://github.com/devfile/devworkspace-operator/pull/598/checks?check_run_id=3654476483

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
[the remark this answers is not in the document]
sleshchenko: > I would worry that it's another piece of documentation that has to be maintained (e.g. would we have updated this when we removed the mountvolume setup error?)

That's true. Feel free to just mark this comment as resolved =)

amisevsk:
Squashed the large mess of fixup commits

sleshchenko (approved):
I haven't reviewed changes carefully again but I have tested on OpenShift and it works fine.

Probably logs much be more reach, like what is the whole config piece, but not just what is changed. And what are actually default config are
Screenshot_20210922_174430
But let's move it out of the current PR scope.

Good job!

amisevsk:
e2e tests failed because `ExecCommandInContainer` assumed a hardcoded container name. Should be fixed now.

JPinkney (approved):

PR #600, opened 2021-09-20
Reformat variable using spaces as indentation for directives

amisevsk:
This PR makes _writing_ the Makefiles a little more tedious (since the new convention is mixing tabs and spaces) but maybe easier to read. Not sure if it's worth merging.

### What does this PR do?
Reformat the Makefiles (since I realized Make directives can be safely indented using spaces):
* Nested directives are indented using spaces, so that e.g.
    ```make
    # minikube handling
    ifeq ($(shell $(K8S_CLI) config current-context 2>&1),minikube)
    # check ingress addon is enabled
    ifeq ($(shell minikube addons list -o json | jq -r .ingress.Status), disabled)
    $(error ingress addon should be enabled on top of minikube)
    endif
    export ROUTING_SUFFIX := $(shell minikube ip).nip.io
    endif
    ```
    becomes
    ```make
    # minikube handling
    ifeq ($(shell $(K8S_CLI) config current-context 2>&1),minikube)
      # check ingress addon is enabled
      ifeq ($(shell minikube addons list -o json | jq -r .ingress.Status), disabled)
        $(error ingress addon should be enabled on top of minikube)
      endif
      export ROUTING_SUFFIX := $(shell minikube ip).nip.io
    endif
    ```
    (note this indentation is done via spaces -- this breaks markdown syntax parsing but should be handled by make okay)
* Reindent regular lines so that first indent is a tab character (as required by Make) and following indentation uses two spaces.

Tested briefly on make `3.81` and indentation seems to cause no issues there either.

Hard to read nested `ifeq`s

### PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ SHELL := bash
 .DEFAULT_GOAL := help
 
 ifndef VERBOSE
-MAKEFLAGS += --silent
+  MAKEFLAGS += --silent
 endif
 
 export NAMESPACE ?= devworkspace-controller
@@ -42,7 +42,7 @@ include build/make/version.mk
 include build/make/olm.mk
 
 ifneq (,$(shell which kubectl 2>/dev/null)$(shell which oc 2>/dev/null))
-include build/make/deploy.mk
+  include build/make/deploy.mk
 endif
 
 OPERATOR_SDK_VERSION = v1.8.0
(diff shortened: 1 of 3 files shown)


amisevsk:
What a silly PR for PR #600 :disappointed:

amisevsk:
Just to get it out of the way:

JPinkney (approved):
I personally like this format better then what we currently have. It might take a bit of time to adjust time but it definitely makes the makefiles more readable. I tested all the makefile commands and everything seems to work for me

sleshchenko (approved):
:+1:

Issue #602, opened 2021-09-20
Propagate workspace started state to DevWorkspaceRouting


amisevsk:
### Description

As pointed out in https://github.com/devfile/devworkspace-operator/pull/598#discussion_r712115641, certain config changes are difficult to propagate to DevWorkspaceRoutings (and thus, to the objects that the routing reconciler manages). This is because once services/ingresses/routes are created, there aren't many events on the cluster that trigger new reconciles. Since we do no cleanup when a workspace is stopped, even restarting a workspace may not trigger any reconciles in the routing controller.

One way to work around this would be to propagate the workspaces `.spec.started` field to the created DevWorkspaceRouting, which would at least trigger reconciles when a workspace is restarted.

PR #604, opened 2021-09-21
Propagate workspace `.spec.started` to DevWorkspaceRouting

amisevsk:
### What does this PR do?
Add field `.spec.started` to DevWorkspaceRouting and propagate the owning workspace's `.spec.started` there. When a workspace is stopped, currently we simply end the routing reconcile early, but in the future this could be used to perform cleanup (e.g. remove services/routes for stopped workspaces).

This change has the effect of always triggering at least one reconcile in the routing controller when a workspace is started.

Closes https://github.com/devfile/devworkspace-operator/issues/602

### Is it tested? How?
Start a workspace, stop a workspace.

### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspacerouting_types.go
+++ b/apis/controller/v1alpha1/devworkspacerouting_types.go
@@ -23,6 +23,8 @@ import (
 type DevWorkspaceRoutingSpec struct {
 	// Id for the DevWorkspace being routed
 	DevWorkspaceId string `json:"devworkspaceId"`
+	// Started is true if the owning DevWorkspace has .spec.started=true
+	Started bool `json:"started"`
 	// Class of the routing: this drives which DevWorkspaceRouting controller will manage this routing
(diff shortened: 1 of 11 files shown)


sleshchenko (approved):

sleshchenko:
Seems it's a breaking chang and DWCO fails to route routing https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/devfile_devworkspace-operator/604/pull-ci-devfile-devworkspace-operator-main-v8-che-happy-path/1440398068900958208/artifacts/che-happy-path/che-happy-path/artifacts/eclipse-che-info/che-operator-65c7f8454d-tw5fp-che-operator.log

and screencast proves it, but happy test is too positive and happy with that ) Or it's just PR check. Will ask @musienko-maxim to discover the reason.

And we probably should revert it and the only safest solution I see - use annotation instead.

sleshchenko:
Should fix the problem with always happy happy path test https://github.com/devfile/devworkspace-operator/pull/609

Issue #578, opened 2021-09-07
Introduce timing out for DevWorkspace to prevent hanging on any of startup phase


sleshchenko:
There may be cases when workspace hangs on some of the startup phase: preparing networking/waiting to deployment/pining main endpoint url.

It would be nice to clean up such devworkspace and move them to FAILED state to save resources which are going to be used forever until user won't interrupt workspace start.

PR #605, opened 2021-09-21
Introduce timeout for hanging workspaces

amisevsk:
**Note:** this PR depends on https://github.com/devfile/devworkspace-operator/pull/598 and so for now includes those changes as well; ignore all but the last three commits (i.e. https://github.com/devfile/devworkspace-operator/commit/05c04413104aa77642784768623c36dc67130815 to the end). Once #598 is merged, this commit will be a lot simpler.

### What does this PR do?
Introduces config setting `.workspace.startProgressTimeout` that controls how long a workspace can be in the starting phase without progressing before it is automatically failed.

"Progressing" in our case means "updating workspace `.status.conditions`" -- if the last condition change (checked by `lastTransitionTime`) is more than `idleTimeout` ago, we assume the workspace has failed.

Default timeout value is 5 minutes, to allow for e.g. an incredibly slow image pull.

Closes https://github.com/devfile/devworkspace-operator/issues/578

### Is it tested? How?
1. Create or update the `DevWorkspaceOperatorConfig` to ignore `FailedScheduling` events (and optionally set a shorter `idleTimeout`):
    ```bash
    kubectl patch dwoc devworkspace-operator-config --type merge -p \
      '{
        "config": {
          "workspace": {
            "ignoredUnrecoverableEvents": [
              "FailedScheduling"
            ],
            "startProgressTimeout": "30s"
          }
        }
      }'
    ```
2. Create a DevWorkspace that cannot be scheduled:
    ```yaml
    cat <<EOF | kubectl apply -f 
    kind: DevWorkspace
    metadata:
      name: test-devworkspace
    spec:
      started: true
      template:
        components:
          - name: tooling
            container:
              image: quay.io/wto/web-terminal-tooling
              memoryRequest: 100Gi
              memoryLimit: 100Gi
    EOF
    ```
3. Wait 30 seconds; workspace should fail to start.

### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -61,6 +61,11 @@ type WorkspaceConfig struct {
 	// requires support in the workspace being started. If not specified, the default
 	// value of "15m" is used.
 	IdleTimeout string `json:"idleTimeout,omitempty"`
+	// StartProgressTimeout determines the maximum duration a DevWorkspace can be in
+	// a "Starting" phase without progressing before it is automatically failed.
+	// Duration should be specified in a format parseable by Go's time package, e.g.
+	// "15m", "20s", "1h30m", etc. If not specified, the default value of "5m" is used.
+	StartProgressTimeout string `json:"startProgressTimeout,omitempty"`
 	// IgnoredUnrecoverableEvents defines a list of Kubernetes event names that should
 	// be ignored when deciding to fail a DevWorkspace startup. This option should be used
 	// if a transient cluster issue is triggering false-positives (for example, if
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -61,11 +61,11 @@ type WorkspaceConfig struct {
 	// requires support in the workspace being started. If not specified, the default
 	// value of "15m" is used.
 	IdleTimeout string `json:"idleTimeout,omitempty"`
-	// StartProgressTimeout determines the maximum duration a DevWorkspace can be in
-	// a "Starting" phase without progressing before it is automatically failed.
+	// ProgressTimeout determines the maximum duration a DevWorkspace can be in
+	// a "Starting" or "Failing" phase without progressing before it is automatically failed.
 	// Duration should be specified in a format parseable by Go's time package, e.g.
 	// "15m", "20s", "1h30m", etc. If not specified, the default value of "5m" is used.
-	StartProgressTimeout string `json:"startProgressTimeout,omitempty"`
+	ProgressTimeout string `json:"progressTimeout,omitempty"`
 	// IgnoredUnrecoverableEvents defines a list of Kubernetes event names that should
 	// be ignored when deciding to fail a DevWorkspace startup. This option should be used
 	// if a transient cluster issue is triggering false-positives (for example, if
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -59,6 +59,10 @@ import (
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 )
 
+const (
+	startingWorkspaceRequeueInterval = 5 * time.Second
+)
+
 // DevWorkspaceReconciler reconciles a DevWorkspace object
 type DevWorkspaceReconciler struct {
 	client.Client
@@ -181,8 +185,16 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	clusterWorkspace := workspace.DeepCopy()
 	timingInfo := map[string]string{}
 	timing.SetTime(timingInfo, timing.DevWorkspaceStarted)
+
 	defer func() (reconcile.Result, error) {
 		r.syncTimingToCluster(ctx, clusterWorkspace, timingInfo, reqLogger)
+		// Don't accidentally suppress errors by overwriting here; only check for timeout when no error
+		// encountered in main reconcile loop.
+		if err == nil {
+			if timeoutErr := checkForStartTimeOut(clusterWorkspace, reqLogger, &reconcileStatus); timeoutErr != nil {
+				reconcileResult, err = r.failWorkspace(workspace, timeoutErr.Error(), reqLogger, &reconcileStatus)
+			}
+		}
 		return r.updateWorkspaceStatus(clusterWorkspace, reqLogger, &reconcileStatus, reconcileResult, err)
 	}()
 
@@ -286,6 +298,9 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			message = routingStatus.Message
 		}
 		reconcileStatus.setConditionFalse(dw.DevWorkspaceRoutingReady, message)
+		if !routingStatus.Requeue && routingStatus.Err == nil {
+			return reconcile.Result{RequeueAfter: startingWorkspaceRequeueInterval}, nil
+		}
 		return reconcile.Result{Requeue: routingStatus.Requeue}, routingStatus.Err
 	}
 	reconcileStatus.setConditionTrue(dw.DevWorkspaceRoutingReady, "Networking ready")
@@ -333,6 +348,9 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// FailStartup is not possible for generating the serviceaccount
 		reqLogger.Info("Waiting for workspace ServiceAccount")
 		reconcileStatus.setConditionFalse(dw.DevWorkspaceServiceAccountReady, "Waiting for DevWorkspace ServiceAccount")
+		if !serviceAcctStatus.Requeue && serviceAcctStatus.Err == nil {
+			return reconcile.Result{RequeueAfter: startingWorkspaceRequeueInterval}, nil
+		}
 		return reconcile.Result{Requeue: serviceAcctStatus.Requeue}, serviceAcctStatus.Err
 	}
 	serviceAcctName := serviceAcctStatus.ServiceAccountName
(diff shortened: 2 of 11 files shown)


JPinkney (approved):
Just tested and everything worked as expected

controllers/workspace/devworkspace_controller.go
```diff
@@ -355,6 +376,9 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 		reqLogger.Info("Waiting on deployment to be ready")
 		reconcileStatus.setConditionFalse(conditions.DeploymentReady, "Waiting for workspace deployment")
+		if !deploymentStatus.Requeue && deploymentStatus.Err == nil {
+			return reconcile.Result{RequeueAfter: startingWorkspaceRequeueInterval}, nil
```
sleshchenko, line 380: Does here the last or the first internal win?

- I mean we wait for deployment initially and scheduled reconciling in 5 secs (1);
- Deployment created pods and it invokes reconcile loop, we scheduling in 5 secs (2);
- Pod is updated several times after containers status are propagates, we scheduled N times in 5 secs(N);

The question is, how many times reconcile loops our `RequeueAfter` will initiate?

controllers/workspace/devworkspace_controller.go
```diff
@@ -59,6 +59,10 @@ import (
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 )
 
+const (
+	startingWorkspaceRequeueInterval = 5 * time.Second
```
sleshchenko, line 63: I wonder if 5 secs is always the best choice or we can make it dynamic and like to minute, 2,3 in case of 5minutes timeout?

sleshchenko (commented):
LGTM I just have some question, which I may find by myself during testing, but I'm not there yet.

controllers/workspace/status.go
```diff
@@ -238,3 +240,29 @@ func updateMetricsForPhase(workspace *dw.DevWorkspace, oldPhase, newPhase dw.Dev
...
+
+// checkForStartTimeOut checks if the provided workspace has not progressed for longer than the configured
+// startup timeout. This is determined by checking to see if the last condition transition time is more
+// than [timeout] duration ago. Workspaces that are not in the "Starting" phase cannot timeout. Returns
+// an error with message when timeout is reached.
+func checkForStartTimeOut(workspace *dw.DevWorkspace, logger logr.Logger, status *currentStatus) error {
+	if workspace.Status.Phase != dw.DevWorkspaceStatusStarting {
+		return nil
+	}
+	timeout, err := time.ParseDuration(config.Workspace.StartProgressTimeout)
```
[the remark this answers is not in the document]
amisevsk: Parsing once would mean it's harder to pick up changes in config while running (we'd probably have to restart the controller).

We could avoid this by using the `Duration` raw in the config, since "parsing" is just converting a string to a `int64`, but that would require the entered value to be in nanoseconds (I believe)

controllers/workspace/devworkspace_controller.go
amisevsk: My understanding is that reconciling due to an event cancels out a requeueAfter, but it's hard to check for sure. In my testing, I don't see bursts of "waiting on deployment" in the logs, which is what would likely happen if we were stacking requeues.

controllers/workspace/devworkspace_controller.go
amisevsk: I mostly picked 5 seconds because it feels long enough that it's not burdening the controller (we can reconcile many times a second) but also short enough to avoid strange issues in setting a timeout (e.g. if we set a 1 minute requeue, what happens if the config specifies a timeout of 1 minute 15 seconds?)

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -61,6 +61,11 @@ type WorkspaceConfig struct {
 	// requires support in the workspace being started. If not specified, the default
 	// value of "15m" is used.
 	IdleTimeout string `json:"idleTimeout,omitempty"`
+	// StartProgressTimeout determines the maximum duration a DevWorkspace can be in
+	// a "Starting" phase without progressing before it is automatically failed.
```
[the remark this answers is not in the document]
amisevsk: Renamed config property to `progressTimeout` and applied it to `Failing` DevWorkspaces too.

controllers/workspace/status.go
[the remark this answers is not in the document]
sleshchenko: > Parsing once would mean it's harder to pick up changes in config while running (we'd probably have to restart the controller).

True, but also we can reevaluate that value on DWOC reconciliation. but don't mind, that's good enough or even better than I have in mind =)

controllers/workspace/devworkspace_controller.go
sleshchenko: > 1 minute 15 seconds

I don't think there is a case for such a precise timeout. Maybe we can declare our precision to like 1minute or 30 seconds.

And actually, I thought about reconciling after (timeout + last transition time - now + 5 sec), which will initiate reconcile loop after 5 sec when potentially we need to kill that.

But I'm OK with any if it does not generate redundant loading.

sleshchenko (approved):
I don't have any reason to delay delivering this valuable feature.
Approving and if I find any way to improve/optimize(if it's really needed) implementation - I'll raise a dedicated issue.

sleshchenko:
Che Happy path fails with
Screenshot_20210928_180850

I assume it's caused by a health check that understands redirect as success.
So, the health endpoint should be added to allowlist on Che side, it should not be a blocker for this PR.
But also we can make DWO respect only 2XX response codes and don't follow redirect links, but it's another story.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -61,6 +61,11 @@ type WorkspaceConfig struct {
...
+	// Duration should be specified in a format parseable by Go's time package, e.g.
+	// "15m", "20s", "1h30m", etc. If not specified, the default value of "5m" is used.
+	StartProgressTimeout string `json:"startProgressTimeout,omitempty"`
```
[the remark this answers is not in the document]
amisevsk: I renamed to `ProgressTimeout` to include Failing dws and forgot this comment -- sorry. Are you okay with that name? My thinking is it's the max duration a DevWorkspace can exist without progressing

controllers/workspace/devworkspace_controller.go
amisevsk: I tested this informally by starting a workspace that will timeout (ignore `FailedScheduling` events and ask for 100Gi in Theia). While the workspace is looping on checking the deployment every 5 seconds, I ran
```
for i in {1..5}; do
  kubectl patch dw theia-next --type merge -p '{"metadata": {"labels": {"touch": "false"}}}'
  sleep 0.5
  kubectl patch dw theia-next --type merge -p '{"metadata": {"labels": {"touch": "true"}}}'
  sleep 0.5
done
```
to trigger 10 reconciles to the object, with the assumption that each of those would _also_ start a `RequeueAfter`. However, after the 10 quick reconciles are completed, the controller goes back to queuing reconciles every five seconds, rather than 10 reconciles every 5 seconds, so it seems like RequeueAfter is cancelled if an event triggers a reconcile.

I agree that 5 seconds may be the wrong value here; we should tweak this later.

Divine1:
hi team

i have installed eclipse-che in k8s cluster. i'm facing below timeout error.
Please let me know how can i increase the timeout limit here? 
should i add `StartProgressTimeout` in `devfile.yaml`? please help me on this

image



i tried increasing the timeout in `checluster`, but it has no effect. 

```
apiVersion: org.eclipse.che/v2
kind: CheCluster
metadata:
  name: eclipse-che
spec:
  components:
    cheServer:
      extraProperties:
        CHE_INFRA_KUBERNETES_WORKSPACE__START__TIMEOUT__MIN: "15"
    database:
      externalDb: true
      postgresHostName: sc2-10-186-67-195.eng.vmware.com
      postgresPort: "5432"
  networking:
    auth:
      identityProviderURL: https://dexvmware.com
      oAuthClientName: eclipse-che
      oAuthSecret: sdsdsdsdsd
```

amisevsk:
Hi @Divine1 -- that error will occur when the DevWorkspace does not reach a "Running" state before the start timeout (introduced here). By default this is 5 minutes but can be increased by setting the `.config.workspace.progressTimeout` field in the DevWorkspaceOperatorConfig, e.g.:
```yaml
metadata:
  namespace: <devworkspace/che installation namespace>
config:
  workspace:
    progressTimeout: 15m # This sets the timeout to 15 minutes
```

However, if it's taking longer than 5 minutes for your workspace to start, that suggests a different problem. Do you have any information about the workspace or its pods while it's starting? Some things to check:

* What `phase` does the workspace get stuck on (`kubectl get dw <workspace-name>`)
* Is the workspace deployment running/ready? If so, is there anything in the logs there?

From here, we can start to narrow down what's causing the issue.

tolusha:
@amisevsk  @Divine1 
So, the problem is the following. It takes more than 5m to pull images which causes workspace fail to start.
What are possible solutions?

Divine1:
@tolusha  Thank you for checking this


i have posted logs with shows failing to pull images within 5m. but is it the only issue? or is there any issue with `project-clone` init-container?


detailed logs are in this link https://github.com/eclipse/che/issues/21613#issuecomment-1219411917




@amisevsk

tolusha:
I think I found the problem. `5m` is hardcoded timeout on dashboard

https://github.com/eclipse-che/che-dashboard/blob/f4b2988abb72cb14068d6f81f649783a5b1d132b/packages/dashboard-frontend/src/containers/Loader/const.tsx#L36
https://github.com/eclipse-che/che-dashboard/blob/f4b2988abb72cb14068d6f81f649783a5b1d132b/packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx#L158

Divine1:
@tolusha but why is the workspace hanging and i'm able to see errors in `kubectl get events -n workspace-namespace` logs ? if increasing timeout is the solution, then there shouldn't be other errors in the `kubectl get events -n workspace-namespace` logs.


One more doubt... i noticed, the `workspace deployment component` is applied immediately after the `PVC` is issued, but more seconds is needed to  provision a `PV`. Should there be any delay between applying `workspace deployment` and `PV` provisioning?

amisevsk:
If the issue is that pulling images takes longer than 5 minutes, then the solutions would be either

1. Increase the timeout as specified in https://github.com/devfile/devworkspace-operator/pull/605#issuecomment-1179301125 (I don't believe the dashboard is failing workspaces, and only the duration in the error is hard-coded)
2. Install and configure the [Kubernetes Image Puller](https://github.com/che-incubator/kubernetes-image-puller) in the cluster to automatically pull images before workspaces are started.
3. Manually ensure every node in the cluster has the images required (the approach the image puller uses is to create a daemonset to run a pod with those images on every node)

PR #610, opened 2021-09-23
Revert "Propagate workspace `.spec.started` to DevWorkspaceRouting"

amisevsk:
Reverts devfile/devworkspace-operator#604

#604 adds the started field as a required field for devworkspaceroutings, which is an breaking change to the dwr spec.


JPinkney (approved):

sleshchenko (approved):
Actually, that's a good change but it's too late to add it as CRD field.
So, as discussed alternative way: drop it into DWR annotation.

PR #617, opened 2021-09-24
Propagate DevWorkspace .spec.started status to an annotation on routings

### LINKED ISSUE #602: Propagate workspace started state to DevWorkspaceRouting

amisevsk:
This is a backwards-compatible re-do of https://github.com/devfile/devworkspace-operator/pull/604

### What does this PR do?
Add annotation 'controller.devfile.io/devworkspace-started' to be applied to DevWorkspaceRoutings. If a workspace is stopped, the annotation is set to "false", otherwise it is "true". If the DevWorkspaceRouting reconciler encounters a DevWorkspaceRouting with the annotation set to false, it ends the reconcile early.

This ensures that every workspace start/stop event triggers a reconcile, and allows the routing to clean up resources if necessary.

Re-closes #602 

### Is it tested? How?
Check that DevWorkspaceRoutings get the annotation, *and that this doesn't break the Che operator*.

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -21,6 +21,7 @@ import (
 
 	devfilevalidation "github.com/devfile/api/v2/pkg/validation"
 
+	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/controllers/workspace/metrics"
 	"github.com/devfile/devworkspace-operator/pkg/common"
@@ -414,6 +415,28 @@ func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.
 			return true, nil
 		}
 		return false, err
+	}
+
+	// Update DevWorkspaceRouting to have .spec.started=false
+	routing := &v1alpha1.DevWorkspaceRouting{}
+	routingRef := types.NamespacedName{
+		Name:      common.DevWorkspaceRoutingName(workspace.Status.DevWorkspaceId),
+		Namespace: workspace.Namespace,
+	}
+	err = r.Get(context.TODO(), routingRef, routing)
+	if err != nil {
+		if !k8sErrors.IsNotFound(err) {
+			return false, err
+		}
+	} else if routing.Annotations != nil && routing.Annotations[constants.DevWorkspaceStartedStatusAnnotation] != "false" {
+		routing.Annotations[constants.DevWorkspaceStartedStatusAnnotation] = "false"
+		err := r.Update(context.TODO(), routing)
+		if err != nil {
+			if k8sErrors.IsConflict(err) {
+				return false, nil
+			}
+			return false, err
+		}
 	}
 
 	replicas := workspaceDeployment.Spec.Replicas
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -417,7 +417,7 @@ func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.
 		return false, err
 	}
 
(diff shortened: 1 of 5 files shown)


sleshchenko:
Che Happy path test fails with the following error on DWO side:
```
{"level":"info","ts":1632548602.532224,"logger":"controllers.DevWorkspace","msg":"Reconciling Workspace","Request.Namespace":"che-user-che","Request.Name":"spring-petclinic-ewgp","devworkspace_id":"workspace94dadae88e684355"}
panic: assignment to entry in nil map

goroutine 1235 [running]:
github.com/devfile/devworkspace-operator/pkg/provision/workspace.getSpecRouting(0xc000224900, 0xc0004ed1f0, 0xc0009f6880, 0x1cce8df, 0x1ca63c0)
	/devworkspace-operator/pkg/provision/workspace/routing.go:157 +0x1a5
github.com/devfile/devworkspace-operator/pkg/provision/workspace.SyncRoutingToCluster(0xc000224900, 0x1f235e0, 0xc0005e1270, 0xc0004ed1f0, 0x1f17860, 0xc000230e20, 0x1f0da20, 0xc001952120, 0x0, 0x0, ...)
	/devworkspace-operator/pkg/provision/workspace/routing.go:69 +0x77
github.com/devfile/devworkspa
```

controllers/workspace/devworkspace_controller.go
```diff
@@ -416,6 +417,28 @@ func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.
 		return false, err
 	}
 
+	// Update DevWorkspaceRouting to have .spec.started=false
```
sleshchenko, line 420:
```suggestion
	// Update DevWorkspaceRouting to have annotation for started setting.
```

pkg/provision/workspace/routing.go
```diff
@@ -153,6 +154,7 @@ func getSpecRouting(
 	if val, ok := workspace.Annotations[constants.DevWorkspaceRestrictedAccessAnnotation]; ok {
 		annotations = maputils.Append(annotations, constants.DevWorkspaceRestrictedAccessAnnotation, val)
 	}
+	annotations[constants.DevWorkspaceStartedStatusAnnotation] = "true"
```
sleshchenko, line 157: it seems to be a place which try to write nil map.
```suggestion
	annotations = maputils.Append(annotations, constants.DevWorkspaceStartedStatusAnnotation, "true")
```

sleshchenko (commented):
LGTM.

Probably it's too late to name that alternative but we could do spec.started as well, with help of mutating webhooks and restoring value from the existing state. It's the same we do for creator annotation.
I have no idea why it did not appear for me earlier.

But I don't have a strong opinion, I let you decide which you like more.

sleshchenko (approved):

sleshchenko:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/devfile_devworkspace-operator/617/pull-ci-devfile-devworkspace-operator-main-v8-che-happy-path/1442756715840802816/artifacts/che-happy-path/che-happy-path/artifacts/e2e/Workspace_creation_via_factory_url_Open_factory_URL_Register_running_workspace/screenshot-Register_running_workspace.png

Che Happy path failure is the same as in https://github.com/devfile/devworkspace-operator/pull/605#issuecomment-929325801
So, not really related to this PR.

JPinkney (approved):

Issue #623, opened 2021-09-29
DWO can only auto-discover clusterRoutingSuffix config on startup (OpenShift)


amisevsk:
### Description
DWO can only auto-detect the appropriate `clusterRoutingSuffix` (on OpenShift) during startup. This means that if DWO is started with a config that has the value `config.routing.clusterHostSuffix` set (e.g. accidentally), it's not possible to return to the default behavior of auto-detecting it without restarting the controller. 

### How To Reproduce
On OpenShift, run `make install` with the env var `CLUSTER_HOST_SUFFIX` set to a wrong value:
```bash
export CLUSTER_HOST_SUFFIX=nonsense
make install
```
After startup, workspaces use the `nonsense` value for routingSuffix and fail to start fully. If you attempt to resolve the issue by updating the config to not include `clusterHostSuffix`, the controller will loop on an error that `clusterHostSuffix` must be set.

### Expected behavior
Not really sure what we should do here. Perhaps we should autodetect clusterHostSuffix always on OpenShift and use that as a "default" value?

PR #624, opened 2021-09-29
Always get default clusterHostSuffix and store it as a default on OpenShift

amisevsk:
### What does this PR do?
* Always gets `clusterRoutingSuffix` when possible and stores it in default, to avoid an issue when the controller is started with the value overwritten (https://github.com/devfile/devworkspace-operator/issues/623)
* Fail workspaces if `clusterRoutingSuffix` is required and unset.

Fixes https://github.com/devfile/devworkspace-operator/issues/623

### Is it tested? How?
Use reproducer in https://github.com/devfile/devworkspace-operator/issues/623

### PR Checklist

### DIFF
--- a/pkg/config/sync.go
+++ b/pkg/config/sync.go
@@ -71,14 +71,13 @@ func SetupControllerConfig(client crclient.Client) error {
 	} else {
 		syncConfigFrom(config)
 	}
+	defaultRoutingSuffix, err := discoverRouteSuffix(client)
+	if err != nil {
+		return err
+	}
+	DefaultConfig.Routing.ClusterHostSuffix = defaultRoutingSuffix
 	if internalConfig.Routing.ClusterHostSuffix == "" {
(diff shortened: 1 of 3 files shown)


pkg/config/sync.go
```diff
@@ -71,14 +71,13 @@ func SetupControllerConfig(client crclient.Client) error {
 	} else {
 		syncConfigFrom(config)
 	}
+	defaultRoutingSuffix, err := discoverRouteSuffix(client)
+	if err != nil {
+		return err
+	}
+	DefaultConfig.Routing.ClusterHostSuffix = defaultRoutingSuffix
 	if internalConfig.Routing.ClusterHostSuffix == "" {
```
sleshchenko, line 79: I just wonder if there is a way to avoid evaluation ClusterHostSuffix on each reconcile loop?
It may be overhead but we may use configStatus for such stuff.

PR #636, opened 2021-10-08
WIP: Remove Web Terminal Routing

JPinkney:
### What does this PR do?
This PR removes web terminal routing and all other web terminal references

### Is it tested? How?

Untested for now

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [ ] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-che-happy-path`: Happy path for verification integration with Che

PR #637, opened 2021-10-08
Remove internal registry

JPinkney:
### What does this PR do?
As far as I can tell, the internal registry is no longer needed so it's probably time to remove it. This PR is split off from https://github.com/devfile/devworkspace-operator/pull/636 so that we can review the internal registry changes now

### PR Checklist


amisevsk:
I believe the last bits that used the internal registry were removed in https://github.com/devfile/devworkspace-operator/pull/598

PR #647, opened 2021-10-14
feat: Add WorkspaceEngineFailure reason to the devworkspace_fai…

dkwon17:
…l_total Prometheus metric

### What does this PR do?
Adds the `WorkspaceEngineFailure` to the list of failure reasons.
image
image

The PR does this by adding a prefix to the workspace status when workspace is phase is set to `Failing` from the main reconciliation loop.

https://github.com/devfile/devworkspace-operator/pull/597

### Is it tested? How?
Used these steps: https://github.com/devfile/devworkspace-operator/tree/main/doc/grafana#testing-devworkspace-operator-metrics-locally.

To simulate a timeout error:
1. Set `config.workspace. progressTimeout` to `1s` in the `DevWorkspaceOperatorConfig` instance.
2. Create a new DevWorkspace

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -379,7 +379,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	deploymentStatus := wsprovision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI)
 	if !deploymentStatus.Continue {
 		if deploymentStatus.FailStartup {
-			failureReason := wsprovision.DetermineStartupFailureReason(deploymentStatus.Info())
+			failureReason := metrics.DetermineProvisioningFailureReason(deploymentStatus)
 			return r.failWorkspace(workspace, deploymentStatus.Info(), failureReason, reqLogger, &reconcileStatus)
 		}
 		reqLogger.Info("Waiting on deployment to be ready")
@@ -493,7 +493,7 @@ func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.
 func (r *DevWorkspaceReconciler) failWorkspace(workspace *dw.DevWorkspace, msg string, reason metrics.FailureReason, logger logr.Logger, status *currentStatus) (reconcile.Result, error) {
 	logger.Info("DevWorkspace failed to start: " + msg)
 	status.phase = devworkspacePhaseFailing
-	status.setConditionTrueWithReason(dw.DevWorkspaceFailedStart, msg, reason.CamelCase())
+	status.setConditionTrueWithReason(dw.DevWorkspaceFailedStart, msg, reason.GetReason())
 	if workspace.Spec.Started {
 		return reconcile.Result{Requeue: true}, nil
 	}
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -492,7 +492,7 @@ func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.
 func (r *DevWorkspaceReconciler) failWorkspace(workspace *dw.DevWorkspace, msg string, logger logr.Logger, status *currentStatus) (reconcile.Result, error) {
 	logger.Info("DevWorkspace failed to start: " + msg)
 	status.phase = devworkspacePhaseFailing
-	status.setConditionTrue(dw.DevWorkspaceFailedStart, msg)
+	status.setConditionTrue(dw.DevWorkspaceFailedStart, metrics.WorkspaceEngineFailurePrefix+msg)
 	if workspace.Spec.Started {
 		return reconcile.Result{Requeue: true}, nil
 	}
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -199,7 +199,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// encountered in main reconcile loop.
 		if err == nil {
 			if timeoutErr := checkForStartTimeout(clusterWorkspace); timeoutErr != nil {
-				reconcileResult, err = r.failWorkspace(workspace, timeoutErr.Error(), reqLogger, &reconcileStatus)
+				reconcileResult, err = r.failWorkspace(workspace, timeoutErr.Error(), metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus)
 			}
 		}
 		return r.updateWorkspaceStatus(clusterWorkspace, reqLogger, &reconcileStatus, reconcileResult, err)
@@ -208,7 +208,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if workspace.Annotations[constants.DevWorkspaceRestrictedAccessAnnotation] == "true" {
 		msg, err := r.validateCreatorLabel(clusterWorkspace)
 		if err != nil {
-			return r.failWorkspace(workspace, msg, reqLogger, &reconcileStatus)
+			return r.failWorkspace(workspace, msg, metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus)
 		}
 	}
 
@@ -229,7 +229,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	flattenedWorkspace, warnings, err := flatten.ResolveDevWorkspace(&workspace.Spec.Template, flattenHelpers)
 	if err != nil {
-		return r.failWorkspace(workspace, fmt.Sprintf("Error processing devfile: %s", err), reqLogger, &reconcileStatus)
+		return r.failWorkspace(workspace, fmt.Sprintf("Error processing devfile: %s", err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
 	}
 	if warnings != nil {
 		reconcileStatus.setConditionTrue(conditions.DevWorkspaceWarning, flatten.FormatVariablesWarning(warnings))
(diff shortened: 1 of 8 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -491,7 +491,7 @@ func (r *DevWorkspaceReconciler) doStop(workspace *dw.DevWorkspace, logger logr.
 func (r *DevWorkspaceReconciler) failWorkspace(workspace *dw.DevWorkspace, msg string, logger logr.Logger, status *currentStatus) (reconcile.Result, error) {
 	logger.Info("DevWorkspace failed to start: " + msg)
 	status.phase = devworkspacePhaseFailing
-	status.setConditionTrue(dw.DevWorkspaceFailedStart, msg)
+	status.setConditionTrue(dw.DevWorkspaceFailedStart, metrics.WorkspaceEngineFailurePrefix+msg)
```
amisevsk, line 494: I'm not clear on how this works -- if we're always setting the prefix when a workspace fails (and all workspace failures come through this function at some point), it seems redundant. In which cases do we enter `determinateFailureReason` without having this prefix set?

amisevsk (commented):
A more involved alternative, if you're up to a more extensive refactor: the `WorkspaceCondition` struct already has a `Reason` field that is currently unused. We could set a reason when setting the `Failed` condition to `true`, and then use that to differentiate the different causes.

Otherwise, we can figure something more straightforward out and defer the conditions rework into another PR.

controllers/workspace/devworkspace_controller.go
amisevsk: To clarify this a little, currently the general flow is
```
[Do something in controller]
          V
[Detect a failure state]
          V
[Set the DevWorkspaceFailed condition to true]
          V
[Set the .status.message according to the message in failure condition]
          V
[Parse this message in .status.message to distinguish failure reason]
```

It looks like what we're doing in this PR is adding a prefix to the .status.message field, but the same effect could be achieved by getting information from the failed condition.

amisevsk: Yeah, I think every failure has to go through failWorkspace(). The flow is a little confusing because failure happens over 3+ reconciles:
1. [First reconcile](https://github.com/devfile/devworkspace-operator/blob/45f64af451d429ada28a96a78a5b535b3636eeec/controllers/workspace/devworkspace_controller.go#L492-L500): hit error, go through `failWorkspace` to set `Failing` status, return
2. [Second reconcile](https://github.com/devfile/devworkspace-operator/blob/45f64af451d429ada28a96a78a5b535b3636eeec/controllers/workspace/devworkspace_controller.go#L140-L159): hit early condition to check for `Failing` status, update workspace to be stopped
3. [Third reconcile](https://github.com/devfile/devworkspace-operator/blob/45f64af451d429ada28a96a78a5b535b3636eeec/controllers/workspace/devworkspace_controller.go#L161-L166): hit the `stopWorkspace` check for stopped workspaces

controllers/workspace/devworkspace_controller.go
```diff
@@ -72,6 +72,24 @@ type DevWorkspaceReconciler struct {
...
+
+type errorReasonImpl struct {
+	error
+	reason string
+}
+
+func (e errorReasonImpl) Reason() string {
+	return e.reason
+}
+
```
dkwon17, line 92: This `errorReason` interface was created to help set a failure reason when an error has occurred. For example:
https://github.com/devfile/devworkspace-operator/blob/ba8c5c1335339678a22a76ad8694546deabd61a3/controllers/workspace/status.go#L253
could fail due to an incorrect timeout value, or due to a timeout. 

If this `errorReason` is okay, I think functions like: https://github.com/devfile/devworkspace-operator/blob/ba8c5c1335339678a22a76ad8694546deabd61a3/controllers/workspace/devworkspace_controller.go#L286
could also benefit from it, since this function can fail for different reasons?

controllers/workspace/metrics/failure_reason.go
```diff
@@ -0,0 +1,54 @@
...
+	ReasonInfrastructureFailure,
+	ReasonWorkspaceEngineFailure,
+	ReasonUnknown,
+}
+
+func (f *FailureReason) CamelCase() string {
+	return f.camelCase
+}
+
+func (f *FailureReason) SnakeCase() string {
```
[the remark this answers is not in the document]
dkwon17: I took a look at some [che-server metrics](https://gist.github.com/dkwon17/74b75ebbc88da162c9b9740ec11ab2df) and it seems like the label format is not strict, here are some examples I found:
```
jvm_gc_pause_seconds_count{action="end of major GC",cause="Metadata GC Threshold",} 1.0
jvm_gc_pause_seconds_sum{action="end of major GC",cause="Allocation Failure",} 0.196
disk_free_bytes{path="/etc/hostname (tmpfs)",} 8.239316992E9
jvm_memory_committed_bytes{area="nonheap",id="Compressed Class Space",} 1.3893632E7
executor_completed_tasks_total{name="master-websocket-major-endpoint",} 28.0
executor_completed_tasks_total{name="org.eclipse.che.api.workspace.server.WorkspaceSharedPool",} 2.0
che_workspace_failure_total{area="workspace",while="STARTING",} 0.0
okhttp_requests_seconds_count{host="10.217.4.1",method="DELETE",status="200",uri="none",} 29.0
tomcat_threads_busy_threads{name="http-nio-8080",} 0.0
che_workspace_start_time_seconds_max{area="workspace",result="success",} 0.0
```

Maybe we can stick with camel case?

pkg/provision/workspace/deployment.go
```diff
@@ -60,6 +61,19 @@ var unrecoverablePodEventReasons = []string{
...
+	"ImagePullBackOff",
+}
+
+var infrastructureFailures = []string{
+	"CreateContainerError",
+	"RunContainerError",
+	"FailedScheduling",
+	"FailedMount",
+}
+
```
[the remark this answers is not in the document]
dkwon17: In my latest commit, I've moved this to a new file in the metrics package called `deployment_provisioning.go`.

dkwon17:
Thank you for the reviews! I've made some new changes according to the feedback.

controllers/workspace/metrics/failure_reason.go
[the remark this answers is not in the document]
dkwon17: I found some experimental Kubernetes pod metrics that have a `reason` label in UpperCamelCase, with a few example queries at the bottom of the markdown: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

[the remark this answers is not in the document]
ibuziuk: sounds good, let's opt for the UpperCamelCase for labels in this case :+1:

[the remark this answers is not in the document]
dkwon17: Sounds good, I updated the PR to use UpperCamelCase

ibuziuk (approved):
LGTM :+1:

controllers/workspace/metrics/failure_reason.go
```diff
@@ -0,0 +1,49 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package metrics
+
+type FailureReason struct {
+	reason string
+}
```
amisevsk, line 20:
```suggestion
type FailureReason string
```

amisevsk (commented):
Just one more change

controllers/workspace/metrics/failure_reason.go
dkwon17: Thank you, fixed

controllers/workspace/metrics/failure_reason.go
```diff
@@ -15,15 +15,13 @@
...
-var (
-	ReasonBadRequest             = FailureReason{reason: "BadRequest"}
-	ReasonInfrastructureFailure  = FailureReason{"InfrastructureFailure"}
-	ReasonWorkspaceEngineFailure = FailureReason{"WorkspaceEngineFailure"}
-	ReasonUnknown                = FailureReason{"Unknown"}
+const (
+	ReasonBadRequest             = "BadRequest"
+	ReasonInfrastructureFailure  = "InfrastructureFailure"
+	ReasonWorkspaceEngineFailure = "WorkspaceEngineFailure"
+	ReasonUnknown                = "Unknown"
```
amisevsk, line 24: This keeps the constants as string-type (and requires the workarounds below). It'd be simpler to just use
```suggestion
const (
	ReasonBadRequest             FailureReason = "BadRequest"
	ReasonInfrastructureFailure  FailureReason = "InfrastructureFailure"
	ReasonWorkspaceEngineFailure FailureReason = "WorkspaceEngineFailure"
	ReasonUnknown                FailureReason = "Unknown"
)
```

The purpose of using a type here is to lightly restrict the different failure reasons -- if a function expects a defined FailureReason as a parameter, it prompts you to use one of the defined constants. When eventually the failureReason as-a-string is required, we can do `string(reason)`.

JPinkney (approved):

controllers/workspace/metrics/failure_reason.go
dkwon17: Oops, thanks @amisevsk . I made the changes and did some small refactoring.

amisevsk (approved):

ibuziuk:
@musienko-maxim hi, could you please clarify why the happy path is failing?

amisevsk:
Merging as-is since che-happy-path test is failing for all PRs.

We can issues a follow-up PR if an actual issue is present.

controllers/workspace/devworkspace_controller.go
```diff
@@ -199,7 +199,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// encountered in main reconcile loop.
 		if err == nil {
 			if timeoutErr := checkForStartTimeout(clusterWorkspace); timeoutErr != nil {
-				reconcileResult, err = r.failWorkspace(workspace, timeoutErr.Error(), reqLogger, &reconcileStatus)
+				reconcileResult, err = r.failWorkspace(workspace, timeoutErr.Error(), metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus)
```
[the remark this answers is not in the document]
dkwon17: Done, thank you

Issue #616, opened 2021-09-24
Limit controller object caches to only look at DevWorkspace-related resources


amisevsk:
### Description
The DevWorkspace Operator's controller caches a lot of Kubernetes objects (deployments, jobs, configmaps, secrets, serviceaccounts, pods, services, ingresses, routes). This means that the number of such objects on the cluster (even outside of those created for DevWorkspaces) impacts the memory usage of DWO in large clusters with many resources.

controller-runtime v0.9.0 has the beginnings of a system to restrict the controller cache to specific objects, in order to avoid the issue above. 

### Additional context
https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md

Issue #611, opened 2021-09-23
DevWorkspace Operator crashes with OOMKilled


sleshchenko:
That's a strange bug that happened on DevSanbox Prod.
To reproduce it we may try to redo load testing Tina did.
![Screenshot_20210923_122309](https://user-images.githubusercontent.com/5887312/134531004-3883e2b1-9e69-4900-8810-c4010d269344.png)

```
> kc get pod devworkspace-controller-manager-*** -o=yaml
 - containerID: cri-o://4e196253a9478ddc52a19bedadf162561fcc2b2aa4a6d3d2b62d143601f87266
    image: registry.redhat.io/devworkspace/devworkspace-rhel8-operator@sha256:e68ec2fe7ac27e59641bdfc7794ae99fdfaa60e5b6d0cc0e3f20ab3f7a31bc11
    imageID: registry.redhat.io/devworkspace/devworkspace-rhel8-operator@sha256:e68ec2fe7ac27e59641bdfc7794ae99fdfaa60e5b6d0cc0e3f20ab3f7a31bc11
    lastState:
      terminated:
        containerID: cri-o://4e196253a9478ddc52a19bedadf162561fcc2b2aa4a6d3d2b62d143601f87266
        exitCode: 137
        finishedAt: "2021-09-23T09:22:13Z"
        reason: OOMKilled
        startedAt: "2021-09-23T09:21:33Z"
    name: devworkspace-controller
    ready: false
    restartCount: 7
    started: false
    state:
      waiting:
        message: back-off 5m0s restarting failed container=devworkspace-controller
          pod=devworkspace-controller-manager-6c99cdf947-ch5bg_openshift-operators(d29753b5-6155-4ff9-a920-c27457242e5b)
        reason: CrashLoopBackOff
```

DevWorkspace hanged on Waiting for SA, anything reconcile loop related appeared in DWO logs.

PR #652, opened 2021-10-20
Limit objects cached by DevWorkspace controller to reduce memory usage

amisevsk:
### What does this PR do?
Limits the internal controller cache for objects managed by the DevWorkspace Operator (introduced in controller-runtime `v0.9.0`; see [doc](https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md)). Since we can specify just one selector for limiting the cache, we use

* Existence of the `devworkspace_id` label for most objects (deployments, services, etc.)
* New labels `controller.devfile.io/watch-configmap` and `controller.devfile.io/watch-secret` that *must* be applied to secrets/configmaps we use
    * **This changes the flow for automounting secrets/configmaps and git credentials**
* For roles/rolebindings, we limit the cache to the workspace role/binding based on an internal constant

The downside of doing this is that any objects that do not match the selector cannot be read in the controller. This has two impacts on our design:
1. As mentioned above, it changes the flow for automounting secrets/configmaps, as applying an additional label is required. Without it, the controller won't know those resources exist
2. It requires refactoring how we manage syncing objects to the cluster, as there are two new edge cases to worry about
    * We must make sure any resources we create/update have the appropriate label
    * We need to treat failures to Create an object specially, otherwise we enter a loop if someone removes the required label from an object -- getting an `AlreadyExists` error when trying to create requires us to try and update the object.

To address the second point above, rather than rewrite the reconcile for each place its used, I consolidated all syncing (spec vs cluster) into one package and reworked everywhere else to use this.

The main benefit of this change is drastically reduced memory usage on large clusters: for a cluster with 1000 stopped devworkspaces, ~1000 deployments/services/routes, ~5000 configmaps, ~24000 secets, ~26000 rolebindings, we have
* Current main branch: `~1750Mi` memory usage
* This PR branch: `~90Mi` memory usage

This represents an approximate memory use reduction of 18-19x. In the specific case of the cluster I tested against, it appears (unsurprisingly) that secrets are the main culprit. Testing a variant image that only restricts the cache for secrets reduced memory usage to `~350Mi`. 

I'm opting to restrict the cache for all objects as otherwise memory usage of DWO depends on the objects that exist on the cluster. With the cache restriction, memory use should be governed mainly by how many DevWorkspaces exist on the cluster.

#### Graph of memory usage for DWO while testing various different cases (each image is restarted 5 times as memory usage spikes on startup):
DWO-cache-restriction-metrics

Note the numbers here don't match the ones listed above exactly as these are internal metrics and the numbers above use podmetrics from the cluster. The cases being tested are:
1. Main branch
2. Restrict only objects that use the `devworkspace_id` label (easiest case)
3. Same as 2. but also restrict roles and rolebindings based on name
4. Same as 2. but also restrict secrets and configmaps by new label
5. Restrict everything we can (2 + 3 + 4)
6. Restrict *only* secrets from the cache to see the effect there.

#### Diagram for the new sync object flow:
reconcile-flow-alt

- Closes https://github.com/devfile/devworkspace-operator/issues/616
- Closes https://github.com/devfile/devworkspace-operator/issues/611

### Is it tested? How?
Testing might be tricky:
* We need to make sure I didn't miss adding the required label to any secrets/configmaps.
* We need to make sure all objects that get created can be read (e.g. workspace ServiceAccounts weren't previously getting the devworkspace_id label)
* We ideally need to make sure that updating from a branch that doesn't restrict the cache to the current commit functions as expected and that all owned objects are updated if necessary

I've tried testing the above locally and haven't seen any issues.

### PR Checklist


pkg/provision/sync/sync.go
```diff
@@ -72,7 +72,7 @@ func createObjectGeneric(specObj crclient.Object, api ClusterAPI) error {
 	case err == nil:
 		return NewNotInSync(specObj, CreatedObjectReason)
 	case k8sErrors.IsAlreadyExists(err):
-		// Need to try to update the object to address an edge case where removing a label
+		// Need to try to update the object to address an edge case where removing a labeselectorl
```
sleshchenko, line 82:
```suggestion
		// Need to try to update the object to address an edge case where removing a labeselector
```
?

pkg/provision/sync/sync.go
```diff
@@ -70,6 +74,7 @@ func createObjectGeneric(specObj crclient.Object, api ClusterAPI) error {
 	err := api.Client.Create(api.Ctx, specObj)
 	switch {
 	case err == nil:
+		api.Logger.Info("Created object", "kind", reflect.TypeOf(specObj).Elem().String(), "name", specObj.GetName())
 		return NewNotInSync(specObj, CreatedObjectReason)
```
sleshchenko, line 80: That seems not 100% clear. I don't return to the prev commit where it's introduced and leave thoughts here.
Initially, I understood NewNotInSync (as new object is created, was not in sync).
Then I iterate got the idea but then I checked NotInSyncError:
```go
// NotInSyncError is returned when a spec object is out-of-sync with its cluster counterpart
```

Actually, it may be more precise to specify that is was outdated but then we did a try to make it up to date.
That that's true for `UpdatedObjectReason`, `CreatedObjectReason`, `DeletedObjectReason` but not for `NeedRetryReason`...
Then maybe `NotInSyncError` should not be an error but just reporting the Sync state? and only `UnrecoverableSyncError` is an error that can become SyncError.

Or always  return `SyncState(oldClusterObject, updateClusterObject, madeOperation(create,delete,update,recreate), error)`


Then maybe it would be easier to have a dedicated func to create each of state? `NewObjectCreatedSyncState`, `NewObjectUpdatedSyncState`...

Let me know if it makes sense or not, or clarification is needed, I just put thoughts in not well-organized format )

pkg/cache/cache.go
```diff
@@ -64,9 +65,6 @@ func GetCacheFunc() (cache.NewCacheFunc, error) {
 		&networkingv1.Ingress{}: {
 			Label: devworkspaceObjectSelector,
 		},
-		&routev1.Route{}: {
```
sleshchenko: I saw that on the first commit and started minishift to check it out but now I don't need it :-D

sleshchenko (approved):
I haven't tested through proper load testing flow but with usual testing, it works fine.
Please, after it's merged drop PR for Che operator to update DWO go dep, or create an issue for them.

amisevsk:
> Is it possible to use non-caching client (like here) instead of adding labels just to manage caches?

It would be possible, but goes against the intention of the controller design. With caching, we can avoid making API calls out to the cluster most of the time (only really hitting the API server on create/update calls), so I'd be concerned about performance/behavior if we went that route.

As a rough example, it takes around 10-15 reconciles to start a workspace, and all of these reconciles tend to happen within the first couple of seconds. If during that process we have to list/get secrets/configmaps multiple times (in the case of mounting git credentials, we might make two secrets, two configmaps, and read secrets + configmaps 2-3 times _per_ reconcile), we're looking at potentially 50+ API requests per started workspace that are not present with caching. 

We do use a non-caching client in a few places, but those are reserved for startup or one-time tasks. In load testing, I've seen 200+ reconciles per second in the controller, which could result in thousands of get/list calls per second.

The other concern I'd have is in throughput, as making actual calls to the API will necessarily be slower than reading from the cache. This could directly impact startup time when under load, which we'd prefer to avoid.

nickboldt:
Is this fixed in DWO 0.10? Asking because 

image

amisevsk:
v0.10 was branched before this PR was merged; it'll be included in v0.11

PR #653, opened 2021-10-20
Exclude serviceaccount olm

amisevsk:
### What does this PR do?
Drops the serviceaccount from the OLM bundle manifests, as including it results in scorecard failures down the line.

This is an especially annoying fix, as removing it during the kustomize build is very difficult due to the manager.yaml deployment depending on the `$(SERVICE_ACCOUNT_NAME)` var, which is derived from the serviceaccount object.

Scorecard failures. OLM should create serviceaccount on its own.

### PR Checklist


amisevsk:

Already cherry-pick to 0.10.x branch.

Issue #607, opened 2021-09-22
Allow workspaces to disable/modify automatic project cloning


amisevsk:
### Description
The DevWorkspace Operator automatically adds an init container to ensure projects are cloned before starting workspaces. However, since init containers in Kubernetes run sequentially, it may be possible to improve DevWorkspace startup time by allowing the init container to do multiple things in parallel (e.g. download .vsix extensions for Theia/VSCode).

DWO should have a mechanism to allow devworkspaces to override/disable the default behavior as necessary. There are generally two ways we could do this

1. Add DevWorkspace attribute to disable automatically injecting project clone container into a DevWorkspace
    - Pros: simple, easy to implement, and easy to understand (if you add this attribute, you are responsible for cloning projects)
    - Cons: more work for users, as recreating functionality (clone before start) requires adding their own init container
2. Add DevWorkspace attribute to allow devworkspaces to override the image used for project clone
    - Pros: potentially simpler to use, as you could potentially just extend the current project-clone image with additional functionality
    - Cons: ties the implementation of project clone container to any use case users might want to use an init container for

Personally, I think option 1 is better, since option 2 risks opening project clone functionality into a larger "use this image as an init container" feature. This could mean that making changes to project-clone breaks devfiles in ways we can't test, and couples DevWorkspace instances with the DevWorkspace Operator in a way I'd like to avoid. A particular concern is that we set the memory/cpu limit for project clone based on our known use-case, and it would be hard to accomodate additional functionality in a stable way.

### Additional info
The net effect of adding the project clone container on a DevWorkspace is importing the DevWorkspaceTemplate below, so it should be fairly simple to re-implement custom project cloning as a plugin.
```yaml
kind: DevWorkspaceTemplate
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: project-clone
spec:
  components:
  - name: project-clone
    container:
      image: quay.io/devfile/project-clone:next
      mountSources: true
      memoryLimit: "1Gi"
      memoryRequest: "128Mi"
      cpuLimit: "1000m"
      cpuRequest: "100m"
  commands:
  - id: clone-projects
    apply:
      component: project-clone
  events:
    prestart:
      - clone-projects
```

sleshchenko:
+1 for option №1 from my end.

PR #673, opened 2021-11-15
Allow workspaces to disable project cloning

amisevsk:
### What does this PR do?
Adds top-level attribute `controller.devfile.io/project-clone` to enable workspaces to disable automatic project clone.

Closes https://github.com/devfile/devworkspace-operator/issues/607

### Is it tested? How?
Create a DevWorkspace with 
```
    attributes:
      controller.devfile.io/project-clone: disable
```
and verify that no project-clone init container is created.

E.g.
```yaml
cat <<EOF | oc apply -f -
kind: DevWorkspace
metadata:
  name: theia-next
spec:
  started: true
  template:
    attributes:
      controller.devfile.io/project-clone: disable
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
    commands:
      - id: say-hello
        exec:
          component: theia-ide
EOF
```

### PR Checklist

### DIFF
--- a/pkg/constants/attributes.go
+++ b/pkg/constants/attributes.go
@@ -37,6 +37,14 @@ const (
 	//         value: VAL_2
 	WorkspaceEnvAttribute = "workspaceEnv"
 
+	// ProjectCloneAttribute configures how the DevWorkspace will treat project cloning. By default, an init container
+	// will be added to the workspace deployment to clone projects to the workspace before it starts. This attribute
+	// must be applied to top-level attributes field in the DevWorkspace.
+	// Supported options:
+	// - "disable" - Disable automatic project cloning. No init container will be added to the workspace and projects
+	//               will not be cloned into the workspace on start.
+	ProjectCloneAttribute = "controller.devfile.io/project-clone"
+
 	// PluginSourceAttribute is an attribute added to components, commands, and projects in a flattened
 	// DevWorkspace representation to signify where the respective component came from (i.e. which plugin
 	// or parent imported it)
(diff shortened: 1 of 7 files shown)


pkg/constants/attributes.go
```diff
@@ -17,10 +17,39 @@ package constants
...
+	//         value: VAL_2
+	WorkspaceEnvAttribute = "workspaceEnv"
+
+	// ProjectCloneAttribute configures how the DevWorkspace will treat project cloning. By default, an init container
+	// will be added to the workspace deployment to clone projects to the workspace before it starts. This attribute
+	// must be applied to top-level attributes field in the DevWorkspace.
+	// Supported options:
+	// - "disable" - Disable automatic project cloning. No init container will be added to the workspace and projects
+	//               will not be cloned into the workspace on start.
+	ProjectCloneAttribute = "controller.devfile.io/project-clone"
```
JPinkney, line 46: This is optional but I wonder if it makes more sense to do:
```
"controller.devfile.io/clone-project"
```
that way its naturally a boolean and you don't need to specify if its disabled or not. You can just specify whether its true or false (or doesn't exist at all)

JPinkney (approved):
Just one small comment but otherwise its working for me

dkwon17:
Tested on crc (OpenShift 4.9), works well for me.

pkg/constants/attributes.go
amisevsk: The idea with this naming was to leave it open in case there's a need for using a custom image here -- i.e. this version leaves it open to potentially supporting something like
```yaml
attributes:
    controller.devfile.io/project-clone: quay.io/amisevsk/my-custom-project-clone:latest
```
but now that I think of it, maybe it just makes sense to keep it simple and use a separate label in the off chance we need to support this. I'll update.

Issue #618, opened 2021-09-25
DWO should track workspaces that go from `Running` to `Starting`


amisevsk:
### Description
The DevWorkspace operator currently makes no note of workspaces that go from the `Running` phase to the `Starting` phase. This could be useful information in diagnosing certain issues (e.g. a pod that is continually being updated). In addition, not detecting this event can skew metrics for successful startups, as each time a pod restarts it is counted as a successful start (though with start duration based of when the workspace was initially started).

One factor that should be considered is that sometimes these restarts are expected (if e.g. the user updates their running workspace to use a new plugin).

### Additional context
In load testing DWO, I noticed that some pods were continually being recreated, which suggests some sort of issue but isn't tracked by DWO.

max-cx:
Hi, a question to the assignee of this issue:

Will the outcome require any changes to the relevant content of the _Installation Guide_ or _Administration Guide_ or _End-user Guide_?

Yes/No?

PR #686, opened 2021-11-24
Only track fresh starts of workspaces in workspace started metrics

JPinkney:

### What does this PR do?
This PR makes it so that when an already running workspace is reconciled we don't count that as a workspace start in the metrics

https://github.com/devfile/devworkspace-operator/issues/618

### Is it tested? How?

1. Follow the metrics docs to run prometheus/grafana: https://github.com/devfile/devworkspace-operator/tree/main/doc/grafana.
2. Create a devworkspace and wait for it to start `oc apply -f samples/theia-next.yaml`
3. Observe that the workspace start metric was updated in grafana
4. Change the devworkspace such that its reconciled (I tested by adding a postStart event into the devworkspace)
5. Observe that the workspace started metric is not updated
6. Scale the devworkspace down by setting `started: false`
7. Scale the devworkspace up by setting `started: true`
8. Observe that the devworkspace started metric is updated

### PR Checklist

### DIFF
--- a/controllers/workspace/metrics/update.go
+++ b/controllers/workspace/metrics/update.go
@@ -38,8 +38,11 @@ func WorkspaceStarted(wksp *dw.DevWorkspace, log logr.Logger) {
 // encountered, the provided logger is used to log the error. This function assumes the provided workspace has
 // fully-synced conditions (i.e. the WorkspaceReady condition is present).
 func WorkspaceRunning(wksp *dw.DevWorkspace, log logr.Logger) {
-	incrementMetricForWorkspace(workspaceStarts, wksp, log)
-	incrementStartTimeBucketForWorkspace(wksp, log)
+	_, ok := wksp.GetAnnotations()[constants.DevWorkspaceStartedAtAnnotation]
+	if !ok {
+		incrementMetricForWorkspace(workspaceStarts, wksp, log)
+		incrementStartTimeBucketForWorkspace(wksp, log)
+	}
 }
 
 // WorkspaceFailed updates metrics for workspace entering the 'Failed' phase. If an error is encountered, the provided
--- a/controllers/workspace/metrics/update.go
+++ b/controllers/workspace/metrics/update.go
@@ -22,12 +22,16 @@ import (
 
 	"github.com/devfile/devworkspace-operator/pkg/conditions"
 	"github.com/devfile/devworkspace-operator/pkg/config"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
 )
 
 // WorkspaceStarted updates metrics for workspaces entering the 'Starting' phase, given a workspace. If an error is
 // encountered, the provided logger is used to log the error.
 func WorkspaceStarted(wksp *dw.DevWorkspace, log logr.Logger) {
-	incrementMetricForWorkspace(workspaceTotal, wksp, log)
+	_, ok := wksp.GetAnnotations()[constants.DevWorkspaceStartedAtAnnotation]
+	if !ok {
+		incrementMetricForWorkspace(workspaceTotal, wksp, log)
+	}
 }
 
 // WorkspaceRunning updates metrics for workspaces entering the 'Running' phase, given a workspace. If an error is
--- a/pkg/timing/timing.go
+++ b/pkg/timing/timing.go
@@ -42,7 +42,11 @@ func SetTime(timingInfo map[string]string, event string) {
 	if _, set := timingInfo[event]; set {
 		return
 	}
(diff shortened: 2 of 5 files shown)


pkg/timing/timing.go
```diff
@@ -42,7 +42,11 @@ func SetTime(timingInfo map[string]string, event string) {
 	if _, set := timingInfo[event]; set {
 		return
 	}
-	timingInfo[event] = strconv.FormatInt(time.Now().UnixNano()/1e6, 10)
+	timingInfo[event] = CurrentTime()
+}
+
+func CurrentTime() string {
+	return strconv.FormatInt(time.Now().UnixNano()/1e6, 10)
```
amisevsk, line 49: I know this is a rearrangement of what's already here, but it might make sense to do sometime line
```suggestion
	return strconv.Itoa(time.Now().UnixNano()/time.Millisecond)
```

controllers/workspace/metrics/update.go
```diff
@@ -22,12 +22,16 @@ import (
...
 )
 
 // WorkspaceStarted updates metrics for workspaces entering the 'Starting' phase, given a workspace. If an error is
 // encountered, the provided logger is used to log the error.
 func WorkspaceStarted(wksp *dw.DevWorkspace, log logr.Logger) {
-	incrementMetricForWorkspace(workspaceTotal, wksp, log)
+	_, ok := wksp.GetAnnotations()[constants.DevWorkspaceStartedAtAnnotation]
+	if !ok {
+		incrementMetricForWorkspace(workspaceTotal, wksp, log)
+	}
```
amisevsk, line 34: I think this approach is good, but will only solve half the problem. With this PR, we
1. Set the started-at annotation on the DevWorkspace on start
2. Only increment the started metric if the started-at annotation is not present

This kind of reveals how weirdly we've implemented startup/running metrics :smile: -- `WorkspaceStarted` is called in two places (AFAIK): 
* In a guarded conditional that only executes if the workspace's phase is transitioning from `Stopped` or `Failed/Failing` [here](https://github.com/devfile/devworkspace-operator/blob/9f867aeac1d014d315f53594953e7a4c032a9943/controllers/workspace/devworkspace_controller.go#L186)
* Whenever the Workspace switches to `Starting` from another phase [here](https://github.com/devfile/devworkspace-operator/blob/9f867aeac1d014d315f53594953e7a4c032a9943/controllers/workspace/status.go#L241-L242)

The second case is the one we're limiting with the `started-at` annotation, but _whenever_ the second case is called, the workspace will have the `started-at` annotation -- we may as well just remove the Starting condition from `updateMetricsForPhase`.

One thing that's missing from this PR is that we call [`WorkspaceRunning`](https://github.com/devfile/devworkspace-operator/blob/9f867aeac1d014d315f53594953e7a4c032a9943/controllers/workspace/status.go#L237-L238) every time the workspace goes from `Starting -> Running`, even if that workspace restarted due to e.g. a crashed container. This can result in e.g. `1` workspace Start and `n` workspace Running metrics for a single workspace.

A DevWorkspace that should (:crossed_fingers:) reproduce this is something like
```yaml
kind: DevWorkspace
metadata:
  name: sleep-ten
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "sleep"
           - "10s"
```

JPinkney: I think the new commit should fix this. With the new commit the process is essentially:

- Workspace started metric is only added on [line 186](https://github.com/devfile/devworkspace-operator/pull/686/files?diff=unified&w=0#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R186)
- If the workspace eventually is running, [line 209](https://github.com/devfile/devworkspace-operator/pull/686/files?diff=unified&w=0#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R209) records that the workspace is running (as long as it doesn't have a startedAt annotation yet) then the startAt time is synced to the cluster
- If the workspace is then running -> restarted -> running -> restarted
    - running won't increase because the startedAt annotation is already set the first time it got to running
- If the workspace crash loops because of something like ImagePullBackOff
    - starting gets updated but running doesn't  
- If the workspace is scaled down
    - startedAt is removed
    - restarting it again will trigger line 186 and then repeat all the other actions

amisevsk (approved):

amisevsk:
Well done, this looks like a frustrating one to figure out :smile:

Issue #614, opened 2021-09-24
Support Pod tolerations for DevWorkspace Pods


metlos:
### Description
To support node selection for workspace pods, it would be great if we supported pod tolerations and node selector for DevWorkspace pods.

It would be best if this was a noop but we need to investigate if the same can be done on namespace level - e.g. by labeling namespaces such that addmission controllers can select the nodes automatically, see https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector

sleshchenko:
As far I as understand @l0rd will provide a better requirement for this feature, like should it be defined on namespace object/on a specific workspace level.

l0rd:
I have looked at past issues and here is the [original](https://github.com/eclipse/che/issues/14562) where it's mentioned that a central config is required first. NodeSelector per workspace was considered a successive iteration. 

I also found a customer request:
> node selector would be great to have. If that can be defined in dev file e.g, that would be best. 
So being able to specify that at workspace level is important too (I don't think that we should do that at devfile spec level though as that would make the devfile less portable).

Summing it up:
- Higher priority: being able to have current, global, che-server configuration in devworkpace world too
- Lower priority: being able to specify NodeSelector at the workspace level

max-cx:
Hi, a question to the assignee of this issue:

Will the outcome require any changes to the relevant content of the _Installation Guide_ or _Administration Guide_ or _End-user Guide_?

Yes/No?

amisevsk:
Supporting this makes sense as a part of the main Devfile API; I've created https://github.com/devfile/api/issues/689 and will follow up.

skabashnyuk:
Now pod teleration look like this 
```
# Optionally configures tolerations for workspace Pod. Format is a string representing a JSON Array of taint tolerations,
# or `NULL` to disable it. The objects contained in the array have to follow the
# link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#toleration-v1-core[toleration v1 core specifications].
# Example: `[{"effect":"NoExecute","key":"aNodeTaint","operator":"Equal","value":"aValue"}]`
che.workspace.pod.tolerations_json=NULL
```
 @amisevsk can you tell us how this config should look like so devworkspace controller can understand it?

amisevsk:
[...]

PR #696, opened 2021-12-03
Namespace pod tolerations

amisevsk:
### What does this PR do?
Adds support for namespace-level configuration of pod tolerations and node selector via *namespace* annotations

* `"controller.devfile.io/pod-tolerations"`
* `"controller.devfile.io/node-selector"`

The value of the tolerations annotation must be json-formatted `[]corev1.Toleration`. The value of the node-selector annotation must be `map[string]string`. If these annotations are applied to a namespace, their values are used for *all* pods created for workspaces in that namespace. This includes:
* The workspace pod itself
* The pod created as part of the PVC cleanup job
* The async storage server deployment

Closes https://github.com/devfile/devworkspace-operator/issues/614 (assuming this implementation is sufficient)

### Is it tested? How?
Instructions written for minikube, on other clusters node name should be changed
1. Taint a node and add a label to it
    ```bash
    # Apply a taint to the minikube node
    kubectl taint nodes minikube key1=value1:NoSchedule
    # Apply a new label to the minikube node
    kubectl label nodes minikube test-node-selector=test
    ```
2. Add annotations to the namespace to be used:
    ```yaml
    metadata:
      annotations:
        controller.devfile.io/node-selector: |
          {
            "test-node-selector": "test"
          }
        controller.devfile.io/pod-tolerations: |
          [
            {
              "key": "key1",
              "operator": "Equal",
              "value": "value1",
              "effect": "NoSchedule"
            }
          ]
    ```
    Note: formatting can be finicky, if annotations cannot be unmarshalled, all DWO activity in the namespace fails.
3. Create DevWorkspace and verify it can start in the cluster (and deployment has appropriate nodeSelector/tolerations)
    ```bash
    ```
4. Once workspace is started, delete it and verify cleanup succeeds
    ```
    kubectl delete dw theia-next --wait
    ```

### PR Checklist

### DIFF
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -102,4 +102,12 @@ const (
 	// NamespacedConfigLabelKey is a label applied to configmaps to mark them as a configuration for all DevWorkspaces in
 	// the current namespace.
 	NamespacedConfigLabelKey = "controller.devfile.io/namespaced-config"
+
+	// NamespacePodTolerationsAnnotation is an annotation applied to a namespace to configure pod tolerations for all workspaces
+	// in that namespace. Value should be json-encoded []corev1.Toleration struct.
+	NamespacePodTolerationsAnnotation = "controller.devfile.io/pod-tolerations"
+
+	// NamespaceNodeSelectorAnnotation is an annotation applied to a namespace to configure the node selector for all workspaces
+	// in that namespace. Value should be json-encoded map[string]string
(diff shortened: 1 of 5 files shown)


JPinkney (approved):
I can't really comment on the approach since I'm not familiar with pod tolerations/node selectors but I tried it out and everything is working and from a code perspective it looks fine

Issue #699, opened 2021-12-06
If project-clone container is killed while cloning, repos are left in inconsistent state


amisevsk:
### Description
The project-clone container checks if directories exist for projects in the devfile, and does nothing if they do. This is done to avoid modifying repos on subsequent startups in case local changes have been made, but also means that if the project clone container fails during cloning/unzipping, the repos are left in an inconsistent state.

To avoid this, we should try to make the project-clone container act as atomically as possible, so that it either does nothing, or it sets up projects successfully.

### Additional context
Originally reported as "problem 7" here: https://github.com/eclipse/che/issues/20785. In this example, the project clone container is restarted because a slow OpenShift cluster adds an imagePullSecret to the workspace serviceaccount after creation, triggering an update to the workspace deployment (and thus the creation of a new pod)

PR #700, opened 2021-12-06
Make project-clone work in a temporary directory; copy files at end

amisevsk:
### What does this PR do?
In order to avoid issues where the project-clone container is killed during its initial run, leaving repos in a half-configured state, do all operations in a temporary directory and copy repos to their expected locations at the very end.

Closes https://github.com/devfile/devworkspace-operator/issues/699

### Is it tested? How?
Can be tested using the image `quay.io/amisevsk/project-clone:atomic-clone` (e.g. in the relevant `RELATED_IMAGE` env var or by setting `PROJECT_CLONE_IMG` when installing DWO). To test both git and zip type projects, use the devfile
```yaml
kind: DevWorkspace
metadata:
  name: theia-next
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
      - name: devworkspace-operator-zip
        zip:
          location: https://github.com/devfile/devworkspace-operator/archive/refs/heads/main.zip
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
EOF
```
and check that
1. Projects are setup successfully
2. Project-clone does nothing when the workspace is restarted
3. No temp directories or files are present in `$PROJECTS_ROOT`
4. (If you can figure out how) kill project clone while it's preparing projects and make sure the subsequent run behaves as above

### PR Checklist

### DIFF
--- a/project-clone/internal/git/setup.go
+++ b/project-clone/internal/git/setup.go
@@ -18,46 +18,68 @@ package git
 import (
 	"fmt"
 	"log"
+	"os"
+	"path"
 
-	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 
 	"github.com/devfile/devworkspace-operator/project-clone/internal"
 )
 
-func SetupGitProject(project v1alpha2.Project) error {
+func SetupGitProject(project dw.Project) error {
 	needClone, needRemotes, err := internal.CheckProjectState(&project)
 	if err != nil {
 		return fmt.Errorf("failed to check state of repo on disk: %s", err)
 	}
 	if needClone {
-		err := CloneProject(&project)
-		if err != nil {
-			return fmt.Errorf("failed to clone project: %s", err)
-		}
-		repo, err := internal.OpenRepo(&project)
-		if err != nil {
-			return fmt.Errorf("failed to open existing project in filesystem: %s", err)
-		} else if repo == nil {
-			return fmt.Errorf("unexpected error while setting up remotes for project: git repository not present")
-		}
-		if err := SetupRemotes(repo, &project); err != nil {
-			return fmt.Errorf("failed to set up remotes for project: %s", err)
-		}
-		if err := CheckoutReference(repo, &project); err != nil {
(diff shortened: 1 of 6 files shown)


project-clone/internal/git/setup.go
```diff
@@ -18,46 +18,68 @@ package git
...
 	} else {
 		log.Printf("Project '%s' is already cloned and has all remotes configured", project.Name)
+		return nil
+	}
+}
+
+func doInitialGitClone(project *dw.Project) error {
+	// Clone into a temp dir and then move set up project to PROJECTS_ROOT to try and make clone atomic in case
+	// project-clone container is terminated
+	tmpClonePath := path.Join(internal.CloneTmpDir, internal.GetClonePath(project))
```
benoitf, line 47: how much seconds does it take to move from tmp to the cloned directory ?

amisevsk: Depends on a variety of factors, but it's a linux move at the directory level so presumably almost no time -- it's just modifying the path.

benoitf: ok I see, temp directory is inside the PROJECTS_ROOT folder and not like in default temporary directory of the system

JPinkney (approved):

project-clone/internal/git/setup.go
amisevsk: Yeah, I had to keep it that way otherwise we're copying files over one-by-one and risk getting stuck in a half-setup repo as before. The new flow is
1. Create `$PROJECTS_ROOT/project-clone-<random-string>` directory to hold temporary files
2. Trap on exit to remove temp dir, so that if we get killed early no changes are made to the PVC
3. Clone/unzip to `$PROJECTS_ROOT/project-clone-<random-string>/project-name`
4. Move `project-name` to `$PROJECTS_ROOT`
5. Temp dir is removed on exit

amisevsk:
Validation is failing as `os.MkDirTemp` was only added in Go 1.16 :(

amisevsk:
Blocked by https://github.com/devfile/devworkspace-operator/pull/705

Issue #694, opened 2021-12-02
Support user name and email in gitconfig


metlos:
### Description
To achieve feature parity with the Che server, DWO should support setting the user name and email in the git config.

For Devworkspace, it seems reasonable to use the same approach as with other parts of the git configuration and have a dedicated label for marking a config map with this information. E.g.

```yaml
kind: ConfigMap
metadata:
  labels:
    "controller.devfile.io/git-user-config": "true"
    "controllder.devfile.io/watch-configmap": "true"
data:
  name: Will E. Coyote
  email: wcoyote@acme.com
```

UPDATE:
This issue has the following subtasks:
- [x] Allow injecting gitconfigs using a configmap ([PR](https://github.com/devfile/devworkspace-operator/pull/710)) (part of milestone 3)
- [ ] Retrieving user email and name tasks (not part of milestone 3):
  - [ ] https://github.com/eclipse/che/issues/21115 
  - [ ] https://github.com/eclipse/che/issues/20937
  - [ ] https://github.com/eclipse/che/issues/20938

ibuziuk:
@l0rd @metlos the issue has been added to the endgame plan recently. is it expected to be there or could it sleep for the future releases?

metlos:
I'll let @l0rd judge this. I added it to the list just so that it is not forgotten about before we declare feature parity.

l0rd:
We have discussed this issue today. We have identified 2 distinct problems:

1. Retrieve the user name and email before a workspace is started (from the OIDC provider, git services APIs, asking the user etc...)
2. Use a `ConfigMap` to define user git configuration that works for both command line git and editors git UI

If 2 is implemented as in https://github.com/devfile/devworkspace-operator/pull/708 (mounted as `/etc/gitconfig` in all containers of the workspace pod) then the behavior is correct:
- If `/etc/gitconfig` exists, Theia loads values from it at statup time
- Setting Theia git username and email preferences adds them into `~/.gitconfig` in theia-ide container (that have priority over `/etc/gitconfig`)

For 1. in priority order, `che-server/che-operator` should try to get username and email:
- from the OIDC provider when creating the namespace
- from the git service provider (GitHub, Gitlab, Bitbucket) when creating a workspace
- from the git service provider (GitHub, Gitlab, Bitbucket) triggered by the user clicking on a new button in Dashboard("login with <service-provider>" button)

l0rd:
[...]

PR #708, opened 2021-12-10
[feat] Provision user and email credentials to gitconfig

JPinkney:

### What does this PR do?
This PR automatically provisions user and email credentials to the gitconfig whenever the label: controller.devfile.io/git-user-credential is true.

If you want to provision git credentials to your gitconfig you can use:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: sample-user-credentials
  labels:
    controller.devfile.io/git-user-credential: "true"
data:
  name: sampleName
  email: sampleName@sampleName.com
```

https://github.com/devfile/devworkspace-operator/issues/694

### Is it tested? How?

Tested manually using the above configmap, but also added unit tests

### PR Checklist

### DIFF
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -54,8 +54,8 @@ const (
 	DevWorkspaceGitTLSLabel = "controller.devfile.io/git-tls-credential"
 
 	// DevWorkspaceGitUserLabel is the label key to specify if the configmap is credentials for accessing a git server.
-	// Configmap must contain the following data:
-	// username: the username of the user in git
+	// Configmap can contain any of the following data:
+	// name: the name of the user in git
 	// email: the email of the user in git
 	DevWorkspaceGitUserLabel = "controller.devfile.io/git-user-credential"
 
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -53,6 +53,12 @@ const (
 	// If the git host is not defined then the certificate will be used for all http repositories.
 	DevWorkspaceGitTLSLabel = "controller.devfile.io/git-tls-credential"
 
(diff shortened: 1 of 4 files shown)


l0rd:
This is good as first implementation. But rather than a specialized `ConfigMap` for git config username and email we may want to support a new annotation `controller.devfile.io/sub-path` for generic `ConfigMaps` labelled with `controller.devfile.io/mount-to-devworkspace: "true"`. 

That would allow to mount a complete gitconfig (not only user name and email) and to mount any file in an existing folder:. For example:

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: git-config-cm
  annotations:
    controller.devfile.io/mount-path: /etc/gitconfig
    controller.devfile.io/sub-path: gitconfig
  labels:
data:
  gitconfig: |
    [user]
      email = jojo@gmail.com
      name = Jojo
    [core]
      editor = vim
    [alias]
      ...
```

amisevsk:
> But rather than a specialized ConfigMap for git config username and email we may want to support a new annotation controller.devfile.io/sub-path for generic ConfigMaps labelled with controller.devfile.io/mount-to-devworkspace: "true".

I beat you to the punch by about an hour :D https://github.com/devfile/devworkspace-operator/pull/710

l0rd:
> I beat you to the punch by about an hour :D #710

I saw that earlier today! I was probably still thinking at the right way to inject a `gitconfig` when you submitted that PR. And you addressed the problem in a simpler way too. That's magic 🧙 🪄  Unfortunately I noticed it too late and I've lost one hour yesterday 😄

ibuziuk:
@amisevsk I believe we can close this one, right?

amisevsk:
I think so, I'll leave the branch in place in case it's needed later as this was initially requested by the Che server IIRC.

PR #710, opened 2021-12-13
Add automount subpaths option

amisevsk:
### What does this PR do?
Add `subpath` option to the `controller.devfile.io/mount-as` annotation, enabling configmaps/secrets to be mounted as subpath volumemounts in the devworkspace.

This change also makes it so that the DevWorkspace Operator does not create the `devworkspace-gitconfig` map when it would be empty, as this interfered with what I was trying to test (mounting my gitconfig into workspaces)

The main use-case here is mounting files from a configmap/secret to an already-populated directory in the container. The benefit of using "subpath" is that files can be mounted into existing directories, with the downside that changes on the cluster side are not propagated into running workspaces.

### Is it tested? How?
E.g. provisioning your local `~/.gitconfig` into all DevWorkspaces in a namespace:
1. Create configmap with gitconfig contents:
    ```bash
    kubectl create configmap git-config \
      --from-file=gitconfig=${HOME}/.gitconfig
    
    kubectl patch cm git-config --type merge -p '
    metadata:
      labels:
      annotations:
        controller.devfile.io/mount-path: "/etc/"
        controller.devfile.io/mount-as: "subpath"
    '
    ```
2. Start a DevWorkspace
3. Exec into the workspace and execute `git config --list --show-origin`

### PR Checklist

### DIFF
--- a/docs/additional-configuration.md
+++ b/docs/additional-configuration.md
@@ -92,8 +92,11 @@ By default, resources will be mounted based on the resource name:
 Mounting resources can be additionally configured via **annotations**:
 * `controller.devfile.io/mount-path`: configure where the resource should be mounted
 * `controller.devfile.io/mount-as`: for secrets and configmaps only, configure how the resource should be mounted to the workspace
-    * If `controller.devfile.io/mount-as: file`, the configmap/secret will be mounted as files within the mount path. This is the default behavior
-    * If `controller.devfile.io/mount-as: env`, the keys and values in the configmap/secret will be mounted as environment variables in all containers in the DevWorkspace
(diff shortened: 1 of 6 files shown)


JPinkney (approved):

amisevsk:
> What's controller.devfile.io/watch-configmap: "true" for?

It's a required label to enable the configmap to be cached by the Operator. In https://github.com/devfile/devworkspace-operator/pull/652 we limited the controller's objects cache based on labels in order to reduce memory usage (on large clusters, not doing this means caching e.g. every secret/configmap in-memory in the controller).

For configmaps and secrets, there's no single label that's used for all use-cases, so we had to add explicit labels `watch-configmap` and `watch-secret` to enable the DevWorkspace Operator to track them. Without these labels, secrets and configmaps are invisible to the DevWorkspace Operator.

Issue #678, opened 2021-11-18
Enable proxy support for DevWorkspaces


amisevsk:
### Description
Currently, DevWorkspaces have no way to specify proxy settings, so built-in containers such as project-clone will fail to clone projects if running behind a proxy. We need to define a way to specify proxy settings and figure out how to propagate these down to the workspace pods.

### Additional context
Originally brought as a CRW integration issue: https://issues.redhat.com/browse/CRW-2473

amisevsk:
cc: @ibuziuk @l0rd this is likely a milestone 3 issue

skabashnyuk:
in Eclipse Che we  have this https://github.com/eclipse-che/che-operator/pull/1027

metlos:
yes, mounting of proxy settings to workspace containers should already work. This issue is about supporting that also in built-in containers, is that right? I am al little bit out of the loop on DWO capabilities, so I didn't realize project clone happens in DWO and so is affected by this, too.

l0rd:
@metlos cloning happens in initContainers, the mechanism that works for regular containers should work for other containers too.

PR #711, opened 2021-12-14
Enable running DWO workspaces behind a proxy

amisevsk:
### What does this PR do?
* Add support for proxy configuration in DWO. Values are propagated as env vars to containers in workspaces
* Add support for reading cluster Proxy configuration on OpenShift
* Extend DevWorkspaceOperatorConfig CRD to include proxy configuration overrides

The git history for this PR is a little bit of a mess, so it may not make a huge amount of sense to go through it commit by commit, unless you want to follow my thought process.

Closes https://github.com/devfile/devworkspace-operator/issues/678

### Is it tested? How?
1. Set up `crc` with a proxy (see below)
2. Start DWO and verify that proxy configuration is read from the cluster -- should have log `Resolved proxy configuration` with the current configuration
3. Update the DWOC to add proxy settings, check that in-controller configuration is updated
4. Revert DWOC proxy settings and check that defaults from cluster are restored

1. Check behavior on minikube; should be able to set proxy settings in DWOC and defaults should be empty as there's no proxies.config.openshift.io CRD

### Setting up `crc` with a proxy

<details>
  <summary>Click to expand</summary>

This is just the process I followed, there's likely a better way. This assumes that initially crc is not started (might also need to `crc delete`) and that `crc` is running locally (i.e. not remotely)
1. Start Squid proxy in docker:
    ```bash
    docker run --detach \
      --name=squid \
      --restart=always \
      --publish 3128:3128 \
      --volume /srv/docker/squid/cache:/var/spool/squid3 \
      docker.io/sameersbn/squid:3.3.8-23
    ```

    Note: if the container above is stuck restarting, you may need to run `sudo chcon -Rt svirt_sandbox_file_t /srv/docker/squid/cache`
2. Start following Squid access logs to verify it's being used:
    ```bash
    docker exec -it squid tail -f /var/log/squid3/access.log
    ```
3. Configure `crc` to use proxy
    ```
    IP_ADDR=<address of current machine on lan>
    CRC_IP_ADDR=$(crc ip)
    crc config set http-proxy "http://${IP_ADDR}:3128"
    crc config set https-proxy "http://${IP_ADDR}:3128"
    crc config set no-proxy ${IP_ADDR}/24,${CRC_IP_ADDR},api.crc.testing,canary-openshift-ingress-canary.apps-crc.testing,console-openshift-console.apps-crc.testing,default-route-openshift-image-registry.apps-crc.testing,downloads-openshift-console.apps-crc.testing,oauth-openshift.apps-crc.testing
    ```
    (some of these no-proxy entries are probably unnecessary)
4. `crc start`

</details>

### PR Checklist


amisevsk:
I forgot to make DWO itself use the proxy, working on it.

amisevsk:
Updated DWO to use proxy settings for its own HTTP clients. However, due to how the proxy is configured, these settings are only set once when the controller is being started, and changes to the proxy settings require a restart of the controller deployment. We can't restart the controller from itself easily, as the usual way of triggering a rollout (editing the pod's spec in the deployment) does not work in OLM, as the CSV owns the deployment and reverts changes.

I've updated the docs on the DevWorkspaceOperatorConfigs struct to reflect this.

JPinkney (approved):
Just tried on a cluster bot cluster and I'm able to see that the proxy configuration is resolved but when I look at the devworkspaceoperatorconfiguration I just have:
```
config:
  routing:
    defaultRoutingClass: basic
  workspace:
    imagePullPolicy: Always
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: >
      {"apiVersion":"controller.devfile.io/v1alpha1","config":{"routing":{"clusterHostSuffix":null,"defaultRoutingClass":"basic"},"workspace":{"imagePullPolicy":"Always"}},"kind":"DevWorkspaceOperatorConfig","metadata":{"annotations":{},"name":"devworkspace-operator-config","namespace":"devworkspace-controller"}}
```

I'm assuming that the proxy settings are just stored internally in the operator unless they're overridden by the devworkspaceoperatorconfig?

Also the proxy settings seem to be propagated correctly to a devworkspace as well :+1:

amisevsk:
> I'm assuming that the proxy settings are just stored internally in the operator unless they're overridden by the devworkspaceoperatorconfig?

Yes -- in fact the DevWorkspaceOperatorConfig only stores manual overrides and does not need to exist at all for DWO to function. Its existence when testing changes is a side-effect of us wanting to automate config via env vars in the `make` rules resulting in it always being created even if defaults are used. If you check logs when starting DWO with the above DWOC, it should log something like `Updated configuration to [(default config)]`.

PR #712, opened 2021-12-15
Add polling to the custom catalog source to enable finding new bundles

amisevsk:
### What does this PR do?
Adds polling to the custom catalogsource to enable it to find new bundles as they are pushed.

### PR Checklist

### DIFF
--- a/catalog-source.yaml
+++ b/catalog-source.yaml
@@ -8,3 +8,6 @@ spec:
   image: quay.io/devfile/devworkspace-operator-index:next
   publisher: Red Hat
   displayName: DevWorkspace Operator Catalog
+  updateStrategy:
+    registryPoll:
+      interval: 1d


JPinkney (approved):

AObuchow:
@amisevsk I recently had to test installing DWO with OLM and found that I had to remove the `updateStrategy` field for the catalog source's status to become `READY`. I'm not sure if this should be considered a bug as it might only affect development, but I wanted to bring it up. I can make a bug about this, but wanted to ping you about it first.

amisevsk:
I usually end up patching the catalogsource to something like 5s if I get frustrated; I've opened  https://github.com/devfile/devworkspace-operator/pull/910 to hopefully strike a good middle ground (as far as I know, the catalogsource should be readied immediately, so I don't know why shortening the polling interval in the first place is required)

Issue #707, opened 2021-12-09
Async storage sidecar should not use automount resources or workspaceEnv


amisevsk:
### Description
The async storage sidecar, as provisioned when async storage is used, has one role in the workspace (rsync files to the async server). As such, it should not be impacted by any other workspace configuration:
* The workspaceEnv should not apply to it
* Automatically mounted volumes, configmaps, and secrets should not be mounted to it

Without these changes, attempting to e.g. automount ssh keys to `/etc/ssh/private` will cause workspaces to fail to start when async storage is used, as this will conflict with the existing mount used by the async sidecar.

### Additional context
Original discussion here: https://github.com/eclipse-che/che-server/pull/192

PR #722, opened 2021-12-24
Disable automounting volumes to internally-used containers

amisevsk:
### What does this PR do?
Do not apply automounted configmaps, secrets, and pvcs to internally-used containers (e.g. project clone, async sidecar). This requires reworking how automount works somewhat, as we need to add automount volumes earlier in the reconcile loop (as opposed to when we create the workspace deployment)

There's a chance that automounted volumes interfere with internal containers. The main example here would be mounting SSH keys to `/etc/ssh` in the async-storage sidecar container, which has its own SSH key it needs to use.

### Is it tested? How?
1. Create an automounted volume, e.g.
    ```bash
    kubectl create configmap git-config \
      --from-file=gitconfig=${HOME}/.gitconfig
    
    kubectl patch cm git-config --type merge -p '
    metadata:
      labels:
      annotations:
        controller.devfile.io/mount-path: "/etc/"
        controller.devfile.io/mount-as: "subpath"
    '
    ```
2. Start a workspace and verify that automounted volume is not applied to project-clone container.

### PR Checklist


amisevsk:
Fixes https://github.com/devfile/devworkspace-operator/issues/707

JPinkney (approved):

PR #747, opened 2022-01-19
feat: [Draft] Allow DevWorkspaceID to be overridden via annotation

amisevsk:
### What does this PR do?
Allow overriding the default DevWorkspace ID (derived from the DevWorkspace CR's UUID) via the annotation
```
controller.devfile.io/devworkspace_id_override
```
with the caveats:
* DWO attempts to check if the DevWorkspaceID is already in use for another workspace in the current namespace and fails the workspace if it does
* The annotation is only read on workspace creation and ignored once workspace ID is set to avoid losing data
* Controller webhooks are updated to explicitly block updates to `.status.devworkspaceId`

This PR is in draft status for testing if this feature is useful for migrating Che workspaces to the DevWorkspace controller.

Before merging, this PR needs a polishing pass. We should also decide on a max length for overridden workspace IDs, as the controller implicitly assumes the length of the ID is fixed in a few places (where workspace ID is used as part of a k8s object name, for example).

Changes can be tested using the image `quay.io/amisevsk/devworkspace-controller:override-dw-id` (for both controller and webhooks server)

N/A for now.

### Is it tested? How?
No changes if annotation is not used; otherwise, workspace ID should be set to value of annotation.

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -132,9 +132,11 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// Ensure workspaceID is set.
 	if workspace.Status.DevWorkspaceId == "" {
-		workspaceId, err := getWorkspaceId(workspace)
+		workspaceId, err := r.getWorkspaceId(ctx, workspace)
 		if err != nil {
-			return reconcile.Result{}, err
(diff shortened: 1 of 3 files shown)


ibuziuk:
> We should also decide on a max length for overridden workspace IDs

I would propose 25 chars (current value of the ID + the same length is used for che-server devfile v1 workspaces)

ibuziuk:
Related issue for UD

Issue #718, opened 2021-12-23
Enable configuration of pod SecurityContext on Operator level in Kubernetes


amisevsk:
### Description
Currently, on Kubernetes we arbitrarily assign a [security context](https://github.com/devfile/devworkspace-operator/blob/d553f890cd23b574d60cf75384ea2bd9f4b1f755/pkg/provision/workspace/deployment.go#L234-L246) to workspace pods so that everything runs as uid 1234, gid 0.

We should consider making these values configurable at the operator level by adding fields to the DevWorkspaceOperatorConfig CR. If unconfigured, the current values could be a default. On OpenShift, this config setting would be ignored.

### Additional context
Prompted by https://github.com/eclipse/che/issues/20965, where the storage driver is by default creating volume mount subpaths with `755` permissions, preventing e.g. project clone and cleanup

PR #748, opened 2022-01-20
feat: Allow configuration of pod's security context for DevWorkspaces on Kubernetes

amisevsk:
### What does this PR do?
Adds field `podSecurityContext` to the `DevWorkspaceOperatorConfig` CR to allow overriding securityContext in all workspace-related pods when running on Kubernetes.

Also sets a default `fsGroup = 1234` to avoid issues on some k8s clusters (see: https://github.com/eclipse/che/issues/20965). Note that this is what is done in Eclipse Che as well as far as I can tell so should not cause issues.

Configuration value is ignored on OpenShift.

Closes https://github.com/devfile/devworkspace-operator/issues/718
Related: https://github.com/eclipse/che/issues/20965
Related: https://github.com/eclipse/che/issues/20963

### Is it tested? How?
1. On Kubernetes, start a DevWorkspace with default config. Pods created for workspace (including cleanup job pods) should use `fsGroup=1234`
2. Configure `.config.workspace.podSecurityContext` in the `devworkspace-operator-config` `DevWorkspaceOperatorConfig` and make sure change is propagated to DevWorkspaces (may need to trigger a reconcile for running workspaces)

### PR Checklist


nils-mosbach:
Thanks! 
Kubernetes 20 supports a property that allows changing owners only once if the root node does not match: `fsGroupChangePolicy: "OnRootMismatch"`

I haven't used this before, but that could speed up things if projects contain a lot of npm dependencies. What do you think?

See: https://kubernetes.io/blog/2020/12/14/kubernetes-release-1.20-fsgroupchangepolicy-fsgrouppolicy/

amisevsk:
@nils-mosbach Thanks for that link, I wasn't aware of this change! This is something I believe might solve a longterm complaint about workspace startup time for us (and Eclipse Che), as our use cases (especially with npm projects ;)) generally involve _many_ files in the PVC.

The DevWorkspace Operator depends on the Kubernetes 1.21.3 API, which means that this PR implicitly supports setting `fsGroupChangePolicy`:
```yaml
metadata:
config:
  workspace:
    podSecurityContext:
      fsGroupChangePolicy: "OnRootMismatch" 
      # Other fields need to be set too
```

The impact of this on DevWorkspace startup time could be tested using the `next` version of DWO as soon as this PR is merged. However, I don't have access to non-toy Kubernetes clusters (only OpenShift, where I believe this option would be ignored by the cluster). If you have a chance to test it out on your cluster, I'd love to hear the result!

ibuziuk (approved):

nils-mosbach:
Great, thanks a lot! I'll give it at go as soon as we manage to get devworkspaces running under Kubernetes. A version that I patched myself seemed to resolve the issue.

nils-mosbach:
I've just updated devworkspace controller to `quay.io/devfile/devworkspace-controller:next`. 
[#20965](https://github.com/eclipse/che/issues/20963#issue-1086871141) and [#20963](https://github.com/eclipse/che/issues/20965#issue-1086897120) are resolved in my case. I'll close the issues. Thanks, a lot!

amisevsk:
I'm glad to hear it :smile:

Issue #775, opened 2022-02-08
Workspaces can fail to start on OpenShift when waiting for serving certificate from service-ca operator 


amisevsk:
### Description
With certain routingClasses (e.g. `cluster-tls` and `web-terminal`), the DevWorkspace Operator depends on the service-ca operator to [inject](https://docs.openshift.com/container-platform/4.9/security/certificates/service-serving-certificate.html) service certs into containers. This results in a volume being added to the pod:
```yaml
  - name: devworkspace-serving-cert-<workspaceID>-service
    secret:
      defaultMode: 420
      secretName: <workspaceID>-service
```

However, in some cases, it may take the cluster some time to create the secret referenced in this pod, resulting in the workspace failing to start due to the mounted volume not being found

### How To Reproduce
Create a *new* DevWorkspace with routingClass `cluster-tls`. It has to be a new (not existing) workspace as the second time starting the workspace will generally succeed as the secret has been created by that point. 

This issue does not reproduce consistently, but is more likely on larger/more loaded clusters.

### Expected behavior
DWO should ideally ignore volumeMount events when they concern the injected secret.

### Additional context
WTO Jira issue: https://issues.redhat.com/browse/WTO-123

PR #776, opened 2022-02-08
Ignore FailedMount events if they're due to serving certs

amisevsk:
### What does this PR do?
Ignore FailedMount events if the involved object is the secret created by the service-ca operator (i.e. `devworkspace-serving-cert-<workspaceID>-service`. These FailedMount events, if they occur, are generally transient (and workspace startup should continue if they occur).

Closes https://github.com/devfile/devworkspace-operator/issues/775

### Is it tested? How?
Very difficult to test as #775 does not reproduce consistently (and seemingly almost never in testing clusters), and even if it does occur, this fix will result in the workspace being started as normal.

To help with testing this PR, I've built `quay.io/amisevsk/devworkspace-controller:service-ca`. This image is not built from this PR but rather from the inverse of this fix:

* If we successfully detect the `service-ca` volume FailedMount event, the workspace is _intentionally failed_ with message `Found failed mount for service cert volume` instead of ignoring the event.
* If we miss the FailedMount event, the workspace fails with the previous message:
    > `Detected unrecoverable event FailedMount: MountVolume.SetUp failed for volume “devworkspace-serving-cert-<workspaceID>-service” : secret “<workspaceID>-service” not found`

To use this image to verify the changes in this PR, it's necessary to confirm that the first failure _can_ occur, and that the second failure _cannot_.

The diff for this testing image is https://github.com/amisevsk/devworkspace-operator/compare/service-ca-volume-mount...amisevsk:service-ca-volume-mount-testing

### PR Checklist

### DIFF
--- a/pkg/provision/workspace/deployment.go
+++ b/pkg/provision/workspace/deployment.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	nsconfig "github.com/devfile/devworkspace-operator/pkg/provision/config"
 	"github.com/devfile/devworkspace-operator/pkg/provision/sync"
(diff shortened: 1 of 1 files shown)


ibuziuk (approved):

amisevsk:
/test v8-devworkspace-operator-e2e

After some attempting, I was able to verify the fix using image `quay.io/amisevsk/devworkspace-controller:service-ca`

PR #777, opened 2022-02-09
feat: Use `claim-che-workspace` PVC if it exists

amisevsk:
### What does this PR do?
Makes DWO automatically reuse PVCs named `claim-che-workspace` if they are present in a DevWorkspace's namespace to make migration of devfile v1 workspaces created by Eclipse Che easier. While not an ideal solution, this allows DWO to continue using its own PVC in new namespaces while supporting existing workspaces created by Eclipse Che. The alternative (configuring PVC name in the devworkspace operator config CR) results in all DevWorkspaces using the alternate name.

This PR is a draft until we decide this is the route we want to take. Merging this PR comes with some caveats:
* It makes this repo have an explicit reference to an Eclipse Che naming convention
* Once merged, DWO must support `claim-che-workspace` ongoing until there is a safe migration strategy (i.e. this feature cannot be easily dropped).
* This change makes DWO use `claim-che-workspace` by default regardless of the value in the DevWorkspaceOperatorConfig, meaning this feature cannot be disabled if a `claim-che-workspace` PVC exists.

Part of https://github.com/eclipse/che/issues/21038
Full solution would also require https://github.com/devfile/devworkspace-operator/pull/747

### Is it tested? How?
To test:
1. Create claim-che-workspace PVC in namespace, e.g. by copying the config of the claim-devworkspace PVC if it exists:
    ```
    kubectl get pvc claim-devworkspace -o json | \
      jq '{
          "metadata": {
            "name": "claim-che-workspace"
          }, apiVersion, kind, spec
        } | del(.spec.volumeMode) | del(.spec.volumeName)' | \
      kubectl apply -f -
    ```
2. Create a DevWorkspace: `kubectl apply -f samples/theia-next.yaml`
3. Verify that claim-che-workspace PVC is mounted instead of claim-devworkspace

### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -76,6 +76,10 @@ type WorkspaceConfig struct {
 	// PVCName defines the name used for the persistent volume claim created
 	// to support workspace storage when the 'common' storage class is used.
 	// If not specified, the default value of `claim-devworkspace` is used.
+	// Note that changing this configuration value after workspaces have been
+	// created will disconnect all existing workspaces from the previously-used
+	// persistent volume claim, and will require manual removal of the old PVCs
+	// in the cluster.
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +kubebuilder:validation:MaxLength=63
 	PVCName string `json:"pvcName,omitempty"`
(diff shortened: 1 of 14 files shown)


amisevsk:
To test both this PR and https://github.com/devfile/devworkspace-operator/pull/747 at the same time, I've built image `quay.io/amisevsk/devworkspace-controller:che-compatibility` which includes the changes from both.

ibuziuk (approved):

Issue #740, opened 2022-01-14
Make default PVC size configurable


amisevsk:
### Description
PVCs created by the operator by default request 1Gi of storage. The devfile API spec includes a "size" field in volume components, but it would be helpful to be able to configure the default request for creating PVCs as well.

This task is complicated by:
* It's not clear what the `size` field in a volume component means when a `common` storage type is used (e.g. all volumes for all workspaces mount the same PVC
* It's unclear how to manage PVC size in general, since most PVCs cannot be dynamically resized.

benoitf:
even trying a custom 'projects' component with a size field didn't worked in my devfile. It was still a 1Gi PV ( from common)

PR #818, opened 2022-04-13
feat: 'one PVC per-workspace' storage class

### LINKED ISSUE #792: Support 'per-workspace' storage-class (one PVC per workspace)
### Description
The current default storage class (`common`) has the limitation that only one workspace can be run per namespace if the underling persistent volume is RWO. To avoid this (and to clear up some semantics on PVC sizing), it would be useful to also support a storage class where each DevWorkspace gets its own PVC. For cleanup, we could rely on cluster garbage collection when the workspace is deleted.

### Additional context
Downstream Che issue: https://github.com/eclipse/che/issues/21185

AObuchow:
### What does this PR do?
This PR adds a new storage class called 'perWorkspace'. When in use in the devfile, every workspace gets its own PVC.

Fix #792 

### Is it tested? How?
In order to enable the 'perWorkspace' storage class, the `controller.devfile.io/storage-type` devfile attribute must be set to `per-workspace`.

Eg.
```YAML
spec:
  template:
    attributes:
```

**Manual testing:**
- Start up DWO (i.e. `make run`)
- Apply the sample devfile that uses the perWorkspace storage class: `kubectl apply -f ./samples/theia-next_per-workspaceStorage.yaml -n $NAMESPACE`
- Check K8s status (through terminal or dashboard) and ensure a PVC has been created for the workspace. The naming of the PVC should be of the form `storage-[workspaceID]`
- Delete the workspace: `kubectl delete dw theia-next -n $NAMESPACE`
- Ensure the PVC is also deleted shortly after (check K8s dashboard for PVC or use terminal commands)

**Automated testing:**
There is also a new test file, `pkg/provision/storage/perWorkspaceStorage_test.go`.
It is based off of `pkg/provision/storage/CommonStorage_test.go` with slight adaptions. It also ensures that the owner reference is set for the perWorkspace PVC, which is required for automatic PVC deletion with cluster garbage collection.

### PR Checklist

### DIFF
--- a/pkg/common/naming.go
+++ b/pkg/common/naming.go
@@ -90,6 +90,10 @@ func PVCCleanupJobName(workspaceId string) string {
 	return fmt.Sprintf("cleanup-%s", workspaceId)
 }
 
+func PerWorkspacePVCName(workspaceId string) string {
+	return fmt.Sprintf("storage-%s", workspaceId)
+}
+
 func MetadataConfigMapName(workspaceId string) string {
 	return fmt.Sprintf("%s-metadata", workspaceId)
 }
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -73,6 +73,9 @@ const (
 	// EphemeralStorageClassType defines the 'ephemeral' storage policy: all volumes are allocated as emptyDir volumes and
 	// so do not require cleanup. When a DevWorkspace is stopped, all local changes are lost.
 	EphemeralStorageClassType = "ephemeral"
+	// PerWorkspaceStorageClassType defines the 'per-workspace' storage policy: a PVC is provisioned for each workspace within the namespace.
+	// All of the workspace's storage (volume mounts) are mounted on subpaths within the workspace's PVC.
+	PerWorkspaceStorageClassType = "per-workspace"
 
 	// CheCommonPVCName is the name of the common PVC equivalent used by Che. If present in the namespace, this PVC is mounted instead
 	// of the default PVC when the 'common' or 'async' storage classes are used.
--- a/pkg/provision/storage/perWorkspaceStorage.go
+++ b/pkg/provision/storage/perWorkspaceStorage.go
@@ -1,0 +1,197 @@
+//
+// Copyright (c) 2019-2022 Red Hat, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package storage
+
+import (
+	"errors"
+	"fmt"
+
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"github.com/devfile/devworkspace-operator/pkg/common"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
+	devfileConstants "github.com/devfile/devworkspace-operator/pkg/library/constants"
+	nsconfig "github.com/devfile/devworkspace-operator/pkg/provision/config"
+	"github.com/devfile/devworkspace-operator/pkg/provision/sync"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// The PerWorkspaceStorageProvisioner provisions one PVC per workspace and configures all volumes in the workspace
+// to mount on subpaths within that PVC.
+type PerWorkspaceStorageProvisioner struct{}
+
+var _ Provisioner = (*PerWorkspaceStorageProvisioner)(nil)
+
+// This function is used to determine whether a finalizer should be applied to the DevWorkspace.
+// Since the per-workspace PVC are cleaned up/deleted via Owner References when the workspace is deleted,
+// no finalizer needs to be applied
+func (*PerWorkspaceStorageProvisioner) NeedsStorage(workspace *dw.DevWorkspaceTemplateSpec) bool {
+	return false
+}
+
+func (p *PerWorkspaceStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAdditions, workspace *dw.DevWorkspace, clusterAPI sync.ClusterAPI) error {
+	// Add ephemeral volumes
+	if err := addEphemeralVolumesFromWorkspace(workspace, podAdditions); err != nil {
+		return err
+	}
+
+	// If persistent storage is not needed, we're done
+	if !needsStorage(&workspace.Spec.Template) {
+		return nil
+	}
+
+	// Get perWorkspace PVC spec and sync it with cluster
+	perWorkspacePVC, err := syncPerWorkspacePVC(workspace, clusterAPI)
+	if err != nil {
+		return err
+	}
+	pvcName := perWorkspacePVC.Name
+
+	// Rewrite container volume mounts
+	if err := p.rewriteContainerVolumeMounts(workspace.Status.DevWorkspaceId, pvcName, podAdditions, &workspace.Spec.Template); err != nil {
+		return &ProvisioningError{
+			Err:     err,
+			Message: "Could not rewrite container volume mounts",
+		}
+	}
+
+	return nil
+}
+
+// We rely on Kubernetes to use the owner reference to automatically delete the PVC once the workspace is set for deletion.
+func (*PerWorkspaceStorageProvisioner) CleanupWorkspaceStorage(workspace *dw.DevWorkspace, clusterAPI sync.ClusterAPI) error {
+	return nil
+}
+
+// rewriteContainerVolumeMounts rewrites the VolumeMounts in a set of PodAdditions according to the 'per-workspace' PVC strategy
+// (i.e. all volume mounts are subpaths into a PVC used by a single workspace in the namespace).
+//
+// Also adds appropriate k8s Volumes to PodAdditions to accomodate the rewritten VolumeMounts.
+func (p *PerWorkspaceStorageProvisioner) rewriteContainerVolumeMounts(workspaceId, pvcName string, podAdditions *v1alpha1.PodAdditions, workspace *dw.DevWorkspaceTemplateSpec) error {
+	devfileVolumes := map[string]dw.VolumeComponent{}
+
+	// Construct map of volume name -> volume Component
+	for _, component := range workspace.Components {
+		if component.Volume != nil {
+			if _, exists := devfileVolumes[component.Name]; exists {
+				return fmt.Errorf("volume component '%s' is defined multiple times", component.Name)
+			}
+			devfileVolumes[component.Name] = *component.Volume
+		}
+	}
+
+	// Containers in podAdditions may reference e.g. automounted volumes in their volumeMounts, and this is not an error
+	additionalVolumes := map[string]bool{}
+	for _, additionalVolume := range podAdditions.Volumes {
+		additionalVolumes[additionalVolume.Name] = true
+	}
+
+	// Add implicit projects volume to support mountSources, if needed
+	if _, exists := devfileVolumes[devfileConstants.ProjectsVolumeName]; !exists {
+		projectsVolume := dw.VolumeComponent{}
+		projectsVolume.Size = constants.PVCStorageSize
+		devfileVolumes[devfileConstants.ProjectsVolumeName] = projectsVolume
+	}
+
+	// TODO: What should we do when a volume isn't explicitly defined?
+	rewriteVolumeMounts := func(containers []corev1.Container) error {
+		for cIdx, container := range containers {
+			for vmIdx, vm := range container.VolumeMounts {
+				volume, ok := devfileVolumes[vm.Name]
+				if !ok {
+					// Volume is defined outside of the devfile
+					if additionalVolumes[vm.Name] {
+						continue
+					}
+					// Should never happen as flattened Devfile is validated.
+					return fmt.Errorf("container '%s' references undefined volume '%s'", container.Name, vm.Name)
+				}
+				if !isEphemeral(&volume) {
+					containers[cIdx].VolumeMounts[vmIdx].SubPath = vm.Name
+					containers[cIdx].VolumeMounts[vmIdx].Name = pvcName
+				}
+			}
+		}
+		return nil
+	}
+	if err := rewriteVolumeMounts(podAdditions.Containers); err != nil {
+		return err
+	}
+	if err := rewriteVolumeMounts(podAdditions.InitContainers); err != nil {
+		return err
+	}
+
+	podAdditions.Volumes = append(podAdditions.Volumes, corev1.Volume{
+		Name: pvcName,
+		VolumeSource: corev1.VolumeSource{
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: pvcName,
+			},
+		},
+	})
+
+	return nil
+}
+
+func syncPerWorkspacePVC(workspace *dw.DevWorkspace, clusterAPI sync.ClusterAPI) (*corev1.PersistentVolumeClaim, error) {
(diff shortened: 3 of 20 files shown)


pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil, fmt.Errorf("failed to read namespace-specific configuration: %w", err)
+	}
+
+	// TODO: Should we adjust the storage size?
+	pvcSize := constants.PVCStorageSize
+	if namespacedConfig != nil && namespacedConfig.CommonPVCSize != "" {
+		pvcSize = namespacedConfig.CommonPVCSize
+	}
+
+	// We have to call clusterAPI.Client.Scheme() because the fakeClusterAPI.Scheme used in testing returns nil
```
AObuchow, line 58: This is something that might be a bug with the fakeClusterAPI, might be worth fixing in this PR or making another bug for it.

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	return currPVC, nil
+
+}
+
+func getPerWorkspacePVCSpec(workspace *dw.DevWorkspace, scheme *runtime.Scheme, size string) (*corev1.PersistentVolumeClaim, error) {
+	// TODO: Just make these part of the function parameters instead?
+	namespace := workspace.Namespace
+	DevWorkspaceId := workspace.Status.DevWorkspaceId
```
amisevsk, line 93: Generally, vars names aren't capitalized. In Go, capitalized names are exported (except within functions where it doesn't do anything)

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -0,0 +1,231 @@
...
+		},
+	}
+
+	err = controllerutil.SetControllerReference(workspace, pvc, scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	return pvc, nil
+
```
amisevsk, line 126: Nit: unnecessary empty line

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -0,0 +1,231 @@
...
+func (p *perWorkspaceStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAdditions, workspace *dw.DevWorkspace, clusterAPI sync.ClusterAPI) error {
+	// Add ephemeral volumes
+	if err := addEphemeralVolumesFromWorkspace(workspace, podAdditions); err != nil {
+		return err
+	}
+
+	// If persistent storage is not needed, we're done
+	if !p.NeedsStorage(&workspace.Spec.Template) {
+		return nil
+	}
```
amisevsk, line 138: If you decide to follow the suggestion for returning `false` for `NeedsStorage`, this should be replaced with `needStorage` from common.go:

```suggestion
	// If persistent storage is not needed, we're done
	if !needsStorage(&workspace.Spec.Template) {
		return nil
	}
```

pkg/provision/storage/perWorkspaceStorage_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+			if err != nil {
+				t.Fatalf("Failure during setup: %s", err)
+			}
+			workspacePVC.Status.Phase = corev1.ClaimBound
+
+			clusterAPI := sync.ClusterAPI{
+				Client: fake.NewFakeClientWithScheme(scheme, workspacePVC),
+				Logger: zap.New(),
+			}
```
amisevsk, line 69:
```suggestion
			clusterAPI := sync.ClusterAPI{
				Scheme: scheme,
				Client: fake.NewFakeClientWithScheme(scheme, workspacePVC),
				Logger: zap.New(),
			}
```

amisevsk (commented):
First pass at review -- generally looks good!

I'll get around to deeper review and testing later.

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -0,0 +1,231 @@
...
+	currPVC, ok := currObject.(*corev1.PersistentVolumeClaim)
+	if !ok {
+		return nil, errors.New("tried to sync per-workspace PVC to cluster but did not get a PVC back")
+	}
+
+	return currPVC, nil
+
+}
+
+func getPerWorkspacePVCSpec(workspace *dw.DevWorkspace, scheme *runtime.Scheme, size string) (*corev1.PersistentVolumeClaim, error) {
```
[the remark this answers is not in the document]
AObuchow: That sounds good to me, though one thing that came to mind:
`getPerWorkspacePVCSpec` also sets the controller reference, in addition to setting the PVC name.

Do you think it's better to keep a `getPerWorkspacePVCSpec` function that is just a wrapper for `getPVCSpec` (where it takes in the PVC name) but also sets up the controller reference?

Or just call `getPVCSpec` and then set up the controller reference immediately afterwards, whenever needed (both in the test method and in `syncPerWorkspacePVC`?

pkg/provision/storage/perWorkspaceStorage_test.go
AObuchow: Thanks :D Was wondering what was the proper way to fix that

pkg/provision/storage/shared.go
```diff
@@ -221,7 +221,8 @@ func processProjectsVolume(workspace *dw.DevWorkspaceTemplateSpec) (projectsComp
 
 // checkForExistingCommonPVC checks the current namespace for existing PVCs that may be used for workspace storage. If
 // such a PVC is found, its name is returned and should be used in place of the configured common PVC. If no suitable
-// PVC is found,
+// PVC is found, the returned PVC name is an empty string and a nill error is returned. If an error occurs during the lookup,
```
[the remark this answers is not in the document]
AObuchow: No problem :sunglasses:

samples/flattened_theia-next_perWorkspaceStorage.yaml
```diff
@@ -0,0 +1,101 @@
+kind: DevWorkspace
```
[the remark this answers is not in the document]
AObuchow: I made a new one based on `theia-next.yaml` but am having trouble running `theia-next.yaml` locally:
```json
{"level":"info","ts":1649885999.529929,"logger":"controllers.DevWorkspace","msg":"DevWorkspace failed to start: 2 errors occurred:\n\t* devfile contains multiple endpoint entries with same TargetPort: 3100\n\t* devfile contains multiple endpoint entries with same TargetPort: 3100\n\n","Request.Namespace":"devworkspace-controller","Request.Name":"theia-next","devworkspace_id":"workspace6d96e72c3a6742de"}
```

Will have to debug it further.

pkg/constants/constants.go
```diff
@@ -73,6 +73,9 @@ const (
 	// EphemeralStorageClassType defines the 'ephemeral' storage policy: all volumes are allocated as emptyDir volumes and
 	// so do not require cleanup. When a DevWorkspace is stopped, all local changes are lost.
 	EphemeralStorageClassType = "ephemeral"
+	// PerWorkspaceStorageClassType defines the 'per-Workspace' storage policy: a PVC is provisioned for each workspace within the namespace.
```
AObuchow, line 76: Should be 'per-workspace' *

samples/flattened_theia-next_perWorkspaceStorage.yaml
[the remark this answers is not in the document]
AObuchow: Rebasing onto the main branch fixed this - thanks @amisevsk :D

pkg/provision/storage/perWorkspaceStorage.go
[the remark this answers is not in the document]
AObuchow: > Do you think it's better to keep a `getPerWorkspacePVCSpec` function that is just a wrapper for `getPVCSpec` (where it takes in the PVC name) but also sets up the controller reference?

For now, I've done the above, though in theory it's not necessary to have a `getPerWorkspacePVCSpec` function as `provisionStorage` will get called, which should set up the controller reference when necessary. 

The issue I've found with relying on `provisionStorage` to set up the controller reference, is that certain test cases don't lead to the controller reference getting set. For example: when ephemeral storage is used, when storage is not required (`does-nothing-for-no-storage-needed.yaml`) and when the user specifies a project volume manually (`projects-volume-overriding.yaml`).

I was able to write some, kind of ugly workaround's in the test function so that the owner reference isn't checked for the above-mentioned cases. However, I haven't figured out yet how to detect the user-specified project volume case.

Something like:

```go
	ownerReferenceNotNeeded := false
	for _, component := range tt.Input.Workspace.Components {
		if component.Volume != nil && component.Volume.Ephemeral != nil {
			ownerReferenceNotNeeded = ownerReferenceNotNeeded || *component.Volume.Ephemeral
		}
	}

	ownerReferenceNotNeeded = ownerReferenceNotNeeded || !needsStorage(&workspace.Spec.Template)

	if !ownerReferenceNotNeeded {
		assert.NotEmpty(t, workspacePVC.ObjectMeta.OwnerReferences)
		assert.Len(t, workspacePVC.ObjectMeta.OwnerReferences, 1)
		assert.Equal(t, workspacePVC.ObjectMeta.OwnerReferences[0].Kind, "DevWorkspace")
	}

```
Should I go with this approach? I can update my PR with this approach if you would like to test things out.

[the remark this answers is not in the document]
amisevsk: > Do you think it's better to keep a getPerWorkspacePVCSpec function that is just a wrapper for getPVCSpec (where it takes in the PVC name) but also sets up the controller reference?

In general it's better to not have functions optionally do anything (we don't want ownerrefs on commonPVC workspaces ever). This could be worked around by supplying the entire workspace to the `getSpecPVC` function, but it's simpler to just set the controllerRef afterwards (e.g. https://github.com/amisevsk/devworkspace-operator/commit/67f0a59c3d9f3e2817ef1fb614bb54deb0739d9f#diff-a3fc5cfc1fe497d9039f7540c20597bc9e87b4ebbe9ab58d6bc25b84dc647ed6R60-R63). This keeps the purpose of `getSpecPVC` simpler.

In the test method, you're calling `ProvisionStorage` so the PVC should get its ownerref set. To test that the PVC is created as expected, we should just be getting the PVC from the fake k8s client and checking its fields.

pkg/provision/storage/perWorkspaceStorage_test.go
```diff
@@ -0,0 +1,97 @@
...
+			workspace.Spec.Template = *tt.Input.Workspace
+			workspace.Status.DevWorkspaceId = tt.Input.DevWorkspaceID
+			workspace.Namespace = "test-namespace"
+
+			workspacePVC, err := getPerWorkspacePVCSpec(workspace, scheme, "10Gi")
+
+			if err != nil {
+				t.Fatalf("Failure during setup: %s", err)
+			}
+			workspacePVC.Status.Phase = corev1.ClaimBound
```
amisevsk, line 64: It's unclear why this is needed, and why the workspacePVC should already always exist on the cluster before calling provisionStorage. The better approach may be to call `ProvisionStorage` twice in the test (first to create the PVC, second to rewrite container VMs) and verify that each step has the expected output (error the first time, nil error the second, etc.)

pkg/provision/storage/perWorkspaceStorage_test.go
```diff
@@ -0,0 +1,97 @@
...
+			err = perWorkspaceStorage.ProvisionStorage(&tt.Input.PodAdditions, workspace, clusterAPI)
+			if tt.Output.ErrRegexp != nil && assert.Error(t, err) {
+				assert.Regexp(t, *tt.Output.ErrRegexp, err.Error(), "Error message should match")
+			} else {
+				if !assert.NoError(t, err, "Should not return error") {
+					return
+				}
+
+				pvcName := common.PerWorkspacePVCName(workspace.Status.DevWorkspaceId)
+				namespacedName := types.NamespacedName{Name: pvcName, Namespace: workspace.Namespace}
```
amisevsk, line 81:
```suggestion
				namespacedName := types.NamespacedName{Name: common.PerWorkspacePVCName(tt.Input.DevWorkspaceID), Namespace: workspace.Namespace}
```

pkg/provision/storage/perWorkspaceStorage_test.go
AObuchow: I agree -  `ProvisionStorage` should be good to create the PVC rather than creating it on the cluster before calling `ProvisionStorage`.

However, in my testing, it seems that for certain cases,  `ProvisionStorage` will return early because a PVC is not required (https://github.com/devfile/devworkspace-operator/pull/818/files#diff-a3fc5cfc1fe497d9039f7540c20597bc9e87b4ebbe9ab58d6bc25b84dc647ed6R115)

I'm not sure if I should:
- Remove those test cases (most likely not)
- Skip the check for PVC creation, and owner reference check for these test cases
- Or something else that I haven't considered yet.

I'm going to submit the updated PR, so you can see what I mean

pkg/provision/storage/perWorkspaceStorage_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+			clusterAPI := sync.ClusterAPI{
+				Scheme: scheme,
+				Client: fake.NewClientBuilder().WithScheme(scheme).Build(),
+				Logger: zap.New(),
+			}
+
+			if needsStorage(&workspace.Spec.Template) {
+				err := perWorkspaceStorage.ProvisionStorage(&tt.Input.PodAdditions, workspace, clusterAPI)
+				if !assert.Error(t, err, "Should get an error when creating PVC") {
```
amisevsk, line 67: Minor edit for clarity:
```suggestion
				if !assert.Error(t, err, "Should get a NotReady error when creating PVC") {
```

pkg/provision/storage/shared.go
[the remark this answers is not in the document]
amisevsk: Looks like the `nill -> nil` got lost in a rebase

amisevsk (approved):
Mostly just minor nitpicks and polish at this point. Well done!

AObuchow:
> Mostly just minor nitpicks and polish at this point. Well done!
> 
> Tested on OpenShift 4.10 and works as expected +1

Awesome, thank you so much for the help & reviews @amisevsk :grin: 
I made the requested changes.

amisevsk (approved):

amisevsk:
Well done :+1:

pkg/common/naming.go
```diff
@@ -90,6 +90,10 @@ func PVCCleanupJobName(workspaceId string) string {
 	return fmt.Sprintf("cleanup-%s", workspaceId)
 }
 
+func PerWorkspacePVCName(workspaceId string) string {
+	return fmt.Sprintf("storage-%s", workspaceId)
```
ibuziuk, line 94: maybe let's call it claim?

ibuziuk, line 94:
```suggestion
	return fmt.Sprintf("claim-%s", workspaceId)
```

ibuziuk: would be more consistent IMO:
- claim-che-workspace (old PVCs)
- claim-devworkspace (new common PVC)
- claim-workspaceid (new per-workspace PVC)

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -0,0 +1,197 @@
...
+func syncPerWorkspacePVC(workspace *dw.DevWorkspace, clusterAPI sync.ClusterAPI) (*corev1.PersistentVolumeClaim, error) {
+	namespace := workspace.Namespace
+
+	namespacedConfig, err := nsconfig.ReadNamespacedConfig(namespace, clusterAPI)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read namespace-specific configuration: %w", err)
+	}
+
+	// TODO: Determine the storage size that is needed by iterating through workspace volumes,
+	// adding the sizes specified and figuring out overrides/defaults
```
ibuziuk, line 159: This TODO is planned to be addressed as part of https://github.com/devfile/devworkspace-operator/issues/740, right?

AObuchow: Correct, or at the least, it might end up being a separate but related issue

pkg/common/naming.go
amisevsk: We're flip-flopping on our suggestions :) 

https://github.com/devfile/devworkspace-operator/pull/818#discussion_r849648541

ibuziuk (approved):
Great job :+1:

pkg/common/naming.go
AObuchow: I ended up going with `storage` for now since it seems to be a bit more representative of what it's purpose is, though I'm not going to make a fuss :P 

Should be good to merge after finally CI run.

PR #827, opened 2022-04-26
Configurable default PVC size for common and per-workspace storage classes

### LINKED ISSUE #792: Support 'per-workspace' storage-class (one PVC per workspace)

AObuchow:
### What does this PR do?
- Allows for (optionally) configuring the default PVC size for common and per-workspace storage classes using `DevWorkspaceOperatorConfig` CRD
- Sets default PVC sizes for common/async and per-workspace storage classes

In order to configure the default PVC size for a storage class, a `DevWorkspaceOperatorConfig` CRD must be applied to the cluster with the `DefaultStorageSize` `workspace` property set. This property takes a struct where the [`resource.Quantity`](https://pkg.go.dev/k8s.io/apimachinery@v0.21.3/pkg/api/resource?utm_source=gopls#Quantity) fields define the PVC sizes for `common`/`async` and `per-workspace` storage classes.

Fix #792.

### Is it tested? How?
- Manual testing of PVC size defaults has been done by starting up a cluster, applying a devfile that uses the `controller.devfile.io/storage-type` attribute set to `per-workspace` and verifying the newly created PVC size is set to the default size for per-workspace storage.
	- Manually testing the default PVC size for `common` storage still needs to be done, however the "new" default is currently the same as it used to be (`10Gi`) 
- The `commonStorage` test has been modified to support this new change. Both the `common` and `per-workspace` storage class tests are passing with this new change.
- The `sync` tests have been modified to support this new change.

### Additional WIP notes
- **Setting the storage class sizes with a CRD still needs to be manually tested**

### PR Checklist


apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -83,9 +84,15 @@ type WorkspaceConfig struct {
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +kubebuilder:validation:MaxLength=63
 	PVCName string `json:"pvcName,omitempty"`
-	// StorageClassName defines and optional storageClass to use for persistent
+	// StorageClassName defines an optional storageClass to use for persistent
 	// volume claims created to support DevWorkspaces
 	StorageClassName *string `json:"storageClassName,omitempty"`
+	// StorageClassSize defines an optional map of sizes of persistent volume claims for each storage
+	// class used by DevWorkspaces.
+	// The available keys to the map are: "common", "per-workspace".
```
amisevsk, line 92: Maybe worth noting that `common` is used for the `async` storageClass as well.

pkg/config/defaults.go
```diff
@@ -29,6 +31,11 @@ var defaultConfig = &v1alpha1.OperatorConfiguration{
 	Workspace: &v1alpha1.WorkspaceConfig{
 		ImagePullPolicy: "Always",
 		PVCName:         "claim-devworkspace",
+		StorageClassSize: map[string]resource.Quantity{
+			// TODO: Verify these are reasonable defaults
+			constants.CommonStorageClassType:       resource.MustParse("10Gi"),
+			constants.PerWorkspaceStorageClassType: resource.MustParse("3Gi"),
+		},
```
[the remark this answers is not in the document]
AObuchow: Yes good catch!

amisevsk:
You'll need to rebase onto `main` (dropping the first commit)

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
AObuchow: Done :+1:  Note that the documentation maybe could be better, not sure about it yet. Also, it's subject to change once I find proper default sizes for PVC's

pkg/config/defaults.go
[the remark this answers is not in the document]
AObuchow: Done, https://github.com/devfile/devworkspace-operator/pull/827/commits/93e9bafe79078cabe5fa287adfa515dc6cb129e8#

However.. I forgot to also make this information logged in `sync.go`'s `logCurrentConfig` :man_facepalming: Will add that.

deploy/deployment/kubernetes/combined.yaml
```diff
@@ -88,6 +88,19 @@ spec:
...
+                  StorageClassSize:
+                    additionalProperties:
+                      anyOf:
+                      - type: integer
+                      - type: string
+                      pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                      x-kubernetes-int-or-string: true
+                    description: 'StorageClassSize defines an optional map of sizes
+                      of persistent volume claims for each storage class used by DevWorkspaces.
+                      The available keys to the map are: "common", "per-workspace".
```
AObuchow, line 100: I'm not sure why these files don't have the mention of `async` using the `common` storage class PVC size that I added?

pkg/config/sync.go
```diff
@@ -281,6 +281,11 @@ func logCurrentConfig() {
 			config = append(config, fmt.Sprintf("workspace.ignoredUnrecoverableEvents=%s",
 				strings.Join(Workspace.IgnoredUnrecoverableEvents, ";")))
 		}
+		for storageClassName, PVCSize := range Workspace.StorageClassSize {
+			if PVCSize != defaultConfig.Workspace.StorageClassSize[storageClassName] {
+				config = append(config, fmt.Sprintf("workspace.storageClassSize[%s]=%s", storageClassName, &PVCSize))
```
AObuchow, line 286: I'm not sure if this is the best format (i.e. with the `[]`) to use for displaying non-default PVC storage sizes, though it is consistent with its internal representation as a map.

Also, maybe this logged information should occur right after `PVCName` is logged, for consistency with the CRD?

deploy/deployment/kubernetes/combined.yaml
AObuchow: The CI is also complaining about this :P

amisevsk: You have to run `make generate` to regenerate CRDs after changing the go definition. I use
```
make generate manifests fmt generate_default_deployment generate_olm_bundle_yaml
```
to make sure all autogenerated files are in sync.

pkg/config/sync.go
amisevsk: I think you may be getting thrown off by using a `map[string]resource.Quantity` here -- there's no yaml-level difference in defining `storageClassSize` as a `map[string]resource.Quantity` and something like
```go
storageClassSize struct {
  perWorkspaceSize resource.Quantity `json:"per-workspace,omitempty"`
  commonSize resource.Quantity `json:"common,omitempty"`
}
```
(except that we're defining all the valid keys in the map)

In short, the format I'd expect is 
```suggestion
				config = append(config, fmt.Sprintf("workspace.storageClassSize.%s=%s", storageClassName, &PVCSize))
```

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -83,9 +84,16 @@ type WorkspaceConfig struct {
...
+	// Note that the "async" storage class also uses the PVC size set for the "common" storage class.
+	// If not specified, the storage class size defaults are:
+	// 10Gi for "common" (and "async"), 3Gi for "per-workspace".
+	StorageClassSize map[string]resource.Quantity `json:"StorageClassSize,omitempty"`
```
amisevsk, line 96: Json keys should not be capitalized
```suggestion
	StorageClassSize map[string]resource.Quantity `json:"storageClassSize,omitempty"`
```
Also, given future plans around making these sizes configurable in each DevWorkspace (i.e. start looking at the `size` field in Volume components, maybe something like 
```suggestion
	DefaultStorageSize map[string]resource.Quantity `json:"defaultStorageSize,omitempty"`
```
is more suitable.

pkg/config/sync.go
```diff
@@ -242,6 +242,11 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
 		if from.Workspace.PodSecurityContext != nil {
 			to.Workspace.PodSecurityContext = from.Workspace.PodSecurityContext
 		}
+		for storageClassName, PVCSize := range from.Workspace.StorageClassSize {
+			if PVCSize != defaultConfig.Workspace.StorageClassSize[storageClassName] {
+				to.Workspace.StorageClassSize[storageClassName] = PVCSize
```
amisevsk, line 247: `to.Workspace.StorageClassSize` may be `nil` here, resulting in a panic. The merge code for storageClassSize should look something like 
```go
		if from.Workspace.StorageClassSize != nil {
			if to.Workspace.StorageClassSize == nil {
				to.Workspace.StorageClassSize = map[string]resource.Quantity{}
			}
			for storageClassName, PVCSize := range from.Workspace.StorageClassSize {
				to.Workspace.StorageClassSize[storageClassName] = PVCSize
			}
		}
```

pkg/config/sync.go
AObuchow: Thank you for the clarification :) Your suggestion is what I initially was going to write, but wasn't sure if it was correct

AObuchow:
I opened che-server and built it on the dog fooding cluster, then checked the workspace's associated PVC usage from the OpenShift dashboard.

I also made sure that any other workspace I had (other than che-server) was deleted.

It says 545.4 MiB were used:
image

AObuchow:
I opened che-code and built it on the dog fooding cluster.

After initial workspace setup (git clone and plugins, which were automatically setup), disk usage was `981.3MiB`.
After running `yarn prepare`, disk usage was just under `3GiB` (`2.7GiB` IIRC)
After running `yarn build`, disk usage is `3.39GiB`.
Finally, I ran `yarn watch` and in another terminal, `yarn server` which brought up the disk usage to `3.49GiB`
image

amisevsk:
I'm :+1: on setting the default per-workspace storage size to `5Gi` to be on the safe side (`4Gi` isn't much wiggle room for a project like che-code). One of the benefits of this PR is that these defaults become configurable, so we should aim for something that always works as a default, even if it overprovisions storage in most cases.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -87,13 +97,9 @@ type WorkspaceConfig struct {
...
-	// DefaultStorageSize defines an optional map of sizes of persistent volume claims for each storage
-	// class used by DevWorkspaces.
-	// The available keys to the map are: "common", "per-workspace".
-	// Note that the "async" storage class also uses the PVC size set for the "common" storage class.
-	// If not specified, the storage class size defaults are:
-	// 10Gi for "common" (and "async"), 3Gi for "per-workspace".
-	DefaultStorageSize map[string]resource.Quantity `json:"defaultStorageSize,omitempty"`
+	// DefaultStorageSize defines an optional struct with fields to specify the sizes of Persistent Volume Claims for storage
+	// classes used by DevWorkspaces.
+	DefaultStorageSize *StorageSizes `json:"defaultStorageSize,omitempty"`
```
AObuchow, line 102: maybe this should be `DefaultStorageSize *corev1.StorageSizes` instead?

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -67,6 +68,16 @@ type Proxy struct {
 	NoProxy string `json:"noProxy,omitempty"`
 }
 
+type StorageSizes struct {
```
AObuchow, line 71: Not sure if this should have been called `DefaultStorageSize` instead? For consistency with `WorkspaceConfig`

pkg/config/sync.go
```diff
@@ -276,6 +290,14 @@ func logCurrentConfig() {
 			config = append(config, fmt.Sprintf("workspace.ignoredUnrecoverableEvents=%s",
 				strings.Join(Workspace.IgnoredUnrecoverableEvents, ";")))
 		}
+		if Workspace.DefaultStorageSize != nil {
+			if Workspace.DefaultStorageSize.Common != nil {
+				config = append(config, fmt.Sprintf("workspace.storageClassSize.%s=%s", constants.CommonStorageClassType, Workspace.DefaultStorageSize.Common.String()))
```
AObuchow, line 295: Should now be: `"workspace.defaultStorageSize.%s=%s"` instead of `"workspace.storageClassSize.%s=%s"`

pkg/config/sync.go
```diff
@@ -276,6 +290,14 @@ func logCurrentConfig() {
...
+			}
+			if Workspace.DefaultStorageSize.PerWorkspace != nil {
+				config = append(config, fmt.Sprintf("workspace.storageClassSize.%s=%s", constants.PerWorkspaceStorageClassType, Workspace.DefaultStorageSize.PerWorkspace.String()))
```
AObuchow, line 298: Should now be: `"workspace.defaultStorageSize.%s=%s"` instead of `"workspace.storageClassSize.%s=%s"`

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -67,6 +68,16 @@ type Proxy struct {
...
+	// The default Persistent Volume Claim size for the "common" storage class.
+	// Note that the "async" storage class also uses the PVC size set for the "common" storage class.
+	// If not specified, the "common" and "async" Persistent Volume Claim sizes are set to 10Gi
+	Common *resource.Quantity `json:"common,omitempty"`
+	// The default Persistent Volume Claim size for the "per-workspace" storage class.
+	// If not specified, the "per-workspace" Persistent Volume Claim size is set to 5Gi
+	PerWorkspace *resource.Quantity `json:"perWorkspace,omitempty"`
```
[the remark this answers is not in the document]
amisevsk: General API guidelines for k8s suggest `perWorkspace` here, IIRC, so the current format is good.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
amisevsk: You can't add stuff to the corev1 package, that's owned by kubebuilder (usually we use `corev1` as an alias for `"k8s.io/api/core/v1"` since otherwise the import is `v1` which collides with everything :)

E.g.
```go
import (
	corev1 "k8s.io/api/core/v1"
)
```
without an alias, Go will use the last section of the import (`v1` in both cases) so we have to alias them to `corev1` and `metav1`

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
amisevsk: `StorageSizes` is fine here, it's just the name of the internal struct.

pkg/config/sync.go
```diff
@@ -245,12 +246,15 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
...
+				to.Workspace.DefaultStorageSize = &v1alpha1.StorageSizes{}
+
 			}
-			for storageClassName, PVCSize := range from.Workspace.DefaultStorageSize {
-				to.Workspace.DefaultStorageSize[storageClassName] = PVCSize
+			if from.Workspace.DefaultStorageSize.Common != nil {
+				to.Workspace.DefaultStorageSize.Common = from.Workspace.DefaultStorageSize.Common
+			}
+			if from.Workspace.DefaultStorageSize.PerWorkspace != nil {
+				to.Workspace.DefaultStorageSize.PerWorkspace = from.Workspace.DefaultStorageSize.PerWorkspace
```
amisevsk, line 256: Since `resource.Quantity` isn't a base type, it's maybe safer to DeepCopy these fields rather than copying the reference
```suggestion
			if from.Workspace.DefaultStorageSize.Common != nil {
        commonSizeCopy := from.Workspace.DefaultStorageSize.Common.DeepCopy()
				to.Workspace.DefaultStorageSize.Common = &commonSizeCopy
			}
			if from.Workspace.DefaultStorageSize.PerWorkspace != nil {
        perWorkspaceSizeCopy := from.Workspace.DefaultStorageSize.PerWorkspace.DeepCopy()
				to.Workspace.DefaultStorageSize.PerWorkspace = &perWorkspaceSizeCopy
```

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
AObuchow: Okay makes sense :) So in other words, just having it has `DefaultStorageSize *StorageSizes` is okay?

AObuchow:
Not sure yet why the `check sources` CI is suddenly failing:
```bash
go: downloading golang.org/x/sys v0.0.0-20220502124256-b6088ccd6cba
Error: ../../../go/pkg/mod/golang.org/x/tools@v0.1.10/cmd/goimports/goimports.go:14:2: //go:build comment without // +build comment
Error: Process completed with exit code 1.
```

amisevsk:
https://github.com/devfile/devworkspace-operator/pull/831 should fix PR check

pkg/config/sync.go
```diff
@@ -22,13 +22,15 @@ import (
...
+	"github.com/devfile/devworkspace-operator/pkg/constants"
 	routeV1 "github.com/openshift/api/route/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	controller "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
```
amisevsk, line 34: This is a duplicate import, you should use `&controller.StorageSizes{}`, etc. (instead of e.g. `&v1alpha1.StorageSizes{}`) in the code below.

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -157,9 +159,12 @@ func syncPerWorkspacePVC(workspace *dw.DevWorkspace, clusterAPI sync.ClusterAPI)
 
 	// TODO: Determine the storage size that is needed by iterating through workspace volumes,
 	// adding the sizes specified and figuring out overrides/defaults
-	pvcSize := constants.PVCStorageSize
+	pvcSize := *config.Workspace.DefaultStorageSize.PerWorkspace
 	if namespacedConfig != nil && namespacedConfig.CommonPVCSize != "" {
-		pvcSize = namespacedConfig.CommonPVCSize
+		pvcSize, err = resource.ParseQuantity(namespacedConfig.CommonPVCSize)
```
amisevsk, line 164: We should maybe drop this whole `namespacedConfig` check, since using the per-namespace `CommonPVCSize` doesn't really make sense here.

amisevsk (approved):
LGTM with two small notes.

pkg/config/sync.go
```diff
@@ -276,6 +290,14 @@ func logCurrentConfig() {
 			config = append(config, fmt.Sprintf("workspace.ignoredUnrecoverableEvents=%s",
 				strings.Join(Workspace.IgnoredUnrecoverableEvents, ";")))
 		}
+		if Workspace.DefaultStorageSize != nil {
+			if Workspace.DefaultStorageSize.Common != nil {
+				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.%s=%s", constants.CommonStorageClassType, Workspace.DefaultStorageSize.Common.String()))
+			}
+			if Workspace.DefaultStorageSize.PerWorkspace != nil {
+				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.%s=%s", constants.PerWorkspaceStorageClassType, Workspace.DefaultStorageSize.PerWorkspace.String()))
```
amisevsk, line 298: We should hard-code the config names, this will result in logging e.g.
```
workspace.defaultStorageSize.per-workspace=5Gi
```
even though the actual config field is
```
workspace.defaultStorageSize.perWorkspace
```

```suggestion
				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.common=%s", Workspace.DefaultStorageSize.Common.String()))
			}
			if Workspace.DefaultStorageSize.PerWorkspace != nil {
				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.perWorkspace=%s", Workspace.DefaultStorageSize.PerWorkspace.String()))
```

amisevsk (approved):

pkg/config/sync.go
AObuchow: Sounds good, I went with `per-workspace` instead of `perWorkspace`. Let me know if you rather it be `perWorkspace`

amisevsk: It should match the actual field in the config though -- the dot delimited format is meaningful here. If you try to set `workspace.defaultStorageSize.per-workspace` in the DevWorkspaceOperatorConfig, it will be invalid.

ibuziuk (commented):
@AObuchow could you please resolve the conficts?

AObuchow:
> @AObuchow could you please resolve the conficts?

Sorry about that - will be done in a moment.

ibuziuk (approved):
LGTM :+1:

pkg/config/sync.go
```diff
@@ -276,6 +289,14 @@ func logCurrentConfig() {
 			config = append(config, fmt.Sprintf("workspace.ignoredUnrecoverableEvents=%s",
 				strings.Join(Workspace.IgnoredUnrecoverableEvents, ";")))
 		}
+		if Workspace.DefaultStorageSize != nil {
+			if Workspace.DefaultStorageSize.Common != nil {
+				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.common=%s", Workspace.DefaultStorageSize.Common.String()))
+			}
+			if Workspace.DefaultStorageSize.PerWorkspace != nil {
+				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.per-workspace=%s", Workspace.DefaultStorageSize.PerWorkspace.String()))
```
amisevsk, line 297:
```suggestion
				config = append(config, fmt.Sprintf("workspace.defaultStorageSize.perWorkspace=%s", Workspace.DefaultStorageSize.PerWorkspace.String()))
```

AObuchow: Done :) Will autosquash when ready for merge

AObuchow: Should be good for merge assuming nothing else needs to be changed :sunglasses:

PR #834, opened 2022-05-09
feat: cleanup shared PVC when no longer in use

### LINKED ISSUE #826: DWO should (optionally) delete shared PVCs when no workspaces are left that require it
### Description
Currently, starting a workspace that uses the `common` or `async` storage-type results in the creation of a PVC in that workspace's namespace. This PVC exists outside that workspace's lifecycle and is shared with other workspaces in that namespace. However, even after all workspaces are deleted and the given k8s namespace is no longer used for workspaces, the PVC is left untouched.

When DWO detects that no workspaces require the common PVC anymore, it should delete that PVC to avoid leaving unused resources in the cluster.

### Additional context
We would need to verify that this approach doesn't have too heavy of a performance cost, if e.g. garbage collection on the common PVC takes a long time and blocks starting new workspaces.

AObuchow:
### What does this PR do?
When the common PVC is cleaned up after workspace deletion, a check occurs to see if other workspaces using the `common` or `shared` storage-class attribute exist in the same namespace (i.e. other workspaces are using the same shared PVC). If there are no workspaces using the shared PVC, then the PVC is marked for deletion.

Fix #826

### Is it tested? How?
- Create 2 or more workspaces using a shared PVC (eg. with the storage class set to `async`, `common`, or unset)
- Ensure that the `claim-devworkspace` PVC is created
- Delete all workspaces
- Ensure that the `claim-devworkspace` PVC is deleted shortly afterwards.

### PR Checklist

### DIFF
--- a/pkg/provision/storage/commonStorage.go
+++ b/pkg/provision/storage/commonStorage.go
@@ -19,9 +19,12 @@ import (
 	"fmt"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/pkg/provision/sync"
 
 	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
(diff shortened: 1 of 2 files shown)


pkg/provision/storage/commonStorage.go
```diff
@@ -71,8 +74,30 @@ func (p *CommonStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAd
...
+		return err
+	}
+
+	// If the number of common + async workspaces that exist (started or stopped) is zero,
+	// delete common PVC instead of running cleanup job
+	if totalWorkspaces <= 1 {
+		existingPVC := &corev1.PersistentVolumeClaim{}
+		namespacedName := types.NamespacedName{Name: config.Workspace.PVCName, Namespace: workspace.Namespace}
+		err := clusterAPI.Client.Get(clusterAPI.Ctx, namespacedName, existingPVC)
+		if err != nil {
```
AObuchow, line 89: I'm not so sure about the error handling logic here. Currently, if a `k8sErrors.IsNotFound` occurs then we will still return the error. Maybe if the PVC can't be found, we just assume the PVC is already deleted (in other words, we should ignore `k8sErrors.IsNotFound`)?

amisevsk: It's important to handle this correctly as there might be two in-flight reconciles that attempt to delete the PVC at the same time. Generally, for `client.Get` requests, we treat `k8sErrors.IsNotFound` as "assume object is already deleted", and propagate other errors. In this case, this would mean returning `nil`, as we don't need to call `client.Delete` below.

pkg/provision/storage/commonStorage.go
```diff
@@ -71,8 +74,30 @@ func (p *CommonStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAd
...
+			return err
+		}
+
+		err = clusterAPI.Client.Delete(clusterAPI.Ctx, existingPVC)
+		if err != nil && !k8sErrors.IsNotFound(err) {
+			return err
+		}
+		return nil
+	} else {
+		return runCommonPVCCleanupJob(workspace, clusterAPI)
```
amisevsk, line 99: In terms of balancing conditional branches, it probably makes sense to move this one up to the top, i.e. have 
```go
if totalWorkspaces > 1 {
  return runCommonPVCCleanupJob(workspace, clusterAPI)
} else {
  // delete PVC
}
```

pkg/provision/storage/commonStorage.go
```diff
@@ -71,8 +74,30 @@ func (p *CommonStorageProvisioner) ProvisionStorage(podAdditions *v1alpha1.PodAd
...
+		namespacedName := types.NamespacedName{Name: config.Workspace.PVCName, Namespace: workspace.Namespace}
+		err := clusterAPI.Client.Get(clusterAPI.Ctx, namespacedName, existingPVC)
+		if err != nil {
+			return err
+		}
+
+		err = clusterAPI.Client.Delete(clusterAPI.Ctx, existingPVC)
+		if err != nil && !k8sErrors.IsNotFound(err) {
+			return err
+		}
```
amisevsk, line 96: Similarly here, need to handle the `IsNotFound` error case (it's not an error, the object has already been deleted).

pkg/provision/storage/shared.go
```diff
@@ -235,3 +236,24 @@ func checkForExistingCommonPVC(namespace string, api sync.ClusterAPI) (string, e
...
+		return 0, 0, err
+	}
+	for _, workspace := range workspaces.Items {
+		storageClass := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAttribute, nil)
+		// Note, if the storageClass attribute isin't set (ie. storageClass == ""), then the storage class being used is "common"
+		if storageClass == constants.AsyncStorageClassType || storageClass == constants.CommonStorageClassType || storageClass == "" {
+			total++
+			if workspace.Spec.Started {
+				started++
+			}
```
amisevsk, line 255: If started isn't currently needed, may as well not count it for now.

amisevsk (approved):

amisevsk (approved):
LGTM!

Tested on minikube.

Please autosquash commits before merging

ibuziuk (commented):
LGTM, but looks like some of the commits are not signed (no Signed-off-by) which makes the PR check fail

PR #837, opened 2022-05-13
feat: configmap option for perWorkspace PVC size

### LINKED ISSUE #836: DWO should support per-namespace per-workspace PVC size configuration
### Description
Currently, it's possible to configure the size of the common PVC by creating a configmap in the workspace's namespace. However, this configmap only supports setting a value for the size of the common PVC, and this size is used for per-workspace PVCs as well. Instead, this configmap should support setting per-workspace PVC size as well.

AObuchow:
### What does this PR do?
Add a configmap configuration option to specify the perWorkspace PVC size, similar to how it is done for the common PVC size.

Fix #836

### Is it tested? How?
- Start up DWO in the appropriate namespace (eg. `export NAMESPACE="devworkspace-controller"`)
- Apply a config map specifying the perWorkspace PVC size in the appropriate namespace:
```bash
cat <<EOF | kubectl apply -n $NAMESPACE -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-nsconfig
  labels:
    controller.devfile.io/namespaced-config: "true"
data:
  perWorkspacePVCSize: 7Gi
EOF
```
- Apply a devfile that uses the perWorkspace storage class attribute to the same namespace, eg. `kubectl apply -f ./samples/theia-next_per-workspaceStorage.yaml -n $NAMESPACE`
- Verify in Kubernetes (eg. the dashboard) that the newly created PVC has the expected size (`7Gi`)
### PR Checklist

### DIFF
--- a/pkg/provision/config/config.go
+++ b/pkg/provision/config/config.go
@@ -30,11 +30,13 @@ import (
 )
 
 const (
-	commonPVCSizeKey = "commonPVCSize"
+	commonPVCSizeKey       = "commonPVCSize"
+	perWorkspacePVCSizeKey = "perWorkspacePVCSize"
 )
 
 type NamespacedConfig struct {
-	CommonPVCSize string
+	CommonPVCSize       string
+	PerWorkspacePVCSize string
 }
 
 // ReadNamespacedConfig reads the per-namespace DevWorkspace configmap and returns it as a struct. If there are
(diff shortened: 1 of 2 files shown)


AObuchow:
Note: It's probably worth waiting for https://github.com/devfile/devworkspace-operator/pull/827 to be merged and then rebase & modify this PR since they both conflict with eachother slightly.

amisevsk (approved):

ibuziuk (commented):
@AObuchow nice improvement, but it looks like as an admin how can I specify the pvc strategy in the general cas?. Is it possible  to setup the PVC config once  on the operator level e.g. use per-worskpaces strategy with 5 GB for all devworkspace? It is nice to have the option with cm that allows granularity, but we should also have the option for common configuration

ibuziuk:
heh, looks like the general config is addressed as part of https://github.com/devfile/devworkspace-operator/pull/827

ibuziuk (approved):

PR #839, opened 2022-05-16
feat: add WorkspaceStopped method to RoutingSolver

dkwon17:

### What does this PR do?
Adds a `func (s *ClusterSolver) WorkspaceStopped(routing *controllerv1alpha1.DevWorkspaceRouting, workspaceMeta DevWorkspaceMetadata) error` method to `RoutingSolver` interface. This function runs when the corresponding devworkspace has stopped. 

This PR was created to help support https://github.com/eclipse-che/che-operator/pull/1386

### Is it tested? How?

1. Check out this branch: https://github.com/dkwon17/devworkspace-operator/tree/workspacestoppedtesting
This branch contains the same commit as this current PR's branch, except that there is a [new commit](https://github.com/dkwon17/devworkspace-operator/commit/506e11e6ff354a97913af565070309c965e44ba2) on top of it with a print statement that can be used for testing.
2. After checking out the above branch, run the code locally by following these instructions: https://github.com/devfile/devworkspace-operator#run-controller-locally
3. Run the following to create a devworkspace in the `devworkspace-controller` namespace from the devworkspace samples:
```
kubectl apply -f samples/theia-latest.yaml -n devworkspace-controller
```
4. Stop the devworkspace by setting `spec.started` to false.

5. In the output from the terminal used to run DWO from step 2, verify that the we see the print statement:
image

### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspacerouting_types.go
+++ b/apis/controller/v1alpha1/devworkspacerouting_types.go
@@ -16,6 +16,7 @@
 package v1alpha1
 
 import (
+	"github.com/devfile/devworkspace-operator/pkg/constants"
 	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -204,4 +205,8 @@ type DevWorkspaceRoutingList struct {
 
 func init() {
 	SchemeBuilder.Register(&DevWorkspaceRouting{}, &DevWorkspaceRoutingList{})
+}
+
+func (d *DevWorkspaceRouting) IsWorkspaceStopped() bool {
+	return d.Annotations != nil && d.Annotations[constants.DevWorkspaceStartedStatusAnnotation] == "false"
 }
--- a/controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
+++ b/controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
@@ -107,8 +107,15 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(ctx context.Context, req ctrl.
 		return reconcile.Result{}, r.finalize(solver, instance)
 	}
 
-	if instance.Annotations != nil && instance.Annotations[constants.DevWorkspaceStartedStatusAnnotation] == "false" {
-		return reconcile.Result{}, nil
+	workspaceMeta := solvers.DevWorkspaceMetadata{
+		DevWorkspaceId: instance.Spec.DevWorkspaceId,
+		Namespace:      instance.Namespace,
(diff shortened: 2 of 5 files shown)


dkwon17:
For more context, the approach for https://github.com/eclipse-che/che-operator/pull/1386 is that the traefik config is updated when the devworkspace stops. The traefik config is updated so that the mainurl for the workspace is routed to the dashboard.

I use the `WorkspaceStopped` method in that PR: https://github.com/eclipse-che/che-operator/blob/a8cbf103af1a108fab4d10d53b8fd1255dbaa7e2/controllers/devworkspace/solver/solver.go#L200-L206 so that I can provision new traefik config when the workspace has stopped.

ibuziuk (approved):
This is a nice improvement :+1:

apis/controller/v1alpha1/devworkspacerouting_types.go
```diff
@@ -205,3 +206,7 @@ type DevWorkspaceRoutingList struct {
 func init() {
 	SchemeBuilder.Register(&DevWorkspaceRouting{}, &DevWorkspaceRoutingList{})
 }
+
+func (d *DevWorkspaceRouting) IsWorkspaceStopped() bool {
+	return d.Annotations != nil && d.Annotations[constants.DevWorkspaceStartedStatusAnnotation] == "false"
+}
```
amisevsk, line 212: I'm not 100% on whether or not the type definition files should include struct methods like this, but this one is probably fine.

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -107,8 +107,15 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(ctx context.Context, req ctrl.
...
-	if instance.Annotations != nil && instance.Annotations[constants.DevWorkspaceStartedStatusAnnotation] == "false" {
-		return reconcile.Result{}, nil
+	workspaceMeta := solvers.DevWorkspaceMetadata{
+		DevWorkspaceId: instance.Spec.DevWorkspaceId,
+		Namespace:      instance.Namespace,
+		PodSelector:    instance.Spec.PodSelector,
+	}
+
+	if instance.IsWorkspaceStopped() {
+		err := solver.WorkspaceStopped(instance, workspaceMeta)
```
amisevsk, line 117: Is there a case where stopping a workspace may require a second reconcile? I'm wonder if we need an error type (or to just reuse `RoutingNotReady`) to signify that a requeue is needed but no actual error occurred.

amisevsk (commented):
Generally looks good, great addition!

amisevsk:
@dkwon17 it's up to you -- if there's no intention of merging this PR at the moment it can be closed and re-implemented later if need be.

dkwon17:
Sounds good, I will close for now

PR #840, opened 2022-05-19
fix: Adding 'KUBERNETES_SERVICE_HOST' env var to the 'no_proxy / NO_PROXY' list

ibuziuk:
Signed-off-by: Ilya Buziuk <ibuziuk@redhat.com>

### What does this PR do?
Adding 'KUBERNETES_SERVICE_HOST' env var to the 'no_proxy / NO_PROXY' list

Hot Fix for https://issues.redhat.com/browse/CRW-2820 

since DWO is a standalone operator it has it is own mechanism for proxy config detection. DWO automatically reads values from cluster [proxies.config.openshift.io](http://proxies.config.openshift.io/) and propagates them as env vars to workspace
At the same time, Che Operator detects proxy configuration in a similar fashion as the DWO, and creates a cm with the right label so that DWO would be able to mount this cm to workspace deployment
the problem is that values from cm are overridden by the proxy env vars that are detected by DWO, e.g. this particular case when KUBERNETES_SERVICE_HOST value is added to the cm on the Che operator level but DWO during the workspace startup sets NO_PROXY env var that it detected

Initial fix on the che-operator end - https://github.com/eclipse-che/che-operator/pull/1387

### Is it tested? How?

 image `quay.io/ibuziuk/devworkspace-controller:crw-2820`

- setup a cluster using cluster bot `launch 4.10 aws,proxy`
- Run the DWO with a fix
  `export DWO_IMG=quay.io/ibuziuk/devworkspace-controller:crw-2820`
  `export NAMESPACE="devworkspace-controller"`
  `make install`
- Create a sample DW `kubectl apply -f ./samples/flattened_theia-next.yaml` - the workspace panel is shown

image

image

### PR Checklist

### DIFF
--- a/pkg/library/env/workspaceenv.go
+++ b/pkg/library/env/workspaceenv.go
@@ -17,6 +17,7 @@ package env
 
 import (
 	"fmt"
+	"os"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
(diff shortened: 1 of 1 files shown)


amisevsk (commented):
I'm not sure I understand these changes -- when I test in a proxy-enabled cluster, the service host is already present in the NO_PROXY env var.

amisevsk:
e.g. 

> $ echo $NO_PROXY
> .cluster.local,.svc,.us-west-1.compute.internal,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,**172.30.0.0/16**,api-int.ci-ln-8kbd79b-76ef8.origin-ci-int-aws.dev.rhcloud.com,localhost

amisevsk:
It seems like the issue is that the cluster is configured to specify no_proxy assuming the client supports CIDR blocks, but Theia does not do this. `172.30.0.1` is included in `172.30.0.0/16`

This should maybe be documented in the code.

amisevsk:
This is known behaviour in the k8s javascript client, apparently: https://github.com/kubernetes-client/javascript/issues/233. 

The in-cluster config attempts to use `$KUBERNETES_SERVICE_HOST` for the api server rather than the default https://kubernetes.default.svc/ DNS name in order to work (I assume) on clusters without DNS configured.

PR #842, opened 2022-05-24
fix: updating che-machine-exec tag in the sample

ibuziuk:
### PR Checklist


amisevsk (approved):
The flattened examples should be removed entirely IMO. They're from a period in time where we didn't yet support plugin components, and so these devworkspaces are completely out of date with the standard Theia workspace now.

Issue #656, opened 2021-10-21
Extend the DevWorkspaceTemplate spec with containers contributions


sleshchenko:
Che Editors (VS Code, Theia, JetBrain IDEs) are defined as `DevWorkspaceTemplate`. But the `DevWorkspaceTemplate` syntax currently lacks a way to express the changes required to the containers that will host the editor. In fact Che Editors are injected in user's containers (which one is covered [here](https://github.com/eclipse/che/issues/21354)) and require changes to the original user's container to work properly: add env variables, volumes, mem and cpu requisites, postStart and postStart events etc...

This feature is critical to allow using the same Devfile or DevWorkspace with different editors. But also to unblock some other features as "universal" idling.

This is a subtask of https://github.com/eclipse/che/issues/21340

Here is an example of a `DevWorkspaceTemplate` that defines VS Code with a `containerContribution` component:

VS Code editor definition:
```yaml
apiVersion: workspace.devfile.io/v1alpha2
kind: DevWorkspaceTemplate
metadata:
  name: che-code-editor
spec:
  components:
    - name: che-code-volume
      volume: {}
    - name: che-code-injector
      container:
        image: quay.io/che-incubator/che-code:insiders
        command:
          - /entrypoint-init-container.sh
        volumeMounts:
          - name: che-code-volume
            path: /checode
        memoryLimit: 128Mi
        memoryRequest: 32Mi
        cpuLimit: 500m
        cpuRequest: 30m
    - name: che-code-host
      containerContribution:
          endpoints:
            - name: che-code
              attributes:
                type: main
                cookiesAuthEnabled: true
                discoverable: false
                urlRewriteSupported: true
                contributed-by: che-code.eclipse.org
              targetPort: 3100
              exposure: public
              path: '?tkn=eclipse-che'
              secure: false
              protocol: https
            - name: code-redirect-1
              attributes:
                discoverable: false
                urlRewriteSupported: true
                contributed-by: che-code.eclipse.org
              targetPort: 13131
              exposure: public
              protocol: http
            - name: code-redirect-2
              attributes:
                discoverable: false
                urlRewriteSupported: true
                contributed-by: che-code.eclipse.org
              targetPort: 13132
              exposure: public
              protocol: http
            - name: code-redirect-3
              attributes:
                discoverable: false
                urlRewriteSupported: true
                contributed-by: che-code.eclipse.org
              targetPort: 13133
              exposure: public
              protocol: http
          volumeMounts:
[...]

PR #844, opened 2022-05-27
Add support for containerContributions

amisevsk:
**This PR is an early draft. Before merging:**
- [x] Decide on format for specifying container contributions cc: @l0rd, @benoitf
  - [x] Is the current design sufficient? Should we define contributions in another way?
  - [x] Does the mechanism need a way to specify which container to merge into, or is "merge into the first available target" sufficient? (This can be added later if necessary)
  - [x] Does this approach cover all use cases for Code-OSS and Theia?
- [x] Write tests to cover feature once design is settled
- [x] Clean up and polish

### What does this PR do?
Adds support for container contributions as described in https://github.com/devfile/devworkspace-operator/issues/656 without needing to add new component types.

This PR adds two new attributes to the operator:

* `controller.devfile.io/container-contribution: true` defines a container component as a "container contribution"
* `controller.devfile.io/merge-contribution: true` defines a container as a target for merging a "container contribution"

If a flattened DevWorkspace has a container component with the `merge-contribution` attribute, then any container contributions are merged into that container component according to

* Resource requirements (memory/cpu limits/requests) are summed
* The image from the contribution component is ignored
* Remaining fields are merged according to a Kubernetes strategic merge patch (e.g. env vars and endpoints are appended)

This approach has the additional benefit of failing safely if it's not obvious how to merge containers, defaulting to using the unmerged image to host the container components that cannot be merged.

If multiple components have the `merge-contribution` attribute, contributions are merged into the first one encountered.

### Background
Eclipse Che uses DWO to create workspaces from a basic devfile, injecting an editor into the Devfile before converting it to a DevWorkspace. However, since Che intends to run editors inside a devfile container component and not as sidecars, it's necessary to _modify_ existing container components to add endpoints, environment variables, and override the args. For example, to update a container component to run Eclipse Theia plugins, Che has to update the component to add the following fields:
```diff
 components:
   - name: tools
+    attributes:
+      app.kubernetes.io/name: tools
+      che-theia.eclipse.org/vscode-extensions:
+        - https://github.com/golang/vscode-go/releases/download/v0.23.0/go-0.23.0.vsix
     container:
+      args:
+        - ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}
+        - -c
+        - sh
       env:
         - name: GOCACHE
           value: /tmp/.cache
         - name: GOPATH
           value: /projects:/home/user/go
+        - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
+          value: /remote-endpoint/plugin-remote-endpoint
+        - name: THEIA_PLUGINS
+          value: local-dir:///plugins/sidecars/tools
       image: quay.io/devfile/universal-developer-image:ubi8-0e189d9
       memoryLimit: 2Gi
       mountSources: true
+      sourceMapping: /projects
+      volumeMounts:
+        - name: plugins
+          path: /plugins
+        - name: remote-endpoint
+          path: /remote-endpoint
   # Theia is added as a plugin here
+  - name: theia-ide-golang-example
+    plugin:
+      kubernetes:
+        name: theia-ide-golang-example
```
The added fields are defined by the plugin (`theia-ide-golang-example` in this case), so it would be useful if plugins had a mechanism for automatically defining these additions. This would also mean that devfiles could be converted into DevWorkspaces by only adding plugin components, leaving original devfile components untouched (as the merging happens within the flatten process, leaving the DevWorkspace as-is).

Closes https://github.com/devfile/devworkspace-operator/issues/656

### Is it tested? How?
There are two DevWorkspaces for testing this feature:

1. Che-Theia + Golang [[link]](https://gist.github.com/amisevsk/46da73704a126d2324e5c633ff077b8c#file-theia-golang-devworkspace-yaml)
    ```
    kubectl apply -f https://gist.githubusercontent.com/amisevsk/46da73704a126d2324e5c633ff077b8c/raw/41dee8cbe9531363ffd8975982bdc1113d7b58c4/theia-golang.devworkspace.yaml
    ```
    Creates a DevWorkspace that references plugin devfile [theia-next.yaml](https://gist.github.com/amisevsk/b342b9ae064b5a79e08945d04d02c208). The contributions in this plugin add fields as described above to allow Theia plugins to run inside the `tools` container component without having to update the `tools` component in the DevWorkspace itself (duplicating the process that Che does when creating a workspace from a devfile). The diff between the referenced plugin and the default plugin at https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml is
    ```diff
    @@ -9,6 +9,26 @@
       preStart:
         - init-container-command
     components:
    +  - name: theia-ide-contributions
    +    attributes:
    +      controller.devfile.io/container-contribution: true
    +    container:
    +      args:
    +        - sh
    +        - '-c'
    +        - '${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}'
    +      env:
    +        - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
    +          value: /remote-endpoint/plugin-remote-endpoint
    +        - name: THEIA_PLUGINS
    +          value: local-dir:///plugins/sidecars/tools
    +      memoryLimit: 512Mi
    +      volumeMounts:
    +        - name: plugins
    +          path: /plugins
    +        - name: remote-endpoint
    +          path: /remote-endpoint
    +      image: quay.io/devfile/universal-developer-image@sha256:53cec58dd190dd1e06100478ae879d7c28abd8fc883d5fdf5be3eb6e943fe5e7
       - name: theia-ide
         container:
           image: quay.io/eclipse/che-theia:next
    ```

2. Che-Code + Golang [[link]](https://gist.github.com/amisevsk/46da73704a126d2324e5c633ff077b8c#file-code-golang-devfile-yaml)
    ```
    kubectl apply -f https://gist.githubusercontent.com/amisevsk/46da73704a126d2324e5c633ff077b8c/raw/9af5c2d5ffa2012e760acdcd6fa4427fb611af1c/code-golang.devfile.yaml
    ```
    Creates a DevWorkspace that references the plugin devfile [che-code.yaml](https://gist.github.com/amisevsk/53f4f8281c1339599a553f3be2d5c650). This plugin is a modified version of the current plugin hosted [here](https://che-plugin-registry-main.surge.sh/v3/plugins/che-incubator/che-code/insiders/devfile.yaml), which converts the `che-code-runtime-description` into a container contribution. This workspace results in a deployment with two init containers (`che-code-injector` and `project-clone`) and only one running container (`tools`). The changes to the plugin used here compared to the one in the registry are:
    ```diff
    @@ -13,8 +13,9 @@
         attributes:
           app.kubernetes.io/component: che-code-runtime
           app.kubernetes.io/part-of: che-code.eclipse.org
    +      controller.devfile.io/container-contribution: true
         container:
    -      image: quay.io/devfile/universal-developer-image@sha256:53cec58dd190dd1e06100478ae879d7c28abd8fc883d5fdf5be3eb6e943fe5e7
    +      image: quay.io/devfile/universal-developer-image:ubi8-0e189d9
           command:
             - /checode/entrypoint-volume.sh
           volumeMounts:
    ```
    (I updated the image to match the one from the devfile to avoid having to pull two separate UDI images)

### PR Checklist

### DIFF
--- a/pkg/library/flatten/merge.go
+++ b/pkg/library/flatten/merge.go
@@ -18,8 +18,10 @@ package flatten
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/api/v2/pkg/attributes"
 	"github.com/devfile/api/v2/pkg/utils/overriding"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -161,7 +163,6 @@ func mergeContainerContributions(flattenedSpec *dw.DevWorkspaceTemplateSpec) err
 			if err != nil {
 				return fmt.Errorf("failed to merge container contributions: %w", err)
 			}
-			delete(mergedComponent.Attributes, constants.ContainerContributionAttribute)
 			newComponents = append(newComponents, *mergedComponent)
 			mergeDone = true
 		} else {
@@ -186,9 +187,11 @@ func mergeContributionsInto(mergeInto *dw.Component, contributions []dw.Componen
 	}
 
 	// We don't want to reimplement the complexity of a strategic merge here, so we set up a fake plugin override
-	// and use devfile/api overriding functionality. For specific fields that have to be handled specifically (memory
+	// and use devfile/api overriding functionality. For specific fields that have to be handled specially (memory
 	// and cpu limits, we compute the value separately and set it at the end
 	var toMerge []dw.ComponentPluginOverride
+	// Store names of original plugins to allow us to generate the merged-contributions attribute
+	var mergedComponentNames []string
 	for _, component := range contributions {
 		if component.Container == nil {
 			return nil, fmt.Errorf("attempting to merge container contribution from a non-container component")
@@ -197,6 +200,11 @@ func mergeContributionsInto(mergeInto *dw.Component, contributions []dw.Componen
 		component.Name = mergeInto.Name
 		// Unset image to avoid overriding the default image
 		component.Container.Image = ""
+		// Store original source attribute's value and remove from component
+		if component.Attributes.Exists(constants.PluginSourceAttribute) {
+			mergedComponentNames = append(mergedComponentNames, component.Attributes.GetString(constants.PluginSourceAttribute, nil))
+			delete(component.Attributes, constants.PluginSourceAttribute)
+		}
 		if err := addResourceRequirements(totalResources, &component); err != nil {
 			return nil, err
 		}
@@ -232,6 +240,13 @@ func mergeContributionsInto(mergeInto *dw.Component, contributions []dw.Componen
 
 	mergedComponent := mergedSpecContent.Components[0]
 	applyResourceRequirementsToComponent(mergedComponent.Container, totalResources)
+
+	if mergedComponent.Attributes == nil {
+		mergedComponent.Attributes = attributes.Attributes{}
+	}
+	mergedComponent.Attributes.PutString(constants.MergedContributionsAttribute, strings.Join(mergedComponentNames, ","))
+	delete(mergedComponent.Attributes, constants.MergeContributionAttribute)
+	delete(mergedComponent.Attributes, constants.ContainerContributionAttribute)
 
 	return &mergedComponent, nil
 }
--- a/pkg/library/flatten/merge.go
+++ b/pkg/library/flatten/merge.go
@@ -16,10 +16,12 @@
 package flatten
 
 import (
+	"encoding/json"
 	"fmt"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/api/v2/pkg/utils/overriding"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
@@ -117,4 +119,119 @@ func mergeVolume(into, from *dw.VolumeComponent) error {
 		into.Size = from.Size
 	}
 	return nil
+}
+
+func needsContainerContributionMerge(flattenedSpec *dw.DevWorkspaceTemplateSpec) bool {
+	hasContribution, hasTarget := false, false
+	for _, component := range flattenedSpec.Components {
+		if component.Container == nil {
+			// Ignore attribute on non-container components as it's not clear what this would mean
+			continue
+		}
+		if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) {
+			hasContribution = true
+		}
+		if component.Attributes.GetBoolean(constants.MergeContributionAttribute, nil) {
+			hasTarget = true
+		}
+	}
+	return hasContribution && hasTarget
+}
+
+func mergeContainerContributions(flattenedSpec *dw.DevWorkspaceTemplateSpec) error {
+	var contributions []dw.Component
+	for _, component := range flattenedSpec.Components {
+		if component.Container != nil && component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) {
+			contributions = append(contributions, component)
+		}
+	}
+
+	var newComponents []dw.Component
+	mergeDone := false
+	for _, component := range flattenedSpec.Components {
+		if component.Container == nil {
+			newComponents = append(newComponents, component)
+			continue
+		}
+		if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) {
+			// drop contributions from updated list as they will be merged
+			continue
+		} else if component.Attributes.GetBoolean(constants.MergeContributionAttribute, nil) && !mergeDone {
+			mergedComponent, err := mergeContributionsInto(&component, contributions)
+			if err != nil {
+				return fmt.Errorf("failed to merge container contributions: %w", err)
+			}
+			delete(mergedComponent.Attributes, constants.ContainerContributionAttribute)
+			newComponents = append(newComponents, *mergedComponent)
+			mergeDone = true
+		} else {
+			newComponents = append(newComponents, component)
+		}
+	}
+
+	if mergeDone {
+		flattenedSpec.Components = newComponents
+	}
+
+	return nil
+}
+
+func mergeContributionsInto(mergeInto *dw.Component, contributions []dw.Component) (*dw.Component, error) {
+	if mergeInto == nil || mergeInto.Container == nil {
+		return nil, fmt.Errorf("attempting to merge container contributions into a non-container component")
+	}
+	totalResources, err := parseResourcesFromComponent(mergeInto)
+	if err != nil {
+		return nil, err
+	}
+
+	// We don't want to reimplement the complexity of a strategic merge here, so we set up a fake plugin override
+	// and use devfile/api overriding functionality. For specific fields that have to be handled specifically (memory
+	// and cpu limits, we compute the value separately and set it at the end
+	var toMerge []dw.ComponentPluginOverride
+	for _, component := range contributions {
+		if component.Container == nil {
+			return nil, fmt.Errorf("attempting to merge container contribution from a non-container component")
+		}
+		// Set name to match target component so that devfile/api override functionality will apply it correctly
+		component.Name = mergeInto.Name
+		// Unset image to avoid overriding the default image
+		component.Container.Image = ""
+		if err := addResourceRequirements(totalResources, &component); err != nil {
+			return nil, err
+		}
+		component.Container.MemoryLimit = ""
+		component.Container.MemoryRequest = ""
+		component.Container.CpuLimit = ""
+		component.Container.CpuRequest = ""
+		// Workaround to convert dw.Component into dw.ComponentPluginOverride: marshal to json, and unmarshal to a different type
+		// This works since plugin overrides are generated from components, with the difference being that all fields are optional
+		componentPluginOverride := dw.ComponentPluginOverride{}
+		tempJSONBytes, err := json.Marshal(component)
+		if err != nil {
+			return nil, err
+		}
+		if err := json.Unmarshal(tempJSONBytes, &componentPluginOverride); err != nil {
+			return nil, err
+		}
+		toMerge = append(toMerge, componentPluginOverride)
+	}
+
+	tempSpecContent := &dw.DevWorkspaceTemplateSpecContent{
+		Components: []dw.Component{
+			*mergeInto,
+		},
+	}
+
+	mergedSpecContent, err := overriding.OverrideDevWorkspaceTemplateSpec(tempSpecContent, dw.PluginOverrides{
+		Components: toMerge,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	mergedComponent := mergedSpecContent.Components[0]
+	applyResourceRequirementsToComponent(mergedComponent.Container, totalResources)
+
+	return &mergedComponent, nil
 }
--- a/pkg/library/flatten/merge.go
+++ b/pkg/library/flatten/merge.go
@@ -123,21 +123,37 @@ func mergeVolume(into, from *dw.VolumeComponent) error {
 	return nil
 }
 
-func needsContainerContributionMerge(flattenedSpec *dw.DevWorkspaceTemplateSpec) bool {
+// needsContainerContributionMerge returns whether merging container contributions is necessary for this workspace. Merging
+// is necessary if at least one component has the merge-contribution: true attribute and at least one component has the
+// container-contribution: true attribute. If either attribute is present but cannot be parsed as a bool, an error is returned.
+func needsContainerContributionMerge(flattenedSpec *dw.DevWorkspaceTemplateSpec) (bool, error) {
 	hasContribution, hasTarget := false, false
+	var errHolder error
 	for _, component := range flattenedSpec.Components {
 		if component.Container == nil {
 			// Ignore attribute on non-container components as it's not clear what this would mean
 			continue
 		}
-		if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) {
-			hasContribution = true
+		// Need to check existence before value to avoid potential KeyNotFoundError
+		if component.Attributes.Exists(constants.ContainerContributionAttribute) {
+			if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, &errHolder) {
+				hasContribution = true
+			}
+			if errHolder != nil {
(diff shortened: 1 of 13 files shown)


amisevsk:
@dkwon17 @AObuchow I'm not able to request review from you yet, but please take a look as well :)

amisevsk:
To add, this PR enables more easily switching editors in the Che case, as only the plugin needs to be swapped out (and potentially the annotation saying which vsix plugins need to be installed). With the existing approach in Che, the changes to the devfile have to be undone before another editor can be applied

l0rd:
@amisevsk thanks I will test that later. To help me understanding the approach can you please confirm that it requires that:

1. The DevWorkspaceTemplate that defines the IDE include the attribute `controller.devfile.io/container-contribution: true`
2. The DevWorkspace that references the IDE DWT (as a plugin) includes a component with the attribute `controller.devfile.io/merge-contribution: true`

Or are the attributes inferred automatically (for instance if the component is a plugin then, when flattened, it will have `container-contribution: true`)?

amisevsk:
At the moment, attributes are explicit and have to be defined in the DevWorkspace/plugins. I took this approach to avoid breaking devfiles, as e.g. the che-docs devfile does not use an UDI image and injecting the editor stuff there might break things. The `merge-contribution` attribute could be added when plugins are added (e.g. in the Che dashboard), and the `container-contribution` attribute would be included in the Theia/che-code plugin devfile.

Part of my hope with how this works is that plugins with container contributions can use a fallback image in case injecting is not possible. For example, che-code could specify an image that will mostly work to host the editor in the case that a container that can run it is not found -- I believe this is the case with the current plugin definition.

amisevsk:
I've tested the container contribution functionality with the existing Che dashboard:
* For Theia, the dashboard will add env vars, args, and volumeMounts to the devfile container to inject Theia
* These values will be overwritten by the containerContributions from the theia plugin, but this is a no-op since they're the same values (the dashboard gets these values from the plugin devfile anyways)
* Removing the dashboard-added values does not change functionality as the contribution adds them back into the "flattened" devworkspace.

However, in the vscode case, it appears that the dashboard drops the `che-code-runtime-description` component before creating the DevWorkspaceTemplate, breaking injection (since DWO never sees the contribution component). Testing directly by changing the plugin to point at the plugin devfile (instead of the DWT) seems to work fine though.

A note on testing: to test this flow, it's not possible to use standalone devfiles due to https://github.com/eclipse/che/issues/21433. The correct flow appears to be

1. Create a workspace using a factory URL referencing a github repo (I use https://github.com/che-samples/golang-example/tree/devfilev2) and the editor to be tested
2. Update the DevWorkspace in-cluster to add the `merge-contribution` attribute and replace the plugin with the intended one.

benoitf:
Hello,

is that `controller.devfile.io/merge-contribution: true` is optional and then it takes the first container of the Devfile ?

because I don't expect to have users having to change their devfile.

For `controller.devfile.io/container-contribution: true` yes we can add that into the editors definitions in the plugin registry.

amisevsk:
> is that `controller.devfile.io/merge-contribution: true` is optional and then it takes the first container of the Devfile ? Because I don't expect to have users having to change their devfile.

Currently it's not optional, and merging does not happen if it is not present. The reason for this is that it could be a breaking change -- for example, I don't know if the [che-docs](https://github.com/eclipse-che/che-docs) repository would work if the injection happened automatically, since it does not use the common UDI image. If automatically merging contributions breaks your devfile, then you'd have to change your devfile to make it work again, as opposed to changing it to make it work in a different way.

The upshot is that it should be possible to write the editor plugins so that the workspace starts and shows an editor even if merging does not happen, as the contribution component could bring its own default image.

If we want the merge-contribution feature to be inferred, this could also be done in the Che Dashboard, where we currently manually do the injection. For example, in the case of che-code, instead of taking the fields from the `che-code-runtime-description` component and applying them to the main developer container, the dashboard could just add the `merge-contribution` attribute when it adds a the plugin.

benoitf:
@dkwon17 you may have to add `CODE_HOST` env variable set to `0.0.0.0` 
https://github.com/che-incubator/che-code/pull/41

amisevsk:
@dkwon17 Shouldn't be happening :thinking:. Florent is right (in case your DevWorkspace doesn't have the CODE_HOST override, though that is in the sample devfile so if you used that it might be something else.

dkwon17:
I see, I'm still not able to open the editor even though I'm using the sample devfile that @amisevsk provided. Here is the flattened devfile I retrieved by running `cat $DEVWORKSPACE_FLATTENED_DEVFILE` from the `tools` container:
https://gist.githubusercontent.com/dkwon17/a2c6cccbeacdf984eb9ab7230d507117/raw/1ec8dbe13df965aa241be0168c42226b5c9843a9/flattened-code-golang.yaml

I also confirmed that the `CODE_HOST` environment variable is also set in the tools container:
```
sh-4.4$ echo $CODE_HOST
0.0.0.0
```
Also, the devworkspace seems to enter the running state without any issues:
image

Here is the yaml for the `test-contrib-code` devworkspace in the screenshot ^  https://gist.githubusercontent.com/dkwon17/4b1e768f9873d262219e9f85ed940fdf/raw/fcc074c03962355bc03f68f7c516d40ba8cbfab5/golangcode.yaml

ibuziuk:
hit the same problem as David during the vs code verification. 
from my end, I think it is ok to make the attributes required in the initial PR to make sure that no existing devfiles are broken and promote it to the dogfooding.

l0rd:
After discussion yesterday with @amisevsk +1 to require the explicit attributes: adding the `merge-contribution: true` attribute when devfiles don't have one will be Che dashboard responsibility. It makes sense that who adds the contribution (e.g. Che dashboard) selects the contribution recipient too.

On the other hand, to keep the `DevWorkspace.spec.template` as close to a devfile as possible we have discussed to introduce a new `contributions` field at the `DevWorkspace.spec` level. That would allow to avoid using plugins for contributions. Plugins would continue to be supported with their existing behavior (no impact on existing Che/OCP) and we will consider to deprecate them after Che and the OpenShift Console has been updated to use contributions instead of plugins. 

The [theia-golang example above](https://gist.github.com/amisevsk/46da73704a126d2324e5c633ff077b8c#file-theia-golang-devworkspace-yaml) would change like this:

```diff
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: test-contrib-theia
spec:
  started: true
+ contributions:
+  - name: che-theia-ide
+    uri: https://gist.githubusercontent.com/amisevsk/b342b9ae064b5a79e08945d04d02c208/raw/84cd2df9266b4a608fbc0fd0cc18816bca03e344/theia-contrib.devfile.yaml:
  template:
    attributes:
      controller.devfile.io/storage-type: ephemeral
    components:
    (...)      
-      - name: che-theia-ide
-        plugin:
-          uri: https://gist.githubusercontent.com/amisevsk/b342b9ae064b5a79e08945d04d02c208/raw/84cd2df9266b4a608fbc0fd0cc18816bca03e344/theia-contrib.devfile.yaml
```

AObuchow:
I was able to test out both the che-theia and che-code gists provided with this PR.
The code looks good to me so far as well.

I am also +1 for making the `controller.devfile.io/merge-contribution` attribute required to prevent breaking any existing devfile.

amisevsk:
PR https://github.com/devfile/api/pull/879 would add the proposed containerContributions field if merged. However, in the interest of sequencing PRs/issues, I'm marking this PR as ready to review -- if there's no issues with this approach this PR can be merged and the functionality can be extended/applied to the `contributions` field as well.

amisevsk:
@svor good catch -- updated.

ibuziuk (approved):

amisevsk:
Finalized tests for this PR; ready for review.

amisevsk:
I've tested the issue mentioned by @dkwon17 -- when using just the DevWorkspace Operator on OpenShift, it appears that the `?tkn=eclipse-che` parameter in the default URL causes the application not available as it redirects and strips the `/che-code/` portion of the URL. When accessing

```
https://<workspace-id>.<clusterhost>.com/che-code/?tkn=eclipse-che
```
the response is
```
< HTTP/1.1 302 Found
< set-cookie: vscode-tkn=eclipse-che; Max-Age=604800; SameSite=Lax
< location: https://<workspace-id>.<clusterhost>.com/
```
Accessing instead `https://<workspace-id>.<clusterhost>.com/che-code/` leads to the che-code editor as expected. On Kubernetes (minikube) this issue is not present as far as I can tell.

cc: @benoitf @l0rd any ideas what's happening here?

If you have Eclipse Che installed, you can edit the che-code + golang devworkspace to specify `.spec.routingClass = che`. In that case the workspace can be accessed as intended. In this case, the workspace URL is

```
https://eclipse-che.<clusterhost>.com/<workspaceID>/tools/3100/?tkn=eclipse-che
```
which gets redirected to
```
https://eclipse-che.<clusterhost>.com/<workspaceID>/tools/3100/
```

amisevsk:
To test these changes via OLM, I've build an index image `quay.io/amisevsk/devworkspace-operator-index:container-contributions` this can be used to create a catalogsource on the cluster and deploy the current changes via OLM.

l0rd:
I don't recall why we need to add the suffix `?tkn=eclipse-che` to VS Code URL. @azatsarynnyy may know that?

azatsarynnyy:
@l0rd @amisevsk 
The connection token (passed through the `tkn` query parameter) used to be required by VS Code some time ago.
Currently, it's optional and we don't use it in Che-Code.
Che-Code is launched with `--without-connection-token` startup parameter:
https://github.com/che-incubator/che-code/blob/d6291d3ede49eabd27e331ded40d465b7b5f0385/build/scripts/entrypoint-volume.sh#L76

l0rd:
Thanks @azatsarynnyy for the clarification. I think that then we should replace [that path](https://github.com/eclipse-che/che-plugin-registry/blob/main/che-editors.yaml#L523) with `?folder=${PROJECT_SOURCE}` in the `DevWorkspaceTemplate` to fix #21400. And maybe we can also get rid of our built in extension `che-on-start`?

azatsarynnyy:
> Thanks @azatsarynnyy for the clarification. I think that then we should replace [that path](https://github.com/eclipse-che/che-plugin-registry/blob/main/che-editors.yaml#L523) with `?folder=${PROJECT_SOURCE}` in the `DevWorkspaceTemplate` to fix #21400. And maybe we can also get rid of our built in extension `che-on-start`?

@l0rd please see my comment on https://github.com/eclipse/che/issues/21400#issuecomment-1253648569
Getting rid of `che-on-start` extension means losing such important VS Code functionality as the [multi-root workspaces support](https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces).

amisevsk:
In that case, it appears that this issue isn't a real problem for merging the current PR -- we can figure out the parameters later and it works as intended in the Che dashboard in my testing. I'm still interested in why providing the `?tkn=eclipse-che` parameter causes a wrong redirect, but if there's an issue in DWO there it's likely present in main as well.

pkg/library/flatten/merge.go
```diff
@@ -118,3 +120,118 @@ func mergeVolume(into, from *dw.VolumeComponent) error {
...
+	}
+
+	var newComponents []dw.Component
+	mergeDone := false
+	for _, component := range flattenedSpec.Components {
+		if component.Container == nil {
+			newComponents = append(newComponents, component)
+			continue
+		}
+		if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) == true {
```
AObuchow, line 158: Though this line (and other similar ones) could be just `if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) {...` I think it's okay to be explicit here about wanting the value of the `ContainerContributionAttribute` to be true.

pkg/library/flatten/merge.go
```diff
@@ -118,3 +120,118 @@ func mergeVolume(into, from *dw.VolumeComponent) error {
...
 }
+
+func needsContainerContributionMerge(flattenedSpec *dw.DevWorkspaceTemplateSpec) bool {
+	hasContribution, hasTarget := false, false
+	for _, component := range flattenedSpec.Components {
+		if component.Container == nil {
+			// Ignore attribute on non-container components as it's not clear what this would mean
+			continue
+		}
+		if component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil) == true {
```
AObuchow, line 133: This is nitpicking, but maybe an error holder should be added here and at [line 134](https://github.com/devfile/devworkspace-operator/pull/844/commits/0295922d3f69555ddb6cb47d48227d71d9d76dd5#diff-4656ec13493defa6167603ffae8b53cddca10b1b1805a29660dadd8c57f2a3f3R134). It'll make the code that calls this function in `merge.go` a bit uglier though. 

Future calls of `component.Attributes.GetBoolean(constants.ContainerContributionAttribute, nil)` in `mergeContainerContributions` wouldn't need this error holder since  `needsContainerContributionMerge` is called first.

AObuchow (approved):
Left some minor comments that aren't really necessary for this to be merged.
Code looks good to me, as do the automated test cases. 

Tested (again) the provided samples on Minikube, and they work as expected. Awesome work @amisevsk :grin:

pkg/library/flatten/merge.go
amisevsk: No, this is just me being a dummy :D 

I'll remove it.

l0rd (approved):
@amisevsk I have tested the PR [updating the devworkspace-demo](https://github.com/l0rd/devworkspace-demo/tree/container-contributions) and it worked perfectly. I have also tried to reproduce the scenario of this [downstream issue](https://issues.redhat.com/browse/CRW-3356) and the CPU limits are summed up correctly. Just one note: I have also tired to use the `spec.contributions` in the [devworkspace-demo](https://github.com/l0rd/devworkspace-demo/tree/container-contributions-v2) but it's not working yet: I think that's supposed to be implemented in a separate PR.

amisevsk:
> I have also tired to use the spec.contributions in the [devworkspace-demo](https://github.com/l0rd/devworkspace-demo/tree/container-contributions-v2) but it's not working yet: I think that's supposed to be implemented in a separate PR.

This is correct. This PR does not update the devfile/api dependency to include containerContributions in the DevWorkspace CR. The dependency bump is part of https://github.com/devfile/devworkspace-operator/pull/868, which is currently on hold to ensure we don't cause issues applying the CRDs in other tools. Once that PR is merged, I'll open another PR to also process containerContributions.

pkg/library/flatten/merge.go
amisevsk: Fixed now, thanks :)

pkg/library/flatten/merge.go
amisevsk: This is a good point -- previously I was just interpreting wrong values as false but it makes sense to highlight the case where the attribute _is_ present but cannot be read correctly, e.g. if the value is `ture` or something.

I'll update the PR with that change.

PR #833, opened 2022-05-05
fix: prevent storage class attribute modification

AObuchow:
### What does this PR do?
Add a webhook to prevent storage class attribute modification for V1alpha2 workspaces.

### What issues does this PR fix or reference?
Fix #825

### Is it tested? How?
- Start up a workspace
- If the devfile has it's storage-class attribute (`controller.devfile.io/storage-type`) set, change it with `kubectl edit` or `oc edit` (eg. from `common` to `ephemeral`)
- If the devfile does not have it's storage class attribute set, set it (eg. to `async`)
- After making the modification to the devfile, you should see that the change was denied, e.g.: `error: devworkspaces.workspace.devfile.io "theia-next" could not be patched: admission webhook "mutate.devworkspace-controller.svc" denied the request: DevWorkspace storage-type attribute cannot be changed once the workspace has been created.`

### Extra notes
Extra code could be added to validate that the storage class attribute given is actually valid. This could be done in a way that uses a generic helper function that could be extended for validating other attributes.

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [ ] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-che-happy-path`: Happy path for verification integration with Che

PR #854, opened 2022-06-02
Allow changing workspace storage type in some cases

amisevsk:
### What does this PR do?
This PR is a follow-up to https://github.com/devfile/devworkspace-operator/pull/833 that allows some storage-type modification after creation. It allows workspace storage type to be changed if:

* The original storage type is ephemeral, or
* If the workspace does not yet have a storage finalizer applied

In these cases, there is no data stored in any PVC, and so the storage class can be safely switched. This change allows

* Creating workspaces a ephemeral, and then adding storage after the fact if it is necessary
* Creating workspaces with `started: false` as a placeholder, and then changing the storage class before the first start.

N/A

### Is it tested? How?
To test:
1. Create an ephemeral workspace -- e.g.
    ```yaml
    kind: DevWorkspace
    metadata:
      name: test-ephemeral
    spec:
      routingClass: basic
      started: true
      template:
        attributes:
        components:
        - name: web-terminal
          container:
            command: ["tail", "-f", "/dev/null"]
            image: quay.io/amisevsk/web-terminal-tooling:dev
            memoryLimit: 512Mi
            mountSources: true
    ```
    and update the storage type to e.g. `common`

2. Create a workspace with `started: false` and update its storage type before starting it the first time:
    ```
    yq '.spec.started=false' samples/theia-next.yaml | oc apply -f -
    ```

In both cases, changes should be permitted. Of course, for regular workspaces, the change should still be blocked.

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -269,7 +269,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// Set finalizer on DevWorkspace if necessary
 	// Note: we need to check the flattened workspace to see if a finalizer is needed, as plugins could require storage
 	if storageProvisioner.NeedsStorage(&workspace.Spec.Template) {
-		coputil.AddFinalizer(clusterWorkspace, storageCleanupFinalizer)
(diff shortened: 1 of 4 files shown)


ibuziuk (approved):

dkwon17:
I see, thank you @amisevsk and @AObuchow :)

PR #858, opened 2022-06-02
Don't count terminating workspaces when checking shared PVC count

### LINKED ISSUE #857: Cleanup job is run even if all workspaces are deleted
### Description
Since https://github.com/devfile/devworkspace-operator/issues/826, the shared PVC should be deleted when no workspaces that use it exist. However, if there are multiple workspaces that use the common PVC deleted at the same time, they will each instead run the cleanup job. 

Once all but one cleanup job completes, DWO deletes the shared PVC as expected and removes the finalizer from the last workspace, even if the cleanup job is not complete.

### How To Reproduce
```bash
for i in {1..5}; do
  yq --arg i "$i" '.metadata.name="theia-next-\($i)"' samples/theia-next.yaml | kubectl apply -f -
  sleep 2s
  kubectl patch dw "theia-next-$i" --type merge -p '{"spec": {"started": false}}'
done
kubectl delete dw --all --wait=false && kubectl get jobs -w
```

### Expected behavior
If all workspaces that depend on common storage are being terminated, we can just delete the PVC and remove their finalizers.

amisevsk:
### What does this PR do?
Skips counting terminating workspaces when checking if we can delete the common PVC. This avoids running the cleanup job for each workspaces when running e.g. `kubectl delete dw --all`.

Closes #857

### Is it tested? How?
Can be tested via the script from the issue:
```bash
for i in {1..5}; do
  yq --arg i "$i" '.metadata.name="theia-next-\($i)"' samples/theia-next.yaml | kubectl apply -f -
  sleep 2s
  kubectl patch dw "theia-next-$i" --type merge -p '{"spec": {"started": false}}'
done
kubectl delete dw --all --wait=false && kubectl get jobs -w
```
no jobs should be created.

### PR Checklist

### DIFF
--- a/pkg/provision/storage/shared.go
+++ b/pkg/provision/storage/shared.go
@@ -238,6 +238,7 @@ func checkForExistingCommonPVC(namespace string, api sync.ClusterAPI) (string, e
 
 // getSharedPVCWorkspaceCount returns the total number of workspaces which are using a shared PVC
 // (i.e the workspaces storage-class attribute is set to "common", "async", or unset which defaults to "common")
+// Note that workspaces that are have been deleted (i.e. have a deletion timestamp) are not counted.
 func getSharedPVCWorkspaceCount(namespace string, api sync.ClusterAPI) (total int, err error) {
 	workspaces := &dw.DevWorkspaceList{}
 	err = api.Client.List(api.Ctx, workspaces, &client.ListOptions{Namespace: namespace})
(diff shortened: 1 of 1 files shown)


ibuziuk (approved):

AObuchow (approved):
Code looks good & tested and verified the patch works as expected :)

dkwon17 (approved):

Issue #852, opened 2022-06-02
Allow to specify arbitrary pods fields


l0rd:
### Description

There are a few examples of workspaces Pods annotations (`openshift.io/scc` or `runtimeClassName`) or Pods spec fields (`runtimeClassName`, `schedulerName`) that users want to be able to specify. 

The approach until now has been to use `spec.template.attributes`. But that requires code/spec changes for every new field/annotation that we want to support. To avoid that, and allow users to set arbitrary Pods fields, we could add a `DevWorspace.spec.pods` field for that:

```diff
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: myspacework
spec:
  started: true
+  pods:
+    metadata:
+      annotations:
+        io.openshift.userns: "true"
+        io.kubernetes.cri-o.userns-mode: "auto:size=65536;map-to-root=true"  # <-- user namespace
+        openshift.io/scc: container-build
+    spec:
+      runtimeClassName: kata
+      schedulerName: stork      
  template:
  (...)
```

### Additional context

- https://github.com/eclipse/che/issues/21420
- https://github.com/eclipse/che/issues/21105
- https://github.com/eclipse/che/issues/20459

amisevsk:
Opened Devfile API issue: https://github.com/devfile/api/issues/860

PR #868, opened 2022-06-13
Add support for overriding any pod spec field in DevWorkspace deployment

amisevsk:
### What does this PR do?
Add support for specifying Pod spec overrides in DevWorkspaces. This PR is open as a draft for review/discussion on approach. 

The first commit in this PR updates the devfile/api dependency to point at my fork, to pull in changes from PR https://github.com/devfile/api/pull/872. Before merging this PR, the devfile/api PR should be merged and the devfile/api dependency should be updated normally.

Related: https://github.com/devfile/api/issues/860
Closes https://github.com/devfile/devworkspace-operator/issues/852

### Is it tested? How?
To test the changes, there's a new sample:
```
oc apply -f samples/pod-spec-overrides.yaml
```
This workspace will likely fail to start as it sets the scheduler to `stork` and the runtimeClassName to `kata`, which are likely not available. To verify things are working as expected, check that fields are overridden in the Deployment for that workspace

### PR Checklist


amisevsk:
I've pushed fixes to take `metadata` into account as well -- it turns out that controller-gen is unable to generate embedded metadata within a CRD so we have to copy the struct ourselves -- see https://github.com/devfile/api/pull/872/commits/dfc7f70c79cdd91adbd6e32dee76d64e182e14da.

I've just tested again and it appears that everything works as intended. A container image from the current changes is built and pushed to `quay.io/amisevsk/devworkspace-controller:pod-spec-overrides`

To test it on a cluster, run the following:
```bash
export NAMESPACE=dw #or your choice
export DWO_IMG=quay.io/amisevsk/devworkspace-controller:pod-spec-overrides
make install
```

amisevsk:
PRs in devfile/api are merged and this PR is ready for review. 

Note one issue that is a side-effect of this PR is that the DevWorkspace and DevWorkspaceTemplate CRDs are now too big to apply using `kubectl apply`. This is because using client-side apply adds the `kubectl.kubernetes.io/last-applied-configuration` containing the full json of the object applied, and the new CRD instances are larger than the max length of an annotation. This fails with message
> The CustomResourceDefinition "devworkspaces.workspace.devfile.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes

To avoid this issue, the `make install` rule now uses server-side apply instead, which does not create the annotation and instead sets `.metadata.managedFields` directly. However, the drawback here is that we also have to use the `--force-conflicts` parameter: since the `inject-cabundle` annotation is present on the CRDs, the `caBundle` field is owned by `Go-http-client`, which injects the bundle. If CRDs are already present, server-side apply would fail with

> Apply failed with 1 conflict: conflict with "Go-http-client" using apiextensions.k8s.io/v1: .spec.conversion.webhook.clientConfig.caBundle

Before merging this PR, it will be necessary to update documentation on applying generated yaml files directly and verify that tools which apply them can accomodate the change.

amisevsk:
To test these changes via OLM, I've build an index image `quay.io/amisevsk/devworkspace-operator-index:pod-spec-overrides` this can be used to create a catalogsource on the cluster and deploy the current changes via OLM.

dkwon17 (approved):
LGTM and works well 👍  (tested with OLM)

dkwon17:
I did notice though, if try to test by running DWO via:
```
export NAMESPACE="devworkspace-controller"
make install
oc patch deployment/devworkspace-controller-manager --patch "{\"spec\":{\"replicas\":0}}" -n $NAMESPACE
make run
```

and then when I run:
```
oc apply -f samples/pod-spec-overrides.yaml
```

the `spec.podSpecOverride` disappears from the DevWorkspace. I don't think this should happen? I didn't experience this when testing with OLM.

I had to add the `spec.podSpecOverride` content manually after I created the DevWorkspace. After doing that, the pod spec overriding was working

amisevsk:
Superceded by https://github.com/devfile/devworkspace-operator/pull/944

PR #889, opened 2022-07-05
Limit objects cached in webhooks manager

### LINKED ISSUE #888: Webhooks server memory usage depends on number of pods in cluster.
### Description
The webhooks server is caching all pods on the cluster in-memory, regardless of whether they are DevWorkspace pods or not. This can result in the webhook server reaching its default memory limit (300Mi) and being killed by the cluster, causing all DWO webhooks will stop working. Further, since it's not possible to filter `pods/exec` requests in webhooks (see https://github.com/kubernetes/kubernetes/issues/91732), all `kubectl exec` commands in the cluster will be blocked.

This issue hasn't been seen until now as webhook server memory usage only becomes a problem when there are a large number (>6000) pods on a cluster, and normally this does not occur due to CPU/memory constraints in the cluster. However, some cluster tasks (e.g. s2i on OpenShift) can result in many completed or errored pods being left on the cluster.

### How To Reproduce
**Note: This is hard to reproduce without also killing a small test cluster**

1. Create test pods on a cluster -- note these pods have ~100KB of annotations (to hopefully use more space in memory per pod) and complete immediately after starting:
    ```bash
    for i in {0001..0500}; do
      curl https://gist.githubusercontent.com/amisevsk/2ca0a75f2bfcf785e597df37d8a22221/raw/007c6b8ff7780ec3aacfd7881f4ed9e22f809470/big-pod.yaml \
        | yq -y --arg name "test-pod-$i" '.metadata.name = $name' \
        | oc apply -f -
      sleep 1s
    done
    ```
2. (In another terminal) try to `kubectl exec` into a pod to trigger the webhook server caching pods internally
3. Observe webhooks server memory usage (e.g. on OpenShift: `oc adm top pod <webhooks-server-pod-name>`)

In my testing, once I get to around 500 pods, the webhooks server requires ~230MiB of memory (up from ~30MiB at idle)

### Expected behavior
DWO webhooks server memory usage should not depend on the number of non-DWO-related objects on the cluster.

### Additional context

The webhooks server needs to read pods from the cluster in order to validate `pods/exec` requests for restricted-access workspaces. Since this is a read-only operation (we just need to check pod metadata), this is done via the controller-runtime manager, which implements efficient read operations by asynchronously watching all objects of interest in the cluster.

The controller itself has had a similar problem in the past. See: https://github.com/devfile/devworkspace-operator/pull/652

amisevsk:
### What does this PR do?
Similarly to #652, limit pods cached by the webhook server to be only those with the `controller.devfile.io/devworkspace_id` label. This prevents the webhook server from mirroring _all_ pods on the cluster to memory, causing memory usage to increase when there is a large number of pods in the cluster.

Closes #888

### Is it tested? How?
1. Verify that memory usage does not increase when creating non-DevWorkspace pods in the cluster
2. Verify that webhooks continue to work as expected:
    * pods/exec is permitted except for workspaces that have the restricted-access annotation, in which case only the creator can access them
    * mutating/validating webhooks continue to work as before

### PR Checklist

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -99,3 +99,25 @@ func GetCacheFunc() (cache.NewCacheFunc, error) {
 		SelectorsByObject: selectors,
 	}), nil
 }
+
+// GetWebhooksCacheFunc returns a new cache function that restricts the cluster items we store in the webhook
+// server's internal cache. This avoids issues where the webhook server's memory usage scales with the number
+// of objects on the cluster, potentially causing out of memory errors in large clusters.
+func GetWebhooksCacheFunc() (cache.NewCacheFunc, error) {
+	// The webhooks server needs to read pods to validate pods/exec requests. These pods must have the DevWorkspace ID and restricted
+	// access labels (other pods are automatically approved)
+	devworkspaceObjectSelector, err := labels.Parse(constants.DevWorkspaceIDLabel)
+	if err != nil {
+		return nil, err
+	}
+
+	selectors := cache.SelectorsByObject{
+		&corev1.Pod{}: {
+			Label: devworkspaceObjectSelector,
+		},
+	}
+
+	return cache.BuilderWithOptions(cache.Options{
+		SelectorsByObject: selectors,
+	}), nil
+}
--- a/webhook/main.go
+++ b/webhook/main.go
@@ -26,6 +26,7 @@ import (
 
 	dwv1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha1"
 	dwv2 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/pkg/cache"
 	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
 	"github.com/devfile/devworkspace-operator/version"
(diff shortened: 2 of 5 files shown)


ibuziuk (approved):

amisevsk:
CI actually caught an error for once -- earlier this PR was filtering to pods that had the `controller.devfile.io/restricted-access` label, which is incorrect because it's an annotation (i.e. no pods have it on the cluster). I've force-pushed a change to only filter for pods with the `devworkspace_id` label.

I've verified this PR in the following way:
1. Start a small cluster
2. Create 100 completed pods in 10 namespaces using the pod template in #888 
3. Check webhooks server memory usage for both the current `next` image and an image built from this PR

In the `next` image, I see the webhook server pod consuming ~100Mi, which isn't an OOM situation yet could easily get there (extrapolating 100 pods -> 70Mi memory usage increase, we'd expect to hit OOM at around 380-400 completed pods).

Testing with image `quay.io/amisevsk/devworkspace-controller:webhooks-filtering` (built from this PR), memory usage in the same situation is around 30Mi. 

N.B. if you are testing this, it's necessary to run `oc exec` into a pod at least once after the webhook server restarts. The watches used internally in controller-runtime are initialized on-demand, so pods aren't mirrored to the internal cache until they are requested by the webhook server. This doesn't occur until the webhook server is asked to verify a `pods/exec` request.

PR #893, opened 2022-07-15
Add per-user storage-class option

AObuchow:
The 'per-user' storage type is an alternate storage-type name for the 'common' storage class.
In other words, it is an alias for the 'common' storage-type and behaves in the same manner.

Note that in the PR's current state, the 'per-user' storage-type replaces the 'common' storage-type, and the 'common' storage-type is treated as an alias of the 'per-user' storage-type. It may be better to reverse things, i.e. document that the 'per-user' storage-type is an alias of the 'common' storage-type 

Part of eclipse/che#21405

### Is it tested? How?
There are two parts of this PR to test.
**Ensuring that the 'per-user' storage type works the same as the 'common' storage type:**
1. Start up DWO
2. Create and apply a devfile that uses the 'per-user' storage type, e.g.
```yaml
kind: DevWorkspace
metadata:
  name: theia-next
spec:
  started: true
  template:
    attributes:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```
3. Ensure that the workspace starts up and that the `claim-devworkspace` PVC is created.
4. Delete the workspace and ensure that the `claim-devworkspace` PVC is deleted.

**Ensuring that you can switch between the 'per-user' and 'common' storage-type:**
1. Start up DWO
2. Create and apply 2 devfiles, one with the 'per-user' storage-type and the other with the 'common' storage-type
3. After the workspaces are started up, do a `kubectl edit` or `oc edit` and change their storage-type from `per-user` -> `common` and `common` -> `per-user` respectively
4. Ensure the edit is allowed by the webhook

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -685,7 +685,7 @@ func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Req
 	var reconciles []reconcile.Request
 	for _, workspace := range dwList.Items {
 		storageType := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAttribute, nil)
-		if storageType == constants.CommonStorageClassType || storageType == "" {
+		if storageType == constants.CommonStorageClassType || storageType == constants.PerUserStorageClassType || storageType == "" {
 			reconciles = append(reconciles, reconcile.Request{
 				NamespacedName: types.NamespacedName{
 					Name:      workspace.GetName(),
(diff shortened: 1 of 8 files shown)


amisevsk (approved):

AObuchow:
> LGTM, but could you please add some tests

I added a test to ensure that the `CommonStorageProvisioner` is used given a devfile that uses the `per-user` storage class. I haven't been able to think of a better way to test this feature so far.

PR #905, opened 2022-08-15
WIP: feat: support PVC expansion based on DWOC

### LINKED ISSUE #875: Support automatically resizing workspace PVCs where supported.
### Description
Since PR https://github.com/devfile/devworkspace-operator/pull/827, it's possible to configure the PVC size used for both per-workspace and common PVCs. However, after a PVC is created, the storage size is not updated, and so the only way to reflect changes to the configuration is to first delete all DevWorkspaces in the namespace that use that PVC, allow the PVC to be cleaned up, and then recreate it by creating a new DevWorkspace (see https://github.com/devfile/devworkspace-operator/pull/834).

Some persistent volume storage-classes support resizing PVCs. When this is possible, DWO should attempt to automatically increase PVC sizes if it detects that an existing PVC is smaller than the configured default.

This resize could potentially be performed on-demand, e.g. the next time a workspace is started. We do have to be a little careful as resizing potentially hundreds of PVCs could impact cluster stability/performance.

### Additional context
Further notes:
* Since we currently don't update PVCs once they are created, there is a chance that the PVC has already been configured by users. Likely the implementation should only increase PVC size (if the default is less than the existing PVC's size, we do nothing).
* This feature has a potential performance impact, and it would be necessary to verify that updating the default storage size will not adversely impact performance/stability in large clusters (with potentially hundreds of affected PVCs)

AObuchow:
### What does this PR do?
Allows PVC's provisioned using the `common` (or `per-user`) or `per-workspace` storage-type to be expanded to the size configured in the `workspace.DefaultStorageSize` field of the DWOC. This occurs during reconciliation, and requires `enableExperimentalFeatures` to be set to `true` in the DWOC.

The [Kubernetes storage-class](https://kubernetes.io/docs/concepts/storage/storage-classes/) must have it's `allowVolumeExpansion` field set to `true` for resizing to work. Furthermore, only certain volumes support expansion - according to the [Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims) these are:
```
- azureDisk
- azureFile
- awsElasticBlockStore
- cinder (deprecated)
- [csi](https://kubernetes.io/docs/concepts/storage/volumes/#csi)
- flexVolume (deprecated)
- gcePersistentDisk
- glusterfs
- rbd
- portworxVolume
```

Minikube does **not** support PVC expansion by default, even if its storage-class has the `allowVolumeExpansion` field set to `true`. 

This is an 'experimental' feature primarily for the following 2 reasons:
1. If the volume does not support expansion, then a warning stating that the PVC could not be expanded will be logged. In my testing, once the volume expansion request has been submitted, this warning will no longer be logged in the future for a given PVC, assuming `workspace.DefaultStorageSize` is not modified again (as the PVC spec already contains the requested size, causing `PVCExpansionRequired()` to return false).
2. If a `ResourceQuota` or `LimitRange` prevents the PVC expansion request, then an error will be returned and continuously logged by DWO. For example:
```json
{
   "level":"error",
   "ts":1660253077.8029618,
   "logger":"controller-runtime.manager.controller.devworkspace",
   "msg":"Reconciler error",
   "reconciler group":"workspace.devfile.io",
   "reconciler kind":"DevWorkspace",
   "name":"theia-next-2",
   "namespace":"devworkspace-controller",
   "error":"persistentvolumeclaims \"claim-devworkspace\" is forbidden: maximum storage usage per PersistentVolumeClaim is 17Gi, but request is 20Gi",
   "stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/aobuchow/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.5/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/aobuchow/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.5/pkg/internal/controller/controller.go:214"
}
```

Fix #875

### Is it tested? How?
Some automated tests were added to test this feature.

There are two manual-testing scenario's: on Minikube and on OpenShift.
For both scenarios, the steps to follow are:
1. Deploy DWO onto the cluster. E.g for OpenShift `make docker install` (make sure `$DWO_IMG` is set to your quay repository)
2. **For Minikube only:** Edit the standard storage-class used by minikube to set the `allowVolumeExpansion` field to `true`.  If this is not set, then DWO will silently not attempt to resize the PVC. To do so: `kubectl edit storageClass standard`
```diff
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
 annotations:
   kubectl.kubernetes.io/last-applied-configuration: |
     {"apiVersion":"storage.k8s.io/v1","kind":"StorageClass","metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"true"},"labels":{"addonmanager.kub\
ernetes.io/mode":"EnsureExists"},"name":"standard"},"provisioner":"k8s.io/minikube-hostpath"}
   storageclass.kubernetes.io/is-default-class: "true"
 creationTimestamp: "2022-08-15T14:46:41Z"
 labels:
   addonmanager.kubernetes.io/mode: EnsureExists
 name: standard
 resourceVersion: "1507"
 uid: 7585ccfd-04fd-49f9-89e3-c551fb2e62ed
provisioner: k8s.io/minikube-hostpath
reclaimPolicy: Delete
volumeBindingMode: Immediate
+ allowVolumeExpansion: true
```

3. Apply a devfile that uses the common PVC strategy, as well as a devfile that uses the per-workspace PVC strategy, eg. `kubectl apply -f ./samples/theia-next.yaml -n $NAMESPACE && yq -i -y '.metadata.name |= "theia-next-per-workspace"' ./samples/per-workspace-storage.yaml && kubectl apply -f ./samples/per-workspace-storage.yaml -n $NAMESPACE`
4. Edit the DWOC to set `enableExperimentalFeatures` to `true`, as well as set the default storage sizes for per-workspace and common to something bigger than their defaults (which are 5Gi and 10Gi respectively): `kubectl edit dwoc -n $NAMESPACE`

```diff
apiVersion: controller.devfile.io/v1alpha1
config:
+  enableExperimentalFeatures: true
  routing:
    defaultRoutingClass: basic
  workspace:
+    defaultStorageSize:
+      common: 15Gi
+      perWorkspace: 15Gi
    imagePullPolicy: Always
kind: DevWorkspaceOperatorConfig
...
```
5. Trigger a reconcile in both workspaces to allow expansion to occur, eg. `edit dw theia-next-per-workspace -n $NAMESPACE` and toggle `spec.started` from false to true

In the **OpenShift scenario**, both the common PVC and per-workspace PVC will have their capacity expanded to the amount set in the DWOC (15Gi in the example given). This can be viewed on the Administrator view of the OpenShift console -> Storage -> Persistent Volume Claims.

In the **Minikube scenario**, both the common and per-workspace PVC will fail to have their capacity expanded, and a message similar to the following will be logged:
```json
{
   "level":"info",
   "ts":1660582376.9331381,
	"msg":"PVC expansion error: Unable to expand PVC. More information: Ignoring the PVC: didn't find a plugin capable of expanding the volume; waiting for an external controller to process this PVC."
}
```

### PR Checklist

### DIFF
--- a/deploy/templates/components/rbac/role.yaml
+++ b/deploy/templates/components/rbac/role.yaml
@@ -231,6 +231,14 @@ rules:
   verbs:
   - create
 - apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - workspace.devfile.io
   resources:
   - '*'
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -27,6 +27,7 @@ import (
 
 // defaultConfig represents the default configuration for the DevWorkspace Operator.
 var defaultConfig = &v1alpha1.OperatorConfiguration{
+	EnableExperimentalFeatures: pointer.BoolPtr(false),
 	Routing: &v1alpha1.RoutingConfig{
 		DefaultRoutingClass: "basic",
 		ClusterHostSuffix:   "", // is auto discovered when running on OpenShift. Must be defined by CR on Kubernetes.
--- a/pkg/provision/storage/perWorkspaceStorage.go
+++ b/pkg/provision/storage/perWorkspaceStorage.go
@@ -173,8 +173,7 @@ func syncPerWorkspacePVC(workspace *common.DevWorkspaceWithConfig, clusterAPI sy
 		}
 	}
 
-	storageClass := workspace.Config.Workspace.StorageClassName
-	pvc := getPVCSpec(common.PerWorkspacePVCName(workspace.Status.DevWorkspaceId), workspace.Namespace, storageClass, pvcSize)
+	pvc := getPVCSpec(common.PerWorkspacePVCName(workspace.Status.DevWorkspaceId), workspace.Namespace, workspace.Config.Workspace.StorageClassName, pvcSize)
 
 	namespaceName := types.NamespacedName{
 		Name:      common.PerWorkspacePVCName(workspace.Status.DevWorkspaceId),
--- a/pkg/provision/storage/perWorkspaceStorage.go
+++ b/pkg/provision/storage/perWorkspaceStorage.go
@@ -27,8 +27,11 @@ import (
 	nsconfig "github.com/devfile/devworkspace-operator/pkg/provision/config"
 	"github.com/devfile/devworkspace-operator/pkg/provision/sync"
 	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // The PerWorkspaceStorageProvisioner provisions one PVC per workspace and configures all volumes in the workspace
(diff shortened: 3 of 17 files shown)


deploy/templates/components/rbac/role.yaml
```diff
@@ -27,6 +27,14 @@ rules:
   - serviceaccounts
   verbs:
   - '*'
+- apiGroups:
```
AObuchow, line 30: This cluster role modification was required to prevent the following error:
`E0811 17:20:17.422655 1392784 reflector.go:138] pkg/mod/k8s.io/client-go@v0.21.3/tools/cache/reflector.go:167: Failed to watch *v1.StorageClass: unknown (get storageclasses.storage.k8s.io)`

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -170,10 +173,49 @@ func syncPerWorkspacePVC(workspace *common.DevWorkspaceWithConfig, clusterAPI sy
...
+			msg, err := CheckPVCEvents(*pvc, clusterAPI)
+
+			if err != nil {
+				return nil, err
+			}
+
+			if msg != "" {
+				// TODO: Not sure if this is the best way to log in this case
+				// TODO: The error message could be better? Currently, it'll be something like: "PVC expansion error: Unable to expand PVC. More information: ..."
+				log.Log.Info(fmt.Sprintf("PVC expansion error: %s", msg))
```
amisevsk, line 216:
```suggestion
				clusterAPI.Logger.Info(fmt.Sprintf("PVC expansion error: %s", msg))
```
Though this may result in logging repeatedly (on every reconcile), which would fill up operator logs needlessly. Do we expect transient errors like this? Could this be better suited to a workspace warning condition?

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -170,10 +173,49 @@ func syncPerWorkspacePVC(workspace *common.DevWorkspaceWithConfig, clusterAPI sy
...
+
+	existingPVC := &corev1.PersistentVolumeClaim{}
+	err = clusterAPI.Client.Get(clusterAPI.Ctx, namespaceName, existingPVC)
+
 	if err != nil {
-		return nil, err
+		if !k8sErrors.IsNotFound(err) {
+			return nil, err
+		}
+	} else if workspace.Config.Workspace.DefaultStorageSize.PerWorkspace != nil && PVCExpansionRequired(*existingPVC, workspace.Config.Workspace.DefaultStorageSize.PerWorkspace) && *workspace.Config.EnableExperimentalFeatures {
```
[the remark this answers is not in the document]
AObuchow: +1 this would make things much cleaner

pkg/provision/storage/perWorkspaceStorage.go
AObuchow: IIRC this repeated logging is part of the reason why I've been skeptical of this PR/feature (in its current state). I believe repeated logging here would happen in the following case:

> If a ResourceQuota or LimitRange prevents the PVC expansion request, then an error will be returned and continuously logged by DWO.

A workspace warning condition would be much better, though I have to figure out the "warning return flow" for this case.

pkg/config/defaults.go
```diff
@@ -27,6 +27,7 @@ import (
 
 // defaultConfig represents the default configuration for the DevWorkspace Operator.
 var defaultConfig = &v1alpha1.OperatorConfiguration{
+	EnableExperimentalFeatures: pointer.BoolPtr(false),
```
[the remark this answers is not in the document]
AObuchow: > This should not be necessary (nil is to be interpreted as false).

Sounds good, will make the appropriate changes.

> Do we want to be able to enable experimental features in a DevWorkspaceOperatorConfig that isn't the main one?

IMO, no, we do not. In the current state of things (where Che owns its own DWOC) I don't think it makes sense to be able to toggle experimental features from Che, since usage of experimental features shouldn't really be encouraged. Enabling these features should be more of a "deeper" configuration option.

However, I think I understand the other point you're making:

> I think it's not a great idea, since the same option gates internal operator functionality, e.g. logging

I dislike the fact that enabling deeper logging (through `EnableExperimentalFeatures`) would also enable experimental features (e.g. PVC expansion). What if you just want more in-depth logging, without enabling experimental features?

This adds to the complexity of this change: do we update the DWOC API to have a new `verboseLoggingEnabled` option, to separate extra logging from `EnableExperimentalFeatures`? This would be a breaking API change.

AObuchow:
I'm starting to think that this PR should either be abandoned due to the complexity it introduces, while not providing much benefit (you can just manually change the PVC size) OR more work needs to be done to get things _right_: create a ClientSet (which should be doable when running the controller locally and within a pod), pass it down to the PVC provisioning code and use it to check the ResourceQuota and LimitRanges.

PR #909, opened 2022-08-19
feat: Allow external DWOC to be merged with internal DWOC on a per-workspace basis

AObuchow:
### What does this PR do?
This PR aims to make 2 changes to DWO:
1. Refactor (almost) all uses of the DWOC, so that, internally, each workspace instance has its own configuration. Currently, this per-workspace devworkspace-operator configuration is not visible to users, and was added only to support changed n.2
2. Allow for specifying an external DWOC (that exists on the cluster), which gets merged with the workspace's internal DWOC

The external DWOC's name and namespace are specified with the `controller.devfile.io/devworkspace-config` DevWorkspace attribute, which has the following structure:
    
```YAML
attributes:
    name: <string>
    namespace: <string>
```

Part of eclipse/che#21405

### Is it tested? How?
I added 2 automated tests.

For manual testing, perform the following steps:
1. Deploy DWO
2. Create a devfile with the `controller.devfile.io/devworkspace-config` attribute set, e.g.

```YAML
kind: DevWorkspace
metadata:
  name: theia-next-external-dwoc
spec:
  started: true
  template:
    attributes:
          name: external-dwoc-test
          namespace: default
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```
3. Create a DWOC file, for reference, we'll call it `test-external-dwoc.yaml`. Also, ensure its name field is set to the same value as the `controller.devfile.io/devworkspace-config-name` attribute in the devfile from step 2:
```YAML
config:
  routing:
    defaultRoutingClass: basic
  workspace:
    defaultStorageSize:
      common: 20Gi
      perWorkspace: 29Gi
    imagePullPolicy: Always
metadata:
  name: external-dwoc-test
  namespace: default
```
4. Apply the (external) DWOC file you just created, ensuring it's in the namespace you specified in the `controller.devfile.io/devworkspace-config-namespace` (in my example, the default namespace is used): `kubectl apply -f test-external-dwoc.yaml`
5. Apply the devfile created in step 2. 
6. Verify that the internal DWOC has been modified. I recommend checking the size of the per-workspace PVC that is created and ensuring it is `29Gi` instead of the default `10Gi`
7. Afterwards, you can ensure that workspace's that do not have the `controller.devfile.io/devworkspace-config` attribute set are not affected by the external DWOC, and instead, they use the global/internal DWOC. For instance, start up the per-workspace example and ensure the PVC created for it is `10Gi` big: `kubectl apply -f per-workspace-storage.yaml -n $NAMESPACE`.

### Some notes and limitations:
1. The HTTP client / Proxy settings cannot be affected by the externally applied DWOC, as [the HTTP client is created and configured on DWO startup rather than workspace startup](https://github.com/devfile/devworkspace-operator/pull/711#issuecomment-995360107). 
2. There is [a handler](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L680) that checks if the common PVC was deleted. Currently, it filters for PVC's related to DWO by checking the PVC name against the global configs PVC name. In the future, this limitation can be worked around by applying a DWO-specific annotation to all PVC's created by DWO, and filtering by that annotation. 
3. This change could have weird implications for anything configurable that is shared across workspaces. Most notably, the common PVC. For example, if the common PVC does not yet exist on the cluster, and an external DWOC that has the common PVC size set to `20Gi` is applied for a specific workspace, then other (future) workspaces that use the common storage strategy will also be using a PVC of `20Gi`, even if they aren't using an external DWOC. 

### PR Checklist

### DIFF
--- a/controllers/controller/devworkspacerouting/solvers/basic_solver.go
+++ b/controllers/controller/devworkspacerouting/solvers/basic_solver.go
@@ -56,12 +56,7 @@ func (s *BasicSolver) Finalize(*controllerv1alpha1.DevWorkspaceRouting) error {
 func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.DevWorkspaceRouting, workspaceMeta DevWorkspaceMetadata) (RoutingObjects, error) {
 	routingObjects := RoutingObjects{}
 
-	routingAnnotationPrefix := string(routing.Spec.RoutingClass) + constants.RoutingAnnotationInfix
-	if _, set := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]; !set {
-		return routingObjects, &RoutingInvalid{"basic routing requires .config.routing.clusterHostSuffix to be set in operator config"}
-	}
-
-	routingSuffix := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]
+	routingSuffix := routing.Annotations[constants.ClusterHostSuffixAnnotation]
 	if routingSuffix == "" {
 		return routingObjects, &RoutingInvalid{"basic routing requires .config.routing.clusterHostSuffix to be set in operator config"}
 	}
--- a/controllers/controller/devworkspacerouting/solvers/basic_solver.go
+++ b/controllers/controller/devworkspacerouting/solvers/basic_solver.go
@@ -17,10 +17,8 @@ package solvers
 
 import (
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
-	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
 	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 var routeAnnotations = func(endpointName string) map[string]string {
@@ -58,7 +56,12 @@ func (s *BasicSolver) Finalize(*controllerv1alpha1.DevWorkspaceRouting) error {
 func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.DevWorkspaceRouting, workspaceMeta DevWorkspaceMetadata) (RoutingObjects, error) {
 	routingObjects := RoutingObjects{}
 
-	routingSuffix := config.Routing.ClusterHostSuffix
+	routingAnnotationPrefix := string(routing.Spec.RoutingClass) + constants.RoutingAnnotationInfix
+	if _, set := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]; !set {
+		return routingObjects, &RoutingInvalid{"basic routing requires .config.routing.clusterHostSuffix to be set in operator config"}
+	}
+
+	routingSuffix := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]
 	if routingSuffix == "" {
 		return routingObjects, &RoutingInvalid{"basic routing requires .config.routing.clusterHostSuffix to be set in operator config"}
 	}
--- a/controllers/controller/devworkspacerouting/solvers/basic_solver.go
+++ b/controllers/controller/devworkspacerouting/solvers/basic_solver.go
@@ -20,6 +20,7 @@ import (
 	"github.com/devfile/devworkspace-operator/pkg/config"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
 	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 var routeAnnotations = func(endpointName string) map[string]string {
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -222,7 +222,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// encountered in main reconcile loop.
 		if err == nil {
 			if timeoutErr := checkForStartTimeout(&clusterWorkspaceWithConfig.DevWorkspace, workspaceWithConfig.Config); timeoutErr != nil {
-				reconcileResult, err = r.failWorkspace(&workspaceWithConfig.DevWorkspace, timeoutErr.Error(), metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus)
+				reconcileResult, err = r.failWorkspace(workspaceWithConfig, timeoutErr.Error(), metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus)
 			}
 		}
 		if reconcileStatus.phase == dw.DevWorkspaceStatusRunning {
@@ -236,7 +236,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if workspaceWithConfig.Annotations[constants.DevWorkspaceRestrictedAccessAnnotation] == "true" {
 		msg, err := r.validateCreatorLabel(clusterWorkspace)
 		if err != nil {
-			return r.failWorkspace(&workspaceWithConfig.DevWorkspace, msg, metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus)
+			return r.failWorkspace(workspaceWithConfig, msg, metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus)
 		}
 	}
 
@@ -263,7 +263,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	flattenedWorkspace, warnings, err := flatten.ResolveDevWorkspace(&workspaceWithConfig.Spec.Template, flattenHelpers)
 	if err != nil {
-		return r.failWorkspace(&workspaceWithConfig.DevWorkspace, fmt.Sprintf("Error processing devfile: %s", err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
+		return r.failWorkspace(workspaceWithConfig, fmt.Sprintf("Error processing devfile: %s", err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
 	}
 	if warnings != nil {
 		reconcileStatus.setConditionTrue(conditions.DevWorkspaceWarning, flatten.FormatVariablesWarning(warnings))
@@ -278,13 +278,13 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if components != nil {
 		eventErrors := devfilevalidation.ValidateComponents(components)
 		if eventErrors != nil {
-			return r.failWorkspace(&workspaceWithConfig.DevWorkspace, eventErrors.Error(), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
+			return r.failWorkspace(workspaceWithConfig, eventErrors.Error(), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
 		}
 	}
 
 	storageProvisioner, err := storage.GetProvisioner(&workspaceWithConfig.DevWorkspace)
 	if err != nil {
-		return r.failWorkspace(&workspaceWithConfig.DevWorkspace, fmt.Sprintf("Error provisioning storage: %s", err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
+		return r.failWorkspace(workspaceWithConfig, fmt.Sprintf("Error provisioning storage: %s", err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
 	}
 
 	// Set finalizer on DevWorkspace if necessary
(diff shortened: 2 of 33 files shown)


AObuchow:
This PR is based off of [option 2](https://gist.github.com/amisevsk/2e55f864e5a794a0653b3c450ba932dc#option-2-add-annotationattribute-to-devworkspaces-to-allow-specifying-an-alternate-configuration-per-devworkspace) of @amisevsk's configuring-DWO-from-che proposal

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -79,6 +79,16 @@ type StorageSizes struct {
...
+//TODO: Should this be here or in another file?
+// Allows specifying an external DevWorkspace-Operator configuration
+// which will merged with the workspace's DevWorkspace-Operator configuration.
+type ExternalConfig struct {
+	// The metadata.name of the external DevWorkspace-Operator configuration.
+	Name string `json:"name,omitempty"`
+	// The metadata.namespace of the external DevWorkspace-Operator configuration.
+	Namespace string `json:"namespace,omitempty"`
+}
+
```
amisevsk, line 91: This could just be unmarshalled to a `types.NamespacedName` I believe, removing the need for a specific struct (haven't tested, though).

If it is needed after all, this struct should be defined elsewhere (perhaps in the constants/annotations package?)

pkg/config/sync.go
```diff
@@ -55,6 +61,53 @@ func SetConfigForTesting(config *controller.OperatorConfiguration) {
...
+
+	return nil
+}
+
+func mergeInteralConfigWithExternal(externalConfig *controller.OperatorConfiguration) {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+	mergeConfig(externalConfig, internalConfig)
+	updatePublicConfig()
+}
```
amisevsk, line 109: I think this needs more planning -- this will update the global config for all workspaces based on the last reconciled workspace, which is not what we intend. Instead, we need some way to read the config with a context, which is significantly more complicated and may require refactoring how config is used in general :confused:

pkg/config/sync.go
```diff
@@ -55,6 +61,53 @@ func SetConfigForTesting(config *controller.OperatorConfiguration) {
...
+		Namespace: ExternalDWOCMeta.Namespace,
+	}
+
+	err = client.Get(context.TODO(), namespacedName, externalDWOC)
+	if err != nil {
+		return fmt.Errorf("could not fetch external DWOC with name '%s' in namespace '%s': %w", ExternalDWOCMeta.Name, ExternalDWOCMeta.Namespace, err)
+	}
+
+	// TODO: It might be better to just always merge rather than suffering performance hit in configsAlreadyMerged() (or maintaing configsAlreadyMerged if we don't use reflect.DeepEqual)
+	if !configsAlreadyMerged(externalDWOC.Config, internalConfig) {
```
amisevsk, line 97: I would guess that given the current design, `configsAlreadyMerged` will almost always return false if there are multiple workspaces.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
AObuchow: `types.NamespacedName` seems like it'd work perfectly :) I'll give that a try

pkg/config/sync.go
AObuchow: Good catch - I overlooked the fact that the internal config is shared between all workspaces. It seems like a prerequisite for the current PR is to allow each workspace to have its own configuration.

AObuchow:
@amisevsk Though you are free to take an early look (e.g. checking out the TODO comments), this PR is unfortunately still in progress.

I still need to add some tests, document some stuff and try and make the commit history a bit cleaner.

pkg/config/sync.go
```diff
@@ -130,6 +164,14 @@ func syncConfigFrom(newConfig *controller.DevWorkspaceOperatorConfig) {
 	updatePublicConfig()
 }
 
+func getMergedConfig(from, to *controller.OperatorConfiguration) *controller.OperatorConfiguration {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+	mergedConfig := to.DeepCopy()
+	mergeConfig(from, to)
```
AObuchow, line 171: should be `mergeConfig(from, mergedConfig)`

AObuchow:
The git history is unfortunately still _very_ messy, so it might be better to review this PR as a whole (as difficult as that may be, due to the number of files modified in this PR :/ )

AObuchow:
@amisevsk Should I make note of the new attribute in the a[dditional configuration documentation](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#additional-configuration-for-devworkspaces)? Or do we want to keep this as an internal/undocumented feature for the time being, seeing as it was intended for a specific use case (che)?

controllers/controller/devworkspacerouting/solvers/basic_solver.go
```diff
@@ -57,7 +56,12 @@ func (s *BasicSolver) Finalize(*controllerv1alpha1.DevWorkspaceRouting) error {
 func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.DevWorkspaceRouting, workspaceMeta DevWorkspaceMetadata) (RoutingObjects, error) {
 	routingObjects := RoutingObjects{}
 
-	routingSuffix := config.Routing.ClusterHostSuffix
+	routingAnnotationPrefix := string(routing.Spec.RoutingClass) + constants.RoutingAnnotationInfix
+	if _, set := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]; !set {
+		return routingObjects, &RoutingInvalid{"basic routing requires .config.routing.clusterHostSuffix to be set in operator config"}
+	}
+
+	routingSuffix := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]
```
amisevsk, line 64: Might be cleaner to do
```go
	routingSuffix := routing.Annotations[routingAnnotationPrefix+constants.ClusterHostSuffixAnnotationSuffix]
  if routingSuffix == "" {
		return routingObjects, &RoutingInvalid{"basic routing requires .config.routing.clusterHostSuffix to be set in operator config"}
  }
```

amisevsk (commented):
First pass at reviewing. Looking okay so far but I'm still not 100% on the approach we're taking -- there's some mess and inconsistency on whether we pass the extended struct (WorkspaceWithConfig) or whether we pass the config separately (e.g. in library packages) and I'm not sure which approach is better.

pkg/config/sync.go
```diff
@@ -130,6 +167,14 @@ func syncConfigFrom(newConfig *controller.DevWorkspaceOperatorConfig) {
 	updatePublicConfig()
 }
 
+func getMergedConfig(from, to *controller.OperatorConfiguration) *controller.OperatorConfiguration {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+	mergedConfig := to.DeepCopy()
+	mergeConfig(from, mergedConfig)
+	return mergedConfig
```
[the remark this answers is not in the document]
AObuchow: Done in [`8bcd04e` (#909)](https://github.com/devfile/devworkspace-operator/pull/909/commits/8bcd04ee4af917962087c3fb9a3fdfdb0c92cf39)

controllers/workspace/devworkspace_controller.go
```diff
@@ -235,74 +248,77 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	reconcileStatus.setConditionTrue(conditions.DevWorkspaceResolved, "Resolved plugins and parents from DevWorkspace")
 
 	// Verify that the devworkspace components are valid after flattening
-	components := workspace.Spec.Template.Components
+	components := workspaceWithConfig.Spec.Template.Components
 	if components != nil {
 		eventErrors := devfilevalidation.ValidateComponents(components)
 		if eventErrors != nil {
-			return r.failWorkspace(workspace, eventErrors.Error(), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
+			return r.failWorkspace(&workspaceWithConfig.DevWorkspace, eventErrors.Error(), metrics.ReasonBadRequest, reqLogger, &reconcileStatus)
```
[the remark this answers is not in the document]
AObuchow: Done in [`c1c254d` (#909)](https://github.com/devfile/devworkspace-operator/pull/909/commits/c1c254dd5906180b58cbb9cbbb08e622610f235a)

pkg/config/sync.go
```diff
@@ -47,12 +50,46 @@ var (
...
 	internalConfig = defaultConfig.DeepCopy()
 	mergeConfig(config, internalConfig)
 	updatePublicConfig()
+	return internalConfig
+}
+
+// Checks for a external DevWorkspace-Operator configuration specified by the optional workspace attribute "controller.devfile.io/devworkspace-config"
+// If the external DWOC is found, it is merged with the internal/global DWOC and returned.
+// If the exteranl DWOC is not found, or the "controller.devfile.io/devworkspace-config" attribute is not set, then the internal/global DWOC is returned.
+func ResolveConfigForWorkspace(workspace *dw.DevWorkspace, client crclient.Client) (*controller.OperatorConfiguration, error) {
```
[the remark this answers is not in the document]
AObuchow: Done in [`bfbd54e` (#909)](https://github.com/devfile/devworkspace-operator/pull/909/commits/bfbd54e0a27e658ca7bb1443b4af9a705ea132b8)

controllers/workspace/devworkspace_controller.go
```diff
@@ -235,74 +248,77 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	if wsDefaults.NeedsDefaultTemplate(workspace) {
-		wsDefaults.ApplyDefaultTemplate(workspace)
+	if wsDefaults.NeedsDefaultTemplate(workspaceWithConfig) {
+		wsDefaults.ApplyDefaultTemplate(workspaceWithConfig)
 	}
 
-	flattenedWorkspace, warnings, err := flatten.ResolveDevWorkspace(&workspace.Spec.Template, flattenHelpers)
+	// Apply devworkspace routing annotation(s) to pass DWOC routing settings to DWR
+	solvers.AddDWOCRoutingAnnotations(*workspaceWithConfig)
```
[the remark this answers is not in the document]
AObuchow: Fixed this in [`62b3424` (#909)](https://github.com/devfile/devworkspace-operator/pull/909/commits/62b3424566ac8781d556fea0aff22955e27ec884)

controllers/workspace/devworkspace_controller.go
```diff
@@ -109,8 +110,8 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// Fetch the Workspace instance
-	workspace := &dw.DevWorkspace{}
-	err = r.Get(ctx, req.NamespacedName, workspace)
+	workspaceWithConfig := &common.DevWorkspaceWithConfig{}
```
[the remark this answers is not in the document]
AObuchow: Fixed this in [`24a9138` (#909)](https://github.com/devfile/devworkspace-operator/pull/909/commits/24a9138274b9f93ee103be38a71d57a31e25a0d3) and [`cd8ec7a` (#909)](https://github.com/devfile/devworkspace-operator/pull/909/commits/cd8ec7a7c070b3fce90886969c37fb2c5309526f)

controllers/workspace/devworkspace_controller.go
```diff
@@ -175,146 +175,145 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	defer func() (reconcile.Result, error) {
-		r.syncTimingToCluster(ctx, &clusterWorkspaceWithConfig.DevWorkspace, timingInfo, reqLogger)
+		r.syncTimingToCluster(ctx, &clusterWorkspace.DevWorkspace, timingInfo, reqLogger)
 
 		// Don't accidentally suppress errors by overwriting here; only check for timeout when no error
 		// encountered in main reconcile loop.
 		if err == nil {
-			if timeoutErr := checkForStartTimeout(&clusterWorkspaceWithConfig.DevWorkspace, workspaceWithConfig.Config); timeoutErr != nil {
-				reconcileResult, err = r.failWorkspace(workspaceWithConfig, timeoutErr.Error(), metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus)
+			if timeoutErr := checkForStartTimeout(&clusterWorkspace.DevWorkspace, clusterWorkspace.Config); timeoutErr != nil {
```
AObuchow, line 222: It's not really necessary, but I decided to call `checkForStartTimeout()` with the `clusterWorkspace.Config` rather than `workspace.Config`

amisevsk (commented):
I'm still having some issues with how we can do this cleanly in the code. Currently, we have a split between functions that take DevWorkspaceWithConfig and those that take DevWorkspace alone, and it leads to a really confusing interaction:

* There are two types of functions -- those that take DevWorkspaceWithConfig (_DWWC functions_) and those that take DevWorkspace alone (_DW functions_).
* A DW function cannot call a DWWC function
  * In other words -- for any DWWC function, all functions up the call chain to `Reconcile` must also be DWWC functions
* If a DWWC function calls a DW function, it must unpack the DevWorkspaceWithConfig into just a DevWorkspace

This leads to strange blocks, like

```go
  switch newPhase {
	case dw.DevWorkspaceStatusRunning:
		metrics.WorkspaceRunning(workspace, logger) // Use DWWC here
	case dw.DevWorkspaceStatusFailed:
		metrics.WorkspaceFailed(&workspace.DevWorkspace, logger) // Have to unpack the DWWC here
```

it also removes the convenience of having a "common" config that can be read where needed -- if a function decides it wants its behavior to depend on configuration, it and all functions that call it recursively have to be updated to take DWWC instead of DW.

I'm thinking about approaches:

1. We can continue down this path, but consider updating our internal APIs to _always_ take DevWorkspaceWithConfig and only use DevWorkspaces when necessary (i.e. in k8s API calls). This would look something like https://github.com/amisevsk/devworkspace-operator/commit/e5465c0bd08aa54fe689e5904fe77d3aceeca247 though note I haven't tested it thoroughly.
    * Diff against main (including existing PR changes): https://github.com/devfile/devworkspace-operator/compare/main...amisevsk:devworkspace-operator:use-dwwc
    * This makes the API less convenient to use -- there's generally more friction and we're passing around a struct that may just never be used, and it always has to be present.
    * On the other hand, the changes in the PR could be summarized to
        * Update all functions to take DWWC instead of DW (i.e. don't use plain DevWorkspaces _anywhere_)
        * Update all k8s API calls to unpack DWWC to DW

      rather than the current issue which is some functions are half-updated

2. We can take a step back and pass the config separately where it is needed. This has the same problems as above, but at least makes where the config is necessary explicit

3. We could also completely abandon the idea of an anywhere-accessible config for the DevWorkspace and instead propagate specific values as required, but this is a significant change and I have no idea how it would play out in the repo.

AObuchow:
This PR has been reworked and moved to https://github.com/devfile/devworkspace-operator/pull/918

Issue #915, opened 2022-09-01
Update how personal access token git credentials are mounted into workspaces


amisevsk:
### Description
The DevWorkspace Operator allows mounting a Personal Access Token to workspaces, enabling access to private repos in cases where SSH keys aren't convenient. This is currently done by
1. Collecting all secrets labelled `controller.devfile.io/git-credential` in the current namespace
2. Merging them into a secret called `devworkspace-merged-git-credentials` (in order to allow multiple credentials to be specified)
3. Mounting this secret using a subpath volume mount to a mount path specified by the original secrets

The downside of this approach is that using a subpath volume mount prevents changes to the on-cluster secret from being propagated to the workspace, which can cause issues if PATs expire.

Instead, we should consider updating the DevWorkspace Operator to:
1. Mount the merged git credentials secret as files (rather than subpaths) to allow changes to be propagated into the workspace.
2. Ignore mount paths on git-credential secrets, and mount the credentials file to a fixed directory (this is required for the point above, as otherwise mounting to `/` will overwrite the entire filesystem).

### Additional context
Originally brought up as an issue in https://github.com/eclipse/che/issues/21291. Note the GitLab PATs expire after 2 hours, so even if they are refreshed as expected, the workspace would have to be restarted frequently.

Details on how `controller.devfile.io/mount-path` is currently used:
1. If no secret specifies a mount path, the default `/` is used and the credentials file is mounted to `/credentials`
4. If any secret specifies a mount path, that is used instead. Multiple secrets can specify the same mount path
5. If two or more secrets specify conflicting mount paths, it is treated as an error

PR #916, opened 2022-09-01
Ignore mount path for git-credentials secrets and mount as file

amisevsk:
### What does this PR do?
Ignore any mount-path annotations on secrets labelled 'controller.devfile.io/git-credential'. Instead, always mount the merged git credentials secret to `/.git-credentials/`.

Additionally, mount the credentials file as files rather than using subpath mounts, in order to ensure changes to the on-cluster secret can be propagated to the running workspace without requiring a restart.

Closes https://github.com/devfile/devworkspace-operator/issues/915

### Is it tested? How?
1. Create a secret with git-credentials label:
    ```yaml
    kind: Secret
    apiVersion: v1
    metadata:
      name: git-credentials-test-secret
      labels:
        controller.devfile.io/git-credential: 'true'
        controller.devfile.io/watch-secret: 'true'
    data:
      credentials: aGVsbG8gd29ybGQK # "hello world"
    type: Opaque
    ```
2. Start a workspace and verify file `/.git-credentials/credentials` is mounted and contains text "hello world"
3. Edit the secret in the cluster, changing the value of the `credentials` key
4. Update the workspace in any way to trigger a reconcile (e.g. add an annotation)
5. Check that file `/.git-credentials/credentials` reflects the new value (this may take some time to propagate down)

### PR Checklist

### DIFF
--- a/pkg/provision/automount/gitconfig.go
+++ b/pkg/provision/automount/gitconfig.go
@@ -24,6 +24,8 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const mergedGitCredentialsMountPath = "/.git-credentials/"
+
 // ProvisionGitConfiguration takes care of mounting git credentials and a gitconfig into a devworkspace.
 func ProvisionGitConfiguration(api sync.ClusterAPI, namespace string) (*Resources, error) {
 	credentialsSecrets, tlsConfigMaps, err := getGitResources(api, namespace)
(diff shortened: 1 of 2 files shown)


AObuchow (approved):
Tested (following the PR instructions) and works as expected:

```
bash-5.1$ cd  /.git-credentials/
bash-5.1$ cat credentials 
hello world
```

Modified the secret credentials data to: `fafghe9gd29ybGQK`, added an annotation to the workspace, then:
```
bash-5.1$ cat credentials 
}����`world
```

dkwon17 (approved):

l0rd:
Should we update the [documentation](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#adding-git-credentials-to-a-workspace) and remove the reference to `controller.devfile.io/mount-path`?

PR #918, opened 2022-09-02
Allow external DWOC to be merged with internal DWOC on a per-workspace basis

AObuchow:
### What does this PR do?
This PR makes 2 changes to DWO:
1. Refactor (almost) all uses of the global DWOC, so that, internally, each workspace instance may have its own configuration. Currently, this per-workspace DWOC is not visible to users, and was added only to support change n.2
2. Allow for specifying an external DWOC (that exists on the cluster), which gets merged with the workspace's internal DWOC.

The external DWOC's name and namespace are specified with the `controller.devfile.io/devworkspace-config` DevWorkspace attribute, which has the following structure:
    
```YAML
attributes:
    name: <string>
    namespace: <string>
```

Part of eclipse/che#21405

### Is it tested? How?
 2 automated tests were added.

For manual testing, perform the following steps:
1. Deploy DWO
2. Create a devfile with the `controller.devfile.io/devworkspace-config` attribute set, for reference, we'll call it `external-dwoc-workspace.yaml,` e.g.

```YAML
kind: DevWorkspace
metadata:
  name: theia-next-external-dwoc
spec:
  started: true
  template:
    attributes:
          name: external-dwoc-test
          namespace: default
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```
3. Create a DWOC file, for reference, we'll call it `test-external-dwoc.yaml`. Also, ensure its name field is set to the same value as the `controller.devfile.io/devworkspace-config-name` attribute in the devfile from step 2:
```YAML
config:
  routing:
    defaultRoutingClass: basic
  workspace:
    defaultStorageSize:
      common: 20Gi
      perWorkspace: 29Gi
    imagePullPolicy: Always
metadata:
  name: external-dwoc-test
  namespace: default
```
4. Apply the (external) DWOC file you just created, ensuring it's in the namespace you specified in the `controller.devfile.io/devworkspace-config-namespace` (in my example, the default namespace is used): `kubectl apply -f test-external-dwoc.yaml`
5. Apply the devfile created in step 2. e.g. `kubectl apply -f external-dwoc-workspace.yaml -n $NAMESPACE`
6. Verify that the internal DWOC has been modified. I recommend checking the size of the per-workspace PVC that is created and ensuring it is `29Gi` instead of the default `10Gi`
7. Afterwards, you can ensure that workspace's that do not have the `controller.devfile.io/devworkspace-config` attribute set are not affected by the external DWOC, and instead, they use the global/internal DWOC. For instance, start up the per-workspace example and ensure the PVC created for it is `10Gi` big: `kubectl apply -f per-workspace-storage.yaml -n $NAMESPACE`.

### Some notes and limitations:
1. The HTTP client / Proxy settings cannot be affected by the externally applied DWOC, as [the HTTP client is created and configured on DWO startup rather than workspace startup](https://github.com/devfile/devworkspace-operator/pull/711#issuecomment-995360107). 
2. The `basic` DevWorkspace Router is still using the cluster host suffix from the global DWOC. If required in the future, it is possible to allow the cluster host suffix to be retrieved from the workspace's configuration. [Here's an example of how this may be done](https://github.com/devfile/devworkspace-operator/pull/909/commits/62b3424566ac8781d556fea0aff22955e27ec884).
3. There is [a handler](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L680) that checks if the common PVC was deleted. Currently, it filters for PVC's related to DWO by checking the PVC name against the global configs PVC name. In the future, this limitation can be worked around by applying a DWO-specific annotation to all PVC's created by DWO, and filtering by that annotation. 
4. This change could have weird implications for anything configurable that is shared across workspaces. Most notably, the common PVC. For example, if the common PVC does not yet exist on the cluster, and an external DWOC that has the common PVC size set to `20Gi` is applied to a specific workspace (resulting in creation of the common PVC), then other (future) workspaces that use the common storage strategy will also be using a PVC of `20Gi`, even if they aren't using an external DWOC. 

### PR Checklist


AObuchow:
Something that may be currently missing from this PR is documenting the new `controller.devfile.io/devworkspace-config` attribute.

AObuchow:
> Something that may be currently missing from this PR is documenting the new `controller.devfile.io/devworkspace-config` attribute.

Additional documentation has now been updated.

controllers/workspace/devworkspace_controller.go
```diff
@@ -123,7 +123,10 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return reconcile.Result{}, err
 	}
 
-	config := wkspConfig.GetGlobalConfig()
+	config, err := wkspConfig.ResolveConfigForWorkspace(rawWorkspace, clusterAPI.Client)
+	if err != nil {
+		reqLogger.Error(err, "Error applying external DevWorkspace-Operator configuration")
+	}
 	configString := wkspConfig.GetCurrentConfigString()
```
amisevsk: `wkspConfig.GetCurrentConfigString()` should be updated to take a config as a parameter, to make it easier to print the resolved config. As it stands, using a custom config doesn't change the log below:
```go
reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
```

pkg/config/sync.go
```diff
@@ -49,6 +51,40 @@ func GetGlobalConfig() *controller.OperatorConfiguration {
 	return internalConfig.DeepCopy()
 }
 
+// TODO: Improve this function description?
+// Checks for an external DevWorkspace-Operator configuration specified by the optional workspace attribute `controller.devfile.io/devworkspace-config`
+// If the `controller.devfile.io/devworkspace-config` attribute is not set, the internal/global DWOC is returned
+// If the `controller.devfile.io/devworkspace-config` attribute is incorrectly set, the internal/global DWOC and an error are returned.
+// If the external DWOC is found on the cluster, it is merged with the internal/global DWOC and returned.
+// If the external DWOC is not found on the cluster, the internal/global DWOC and an error are returned.
+func ResolveConfigForWorkspace(workspace *dw.DevWorkspace, client crclient.Client) (*controller.OperatorConfiguration, error) {
```
amisevsk, line 60: Exported function docs should start with the name of the function (see: https://go.dev/doc/comment). In addition, returning a non-nil OperatorConfiguration on error is a bit of an anti-pattern as it allows callers to easily ignore errors.
```go
// ResolveConfigForWorkspace returns the OperatorConfiguration that should be used for the provided workspace. If the
// workspace specifies a custom configuration in its attributes, that configuration is merged into the global operator
// configuration and returned. Otherwise, the global operator configuration is returned. Returns an error if the custom
// configuration cannot be resolved.
```

pkg/config/sync.go
```diff
@@ -49,6 +51,40 @@ func GetGlobalConfig() *controller.OperatorConfiguration {
...
+	internalConfigCopy := internalConfig.DeepCopy()
+	if workspace.Spec.Template.Attributes.Exists(constants.ExternalDevWorkspaceConfiguration) {
+		namespacedName := types.NamespacedName{}
+
+		err := workspace.Spec.Template.Attributes.GetInto(constants.ExternalDevWorkspaceConfiguration, &namespacedName)
+		if err != nil {
+			return internalConfigCopy, fmt.Errorf("failed to read attribute %s in DevWorkspace attributes: %w", constants.ExternalDevWorkspaceConfiguration, err)
```
amisevsk, line 67: Should return `nil, [error]` (here and elsewhere in the function to force the error to be addressed (and not hidden/ignored). Then, call sites should be something like 
```go
	config, err := wkspConfig.ResolveConfigForWorkspace(rawWorkspace, clusterAPI.Client)
	if err != nil {
		reqLogger.Error(err, "Error applying external DevWorkspace-Operator configuration")
		config = wkspConfig.GetGlobalConfig()
	}
```
i.e. _explicitly_ use the global configuration if we fail to read a custom one.

pkg/config/sync_test.go
```diff
@@ -83,6 +89,96 @@ func TestSetupControllerMergesClusterConfig(t *testing.T) {
...
+	externalConfig.Config.Workspace.ImagePullPolicy = "Always"
+	externalConfig.Config.Workspace.PVCName = "test-PVC-name"
+	storageSize := resource.MustParse("15Gi")
+	externalConfig.Config.Workspace.DefaultStorageSize = &v1alpha1.StorageSizes{
+		Common:       &storageSize,
+		PerWorkspace: &storageSize,
+	}
+
+	clusterConfig := buildConfig(defaultConfig.DeepCopy())
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(clusterConfig).WithObjects(externalConfig).Build()
```
amisevsk, line 140: `WithObjects` takes `initObjs ...client.Object`:
```go
 	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(clusterConfig, externalConfig).Build()
```

pkg/constants/attributes.go
```diff
@@ -28,6 +28,21 @@ const (
...
+	// The structure of the attribute value should contain two strings: name and namespace.
+	// 'name' specifies the metadata.name of the external DevWorkspace-Operator configuration.
+	// 'namespace' specifies the metadata.namespace of the external DevWorkspace-Operator configuration.
+	// For example:
+	//
+	//   attributes:
+	//     controller.devfile.io/devworkspace-config:
+	//         name: external-dwoc-name
+	//         namespace: some-namespace
+	ExternalDevWorkspaceConfiguration = "controller.devfile.io/devworkspace-config"
```
amisevsk, line 44: For clarity, correct terms to use are `DevWorkspace Operator` for the name of the operator, `DevWorkspaceOperatorConfig` for the name of the custom resource, and just `configuration` or `operator configuration` for the general concept

docs/additional-configuration.adoc
```diff
@@ -156,6 +156,31 @@ Note: As for automatically mounting secrets, it is necessary to apply the `contr
 
 This will mount a file `/tmp/.git-credentials/credentials` in all workspace containers, and construct a git config to use this file as a credentials store. The mount path specified by the `controller.devfile.io/mount-path` annotation can by omitted, in which case `/` is used (mounting a file `/credentials`).
 
+## Setting an alternate configuration for a workspace
+It is possible to configure a workspace to use an alternate DevWorkspace-Operator configuration.
```
amisevsk, line 160:
```diff
- DevWorkspace-Operator configuration
+ DevWorkspaceOperatorConfig
```

docs/additional-configuration.adoc
```diff
@@ -156,6 +156,31 @@ Note: As for automatically mounting secrets, it is necessary to apply the `contr
...
+  template:
+    attributes:
+      controller.devfile.io/devworkspace-config:
+        name: <string>
+        namespace: <string>
+----
+
+*Note:* the alternate DevWorkspace-Operator configuration will be merged with the global DevWorkspace-Operator configuration; the configuration resulting from the merge will be used by the workspace that has the `controller.devfile.io/devworkspace-config` attribute set.
+During the merge of the alternate and global configuration, the fields which are set in the alternate configuration will overwrite those existing in the global configuration.
+
```
amisevsk, line 183: This kind of just says the same thing twice. 
```asciidoc
*Note:* the alternate DevWorkspace Operator configuration will be 
merged with the default DevWorkspaceOperatorConfig, overriding 
fields in the default configuration. Fields unset in the overridden 
configuration will use the global values.
```

amisevsk (commented):
Looking good but needs a bit more testing. I've pushed a couple of commits to this branch as well:

1. Commit d2a85ca3 - fix an issue where we were trying to patch DevWorkspaceWithConfig
2. Commit 41929c25 - fix the logging to log the resolvedConfig rather than the internal

controllers/workspace/devworkspace_controller.go
amisevsk: Fixed by 41929c25

amisevsk:
Also pushed d81bc716 to fix a bug I found, where DWO wasn't provisioning the common PVC. Finally got around to renaming `checkForExistingCommonPVC`, which was super confusing.

pkg/config/sync.go
AObuchow: I didn't want to return an error in the case that the attribute is not set in the workspace (cause then it would get logged constantly) so I added another function [`ExternalConfigSet`](https://github.com/devfile/devworkspace-operator/pull/918/commits/5d11520216ec35a96bb1a57d0be6041b287c4528#diff-44d82704651db22ff5ebfa1b636e76edf8e97a075c3cba2ea1714be914d6af74R55) to check before trying to resolve the workspace config. However, I dislike the [if/else case](https://github.com/devfile/devworkspace-operator/pull/918/commits/5d11520216ec35a96bb1a57d0be6041b287c4528#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R129) I had to resort to - maybe this could be done more elegantly

AObuchow: This has now been fixed

controllers/workspace/devworkspace_controller.go
```diff
@@ -121,8 +122,19 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+
+	config, err := wkspConfig.ResolveConfigForWorkspace(rawWorkspace, clusterAPI.Client)
+	if err != nil {
+		reqLogger.Error(err, "Error applying external DevWorkspace-Operator configuration")
```
amisevsk, line 128: We should reserve error logs for issues with the Operator itself; issues encountered here would be due to an issue in how the workspace is defined and so should be logged at info level with a warning:
```suggestion
		reqLogger.Info(fmt.Sprintf("Failed to apply external DevWorkspace-Operator configuration: %s", err)
```

As a future TODO, this should also probably set a warning condition on the workspace, but that's a task for another day.

pkg/common/types.go
```diff
@@ -0,0 +1,26 @@
...
+
+package common
+
+import (
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+)
+
+//TODO: Rename?
+// TODO: Move
```
amisevsk, line 22: Not necessary, but remember to remove all TODOs before we merge.

amisevsk (commented):
Pretty much looks good; will do one more quick round of testing tomorrow and then we should be good to merge.

amisevsk:
Rebased + squashed fixup commits

amisevsk (approved):
LGTM

PR #927, opened 2022-09-16
fix: Allow switching to per-user or common strategy if the storage type is not defined

ibuziuk:
…
Signed-off-by: Ilya Buziuk <ibuziuk@redhat.com>

### What does this PR do?
Allow switching to per-user or common strategy if the storage type is not defined

https://github.com/eclipse/che/issues/21707

### Is it tested? How?

Tested via `oc apply -f ./samples/git-clone-sample.yaml -n <namespace>` on dogfodding and the cluster with the fix.

On Dogfooding after patching the DevWorkspace ^ which does not contain attributes with:

```
  template:
    attributes:
        name: devworkspace-config
        namespace: dogfooding
``` 

The webhook error is shown `admission webhook "mutate.devworkspace-controller.svc" denied the request: DevWorkspace storage-type attribute cannot be changed once the workspace has been created.`

On a cluster, with a fix, it is possible to patch the DevWorkspace with the storage attributes, and no error is shown

### PR Checklist

### DIFF
--- a/webhook/workspace/handler/workspace.go
+++ b/webhook/workspace/handler/workspace.go
@@ -116,6 +116,9 @@ func (h *WebhookHandler) MutateWorkspaceV1alpha2OnUpdate(ctx context.Context, re
 		case oldStorageType == constants.EphemeralStorageClassType:
 			// Allow switching from ephemeral to a persistent storage type
 			break
+		case oldStorageType == "" && (newStorageType == constants.CommonStorageClassType || newStorageType == constants.PerUserStorageClassType):
+			// Allow switching to per-user or common persistent storage type if the oldStorageType is empty (if empty, the common PVC strategy is used by design)
+			break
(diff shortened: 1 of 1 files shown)


ibuziuk:
@AObuchow @amisevsk folks, if all is good this PR should be cherry-picked to 0.16.x
I have run the pre-release already, but have not run the release action for 0.16.0

AObuchow:
@ibuziuk The PR looks good to me (though not tested yet).

One thing that comes to mind, which is probably OK (as it is not the default case), is that this issue will still occur if the CheCluster CR gets configured to use the per-workspace storage strategy, and a workspace is opened that previously did not have the storage-type attribute set. 

However, yesterday when I was testing things out on Che, I was able to switch from per-workspace/per-user by restarting the workspace from the dashboard UI. So I imagine that should suffice as a workaround (assuming that still works).

ibuziuk:
> if the CheCluster CR gets configured to use the per-workspace storage strategy, and a workspace is opened that previously did not have the storage-type

need to analyze this use case more, but this does not look like a blocker at the first glance. In general, I'm thinking maybe we even should consider removing the webhook entirely, but somehow urge user that switching the storage type could lead to data loss and should be avoided.

amisevsk (approved):

PR #928, opened 2022-09-16
fix: Allow switching to per-user or common strategy if the storage type is not defined

ibuziuk:
### What does this PR do?
Cherry-picking https://github.com/devfile/devworkspace-operator/pull/927 to 0.16.x


amisevsk:
Didn't see this PR until now -- apologies; #927 is included in https://github.com/devfile/devworkspace-operator/pull/933.

Issue #929, opened 2022-09-16
Add code coverage GitHub action


AObuchow:
### Description
The devworkspace-operator repo should have a github action to check and display test coverage, similar to [che-operator](https://github.com/eclipse-che/che-operator/blob/main/.github/workflows/codecov.yml)

Related: https://github.com/devfile/devworkspace-operator/issues/735

PR #943, opened 2022-10-04
Add code coverage GitHub Action 

AObuchow:
### What does this PR do?
Enables the creation of code coverage reports for the repository.
A new CI action will run for PR's and for pushes to the main branch. The latter is required to have coverage reports which PR coverage reports can be compared against.  

I've [disabled](https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations) the annotations which point out which lines of code are not tested, as I personally find them to make PR's harder to read.

I also added the codecov badge to the README. 

One thing to note: the [codecov GH app](https://github.com/apps/codecov) must be enabled. I sent a request for this yesterday and saw today that it seems to be enabled. So there shouldn't be any issue there.

https://github.com/devfile/devworkspace-operator/issues/929

### Is it tested? How?
~~Testing is a bit tricky as this GH actions can only really be tested after they are merged.~~ 
**Edit:** Apparently, the code coverage action is running in this PR.. though it is likely to not work correctly as the main branch does not have a code coverage report uploaded to codecov.io yet.

I did test out the action [on my own fork of the repo](https://github.com/AObuchow/devworkspace-operator/pull/7#issuecomment-1267178317), and it seems to be working as expected.

I did not however test that the badge is working as expected, this will have to be verified after this PR is merged.

### PR Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+github_checks:
+    annotations: false
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,50 @@
+name: Code Coverage Report
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  coverage-report:
+    name: Check Code Coverage
+    runs-on: ubuntu-20.04
+    steps:
+    -
+      name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16.6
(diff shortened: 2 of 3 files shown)


codecov.yml
```diff
@@ -0,0 +1,2 @@
+github_checks:
+    annotations: false
```
amisevsk, line 2: Trailing newline

amisevsk (approved):
Looks fine to me :+1:

ibuziuk (approved):

PR #953, opened 2022-10-18
Add container security context field to DevWorkspace Operator Config

### LINKED ISSUE #951: Allow configuring default container security context
### Description
The DevWorkspace Operator Config allows configuration of the pod SecurityContext used for workspace pods. However, some securityContext fields are only available in the container-level securityContext. DWO should support configuring the container security context as well.

### LINKED ISSUE #952: Allow configured podSecurityContext to apply to OpenShift as well
### Description
The DevWorkspace Operator Config `.config.workspace.podSecurityContext` field is currently ignored on OpenShift (as by default you cannot set e.g. a custom runAsUser). However, other fields in podSecurityContext may be useful, so it would be convenient to be able to configure them on OpenShift.

amisevsk:
### What does this PR do?
* Add `.config.workspace.containerSecurityContext` field to DevWorkspaceOperatorConfig to allow configuring container-level security context for workspace pods
* Update how podSecurityContext is handled to allow specifying a configured value in OpenShift as well (previously it was ignored)
* Merge podSecurityContext values into default configuration rather than overwriting it entirely
* Minor fixes:
  * Add config package test that fuzzes config to ensure all fields are considered by merging
  * Update CRD documentation
  * Allow specifying different defaults for OpenShift and Kubernetes for podSecurityContext
  * Log when a non-default container/pod security context is used

Closes #951 
Closes #952 

### Is it tested? How?
To test:
* Verify that functionality is unchanged on OpenShift when no values are configured
* Verify that it's possible to configure both container and pod security context
* Verify that pod security context does not overwrite unset fields

On Kubernetes, the following DevWorkspaceOperatorConfig fields can be used to test:
```yaml
config:
  workspace:
    containerSecurityContext:
      capabilities:
        add:
        - SETUID
        - SETGID
    podSecurityContext:
      runAsUser: 2048
```
This should result in the workspace pod having `runAsUser: 2048` and pod containers having the `SETUID` and `SETGID` capabilities.

On OpenShift, the above DWOC will likely cause the workspace to fail to start. One permissible configuration is to specify a `runAsUser` within the UID range assigned the current project, e.g.
```yaml
config:
  workspace:
    containerSecurityContext:
      runAsUser: 1000670002
    podSecurityContext:
      runAsUser: 1000670001
```
(where `1000670000` is the default UID used for running workspace pods).

### PR Checklist

### DIFF
--- a/pkg/config/sync.go
+++ b/pkg/config/sync.go
@@ -17,6 +17,7 @@ package config
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"sync"
@@ -24,9 +25,11 @@ import (
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/pkg/config/proxy"
 	routeV1 "github.com/openshift/api/route/v1"
+	corev1 "k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -291,10 +294,10 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
 			to.Workspace.CleanupOnStop = from.Workspace.CleanupOnStop
 		}
 		if from.Workspace.PodSecurityContext != nil {
-			to.Workspace.PodSecurityContext = from.Workspace.PodSecurityContext
+			to.Workspace.PodSecurityContext = mergePodSecurityContext(to.Workspace.PodSecurityContext, from.Workspace.PodSecurityContext)
 		}
 		if from.Workspace.ContainerSecurityContext != nil {
-			to.Workspace.ContainerSecurityContext = from.Workspace.ContainerSecurityContext
+			to.Workspace.ContainerSecurityContext = mergeContainerSecurityContext(to.Workspace.ContainerSecurityContext, from.Workspace.ContainerSecurityContext)
 		}
 		if from.Workspace.DefaultStorageSize != nil {
 			if to.Workspace.DefaultStorageSize == nil {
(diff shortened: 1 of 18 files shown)


pkg/config/sync.go
```diff
@@ -314,7 +373,7 @@ func GetCurrentConfigString(currConfig *controller.OperatorConfiguration) string
 	}
 	workspace := currConfig.Workspace
 	if workspace != nil {
-		// Don't include PodSecurityContext for now as it's less easy to compare
+		// Don't include PodSecurityContext or ContainerSecurityContext for now as it's less easy to compare
```
AObuchow, line 376: Isn't this being done with the call to `reflect.DeepEqual`? Or did you mean, don't explicitly state which part of the PodSecurityContext/ContainerSecurityContext was modified and instead state that it was set by the user.

AObuchow:
Code looks good to me. 

I still have a bit of testing to do (had some install issues on my cluster & forgot to check the workspace deployment YAML instead of workspace pod YAML :man_facepalming:) but here are my findings so far (tested on OpenShift 4.11 and Minikube):

> Verify that functionality is unchanged on OpenShift when no values are configured

Things seem fine regarding this point, but I will post an update regarding this once I get a hold of another cluster.
When installed through OperatorHub (i.e previous functionality on OpenShift prior to this patch), the workspace deployment's securityContext should be empty, e.g
`securityContext: {} `

I still have to double-check this stays the same when the patch is applied, though it should.

For the workspace pod, **without the patch applied**, the pod SecurityContext was the following:
 ```YAML
  securityContext:
    seLinuxOptions:
      level: 's0:c20,c10'
    fsGroup: 1000400000
    seccompProfile:
      type: RuntimeDefault
```
And the container SecurityContext:
```YAML
      securityContext:
        capabilities:
          drop:
            - ALL
        runAsUser: 1000400000
        runAsNonRoot: true
        allowPrivilegeEscalation: false
```

For the workspace pod, **with the patch applied**, the pod SecurityContext was the following:
```YAML
  securityContext:
    seLinuxOptions:
      level: 's0:c26,c10'
    fsGroup: 1000670000
    seccompProfile:
      type: RuntimeDefault
```
And the container SecurityContext:
```YAML
      securityContext:
        capabilities:
          drop:
            - ALL
        runAsUser: 1000670000
        runAsNonRoot: true
        allowPrivilegeEscalation: false
```

The differences between the patch applied vs. without the patch are the `seLinuxOptions` `level`, the `fsGroup` and the `runAsUser` UID. I imagine these differences were caused by the cluster and not by the patch, but I am not sure. 

> Verify that it's possible to configure both container and pod security context

This worked on both OpenShift and Minikube. 

**On OpenShift**, I used the example pod and container security contexts provided in the PR. 
The workspace pod's SecurityContext was:
```YAML
  securityContext:
    seLinuxOptions:
      level: 's0:c26,c10'
    runAsUser: 1000670001
    fsGroup: 1000670000
    seccompProfile:
      type: RuntimeDefault
```
And the container security contexts were:
```YAML
      securityContext:
        capabilities:
          drop:
            - ALL
        runAsUser: 1000670002
        runAsNonRoot: true
        allowPrivilegeEscalation: false
```
Note that the values for runAsUser are correctly set.

**On Minikube**, I used the example pod and container security contexts provided in the PR, but also added the KILL capability.
The workspace pod's SecurityContext was: 
```YAML
   securityContext:
    runAsUser: 2048
    runAsGroup: 0
    runAsNonRoot: true
    fsGroup: 1234
```
And the container security contexts were:
```YAML
      securityContext:
        capabilities:
          add:
            - SETUID
            - SETGID
            - KILL
```
Note that the correct capabilities were added (including KILL) and the value of runAsUser was correct. 


> Verify that pod security context does not overwrite unset fields

This seems to work.

On Minikube, any unset fields for the pod security context in the DWOC will result in the [default values being used](https://github.com/devfile/devworkspace-operator/pull/953/files#diff-2658370b0497fa11f0682953637e6e2d237a568851f410b6de8e1ad202045a6cR49).

On OpenShift, fields that were not set in the pod security context (such as fsGroup) did not change.

With pod SecurityContext set in DWOC:
```YAML
  securityContext:
    seLinuxOptions:
      level: 's0:c26,c10'
    runAsUser: 1000670001
    fsGroup: 1000670000
    seccompProfile:
      type: RuntimeDefault
```

Without pod SecurityContext set in DWOC:
```YAML
  securityContext:
    seLinuxOptions:
      level: 's0:c26,c10'
    fsGroup: 1000670000
    seccompProfile:
      type: RuntimeDefault
```

AObuchow (approved):

PR #959, opened 2022-10-25
Use k8s.io/utils/pointer to instantiate pointers

### LINKED ISSUE #958: Use k8s.io/utils/pointer to instantiate pointers throughout the codebase
### Description
Currently, throughout the DWO repo we are [initializing variables](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/defaults.go#L52) so that they can be referenced as pointer values when needed, for example [here](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/defaults.go#L43). 

Instead, we should make use of `k8s.io/utils/pointer` to instantiate pointers and reference them in a single statement.
For example, one can get a pointer that points to a boolean with the value `true` by doing: `pointer.BoolPtr(true)`

### Additional context
Here is where this idea was initially suggested: https://github.com/devfile/devworkspace-operator/pull/953#discussion_r1001981062

PiotrLewandowski323:
### What does this PR do?

Removes some of the variables which are initialized only so that they be referenced as pointer values and replaces their usage with pointers from k8s.io/utils/pointer.

Fixes #958 

### Is it tested? How?

I've run make test and make docker-build. I don't know how I should properly test these changes otherwise.

### PR Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.21.3
 	k8s.io/apimachinery v0.21.3
 	k8s.io/client-go v0.21.3
+	k8s.io/utils v0.0.0-20210722164352-7f3ee0f31471
 	sigs.k8s.io/controller-runtime v0.9.5
 	sigs.k8s.io/yaml v1.2.0
 )
--- a/pkg/config/common_test.go
+++ b/pkg/config/common_test.go
@@ -38,8 +38,7 @@ const (
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	trueBool = true
+	scheme = runtime.NewScheme()
 )
 
 func init() {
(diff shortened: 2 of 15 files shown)


amisevsk (approved):
LGTM

Welcome @PiotrLewandowski323 and thanks for your first contribution to the DevWorkspace Operator :smile:

amisevsk:
In case the Che Operator needs to pull in kubernetes/utils in the future, here is the clearlydefined.io entry for the current version: https://clearlydefined.io/definitions/git/github/kubernetes/utils/7d6a63dca704acaa0679df837cea69eabb2e5f67

AObuchow:
> In case the Che Operator needs to pull in kubernetes/utils in the future, here is the clearlydefined.io entry for the current version: [clearlydefined.io/definitions/git/github/kubernetes/utils/7d6a63dca704acaa0679df837cea69eabb2e5f67](https://clearlydefined.io/definitions/git/github/kubernetes/utils/7d6a63dca704acaa0679df837cea69eabb2e5f67)

I actually found out about kubernetes/utils from the Che-Operator codebase :stuck_out_tongue:  good to have that lying around though :+1:

AObuchow (approved):

AObuchow:
@PiotrLewandowski323 Thank you very much for your PR! :)

Issue #966, opened 2022-11-03
Support specifying patch directives in pod and container overrides


amisevsk:
### Description
When using the pod-overrides or container-overrides attribute in a DevWorkspace, the override bytes are deserialized into a PodSecurityContext/SecurityContext and applied via strategic merge patch. However, this approach prevents _un-setting_ fields within a struct, as empty fields are ignored in the serialize/deserialize procedure (due to omitempty). This means that if a pod is configured to have security context

```yaml
securityContext:
    runAsUser: 1000
    runAsGroup: 3000
    fsGroup: 2000
```
it's not possible to specify a pod override to make this
```yaml
securityContext:
    runAsUser: 1001
```

Kubernetes strategic merge supports patch directives to change the behavior of the patch:
* `$patch: replace` causes fields to be applied as-is rather than merged into existing values
* `$patch: delete` deletes fields altogether without patching anything

### Additional context
Originally brought up in discussion on https://github.com/devfile/api/issues/920

l0rd:
@amisevsk is there some documentation about this or should we add it (in the [dwo additional doc md](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#pod-overrides) file I guess)?

l0rd:
cc @max-cx

amisevsk:
I'm hesitant to add explicit documentation as it's not our functionality so much as it's what is supported by Kubernetes strategic merge patch (documented [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md)).

l0rd:
Right. Linking to the strategic merge path documentation and mentioning that directives modify the behavior would be useful I think.

amisevsk:
Created PR https://github.com/devfile/devworkspace-operator/pull/981 to add a short note and link to docs mentioned above.

PR #967, opened 2022-11-03
Support patch directives in pod/container overrides

amisevsk:
### What does this PR do?
Adds support for Kubernetes strategic merge [patch directives](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#basic-patch-format). This allows deleting/overwriting fields that are not settable otherwise.

For example, if the default pod securityContext is configured to be
```yaml
config:
  workspace:
    podSecurityContext:
      fsGroup: 2000
      runAsGroup: 3000
      runAsUser: 1000
```
setting the pod-overrides attribute to be
```yaml
      pod-overrides:
        spec:
          securityContext:
            runAsUser: 1001
```
will result in the pod using the securityContext 
```yaml
securityContext:
    fsGroup: 2000
    runAsGroup: 3000
    runAsUser: 1001
```
whereas setting it to be
```yaml
      pod-overrides:
        spec:
          securityContext:
            runAsUser: 1001
            $patch: replace
```
will result in the securityContext
```yaml
securityContext:
    runAsUser: 1001
```
To do this, we have to apply the raw yaml specified in the overrides field as the patch rather than deserializing it, since Go will drop unrecognized fields in the struct when deserializing.

Closes https://github.com/devfile/devworkspace-operator/issues/966

### Is it tested? How?
Test by using directives in pod/container overrides.

### PR Checklist

### DIFF
--- a/pkg/library/overrides/testdata/container-overrides/overrides-can-override-securityContext.yaml
+++ b/pkg/library/overrides/testdata/container-overrides/overrides-can-override-securityContext.yaml
@@ -1,0 +1,26 @@
+name: "container overrides can override securityContext"
+
+input:
+  component:
+    name: test-component
+    attributes:
+      container-overrides:
+        securityContext:
+          runAsUser: 1001
+    container:
+      image: test-image
+
+  container:
+    name: test-component
+    image: test-image
+    securityContext:
+      runAsUser: 1000
+      runAsGroup: 2000
+
+output:
+  container:
+    name: test-component
+    image: test-image
+    securityContext:
+      runAsUser: 1001
+      runAsGroup: 2000
--- a/pkg/library/overrides/containers.go
+++ b/pkg/library/overrides/containers.go
@@ -33,19 +33,18 @@ func ApplyContainerOverrides(component *dw.Component, container *corev1.Containe
 	if err := component.Attributes.GetInto(constants.ContainerOverridesAttribute, override); err != nil {
 		return nil, fmt.Errorf("failed to parse %s attribute on component %s: %w", constants.ContainerOverridesAttribute, component.Name, err)
 	}
-	override = restrictContainerOverride(override)
-
-	overrideBytes, err := json.Marshal(override)
-	if err != nil {
-		return nil, fmt.Errorf("error applying container overrides: %w", err)
+	if err := restrictContainerOverride(override); err != nil {
+		return nil, fmt.Errorf("failed to parse %s attribute on component %s: %w", constants.ContainerOverridesAttribute, component.Name, err)
 	}
+
+	overrideJSON := component.Attributes[constants.ContainerOverridesAttribute]
 
 	originalBytes, err := json.Marshal(container)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


pkg/library/overrides/testdata/container-overrides/overrides-can-override-securityContext.yaml
```diff
@@ -0,0 +1,26 @@
+name: "container overrides can override securityContext"
```
AObuchow, line 1: This test name as well as `/container-overrides/overrides-can-use-delete-directive.yaml` and `/container-overrides/overrides-can-use-replace-directive.yaml` starts with a lower-case letter rather than uppercase like the others.

Very unimportant and doesn't need to be fixed, but it's the only remark I could find about this PR :P

AObuchow:
I still need to do some manual testing, but the code changes look good to me.

It might be worth linking https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md?rgh-link-date=2022-11-03T17%3A21%3A38Z#basic-patch-format in the [additional documentation for pod/container overrides](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#fine-grained-configuration-of-workspace-pods-and-containers).

Maybe something as simple as:

> ...The value of the attribute is applied via Kubernetes strategic merge to patch the default configuration according to[ predefined merge rules](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md?rgh-link-date=2022-11-03T17%3A21%3A38Z#basic-patch-format).

dkwon17 (approved):
LGTM, and I had some issues with manipulating the `securityContext:` using `$patch: replace` / `$patch: delete` on OpenShift (I think OpenShift might intervene and set the securityContext itself), but it worked well on Minikube.

AObuchow (approved):
Tested on OpenShift 4.11, everything works as expected :+1: 
Here was the DWOC I used:
```YAML
config:
  routing:
    defaultRoutingClass: basic
  workspace:
    imagePullPolicy: Always
    podSecurityContext:
      fsGroup: 1000670000
      runAsUser: 1000670000
```
Setting the devworkspace's pod-overrides to the following:
```YAML
    attributes:
      pod-overrides:
        spec:
          securityContext:
            runAsUser: 1000670001
```

Resulted in the following security context:
```YAML
      securityContext:
        runAsUser: 1000670001
        fsGroup: 1000670000
```

Here I tested the replace patch directive:
```YAML
    attributes:
      pod-overrides:
        spec:
          securityContext:
            runAsUser: 1000670001
            $patch: replace
```

Which resulted in the following security context (as expected):
```YAML
    securityContext:
        runAsUser: 1000670001
```

And lastly, I tested the delete patch directive:
```YAML
    attributes:
      pod-overrides:
        spec:
          securityContext:
            runAsUser: 1000670001
            $patch: delete
```

Which removed the security context entirely, as expected:
```YAML
securityContext: {}
```

ibuziuk (approved):

PR #970, opened 2022-11-03
Floor the Grafana Prometheus queries for workspace starts/fails

dkwon17:

### What does this PR do?
Floors the devworkspace start and devworkspace start success/failure metrics so that decimal points don't appear in the dashboard

The dashboard can display non integer values for the devworkspace start and devworkspace start success/failure metrics, which does not make much sense:
image

This is because the [increase() function](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase) used to determine the #of starts in a time interval, extrapolates the given data points (# of starts, successes/failures and what time they happened) within the given time interval. Extrapolating here does not make much sense because # of starts, successes/failures should be integer values.

### PR Checklist

### DIFF
--- a/docs/grafana/grafana-dashboard.json
+++ b/docs/grafana/grafana-dashboard.json
@@ -258,7 +258,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum by (source) (increase(devworkspace_started_total[$__interval]))",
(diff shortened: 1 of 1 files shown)


amisevsk (approved):
I'm not sure this is an ideal way to fix this issue (and potentially the solution is to increase the interval instead) but I don't have a great alternative so :+1:

ibuziuk:
@dkwon17 is it still Draft of can we merge?

PR #978, opened 2022-11-15
Add check for pod unschedulable conditon

### LINKED ISSUE #977: Workspace pod FailedScheduling event not being properly caught
### Description
In some cases, a workspace deployment's pod will trigger the `FailedScheduling` event and DWO will not be able to properly catch it. 
So far, this seems to occur when a workspace requests more memory or CPU than is available on any node on an OpenShift cluster. 

What I have observed is that, although DWO sees the pod's `FailedScheduling` event, the `event.count` is set to 0. DWO then [checks to see if the `event.count` is higher](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/status/check.go#L126) than the [threshold required to report an error](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/status/check.go#L46). Since the threshold is 1 (i.e. `event.count` must = 1 for an error to be reported), the error is never reported, and eventually, the workspace times out.

I have been trying to narrow down the exact cause of this bug, but have been unsuccessful thus far.
However, here are my current findings:
- This issue seems to be consistently reproducible on OpenShift 4.11 and 4.12. On OpenShift 4.10, the `FailedScheduling` event is sometimes properly caught, and in other cases, the workspace times out. 
    - When this bug occurs (on OpenShift 4.10, 4.11 & 4.12) the `event.count` is set to 0 and `event.series` is `nil` (so `event.series.count` is unusable)
- This issue does not occur on Minikube v1.27.1 with Kubernetes v1.25.2
    - When testing, the `event.count` was set to 1 upon first encountering the `FailedScheduling`event, and the count increments with each occurrence of the event. The `event.series` is `nil`.
- The `event.count` field has been [deprecated in Kubernetes v1.25](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#event-v125), and instead the `event.series.count` field is supposed to be used. However, in my testing on Minikube, the `event.series` field was still `nil`.
    - The new event API is [`k8s.io/api/events/v1`](https://pkg.go.dev/k8s.io/api/events/v1#Event) (and the deprecated one which is currently in use in DWO comes from [`k8s.io/api/core/v1`](https://pkg.go.dev/k8s.io/api@v0.25.4/core/v1#Event)
        - I have tried using `k8s.io/api/events/v1` but encountered the same behaviour as `k8s.io/api/core/v1`: on affected OpenShift versions, `event.count` is set to 0 and `event.series` is nil. There is also auto-generated code to convert back and forth between core v1 events and events v1 ([here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/zz_generated.conversion.go#L3640) and [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/zz_generated.conversion.go#L3668)) 
- The default Kubernetes scheduler (which is also used by OpenShift, unless configured otherwise) seems to use [the following event recorder implementation](https://github.com/kubernetes/kubernetes/blob/9918aa1e035a00bc7c0f16a05e1b222650b3eabc/staging/src/k8s.io/client-go/tools/events/event_recorder.go#L70), which does **not** initialize the event count field (thus, we'd expect `event.count` to be initialized to 0, but oddly, this is not the case on Kubernetes v1.25).
    - The call stack seems to be: kube-scheduler executes [`runCommand`](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L140) -> `Setup` is called, which executes [`options.Config()`](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L318) -> The [kube-scheduler's event broadcaster is initialized](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/options/options.go#L289) as an event broadcaster adapter. The event broadcaster adapter [checks if the cluster supports `k8s.io/api/events/v1` resources](https://github.com/kubernetes/client-go/blob/master/tools/events/event_broadcaster.go#L376) (i.e. the new events API), and if so, an events/v1 EventBroadcaster will be used (otherwise a core/v1 events EventBroadcaster is used). In my testing on OpenShift 4.11 and Minikube, `k8s.io/api/events/v1` resources were supported, so [the new event API should be in use by the EventBroadcaster](https://github.com/kubernetes/client-go/blob/master/tools/events/event_broadcaster.go#L69). 
    - Continuing the call stack: The kube-scheduler [obtains](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L333) an (event) `recorderFactory` [from the `k8s.io/api/events/v1` EventBroadcaster](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L294). This recorderFactory is [used by the kubernetes scheduler struct](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L339): a map of scheduling profiles to [frameworks](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/interface.go#L533) allows [the association of pods to the `k8s.io/api/events/v1` event recorder](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/schedule_one.go#L309) when a [pod is being scheduled](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/schedule_one.go#L70). Finally, when the scheduler's failure handler is run, the `FailedScheduling` [event is created](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/schedule_one.go#L908) using the `k8s.io/api/events/v1` event recorder.

Other notes:
- @ibuziuk reported that the `FailedScheduling` event was being caught correctly (on CRC 4.10.14 I believe) when it was caused by a disk pressure issue. I am not sure whether there is something particular about disk-pressure warnings that cause them to be caught, or if this is a case of the `FailedScheduling` event being sometimes caught on OpenShift 4.10
- I haven't been able to find any Kubernetes or OpenShift bugs directly related to this issue. However, I did find this https://github.com/elastic/beats/issues/31126 which describes a similar issue with the event count being set to 0. 
    - I considered using a similar approach to [this PR](https://github.com/elastic/beats/pull/29591/files#diff-466985006c0af96104c2ed6bfe6e131aeb8a33f3b1e7e5396d2dc824e0543a44R134), however this could still lead to breaking DWO's detection of the `FailedMount` event,  as [we allow 3 `FailedMount` events to occur](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/status/check.go#L45) before reporting an error. 
    - I'm not sure if it is easy to reproduce the `FailedMount` event for testing purposes, but if its event count is being correctly set and incremented, then the above approach could suffice. 
- We could add extra checks (after checking pod events) for the `pod.status.conditions` to see if the `PodScheduled` condition is set to False with the `Unschedulable` reason, and fail the workspace if this condition is found. However, this seems less fine-grained than checking pod events. Perhaps there's another way to determine if a pod will not be able to be scheduled, other than checking pod events and pod status conditions?
### How To Reproduce
Apply either of the following devworkspaces to an OpenShift 4.11 or 4.12 cluster (on 4.10, this issue may or may not occur), and see that they will time out rather than fail immediately:
 
```YAML
kind: DevWorkspace
metadata:
  name: theia-next-high-cpu
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
                memoryRequest: 2Gi
                memoryLimit: 16Gi
                cpuRequest: 4000m
                cpuLimit: 8000m
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```

```YAML
kind: DevWorkspace
metadata:
  name: theia-next-high-mem
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
                memoryRequest: 32Gi
                memoryLimit: 64Gi
                cpuRequest: 500m
                cpuLimit: 2000m
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```

When checking the workspace pod with kubectl/oc or the OpenShift UI, you can see that the pod is pending and that an event similar to the following has occurred:
`0/6 nodes are available: 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 6 Insufficient cpu. preemption: 0/6 nodes are available: 3 No preemption victims found for incoming pod, 3 Preemption is not helpful for scheduling.`

image

### Expected behaviour
The `FailedScheduling` event should be caught and reported, and the workspace should be failed immediately. 

### Additional context
[Downstream issue](https://issues.redhat.com/browse/CRW-3474)

AObuchow:
### What does this PR do?
This PR adds a check for deployed pods to see whether the unschedulable condition has occurred.
If this condition is detected, an error message is propagated and the workspace is subsequently failed. 

Fix #977

### Is it tested? How?
To properly test that this PR fixes #977, you must test this on an OpenShift 4.11 or 4.12 cluster (the original bug also can occur on  4.10, but not consistently).
1. Build and install a DWO image from this PR: `make docker install`
2. Create the following 2 workspaces (i.e `kubectl apply -f ... `):

```YAML
kind: DevWorkspace
metadata:
  name: theia-next-high-cpu
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
                memoryRequest: 2Gi
                memoryLimit: 16Gi
                cpuRequest: 4000m
                cpuLimit: 8000m
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```

```YAML
kind: DevWorkspace
metadata:
  name: theia-next-high-mem
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: theia
        plugin:
          components:
            - name: theia-ide
              container:
                env:
                  - name: THEIA_HOST
                    value: 0.0.0.0
                memoryRequest: 32Gi
                memoryLimit: 64Gi
                cpuRequest: 500m
                cpuLimit: 2000m
    commands:
      - id: say-hello
        exec:
          component: theia-ide
```
3. Do a `kubectl get dw -n <namespace-where-workspaces-are-deployed> -w` and ensure that both workspaces change to the Failed state. You should see something like the following:

```
NAME                  DEVWORKSPACE ID             PHASE      INFO
theia-next-high-cpu   workspace4f052cedf39c4b99   Failed     Pod is unschedulable: 0/6 nodes are available: 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 6 Insufficient cpu. preemption: 0/6 nodes are available: 3 No preemption victims found for incoming pod, 3 Preemption is not helpful for scheduling.
theia-next-high-mem   workspace925fffa48cf24301   Failed     Pod is unschedulable: 0/6 nodes are available: 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 6 Insufficient memory. preemption: 0/6 nodes are available: 3 No preemption victims found for incoming pod, 3 Preemption is not helpful for scheduling.
```

### PR Checklist

### DIFF
--- a/pkg/library/status/check.go
+++ b/pkg/library/status/check.go
@@ -96,6 +96,11 @@ func CheckPodsState(workspaceID string, namespace string, labelSelector k8sclien
 		if msg, err := CheckPodEvents(&pod, workspaceID, ignoredEvents, clusterAPI); err != nil || msg != "" {
 			return msg, err
 		}
+		if pod.Status.Phase == corev1.PodPending {
+			if msg := checkPodConditions(&pod); msg != "" {
+				return msg, nil
+			}
+		}
 	}
 	return "", nil
 }
@@ -163,4 +168,16 @@ func checkIfUnrecoverableEventIgnored(reason string, ignoredEvents []string) (ig
 		}
 	}
 	return false
+}
+
+func checkPodConditions(pod *corev1.Pod) (msg string) {
+	if pod.Status.Conditions != nil {
+		for _, condition := range pod.Status.Conditions {
+			// Pod unschedulable condition
+			if condition.Type == corev1.PodScheduled && condition.Status == corev1.ConditionFalse && condition.Reason == corev1.PodReasonUnschedulable {
+				return fmt.Sprintf("Pod is unschedulable: %s", condition.Message)
+			}
+		}
+	}
+	return ""
 }


pkg/library/status/check.go
```diff
@@ -96,6 +96,11 @@ func CheckPodsState(workspaceID string, namespace string, labelSelector k8sclien
 		if msg, err := CheckPodEvents(&pod, workspaceID, ignoredEvents, clusterAPI); err != nil || msg != "" {
 			return msg, err
 		}
+		if pod.Status.Phase == corev1.PodPending {
```
AObuchow, line 99: Checking the pod phase could be put inside the `checkPodConditions()` function, or omitted entirely. Not really sure about this one, though in my observation of the original bug, the pod status would always be stuck in the pending phase, so it made sense to add a check for it.

amisevsk (approved):

AObuchow:
@ibuziuk This is the error shown in the Che Dashboard, it took about 10 seconds from starting the workspace to appear:
image

Issue #984, opened 2022-11-23
Give warning when using unsupported devfile features


AObuchow:
### Description
Since there are devfile 2.2.x features that are not currently supported by DWO, we should warn the user when these features are being used in a devworkspace. 

I imagine this would involve checking the workspace after it has been flattened and setting a warning to the workspace status. 
### Additional context
Related: https://github.com/devfile/devworkspace-operator/issues/983

AObuchow:
There are 3 potential approaches to presenting this warning to users that I've thought of so far: setting a warning in the workspace's status conditions, letting the webhook return an error when applying the devworkspace or letting the workspace give a warning when applying the devworkspace.

## Workspace status warning
**Pros:**
- Allows users to still use devworkspaces with unsupported features (though it's possible the workspace won't work as expected)

**Cons:**
- Not very visible to users. AFAIK, the dashboard does not present workspace status warnings. Thus, the only way for a user to see the warning is by doing a `kubectl get dw ...` or a `kubectl describe dw <workspace-name>`.  This could lead to a scenario where the user's workspace isn't working correctly, and they have to a bit of debugging/research to figure out the issue. 

## Blocking workspaces that use unsupported features with webhooks
**Pros:**
- The dashboard is already able to present a UI error when the webhook returns an error
- It is impossible for the user to ignore the warning (since the workspace won't start), preventing potential cases where a user ignores the warning and reports issues with their workspace

**Cons:**  
- There are unsupported features such as `components.volume.size` that won't necessarily break the workspace if they aren't supported by DWO. It could be frustrating to try using a devfile that uses these features (especially if the devfile comes from a repository that the user does not own), since the flow to get the workspace running would be:
	1. Try opening the workspace and receive the error 
	2. Fork the repository
	3. Locally clone the repository
	4. Remove unsupported features from devfile
	5. Commit and push changes
	6. Try opening the workspace again, using the user's fork repository

## Webhook warnings for unsupported features
**Pros:**
- Would allow users to still use devworkspace that have unsupported features (though the workspace may not behave correctly)

**Cons:**
- The dashboard does not support displaying webhook warnings yet. Currently, the only way to see the webhook warning would be to check DWO's logs, which a normal user may not have privilege to do. Checking the logs is also more difficult than doing a `kubectl get dw ...` or a `kubectl describe dw <workspace-name>`


[...]

PR #989, opened 2022-12-02
feat: warn when trying to use unsupported devfile features

AObuchow:
This PR adds a webhook warning when trying to create a devworkspace that uses features that are [unsupported by DWO](https://github.com/devfile/devworkspace-operator/blob/main/docs/unsupported-devfile-api.adoc).

A warning is also given when adding unsupported features to an existing devworkspaces `spec.template`. When this happens, a warning is only given for unsupported features that were not already present in the devworkspace (to reduce noise and make it more obvious which modifications gave new warnings).

#984

### Is it tested? How?
Some automated tests were added, however these only test internal functions and not the webhook create/update functions themselves. 

There are 3 cases to manually test, I recommend following them in order.
 
First install DWO to your cluster with the changes from this PR, i.e: `make docker install`
# Creating a devworkspace that uses unsupported features
1. Apply the following devworkspace, which uses `components.annotation.service` and `events.preStop`:

```YAML
cat <<EOF | kubectl apply -n $NAMESPACE -f -         
kind: DevWorkspace
metadata:
  name: unsupported-features
spec:
  started: true
  routingClass: 'basic'
  template:
    commands:
      - id: init-cache
        exec:
          component: tools
          commandLine: "init cache"
          workingDir: /.m2
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
          annotation:
            service:
              key: value
    events:
      preStop:
        - init-cache
EOF
```
2. You should see a warning like the following:

```
Warning: Unsupported Devfile features are present in this workspace. The following features will have no effect: components[].container.annotation.service, used by components: web-terminal; events.preStop: init-cache
devworkspace.workspace.devfile.io/unsupported-features created
```

# Adding new unsupported features to a devworkspace that already has unsupported features
Add an [unsupported feature](https://github.com/devfile/devworkspace-operator/blob/main/docs/unsupported-devfile-api.adoc) to the `template.spec`, e.g `spec.template.components.container.dedicatedPod`.

Continuing off the last test case: 
1. `kubectl edit dw unsupported-features -n $NAMESPACE`
```diff
apiVersion: workspace.devfile.io/v1alpha2
kind: DevWorkspace
(..)
spec:
  routingClass: basic
  started: true
  template:
    commands:
    - exec:
        commandLine: init cache
        component: tools
        workingDir: /.m2
      id: init-cache
    components:
    - container:
        annotation:
          service:
            key: value
        command:
        - tail
        - -f
        - /dev/null
        image: quay.io/wto/web-terminal-tooling:next
        memoryLimit: 512Mi
        mountSources: true
        sourceMapping: /projects
+       dedicatedPod: true
      name: web-terminal
    events:
      preStop:
      - init-cache
status:
  conditions:
(..)
```
2. You should see a warning that only mentions the newly added unsupported feature(s):
```
Warning: Unsupported Devfile features are present in this workspace. The following features will have no effect: components[].container.dedicatedPod, used by components: web-terminal
devworkspace.workspace.devfile.io/unsupported-features edited
```

# Trying to remove the creator label while adding unsupported features to a devworkspace
Continuing off of the last test case: remove the `spec.template.components.container.dedicatedPod` from the devworkspace (no warnings should be given).

1. Edit the devworkspace and remove the `controller.devfile.io/creator` label, and add an unsupported feature to the `spec.template`.

```diff
apiVersion: workspace.devfile.io/v1alpha2
kind: DevWorkspace
metadata:
(..)
- labels:
-   controller.devfile.io/creator: ""
  name: unsupported-features
  namespace: devworkspace-controller
  resourceVersion: "83951"
  uid: 292207f2-b776-41b7-9d93-3a4603042f04
spec:
  routingClass: basic
  started: true
  template:
    commands:
    - exec:
        commandLine: init cache
        component: tools
        workingDir: /.m2
      id: init-cache
    components:
    - container:
        annotation:
          service:
            key: value
        command:
        - tail
        - -f
        - /dev/null
+       dedicatedPod: true
        image: quay.io/wto/web-terminal-tooling:next
        memoryLimit: 512Mi
        mountSources: true
        sourceMapping: /projects
      name: web-terminal
    events:
      preStop:
      - init-cache
status:
  conditions:
```
2. The edit should result in a warning about the feature you added:
```
Warning: Unsupported Devfile features are present in this workspace. The following features will have no effect: components[].container.dedicatedPod, used by components: web-terminal
devworkspace.workspace.devfile.io/unsupported-features edited
```
3. Checking the devworkspace again (`kubectl describe dw unsupported-features -n $NAMESPACE
`) should show that the `controller.devfile.io/creator` label was re-added.
 
```diff
 Name:         unsupported-features
 Namespace:    devworkspace-controller
+ Labels:       controller.devfile.io/creator=
 Annotations:  controller.devfile.io/started-at: 1669930843394
 API Version:  workspace.devfile.io/v1alpha2
 Kind:         DevWorkspace
Metadata:
(...)
Spec:
  Routing Class:  basic
  Started:        true
  Template:
    Commands:
      Exec:
        Command Line:  init cache
        Component:     tools
        Working Dir:   /.m2
      Id:              init-cache
    Components:
      Container:
        Annotation:
          Service:
            Key:  value
        Command:
          tail
          -f
          /dev/null
+       Dedicated Pod:   true
        Image:           quay.io/wto/web-terminal-tooling:next
        Memory Limit:    512Mi
        Mount Sources:   true
        Source Mapping:  /projects
      Name:              web-terminal
    Events:
      Pre Stop:
        init-cache
```

### PR Checklist

### DIFF
--- a/webhook/workspace/handler/testdata/warning/add-unsupported-features-when-some-present.yaml
+++ b/webhook/workspace/handler/testdata/warning/add-unsupported-features-when-some-present.yaml
@@ -1,0 +1,37 @@
+name: "Add unsupported devfile features to workspace that already contains unsupported features"
+
+input:
+  oldWorkspace:
+    components:
+      - name: testing-container-1
+        container:
+          image: testing-image
+          dedicatedPod: true
+      - name: projects
+        volume:
+          ephemeral: true
+
+  newWorkspace:
+    components:
+      - name: testing-container-1
+        container:
+          image: testing-image
+          dedicatedPod: true
+      - name: projects
+        volume:
+          ephemeral: true
+          size: "10Gi"
+      - name: image-component
+        image:
+          imageName: python-image:latest
+          autoBuild: true
+          dockerfile:
+            uri: docker/Dockerfile
+            args:
+              - 'MY_ENV=/home/path'
+            buildContext: .
+            rootRequired: false
+
+output:
+  expectedWarning: "Unsupported Devfile features are present in this workspace. The following features will have no effect: components[].image, used by components: image-component; components[].volume.size, used by components: projects"
+  newWarningsPresent: true
--- a/webhook/workspace/handler/warning.go
+++ b/webhook/workspace/handler/warning.go
@@ -1,0 +1,179 @@
+//
+// Copyright (c) 2019-2022 Red Hat, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package handler
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	dwv2 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+)
+
+type unsupportedWarnings struct {
+	serviceAnnotations  map[string]bool
+	endpointAnnotations map[string]bool
+	dedicatedPod        map[string]bool
+	imageComponent      map[string]bool
+	customComponent     map[string]bool
+	volumeSize          map[string]bool
+	eventPostStop       map[string]bool
+	eventPreStop        map[string]bool
+}
+
+// Returns an initialized unsupportedWarnings struct
+func newUnsupportedWarnings() *unsupportedWarnings {
+	return &unsupportedWarnings{
+		serviceAnnotations:  make(map[string]bool),
+		endpointAnnotations: make(map[string]bool),
+		dedicatedPod:        make(map[string]bool),
+		imageComponent:      make(map[string]bool),
+		customComponent:     make(map[string]bool),
+		volumeSize:          make(map[string]bool),
+		eventPostStop:       make(map[string]bool),
+		eventPreStop:        make(map[string]bool),
+	}
+}
+
+func checkUnsupportedFeatures(devWorkspaceSpec dwv2.DevWorkspaceTemplateSpec) (warnings *unsupportedWarnings) {
+	warnings = newUnsupportedWarnings()
+	for _, component := range devWorkspaceSpec.Components {
+		if component.Container != nil {
+			if component.Container.Annotation != nil && component.Container.Annotation.Service != nil {
+				warnings.serviceAnnotations[component.Name] = true
+			}
+			for _, endpoint := range component.Container.Endpoints {
+				if endpoint.Annotations != nil {
+					warnings.endpointAnnotations[component.Name] = true
+				}
+			}
+			if component.Container.DedicatedPod != nil && *component.Container.DedicatedPod {
+				warnings.dedicatedPod[component.Name] = true
+			}
+		}
+
+		if component.Image != nil {
+			warnings.imageComponent[component.Name] = true
+		}
+		if component.Custom != nil {
+			warnings.customComponent[component.Name] = true
+		}
+		if component.Volume != nil && component.Volume.Size != "" {
+			warnings.volumeSize[component.Name] = true
+		}
+	}
+	if devWorkspaceSpec.Events != nil {
+		if len(devWorkspaceSpec.Events.PostStop) > 0 {
+			for _, event := range devWorkspaceSpec.Events.PostStop {
+				warnings.eventPostStop[event] = true
+			}
+		}
+		if len(devWorkspaceSpec.Events.PreStop) > 0 {
+			for _, event := range devWorkspaceSpec.Events.PreStop {
+				warnings.eventPreStop[event] = true
+			}
+
+		}
+	}
+	return warnings
+}
+
+func unsupportedWarningsPresent(warnings *unsupportedWarnings) bool {
+	return len(warnings.serviceAnnotations) > 0 ||
+		len(warnings.endpointAnnotations) > 0 ||
+		len(warnings.dedicatedPod) > 0 ||
+		len(warnings.imageComponent) > 0 ||
+		len(warnings.customComponent) > 0 ||
+		len(warnings.volumeSize) > 0 ||
+		len(warnings.eventPostStop) > 0 ||
+		len(warnings.eventPreStop) > 0
+}
+
+func formatUnsupportedFeaturesWarning(warnings *unsupportedWarnings) string {
+	var msg []string
+
+	// Returns warning names in sorted order
+	getWarningNames := func(warningsMap map[string]bool) []string {
+		var warningNames []string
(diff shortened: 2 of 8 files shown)


webhook/workspace/handler/testdata/warning/add-unsupported-features-when-some-present.yaml
```diff
@@ -0,0 +1,37 @@
...
+          autoBuild: true
+          dockerfile:
+            uri: docker/Dockerfile
+            args:
+              - 'MY_ENV=/home/path'
+            buildContext: .
+            rootRequired: false
+
+output:
+  expectedWarning: "Unsupported Devfile features are present in this workspace. The following features will have no effect: components[].image, used by components: image-component; components[].volume.size, used by components: projects"
```
AObuchow, line 36: It might be nicer to change the warning message in the `kubectl edit ..` case to something different than the `kubectl apply ..` case.

Like "Unsupported Devfile features _were added_ to this workspace..."

webhook/workspace/handler/warning.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	if len(warnings.eventPostStop) > 0 {
+		eventPostStopMsg := "events.postStop: " + strings.Join(warnings.eventPostStop, ", ")
+		msg = append(msg, eventPostStopMsg)
+	}
+	if len(warnings.eventPreStop) > 0 {
+		eventPreStopMsg := "events.preStop: " + strings.Join(warnings.eventPreStop, ", ")
+		msg = append(msg, eventPreStopMsg)
+	}
+	// TODO: Improve warning message?
```
amisevsk, line 121: Message is fine

webhook/workspace/handler/warning.go
```diff
@@ -0,0 +1,162 @@
...
+type unsupportedWarnings struct {
+	serviceAnnotations  []string
+	endpointAnnotations []string
+	dedicatedPod        []string
+	imageComponent      []string
+	customComponent     []string
+	volumeSize          []string
+	eventPostStop       []string
+	eventPreStop        []string
+}
```
amisevsk, line 34: Since these are basically sets (should have no duplicate component names in any field), you could consider using `map[string]bool` instead of `[]string` here. This would simplify their usage:

* Adding an element: `serviceAnnotations[component.Name] = true`
* Checking membership: `if serviceAnnotations[component.Name] {/*there is a warning for component.Name*/}
* Doing set subtraction (e.g. in `checkForAddedUnsupportedFeatures` below):
    ```go
    func getNewElements(old, new map) []string{
      var newNames []string
      for name := range new {
        if !old[name] {
          newNames = append(newNames, name)
        }
      }
      return newNames
    }
    ```

webhook/workspace/handler/warning.go
```diff
@@ -0,0 +1,162 @@
...
+	// TODO: Improve warning message?
+	return fmt.Sprintf("Unsupported Devfile features are present in this workspace. The following features will have no effect: %s", strings.Join(msg, "; "))
+}
+
+// Returns unsupported feature warnings that are present in the new workspace
+// but not present in the old workspace
+func checkForAddedUnsupportedFeatures(oldWksp, newWksp *dwv2.DevWorkspace) *unsupportedWarnings {
+	oldWarnings := checkUnsupportedFeatures(oldWksp.Spec.Template)
+	newWarnings := checkUnsupportedFeatures(newWksp.Spec.Template)
+	additionalWarnings := &unsupportedWarnings{}
```
amisevsk, line 130: "Additional" suggests supplementary instead of the diff between the two.
```suggestion
	addedWarnings := &unsupportedWarnings{}
```

webhook/workspace/handler/warning_test.go
```diff
@@ -0,0 +1,100 @@
...
+	Output testOutput `json:"output,omitempty"`
+}
+
+type testInput struct {
+	OldWorkspace *dwv2.DevWorkspaceTemplateSpec `json:"oldWorkspace,omitempty"`
+	NewWorkspace *dwv2.DevWorkspaceTemplateSpec `json:"newWorkspace,omitempty"`
+}
+
+type testOutput struct {
+	// TODO: Maybe NewWarningsExpected is a better name?
```
amisevsk, line 40: `NewWarningsPresent` is fine.

webhook/workspace/handler/workspace.go
```diff
@@ -155,14 +165,22 @@ func (h *WebhookHandler) MutateWorkspaceV1alpha2OnUpdate(ctx context.Context, re
...
 	if newCreator != oldCreator {
 		return admission.Denied(fmt.Sprintf("label '%s' is assigned once devworkspace is created and is immutable", constants.DevWorkspaceCreatorLabel))
 	}
 
-	return admission.Allowed("new workspace has the same devworkspace as old one")
+	response := admission.Allowed("new workspace has the same devworkspace as old one")
+	if warnings != "" {
+		response.WithWarnings(warnings)
+	}
+	return response
```
amisevsk, line 183: Not really important, but if the workspaces are identical here (not sure if that's _actually_ what we're checking...) then warnings should always be empty, right?

webhook/workspace/handler/warning.go
AObuchow: Submitted a fixup to use a `map[string]bool` :+1:

webhook/workspace/handler/workspace.go
AObuchow: The case we're checking here is "No change was made that was denied or required a patch, but features that are unsupported (and cause warnings) may have been added".

We need to modify all the cases where a response is returned because the warnings need to be added _in addition_ to the final response. If we were to immediately return a response with warnings (`admission.allowed().withWarnings(warnings)`), then we could allow other changes to be made that we should be denying (e.g. changing the devworkspace creator label)

AObuchow: I just realized why this looks weird, my bad :facepalm:.
This should be changed to something like:
```go
	if warnings != "" {
		return admission.Allowed("").WithWarnings(warnings)
	}
	return admission.Allowed("new workspace has the same devworkspace as old one")
```

It doesn't make sense to say "new workspace has the same devworkspace as old one" when there are warnings created because of the new devworkspace being changed to have additional, unsupported features.

Though I'm not sure if responding with a blank message, i.e. `admission.Allowed("")` is correct.

AObuchow: I updated my PR according to my comment.

It seems weird to have a blank reason/message, but checking the [implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/response.go#L30) of `admission.Allowed()` shows that [if an empty reason is given, then it is omitted from the response](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/response.go#L76).  So this seems valid.

amisevsk (approved):
LGTM

webhook/workspace/handler/workspace.go
AObuchow: @amisevsk yes exactly - that's what I had realized :P thanks for catching that :+1:

ibuziuk (approved):

PR #994, opened 2022-12-13
Use theia-latest instead of theia-next in samples

### LINKED ISSUE #992: theia-next samples are broken
### Description
The theia-next sample URI (`https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml`) is broken, as `next` builds of Theia are no longer being produced.

I suggest we modify the following samples to use `theia:latest` (https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/latest/devfile.yaml):
- `theia-next.yaml`
- `theia-next-contributions.yaml`
- `pod-overrides.yaml`
- `ephemeral-storage.yaml`
- `per-workspace-storage.yaml`
- `git-clone-sample.yaml`

I'm not sure if we should rename `theia-next.yaml` to `theia-latest.yaml`, only because some bug reports and PRs reference it directly (eg. https://github.com/devfile/devworkspace-operator/pull/973, https://github.com/devfile/devworkspace-operator/issues/871, https://github.com/devfile/devworkspace-operator/issues/956). It might be convenient to keep the filename the same so that commands from these issues & PRs don't break (however... "backwards compatibility" isn't really _that_ big of a deal here).  @amisevsk WDYT?

### How To Reproduce
- Install DWO on your cluster
- Apply one of the samples listed above, e.g. `kubectl apply -f ./samples/theia-next.yaml` 
- Check the workspace status: `kubectl get dw`
- See that the workspace failed to start due to a broken URI for the theia component:

```
theia-next           workspace191ffef7efb64bbc   Failed   Error processing devfile: failed to resolve component theia by URI: could not fetch file from https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml: got status 404
```
### Expected behavior
The workspace should start up normally

AObuchow:
This PR replaces usage of the theia-next devfile URI (which is now broken) with the theia-latest devfile URI for all samples that were using theia-next.

I also modified the names of the sample workspaces so that they would all be unique, preventing any conflicts when applying multiple samples at the same time. This isn't necessary (though I've found it useful in the past) and the commit for it can be removed if desired. 

Fix #992

### Is it tested? How?
1. Deploy DWO
2. `kubectl apply -f ` the following samples:
- `theia-latest.yaml`
- `theia-latest-contributions.yaml`
- `pod-overrides.yaml`
- `ephemeral-storage.yaml`
- `per-workspace-storage.yaml`
- `git-clone-sample.yaml`
3. Ensure they all start up sucessfully with `kubectl get dw -w`

### PR Checklist

### DIFF
--- a/samples/ephemeral-storage.yaml
+++ b/samples/ephemeral-storage.yaml
@@ -15,7 +15,7 @@ spec:
     components:
       - name: theia
         plugin:
-          uri: https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml
+          uri: https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/latest/devfile.yaml
           components:
             - name: theia-ide
               container:
--- a/samples/ephemeral-storage.yaml
+++ b/samples/ephemeral-storage.yaml
@@ -1,7 +1,7 @@
 kind: DevWorkspace
 apiVersion: workspace.devfile.io/v1alpha2
 metadata:
-  name: theia-next
+  name: theia-next-ephemeral
 spec:
   started: true
   template:
--- a/samples/pod-overrides.yaml
+++ b/samples/pod-overrides.yaml
@@ -1,7 +1,7 @@
 kind: DevWorkspace
 apiVersion: workspace.devfile.io/v1alpha2
 metadata:
-  name: theia-next
+  name: theia-next-pod-overrides
 spec:
   started: true
   template:
--- a/samples/pod-overrides.yaml
+++ b/samples/pod-overrides.yaml
@@ -23,7 +23,7 @@ spec:
     components:
       - name: theia
         plugin:
-          uri: https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml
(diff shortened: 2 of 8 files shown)


ibuziuk (approved):

amisevsk (approved):

AObuchow:
No idea why codecov is reporting that code coverage decreases with this PR

AObuchow:
Before I merge this, I also need to update the README since it mentions the `flattened_theia-next` sample

samples/ephemeral-storage.yaml
```diff
@@ -1,7 +1,7 @@
 kind: DevWorkspace
 apiVersion: workspace.devfile.io/v1alpha2
 metadata:
-  name: theia-next
+  name: theia-next-ephemeral
```
AObuchow, line 4: Should be called `theia-latest-ephemeral`

samples/pod-overrides.yaml
```diff
@@ -1,7 +1,7 @@
 kind: DevWorkspace
 apiVersion: workspace.devfile.io/v1alpha2
 metadata:
-  name: theia-next
+  name: theia-next-pod-overrides
```
AObuchow, line 4: Should be `theia-latest-pod-overrides`

Issue #894, opened 2022-07-18
Installation on OKD is not straight-forward (OperatorHub)


a-elie:
### Description
Is it possible to have some kind of walk-trough in the readme ?

- Default `ClusterServiceVersion` mentions `registry.redhat.io/openshift4/ose-kube-rbac-proxy:v4.xx`
In OKD, this image is not available
The installation is successfully completed using `quay.io/openshift/origin-kube-rbac-proxy:4.xx` (edited in yaml)

- `devworkspace-operator-config` have to be created in `openshift-operators` namespace
Is it possible to have a pre-filled config created automatically by the operator ?
Related : https://github.com/devfile/devworkspace-operator/issues/843

Thanks !

### Additional context
devworkspace-operator.v0.16.0-dev.7

amisevsk:
Hi @a-elie, thanks for pointing this out. Currently, our OperatorHub install targets OpenShift and assumes image pull secrets are present (allowing pulling the `ose-kube-rbac-proxy` image). I'll look into making the process there on OKD simpler -- the two images are the same, so it's just a matter of selecting the correct one.

As for the configuration CR, on OKD it shouldn't be strictly necessary and is only required if you need to override default values. We can try to create a blank one on startup to make editing it easier though (it's tedious to have to create the object with a specific name). 

OKD was a blindspot for me until now, I agree some improved documentation is needed.

a-elie:
> it's tedious to have to create the object with a specific name

This !

> I'll look into making the process there on OKD simpler

Thank you @amisevsk, updating is quite difficult with the current edit-by-hand process

cgruver:
I have also encountered this issue while working with upstream Che on OKD.

Since all of the other images needed by the operator are restricted by a pull secret, why not just use `kube-rbac-proxy` from `quay.io`?    Alternatively, you could move `ose-kube-rbac-proxy:v4.xx` to a space in `registry.redhat.io/` that is not restricted by a pull secret.  Like the UBI images.

debovema:
Having the same issue here.
@a-elie :  what did you edit to make it work with quay.io image please ?

cgruver:
You can fix it from the CLI, or from the console.

The key is to edit the `ClusterServiceVersion` `devworkspace-operator.v0.17.0` in the `openshift-operators` namespace.

[...]

Issue #1014, opened 2023-01-06
Update downstream build process to replace kube-rbac-image used in controller deployment


amisevsk:
### Description
The downstream build process (for the version of this operator that is pushed to the Red Hat catalog needs to be updated to use the current ose-kube-rbac-proxy image.

PR #1015, opened 2023-01-06
Update OpenShift deployments to use kube-rbac-proxy from kubebuilder

amisevsk:
### What does this PR do?
Updates deployment template and catalog to use `gcr.io/kubebuilder/kube-rbac-proxy` instead of `registry.redhat.io/openshift4/ose-kube-rbac-proxy` as the latter image requires an OpenShift subscription. Also updates the kube-rbac-proxy image used to v0.13.1 (latest). The image is used only to secure the operator metrics endpoint.

I also considered using https://quay.io/repository/openshift/origin-kube-rbac-proxy?tab=info, however it's unclear where this image is used elsewhere.

We will need to adapt the downstream process to continue using the correct image -- created issue https://github.com/devfile/devworkspace-operator/issues/1014 to track this work.

Closes https://github.com/devfile/devworkspace-operator/issues/894

### Is it tested? How?
To test, deploy DWO from changes (i.e. using the new image) and verify that metrics work as expected.

### PR Checklist

### DIFF
--- a/deploy/bundle/manifests/devworkspace-operator.clusterserviceversion.yaml
+++ b/deploy/bundle/manifests/devworkspace-operator.clusterserviceversion.yaml
@@ -340,7 +340,7 @@ spec:
                 - name: RELATED_IMAGE_devworkspace_webhook_server
                   value: quay.io/devfile/devworkspace-controller:next
                 - name: RELATED_IMAGE_kube_rbac_proxy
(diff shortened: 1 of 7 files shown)


AObuchow (approved):
Works as expected and change looks good to me :+1: 

- The rbac proxy image was updated in the deployment on openshift
- Was able to `curl` the metrics as kube:admin
- Was not able to `curl` the metrics as another user with insufficient priviledges

Issue #914, opened 2022-09-01
DevWorkspace Operator should detect changes to automount volumes/secrets/configmaps and update running workspaces


amisevsk:
### Description
The DevWorkspace Operator processes automount volumes/configmaps/secrets only when reconciling a DevWorkspace. However, since these objects are meant to be shared across workspaces, they are not owned by any one workspace and so editing them does not trigger reconciles. 

This means that if e.g. a new automount configmap is added to a namespace, it will only be propagated to DevWorkspaces on the next reconcile (which could be a while if there are no events).

Instead, the DevWorkspace Operator should watch automount resources and trigger reconciles for any running workspaces in their namespace to ensure they are up-to-date.

### Additional context
This issue also means that updating/refreshing personal access tokens store in a `controller.devfile.io/git-credential` secret will not be propagated to the merged credentials secret until any DevWorkspace is reconciled.

PR #1017, opened 2023-01-13
Detect changes in automount resources and queue reconciles for started workspaces

amisevsk:
### What does this PR do?
Watches namespaces for changes to automounted resources (configmaps, secrets, PVCs -- including git credentials secrets) and queues reconciles for any started workspaces in the same namespace.

There is a new controller test to verify this functionality. To avoid erroneous passing tests, I had to reduce the timeout, which could potentially cause flakiness in the future.

Closes https://github.com/devfile/devworkspace-operator/issues/914

### Is it tested? How?
To test manually:
1. Create a workspace and wait for it to be running. Verify that the controller is no longer reconciling that workspace (i.e. there are no new events being triggered)
2. Create an automount resource and check that it is picked up and mounted to the workspace (causing it to restart)

To verify the specific case in the issue:
1. Create a workspace and wait for it to be running. Verify that the controller is no longer reconciling that workspace (i.e. there are no new events being triggered)
2. Create a git credentials secret (does not need to be valid). Verify that workspace is immediately restarted to include new volume
3. Create a second git-credentials secret (or modify the original) and verify that change is propagated to the `devworkspace-merged-git-credentials` secret immediately.

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -700,69 +700,6 @@ func (r *DevWorkspaceReconciler) getWorkspaceId(ctx context.Context, workspace *
 	}
 }
 
-// Mapping the pod to the devworkspace
-func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
-	labels := obj.GetLabels()
-	if _, ok := labels[constants.DevWorkspaceNameLabel]; !ok {
-		return []reconcile.Request{}
-	}
-
-	//If the dewworkspace label does not exist, do no reconcile
-	if _, ok := labels[constants.DevWorkspaceIDLabel]; !ok {
-		return []reconcile.Request{}
-	}
-
-	return []reconcile.Request{
-		{
-			NamespacedName: types.NamespacedName{
-				Name:      labels[constants.DevWorkspaceNameLabel],
-				Namespace: obj.GetNamespace(),
-			},
-		},
-	}
-}
-
-func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Request {
-	// Check if PVC is owned by a DevWorkspace (per-workspace storage case)
-	for _, ownerref := range obj.GetOwnerReferences() {
-		if ownerref.Kind != "DevWorkspace" {
-			continue
-		}
-		return []reconcile.Request{
-			{
-				NamespacedName: types.NamespacedName{
-					Name:      ownerref.Name,
-					Namespace: obj.GetNamespace(),
-				},
-			},
-		}
-	}
-
-	// TODO: Label PVCs used for workspace storage so that they can be cleaned up if non-default name is used.
-	// Otherwise, check if common PVC is deleted to make sure all DevWorkspaces see it happen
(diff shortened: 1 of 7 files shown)


dkwon17:
I was able to go through the testing steps successfully, but I have a couple of questions:

> 3. Create a second git-credentials secret (or modify the original) and verify that change is propagated to the devworkspace-merged-git-credentials secret immediately.

The devworkspace-merged-git-credentials secret was updated immediately, but the workspace did not restart, is this expected?

Also, I was able to test this PR by mounting a PVC:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-pvc
  labels:
    controller.devfile.io/mount-to-devworkspace: 'true'
  annotations:
    controller.devfile.io/mount-path: /home/user/my-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi 
```

After starting a workspace, I updated the PVC by changing the mount path, which automatically restarted my running workspaces. Just wanted to highlight this because changing a PVC seems to immediately restart workspaces, but changes in  configmap/secret's data does not, is this expected?

AObuchow:
Things seem to work well in my testing.

The git credentials case worked as I expected: I created a git credentials secret, then modified it to change the credentials, and the mounted credential file was updated without restarting the workspace (which I imagine is intended?). I also saw the following in the controller logs:

```json
{
   "level":"info",
   "ts":1673904579.3065548,
   "logger":"controllers.DevWorkspace",
   "msg":"syncing merged git credentials secret: v1.Secret devworkspace-merged-git-credentials is not ready: Updated object",
   "Request.Namespace":"devworkspace-controller",
   "Request.Name":"plain-devworkspace",
   "devworkspace_id":"workspace378ff1947b854bef"
}
```

I also tried creating a configmap with the `controller.devfile.io/mount-to-devworkspace` label. When it was created, the workspace was restarted, and I could see that the configmap was mounted within the workspace's filesystem.

One small thing to note (that I think is outside the scope of this issue):
My configmap set the per-workspace PVC size to a non-default value, and though the workspace restarted, the PVC size was not modified. 

I think this is more related to the way the [per-workspace pvc size configmaps are handled](https://github.com/devfile/devworkspace-operator/pull/837). I guess in the per-workspace storage provisioner, there's no check to see if the existing PVC for the workspace is the correct size. Deleting the PVC could result in data loss, so the current behaviour is probably for the better until https://github.com/devfile/devworkspace-operator/issues/875 is resolved.

AObuchow (approved):
Code looks good to me & great work on adding more controller tests :sunglasses: :pray:

amisevsk:
> The devworkspace-merged-git-credentials secret was updated immediately, but the workspace did not restart, is this expected?

This is expected -- the changes to the merged secret result in no changes to the workspace pod's spec, and so the pod is not restarted. Other changes (e.g. adding an automount PVC) change the pod spec (to mount the new object) and so the pod must be restarted to pick up the changes.

For the merged-git-credentials secret in specific, restarting the pod is not necessary. As the secret is mounted as files in the workspace, any changes to the secret's data will eventually be propagated down into the mounted file within the pod. This allows for rotating PATs automatically without requiring workspace restarts.

dkwon17 (approved):

Issue #947, opened 2022-10-06
Compute required size for per-workspace PVC when all volumes have sizes specified.


amisevsk:
### Description
Devfile volume components support specifying a size parameter, which is currently ignored. This means that the configured default value for PVC size is always used, even if a DevWorkspace specifies the size of all volumes. 

Instead, DWO should have a mechanism for computing PVC size from all volumes in the DevWorkspace, and use that if the request is larger than the default. This would enable supporting workspaces that require a lot of storage without increasing the default request for _all_ workspaces.

AObuchow:
> DWO should have a mechanism for computing PVC size from all volumes in the DevWorkspace, and **use that if the request is larger than the default.**

Is it a strict requirement to use the calculated size only if it's larger than the default size? There's a tradeoff: save space if you don't need to default 5Gi, or ensure you have always headroom if the calculated size is smaller than the default. Not sure which is better.

In my mind, having headroom seems better, but it is up for debate. 

CC: @ibuziuk @l0rd

amisevsk:
Given what we've found with per-workspace PVCs, using whatever is defined and ignoring defaults is probably the right choice. The main concern would be once-working devfiles no longer working right, but its easily remedied (request as much memory as you actually need).

PR #1020, opened 2023-01-16
compute per-workspace PVC size

AObuchow:
### What does this PR do?
When using the per-workspace storage strategy, it is now possible for a workspace to use a PVC other than the default PVC size set in the DWOC. The non-default PVC size is calculated by summing the size of volume components in the devworkspace.

The following rules determine whether the computed PVC size will be used:

- If the PVC size is specified in a configmap, it will take priority and the computed PVC size will not be used.

- If all volumes in a devworkspace specify their size, the computed PVC size will be used.

- If all volumes in a devworkspace either specify their size or are ephemeral, the computed PVC size will be used.

- If at least one volume in a devworkspace specifies its size, and the computed PVC size is greater than the default per-workspace PVC size, the computed PVC size will be used.

In addition to this modification to the per-workspace storage strategy, the webhook warning for unsupported features was modified to stop giving warnings when `volume.size` is used. Additionally, the unsupported devfile API docs were modified to remove the mention of `volume.size`.

#947

### Is it tested? How?
Some automated tests were added. Below are some manual test cases.
Before testing, do a `make docker install` while checking out the branch from this PR.
**Note:** when applying the following devworkspaces, you should notice that no webhook warning is given anymore.

#### DevWorkspace with all volume sizes defined
`kubectl apply -f ...` the following devworkspace:

```YAML
kind: DevWorkspace
metadata:
  name: plain-devworkspace-with-volume
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
      - name: test-volume
        volume:
          size: 1Gi
```

Then check that the PVC provisioned for the workspace has a size of `1Gi`.

#### DevWorkspace with all volume sizes defined except one ephemeral volume

`kubectl apply -f ...` the following devworkspace:

```YAML
kind: DevWorkspace
metadata:
  name: plain-devworkspace-with-ephemeral-volume
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
      - name: test-volume
        volume:
          size: 512Mi
      - name: test-volume-2
        volume:
          size: 512Mi
      - name: ephemeral-volume
        volume:
          ephemeral: true
```

Then check that the PVC provisioned for the workspace has a size of `1Gi`.

#### DevWorkspace where not all volume sizes are defined and computed PVC size is smaller than default

`kubectl apply -f ...` the following devworkspace:

```YAML
kind: DevWorkspace
metadata:
  name: plain-devworkspace-with-volume-size-not-defined
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
      - name: test-volume
        volume:
          size: 512Mi
      - name: test-volume-2
        volume:
          size: 1Gi
      - name: test-volume-3
        volume: {}
```

Then check that the PVC provisioned for the workspace has the **default per-workspace PVC size** of `5Gi`.

#### DevWorkspace where not all volume sizes are defined and computed PVC size is larger than default

`kubectl apply -f ...` the following devworkspace:

```YAML
kind: DevWorkspace
metadata:
  name: plain-dw-missing-volume-size-larger-than-default
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: web-terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
      - name: test-volume
        volume:
          size: 1Gi
      - name: test-volume-2
        volume:
          size: 7Gi
      - name: test-volume-3
        volume: {}
```

Then check that the PVC provisioned for the workspace has a size of `8Gi`.

### PR Checklist

### DIFF
--- a/pkg/provision/storage/perWorkspaceStorage.go
+++ b/pkg/provision/storage/perWorkspaceStorage.go
@@ -155,23 +155,60 @@ func (p *PerWorkspaceStorageProvisioner) rewriteContainerVolumeMounts(workspaceI
 	return nil
 }
 
+func getPVCSize(workspace *common.DevWorkspaceWithConfig, namespacedConfig *nsconfig.NamespacedConfig) (*resource.Quantity, error) {
+	defaultPVCSize := *workspace.Config.Workspace.DefaultStorageSize.PerWorkspace
+
+	// Calculate required PVC size based on workspace volumes
+	allVolumeSizesDefined := true
+	requiredPVCSize := resource.NewQuantity(0, resource.BinarySI)
+	for _, component := range workspace.Spec.Template.Components {
+		if component.Volume != nil {
+			if isEphemeral(component.Volume) {
+				continue
+			}
+
+			if component.Volume.Size == "" {
+				allVolumeSizesDefined = false
+				break
+			}
+
+			volumeSize, err := resource.ParseQuantity(component.Volume.Size)
+			if err != nil {
+				return nil, err
+			}
+			requiredPVCSize.Add(volumeSize)
+		}
+	}
+
+	// Use the calculated PVC size if it's greater than default PVC size
+	if allVolumeSizesDefined || requiredPVCSize.Cmp(defaultPVCSize) == 1 {
+		return requiredPVCSize, nil
+	}
+
+	if namespacedConfig != nil && namespacedConfig.PerWorkspacePVCSize != "" {
+		pvcSize, err := resource.ParseQuantity(namespacedConfig.PerWorkspacePVCSize)
+		if err != nil {
+			return nil, err
+		}
+		return &pvcSize, nil
+	}
+
+	return &defaultPVCSize, nil
+}
+
 func syncPerWorkspacePVC(workspace *common.DevWorkspaceWithConfig, clusterAPI sync.ClusterAPI) (*corev1.PersistentVolumeClaim, error) {
 	namespacedConfig, err := nsconfig.ReadNamespacedConfig(workspace.Namespace, clusterAPI)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read namespace-specific configuration: %w", err)
 	}
-	// TODO: Determine the storage size that is needed by iterating through workspace volumes,
-	// adding the sizes specified and figuring out overrides/defaults
-	pvcSize := *workspace.Config.Workspace.DefaultStorageSize.PerWorkspace
-	if namespacedConfig != nil && namespacedConfig.PerWorkspacePVCSize != "" {
-		pvcSize, err = resource.ParseQuantity(namespacedConfig.PerWorkspacePVCSize)
-		if err != nil {
-			return nil, err
-		}
+
+	pvcSize, err := getPVCSize(workspace, namespacedConfig)
+	if err != nil {
+		return nil, err
 	}
 
 	storageClass := workspace.Config.Workspace.StorageClassName
-	pvc, err := getPVCSpec(common.PerWorkspacePVCName(workspace.Status.DevWorkspaceId), workspace.Namespace, storageClass, pvcSize)
+	pvc, err := getPVCSpec(common.PerWorkspacePVCName(workspace.Status.DevWorkspaceId), workspace.Namespace, storageClass, *pvcSize)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/provision/storage/testdata/perWorkspace-storage/calculates-pvc-size-ignoring-ephemeral-volumes.yaml
+++ b/pkg/provision/storage/testdata/perWorkspace-storage/calculates-pvc-size-ignoring-ephemeral-volumes.yaml
@@ -1,0 +1,76 @@
+name: "Calculates PVC size when all (non-ephemeral) volumes define their size"
+
+input:
+  devworkspaceId: "test-workspaceid"
+  podAdditions:
+    containers:
+      - name: testing-container-1
+        image: testing-image
+        volumeMounts:
+          - name: "projects"
+            mountPath: "/projects-mountpath"
+          - name: "volume-1"
+            mountPath: "/test-1"
+    initContainers:
+      - name: testing-initContainer-1
+        image: testing-image
+        volumeMounts:
+          - name: "plugins"
+            mountPath: "/plugins"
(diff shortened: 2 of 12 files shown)


AObuchow:
> If the PVC size is specified in a configmap, it will take priority and the computed PVC size will not be used.

@amisevsk this was kind of an arbitrary decision and maybe should be changed so that the computed PVC size takes priority over the config map.

pkg/provision/storage/testdata/perWorkspace-storage/calculates-pvc-size-ignoring-ephemeral-volumes.yaml
```diff
@@ -0,0 +1,76 @@
...
+            mountPath: "/plugins"
+          - name: storage-test-workspaceid
+            subPath: "volume-1"
+            mountPath: "/test-3"
+    volumes:
+      - name: storage-test-workspaceid
+        persistentVolumeClaim:
+          claimName: storage-test-workspaceid
+      - name: volume-5
+        emptyDir: {}
```
amisevsk, line 76: Trailing newlines :)

pkg/provision/storage/testdata/perWorkspace-storage/use-default-size-when-calculated-size-smaller-than-default.yaml
```diff
@@ -0,0 +1,70 @@
...
+          size: 1Gi
+      - name: volume-3
+        volume:
+          size: 512Mi
+      - name: volume-4
+        volume:
+          size: 248Mi
+      - name: plugins
+        volume: {}
+      # Calculated PVC size is 2760Mi, which is smaller than default 5Gi, so default should be used
```
amisevsk, line 43: Two issues here:

1. The output does not define a size, so the computed size is not checked for this test (we only check the returned PVC size if it's not nil)
2. If the devfile requests less storage than the configured default, we _do_ want to use the smaller amount. This would allow, for example, attaching a small amount of storage to a web terminal.

I think this test file should be dropped, as the code as written currently behaves according to point 2 above and would fail this test.

amisevsk (approved):
LGTM, well done. Couple of minor fixups.

pkg/provision/storage/testdata/perWorkspace-storage/use-default-size-when-calculated-size-smaller-than-default.yaml
AObuchow: >     1. The output does not define a size, so the computed size is not checked for this test (we only check the returned PVC size if it's not nil)

Unless I'm mistaken, I added [a check](https://github.com/devfile/devworkspace-operator/pull/1020/files#diff-6074799d5832e0da6614320e8779bcacb5c0a169069bee9820670c970f36b636R91) so that if the output does not define a PVC size _and_ we're not expecting an error, we check that the retrieved PVC size is the same as the default PVC size. 

>     2. If the devfile requests less storage than the configured default, we _do_ want to use the smaller amount. This would allow, for example, attaching a small amount of storage to a web terminal.

I'll make the appropriate change, as I forgot that you had [mentioned that](https://github.com/devfile/devworkspace-operator/issues/947#issuecomment-1380570915). Currently, you only can use a smaller amount if all volumes have their size defined (ignoring ephemeral volumes). 

> I think this test file should be dropped, as the code as written currently behaves according to point 2 above and would fail this test.

:+1:

AObuchow: @amisevsk to be clear: if not all volume sizes are specified (but at least one volume size is) and the computed PVC size is smaller than the default, we want to use the computed PVC size?

amisevsk: Ahh I see, I misread the test. Nevermind!

amisevsk:
>> If the PVC size is specified in a configmap, it will take priority and the computed PVC size will not be used.
>
> @amisevsk this was kind of an arbitrary decision and maybe should be changed so that the computed PVC size takes priority over the config map.

I think the configmap is a holdover from long ago -- it shouldn't override here and we should probably deprecate it TBH.

amisevsk (approved):

AObuchow:
> > > If the PVC size is specified in a configmap, it will take priority and the computed PVC size will not be used.
> > 
> > 
> > @amisevsk this was kind of an arbitrary decision and maybe should be changed so that the computed PVC size takes priority over the config map.
> 
> I think the configmap is a holdover from long ago -- it shouldn't override here and we should probably deprecate it TBH.

Ok sounds good, I think it makes sense then for me to lower its priority over the computed PVC size. Will make the modification and then this should be good to merge.

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -155,23 +155,60 @@ func (p *PerWorkspaceStorageProvisioner) rewriteContainerVolumeMounts(workspaceI
...
+
+	// Use the calculated PVC size if it's greater than default PVC size
+	if allVolumeSizesDefined || requiredPVCSize.Cmp(defaultPVCSize) == 1 {
+		return requiredPVCSize, nil
 	}
-	// TODO: Determine the storage size that is needed by iterating through workspace volumes,
-	// adding the sizes specified and figuring out overrides/defaults
-	pvcSize := *workspace.Config.Workspace.DefaultStorageSize.PerWorkspace
+
 	if namespacedConfig != nil && namespacedConfig.PerWorkspacePVCSize != "" {
```
AObuchow, line 188: My last force push just moved this block of code so that the computed PVC size takes priority over the namespaced config PVC size.

ibuziuk (approved):

PR #1013, opened 2023-01-05
feat: implicit container contributions

AObuchow:
### What does this PR do?
This PR introduces 3 changes to the container contribution mechanism:

1. The `controller.devfile.io/merge-contribution` component attribute is no longer required in order to define a target for container contributions. The first **container** component in a DevWorkspace (that is not imported by a plugin or parent DevWorkspace) is automatically selected as a merge target if the `controller.devfile.io/container-contribution` is used on at least one other DevWorkspace component. However, if the `controller.devfile.io/merge-contribution` attribute is set to `true` for a container component, it will be selected as a target for container contributions.

2. Users can opt out a container component from being implicitly selected as a container contribution merge target by setting the `controller.devfile.io/merge-contribution` attribute to false.

3. If a DevWorkspace has multiple container components with the `controller.devfile.io/merge-contribution` attribute set to true, the workspace fails. The webhook was extended to prevent the creation of a DevWorkspace that has multiple container components with the `controller.devfile.io/merge-contribution` attribute set to true, and to prevent updating a DevWorkspace to have multiple container components with the `merge-contribution` attribute set to true.

### What issues does this PR fix or reference?
Fix #993

#### Automated tests
Automated tests were added to check that implicit container contributions (with the `spec.contributions` field and the attribute) work as expected; these are based off of the tests previously created in https://github.com/devfile/devworkspace-operator/pull/844 and https://github.com/devfile/devworkspace-operator/pull/939. A test was also added to ensure that DevWorkspaces with multiple container components with the   `controller.devfile.io/merge-contribution: true` attribute will fail. 

#### Manual testing
First, checkout the changes from this PR and install DWO onto your cluster, e.g set your DWO_IMG environment variable to point to your quay repo and `make docker install` on OpenShift.

The following DevWorkspaces were based off of the [Che-Theia + Golang](https://gist.githubusercontent.com/amisevsk/46da73704a126d2324e5c633ff077b8c/raw/41dee8cbe9531363ffd8975982bdc1113d7b58c4/theia-golang.devworkspace.yaml) example from https://github.com/devfile/devworkspace-operator/pull/844. They should behave the same as the original example, except for the opt-out cases and multiple `merge-contribution` attribute cases.

##### Ensure implicit container contributions work

Apply the following DevWorkspace:

```YAML
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: test-contrib-theia
spec:
  started: true
  template:
    attributes:
[...]

PR #1030, opened 2023-02-01
Idea editor sample

vitaliy-guliy:
### What does this PR do?
Adds a sample with JetBrains IDEA editor

### PR Checklist

### DIFF
--- a/samples/idea-latest.yaml
+++ b/samples/idea-latest.yaml
@@ -1,0 +1,19 @@
+kind: DevWorkspace
+apiVersion: workspace.devfile.io/v1alpha2
+metadata:
+  name: idea-latest
+spec:
+  started: true
+  template:
+    projects:
+      - name: python-hello-world
+        git:
+          remotes:
+            origin: "https://github.com/che-samples/python-hello-world.git"
+    components:
+      - name: dev
+        container:
+          image: quay.io/devfile/universal-developer-image:latest
+  contributions:
+    - name: che-idea
+      uri: https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-idea/latest/devfile.yaml


samples/idea-latest.yaml
```diff
@@ -0,0 +1,36 @@
+kind: DevWorkspace
+apiVersion: workspace.devfile.io/v1alpha2
+metadata:
+  name: idea-latest
+spec:
+  started: false
```
amisevsk, line 6: Should be 
```suggestion
  started: true
```
to avoid having to patch the workspace to start it

samples/idea-latest.yaml
```diff
@@ -0,0 +1,36 @@
...
+  template:
+    projects:
+      - name: web-nodejs-sample
+        git:
+          remotes:
+            origin: "https://github.com/vitaliy-guliy/web-nodejs-sample.git"
+            # origin: "https://github.com/che-samples/web-nodejs-sample.git"
+          checkoutFrom:
+            revision: "devfilev2"
```
amisevsk, line 15: We should consider a different sample, as this is a javascript project. Opening `app.js` in the editor leads to a warning that "IDEA provides other IDEs for editing javascript files"

samples/idea-latest.yaml
```diff
@@ -0,0 +1,36 @@
...
+    components:
+      - name: dev
+        container:
+          image: quay.io/devfile/universal-developer-image:latest
+#        attributes:
+#          controller.devfile.io/merge-contribution: true
```
amisevsk, line 21: No longer required due to https://github.com/devfile/devworkspace-operator/pull/1013

samples/idea-latest.yaml
```diff
@@ -0,0 +1,36 @@
...
+          workingDir: ${PROJECT_SOURCE}/app
+  contributions:
+    - name: che-idea
+      uri: https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-idea/latest/devfile.yaml
+#      components:
+#        - name: che-idea-runtime-description
+#          container:
+#            env:
+#              - name: CODE_HOST
+#                value: 0.0.0.0
```
amisevsk, line 36: Can be dropped for `che-idea`, as it currently listens on `0.0.0.0` by default (this may be an issue in Che, though, as it may bypass the gateway proxy).

amisevsk (approved):
Tested sample in an OpenShift 4.12 cluster, and it works as expected (with a few caveats):

* Setting `.spec.routingClass: che` (with Eclipse Che installed in the cluster) results in the workspace working as expected
* Using the `basic` routing class (default) results in the workspace failing to start. This is due to https://github.com/devfile/devworkspace-operator/issues/1031 and not a problem with this PR; using the correct URL allows accessing the editor as expected.

In my testing, the experience of actually using the editor is fairly slow, with typing into a file taking multiple seconds at times, though I think this is an issue with the editor/setup and not the devfile itself.

samples/idea-latest.yaml
vitaliy-guliy: Yes, it was `true` before, but then changed it to test the dashboard and understand why I meet the error
Screenshot from 2023-02-01 19-12-47

samples/idea-latest.yaml
vitaliy-guliy: https://github.com/che-samples/python-hello-world/tree/devfilev2 would be not bad candidate

samples/idea-latest.yaml
vitaliy-guliy: Thanks!

samples/idea-latest.yaml
vitaliy-guliy: Removed

vitaliy-guliy:
I'm not sure whether it make any sense to have a sample command here as the IDEA knows nothing about devfile-based commands.

amisevsk:
> I'm not sure whether it make any sense to have a sample command here as the IDEA knows nothing about devfile-based commands.

That's a good point -- IDEA would need some sort of plugin or similar to understand devfile commands. They should likely be dropped for now.

samples/idea-latest.yaml
```diff
@@ -0,0 +1,27 @@
+kind: DevWorkspace
+apiVersion: workspace.devfile.io/v1alpha2
+metadata:
+  name: idea-latest
+spec:
+  started: true
+  template:
+    projects:
+      - name: web-nodejs-sample
```
AObuchow, line 9: The project name should probably be changed to something like `python-hello-world`

AObuchow (commented):
I've been trying to test this on OpenShift 4.12, but unfortunately the workspace hasn't been able to start up before the timeout yet after 3 tries.

Asides from minor comment, the actual DevWorkspace YAML looks good to me.

I recommend squashing all the commits for this PR into one once it is ready for a final review/merge.

AObuchow:
> I've been trying to test this on OpenShift 4.12, but unfortunately the workspace hasn't been able to start up before the timeout yet after 3 tries.

Just realized I forgot to try this PR with https://github.com/devfile/devworkspace-operator/pull/1032 applied :sweat_smile:  will re-test

AObuchow (approved):
Was able to get the sample working after applying https://github.com/devfile/devworkspace-operator/pull/1032

Similar to @amisevsk the editor was kinda slow at first, but after a bit things sped up (still somewhat slow, but usable). I also installed the python editor and got the script to run.

amisevsk (approved):
LGTM apart from the project name

amisevsk:
Before we can merge, commits should be squashed as appropriate and all need a signed-off-by footer.

samples/idea-latest.yaml
```diff
@@ -0,0 +1,27 @@
...
+        git:
+          remotes:
+            origin: "https://github.com/che-samples/python-hello-world.git"
+          checkoutFrom:
+            revision: "devfilev2"
```
svor, line 14: don't do checkout to **devfilev2**, this branch is going to be removed https://github.com/eclipse/che/issues/21983
just use the default **main** branch

vitaliy-guliy: Done

samples/idea-latest.yaml
vitaliy-guliy: Done

PR #1035, opened 2023-02-06
[WIP] feat: Set image digest on release

tolusha:
Signed-off-by: Anatolii Bazko <abazko@redhat.com>

feat: Set image digest on release

### PR Checklist

### DIFF
--- a/build/scripts/build_index_image.sh
+++ b/build/scripts/build_index_image.sh
@@ -110,6 +110,9 @@ if [ "$RELEASE" == "true" ]; then
   # shellcheck disable=SC2016
   yq -Yi --arg replaces "$LATEST_BUGFIX" '.spec.replaces = $replaces' \
     deploy/templates/components/csv/clusterserviceversion.yaml
+
+  # Set digests instead of tags, it allows to mirror images in disconnected environment
+  . build/scripts/set_digests.sh -c deploy/templates/components/csv/clusterserviceversion.yaml
 else
   # Remove replaces field from CSVs in nightly builds since the 'next' catalog doesn't currently support
   # an upgrade path.
--- a/build/scripts/set_digests.sh
+++ b/build/scripts/set_digests.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+set -e
+
+usage () {
+	echo "Usage:   $0 -c [/path/to/csv.yaml]"
+	echo "Example: $0 -c devworkspace-operator.clusterserviceversion.yaml"
+}
+
+unset CSV_FILE
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    '-c') CSV_FILE="$2";shift 1;;
+    '--help'|'-h') usage; exit;;
+  esac
+  shift 1
+done
+
+if [[ ! ${CSV_FILE} ]]; then usage; exit 1; fi
+
+set_image_digest() {
+  local JQ_FILTER=${1}
+  local IMAGE=$(yq -r ${JQ_FILTER} "${CSV_FILE}")
+  if [[ ! ${IMAGE} =~ sha256 ]]; then
+    local DIGEST=$(skopeo inspect --tls-verify=false docker://${IMAGE} 2>/dev/null | jq -r '.Digest')
+    local IMAGE_WITH_DIGEST=$(echo "${IMAGE}" | sed -e 's/^\(.*\):[^:]*$/\1/')@"${DIGEST}"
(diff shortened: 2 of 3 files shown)


tolusha:
> Maybe a better approach is to do similar to what we do downstream, and set digests in the CSV during the bundle build, but not actually commit digests to the repo?

That is even better. Where is entrypoint to look at?

amisevsk:
The `build_index_image.sh` [script](https://github.com/devfile/devworkspace-operator/blob/main/build/scripts/build_index_image.sh) is responsible for building the bundle + index image. We could potentially look into replacing tags with digests [before the bundle is generated](https://github.com/devfile/devworkspace-operator/blob/166f43d5f0d3134296d644295aa5e8b6240fff94/build/scripts/build_index_image.sh#L118-L122) by editing the relatedImages in `./deploy/bundle/manifests/devworkspace-operator.clusterserviceversion.yaml`

build/scripts/build_index_image.sh
```diff
@@ -110,6 +110,9 @@ if [ "$RELEASE" == "true" ]; then
   # shellcheck disable=SC2016
   yq -Yi --arg replaces "$LATEST_BUGFIX" '.spec.replaces = $replaces' \
     deploy/templates/components/csv/clusterserviceversion.yaml
+
+  # Set digests instead of tags, it allows to mirror images in disconnected environment
+  . build/scripts/set_digests.sh -c deploy/templates/components/csv/clusterserviceversion.yaml
```
amisevsk, line 115: There is still an issue here, as `build_index_image.sh` is called _before_ building the controller and project clone images, so we'll still end up with empty digests for those :/

There's a fundamental issue here that I'm struggling to work around (see [this branch](https://github.com/amisevsk/devworkspace-operator/tree/use-digests-in-olm-release-wip) for today's efforts). 

* The index is built based off the files in `./olm-catalog`, which need to contain the digest of the controller image (i.e. have to build controller image before committing)
* The controller image refers to the commit its built from in its version information (i.e. have to build controller image after committing)

I'm not sure using digests in the catalog is possible to do with only one repo, if we're using file-based catalogs. 

I'm trying to work out a way to do this as we intend, but it's going to be complicated.

build/scripts/set_digests.sh
```diff
@@ -0,0 +1,64 @@
...
+  i=0
+  local CONTAINERS_LENGTH=$(yq -r '.spec.install.spec.deployments[0].spec.template.spec.containers | length' ${CSV_FILE})
+  while [ "${i}" -lt "${CONTAINERS_LENGTH}" ]; do
+      # RELATED_IMAGE environment variables
+      j=0
+      local ENV_VARS_LENGTH=$(yq -r '.spec.install.spec.deployments[0].spec.template.spec.containers['"${i}"'].env | length' ${CSV_FILE})
+      while [ "${j}" -lt "${ENV_VARS_LENGTH}" ]; do
+          ENV_NAME=$(yq -r '.spec.install.spec.deployments[0].spec.template.spec.containers['"${i}"'].env['"${j}"'].name' ${CSV_FILE})
+          if [[ ${ENV_NAME} =~ ^RELATED_IMAGE_ ]]; then
+              set_image_digest ".spec.install.spec.deployments[0].spec.deploy/templates/components/csv/clusterserviceversion.yamltemplate.spec.containers[${i}].env[${j}].value"
```
amisevsk, line 43: It looks like a path accidentally got copy-pasted into the script

```suggestion
              set_image_digest ".spec.install.spec.deployments[0].spec.deploy.template.spec.containers[${i}].env[${j}].value"
```

build/scripts/build_index_image.sh
tolusha: I found two more issues here:
* References to latest [1] and nightly [2] tags are not good at all.
* Replacing tags on digests in `spec.relatedImages` section is not enough.  Container image and images in evn vas must be replaced as well.

[1] `quay.io/wto/web-terminal-tooling:latest` 
[2] `quay.io/eclipse/che-machine-exec:nightly`

I can suggest creating a dedicated issue and describe all problems we have.

amisevsk: Yes, I think a dedicated issue is warranted, this is more complicated than it appeared at first.

amisevsk: For those images, at least, they can just be dropped: https://github.com/devfile/devworkspace-operator/pull/1039

They were needed earlier on but the functionality to use those images has since been dropped.

amisevsk:
Closing this PR, issue is resolved via https://github.com/devfile/devworkspace-operator/pull/1120

Issue #1021, opened 2023-01-16
Use projected volumes for automount configmaps/secrets when they use the same mount path


amisevsk:
### Description
Currently, attempting to start a workspace in a namespace with automount configmaps/secrets that share the same mountpath results in the workspace failing to start, with message like 

> Failed to process automount resources: auto-mounted volumes from configmap 'configmap-1' and configmap 'configmap-2' have the same mount path

Instead, DWO could mount both configmaps using a projected volume.

max-cx:
sync'd to Red Hat JIRA https://issues.redhat.com/browse/CRW-4195

Issue #835, opened 2022-05-09
Collision between automount volumes results in all workspaces being unable to start.


amisevsk:
### Description
When two automounted PVCs/configmaps/secrets collide on mountPath, all workspaces in the namespace will fail to start.

### How To Reproduce
Create two configmaps with the same `mount-path` annotation.

### Expected behavior
DWO should start the workspace anyways, mounting only one of the configmaps, and set the warning status on the workspace to notify the user there is a configuration problem.

amisevsk:
Closed by https://github.com/devfile/devworkspace-operator/pull/1043

PR #1043, opened 2023-02-13
Use projected volumes for automount resources if they share a mount path

amisevsk:
### What does this PR do?
Merges automount secrets/configmaps into projected volumes if they share a mountPath, avoiding workspace failures when multiple automount resources target the same path.

There are some cases where this won't work (e.g. we can't project a PVC, or correctly handle subpath volumeMounts since they're files). An error is returned in this case.

Closes https://github.com/devfile/devworkspace-operator/issues/1021
Closes https://github.com/devfile/devworkspace-operator/issues/835

### Is it tested? How?
To test, create automount resources and ensure they are mounted as expected. For example, the following three resources should be merged into one in a DevWorkspace
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-secret-file
  labels:
    controller.devfile.io/watch-secret: 'true'
  annotations:
    controller.devfile.io/mount-as: file
    controller.devfile.io/mount-path: /tmp/test-projected/
type: Opaque
data:
  secret-data: aGVsbG8K # hello 
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-cm-1
  labels:
  annotations:
    controller.devfile.io/mount-as: file
    controller.devfile.io/mount-path: /tmp/test-projected/
data:
  configmap-2-key: "hello"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-cm-2
  labels:
  annotations:
    controller.devfile.io/mount-as: file
    controller.devfile.io/mount-path: /tmp/test-projected/
data:
  configmap-1-key: "hello"
```

Need to also test that existing automount flows works as expected (PVCs, non-colliding secrets, envFrom, etc.)

### PR Checklist

### DIFF
--- a/pkg/common/naming.go
+++ b/pkg/common/naming.go
@@ -16,6 +16,7 @@
 package common
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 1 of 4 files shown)


AObuchow (approved):
Code looks good to me, awesome work @amisevsk :) :+1:

Issue #1049, opened 2023-02-21
Checking permissions for using Kubernetes/OpenShift components fails for regular users


amisevsk:
### Description
When verifying RBAC for Kubernetes components in a DevWorkspace, DWO checks whether the user is able to work with this object in the current namespace. To avoid multiple SAR tests per request, the webhook server checks that the user can `*` the relevant resource, equivalent to
```
oc auth can-i '*' <resource> -n <namespace>
```

However, even if the user has `admin` privileges, granting all individual verbs for the resource, the SAR request appears to check if the user explictly has `*` permissions, causing the check to fail.

### How To Reproduce
1. Create a regular user in an OpenShift cluster and create a namespace for testing
2. (optional) Verify the user can use all verbs on pods: 
    ```
    for verb in create delete deletecollection get list patch update watch; do oc auth can-i $verb pods -n $NAMESPACE; done
    ```
3. (optional) Verify the user cannot `*` pods:
    ```
    oc auth can-i '*' pods -n $NAMESPACE
    ```
4. Create a DevWorkspace that has a pod component

### Expected behavior
Check should be improved to avoid this issue.

PR #1050, opened 2023-02-21
Update verbs used to check SAR for users when k8s components are used

amisevsk:
### What does this PR do?
Update SAR checks for user permissions in webhook server to check whether a user can get/create/update/delete the resource rather than checking for '*' permissions. This is required as even if the user has the admin rolebinding, they do not have '*' permissions from the perspective of the cluster.

Closes https://github.com/devfile/devworkspace-operator/issues/1049

### Is it tested? How?
Full testing would include verifying that
* Webhooks server can manage load of doing 4 SAR checks instead of 1, and this doesn't overtax the cluster API
* There's minimal risk of privilege escalation in checking only `get`, `create`, `update`, `delete` instead of `*`
However, the risks of issues with the points above is minimal: 
* DWO only performs SAR checks when the resource is created or updated to include modified k8s components, so unless the component is being modified the checks should only be performed once per DevWorkspace
* While DWO checks whether it has `*` permissions, it only performs the `get`, `create`, `update`, and `delete` verbs.

Basic testing: 
1. Set up user in an OpenShift cluster without any permissions
2. Create a namespace as the new user
3. Verify that user has `admin` permissions but not `*` permissions for pods: `oc auth can-i '*' pods -n <namespace>` should print "no"
4. Create a DevWorkspace that uses a Kubernetes component containing a pod:
    ```yaml
    kind: DevWorkspace
    metadata:
      name: plain-devworkspace
    spec:
      started: true
      routingClass: 'basic'
      template:
        components:
          - name: web-terminal
            container:
              memoryLimit: 512Mi
              mountSources: true
              command:
                - "tail"
                - "-f"
                - "/dev/null"
          - name: test-k8s-component
            kubernetes:
              deployByDefault: true
              inlined: |
                apiVersion: v1
                kind: Pod
                metadata:
                  name: test-pod
                spec:
                  containers:
                    - image: quay.io/wto/web-terminal-tooling:next
                      name: k8s-component-pod
                      command:
                        - "tail"
                        - "-f"
                        - "/dev/null"
                      ports:
                        - name: test-port
                          containerPort: 8080
    ```

### PR Checklist

### DIFF
--- a/webhook/workspace/handler/kubernetes.go
+++ b/webhook/workspace/handler/kubernetes.go
@@ -26,6 +26,13 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+var (
+	// Verbs used in SAR checks when users attempt to use a Kubernetes/OpenShift component. A longer list results in
+	// more SAR checks per request. We cannot use '*' as a verb as the SAR will check for literal '*' permissions
+	// rather than "all verbs"
+	userVerbs = []string{"get", "create", "update", "delete"}
+)
+
 func (h *WebhookHandler) validateKubernetesObjectPermissionsOnCreate(ctx context.Context, req admission.Request, wksp *dwv2.DevWorkspaceTemplateSpec) error {
 	kubeComponents := getKubeComponentsFromWorkspace(wksp)
 	for componentName, component := range kubeComponents {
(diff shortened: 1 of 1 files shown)


AObuchow (approved):
Code looks good to me.

I followed the basic testing instructions on OpenShift 4.12: 
- created a user without any permissions, and made a "test-project" namespace/project as the user
- The result of `oc auth can-i '*' pods -n test-project` was `no`
- Created the DevWorkspace that uses a Kubernetes component containing a pod
- Checked I was able to also `oc get`, `oc describe`, `oc edit` and `oc delete` the workspace

> Webhooks server can manage load of doing 4 SAR checks instead of 1, and this doesn't overtax the cluster API

I may be wrong, but isn't this being tested by the fact the [updated code](https://github.com/devfile/devworkspace-operator/pull/1050/files#diff-ab78424018d3706f07183a548ed353e57ed5e98202daa6a5eef2c33498407caeR108) is working as intended? (I didn't see any error logs regarding this on the webhook server or the controller manager)

ibuziuk (approved):

amisevsk:
> I may be wrong, but isn't this being tested by the fact the [updated code](https://github.com/devfile/devworkspace-operator/pull/1050/files#diff-ab78424018d3706f07183a548ed353e57ed5e98202daa6a5eef2c33498407caeR108) is working as intended? (I didn't see any error logs regarding this on the webhook server or the controller manager)

The question is whether this becomes a problem when there are e.g. 1000 workspaces across 1000 namespaces. The webhook server is now making 5 SAR requests (4 SAR, 1 SSAR) instead of 2. However, we're (hopefully) careful to only check for RBAC when k8s components are added/modified, so it should still be a minimal amount of checks overall.

Issue #1045, opened 2023-02-15
Support configuring rolling updates for workspace deployments


amisevsk:
### Description
DWO currently uses the `Recreate` strategy for workspace deployments. We should investigate whether configuring workspaces to instead use `Rolling` via the DevWorkspaceOperatorConfig would be useful.

### Additional context
Related issue: https://github.com/eclipse/che/issues/21974

PR #1057, opened 2023-03-01
support configuring workspace deploymentStrategy in DWOC

AObuchow:
### What does this PR do?

This PR adds a new DWOC field `config.workspace.deploymentStrategy` which allows for specifying the workspace deployment strategy in the DWOC. The possible choices are `Recreate` (Default) and `RollingUpdate`.

https://github.com/devfile/devworkspace-operator/issues/1045 & https://github.com/eclipse/che/issues/21974 

### Is it tested? How?

Edit the DWOC and set the `deploymentStrategy` field to `RollingUpdate`. Also, ensure that the only accepted options for the `deploymentStrategy` field are `Recreate` and `RollingUpdate`.

```diff
kind: DevWorkspaceOperatorConfig
apiVersion: controller.devfile.io/v1alpha1
config:
  routing:
    clusterHostSuffix: 192.168.39.117.nip.io
    defaultRoutingClass: basic
  workspace:
+    deploymentStrategy: RollingUpdate
    imagePullPolicy: Always
```

Deploy a workspace and modify a field that would change the deployment spec, for example, modify one of the container's CPU or Memory request. 

To try this out, you could apply the following devworkspace:

```YAML
kind: DevWorkspace
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryLimit: 5Gi
          memoryRequest: 256Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

Next, watch the workspace status (as it will change when the rolling update occurs): `kubectl get dw -n devworkspace-controller`

Once the workspace is running, you could then increase the `memoryRequest`:  `kubectl edit dwoc -n plain-devworkspace -n devworkspace-controller`

```diff
apiVersion: workspace.devfile.io/v1alpha2
kind: DevWorkspace
(...)
spec:
  routingClass: basic
  started: true
  template:
    components:
    - container:
        command:
        - tail
        - -f
        - /dev/null
        image: quay.io/wto/web-terminal-tooling:next
        memoryLimit: 5Gi
+        memoryRequest: 1Gi
        mountSources: true
        sourceMapping: /projects
      name: web-terminal
```

Under the workspace's deployment (which could be checked on the OpenShift or Kubernetes dashboard) you should see a second pod get created and started before the old one is terminated. 

You should also see the devworkspace status phase change from Running -> Starting -> Running:

```
$ kubectl get dw -n $NAMESPACE -w
NAME                 DEVWORKSPACE ID             PHASE      INFO
plain-devworkspace   workspacec3a273cb552340a5   Starting   Waiting for workspace deployment
plain-devworkspace   workspacec3a273cb552340a5   Running    Workspace is running
plain-devworkspace   workspacec3a273cb552340a5   Running    Workspace is running
plain-devworkspace   workspacec3a273cb552340a5   Running    Workspace is running
plain-devworkspace   workspacec3a273cb552340a5   Starting   Networking ready
plain-devworkspace   workspacec3a273cb552340a5   Starting   Waiting for workspace deployment
plain-devworkspace   workspacec3a273cb552340a5   Running    Workspace is running
```

#### Extra test case: removing the deploymentStrategy from the DWOC

1. Set the `deploymentStrategy` to `RollingUpdate` in the DWOC
2. Apply a workspace (`kubectl apply -f ./samples/plain-devworkspace -n $NAMESPACE`), ensure the deployments `spec.strategy.type` is `RollingUpdate` 
3. Remove the `deploymentStrategy` field from the DWOC
4. Trigger a reconcile on the workspace (e.g. add an annotation)
5. Check the workspace deployment, the `spec.strategy.type` field should be `Recreate`

In other words, when the `deploymentStrategy` field is absent from the DWOC, the deployment strategy will be `Recreate`

### PR Checklist

### DIFF
--- a/pkg/library/status/check.go
+++ b/pkg/library/status/check.go
@@ -51,6 +51,7 @@ var unrecoverablePodEventReasons = map[string]int32{
 
 var unrecoverableDeploymentConditionReasons = []string{
 	"FailedCreate",
+	"ProgressDeadlineExceeded",
 }
 
 func CheckDeploymentStatus(deployment *appsv1.Deployment, workspace *common.DevWorkspaceWithConfig) (ready bool) {
--- a/pkg/library/status/check.go
+++ b/pkg/library/status/check.go
@@ -53,8 +53,27 @@ var unrecoverableDeploymentConditionReasons = []string{
 	"FailedCreate",
 }
 
-func CheckDeploymentStatus(deployment *appsv1.Deployment) (ready bool) {
+func CheckDeploymentStatus(deployment *appsv1.Deployment, workspace *common.DevWorkspaceWithConfig) (ready bool) {
+	if workspace.Config.Workspace.DeploymentStrategy == appsv1.RollingUpdateDeploymentStrategyType {
+		return checkRollingUpdateDeploymentStatus(deployment)
+	}
 	return deployment.Status.ReadyReplicas > 0
+}
+
+func checkRollingUpdateDeploymentStatus(deployment *appsv1.Deployment) (ready bool) {
+	if deployment.Generation > deployment.Status.ObservedGeneration {
+		// Current deployment spec not observed by cluster
+		return false
+	}
+	if deployment.Status.UpdatedReplicas < 1 {
+		// Replica has not been updated
+		return false
+	}
+	if deployment.Status.AvailableReplicas < 1 {
+		// Updated replica is not available yet
+		return false
+	}
+	return true
 }
 
 func CheckDeploymentConditions(deployment *appsv1.Deployment) (healthy bool, errorMsg string) {
--- a/pkg/provision/workspace/deployment.go
+++ b/pkg/provision/workspace/deployment.go
@@ -93,7 +93,7 @@ func SyncDeploymentToCluster(
 		return DeploymentProvisioningStatus{ProvisioningStatus{Err: err}}
 	}
 	clusterDeployment := clusterObj.(*appsv1.Deployment)
-	deploymentReady := status.CheckDeploymentStatus(clusterDeployment)
+	deploymentReady := status.CheckDeploymentStatus(clusterDeployment, workspace)
 	if deploymentReady {
 		return DeploymentProvisioningStatus{
 			ProvisioningStatus: ProvisioningStatus{
--- a/pkg/provision/workspace/deployment.go
+++ b/pkg/provision/workspace/deployment.go
@@ -19,6 +19,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
+	"time"
 
 	"github.com/devfile/devworkspace-operator/pkg/library/overrides"
 	"github.com/devfile/devworkspace-operator/pkg/library/status"
@@ -208,6 +210,10 @@ func getSpecDeployment(
 			MaxSurge:       &constants.RollingUpdateMaximumSurge,
 		}
 	}
+	progressDeadlineSeconds, err := getProgressDeadlineSeconds(workspace.Config)
+	if err != nil {
+		return nil, err
+	}
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
@@ -223,7 +229,8 @@ func getSpecDeployment(
 					constants.DevWorkspaceIDLabel: workspace.Status.DevWorkspaceId,
 				},
 			},
-			Strategy: deploymentStrategy,
+			Strategy:                deploymentStrategy,
+			ProgressDeadlineSeconds: &progressDeadlineSeconds,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      workspace.Status.DevWorkspaceId,
@@ -306,6 +313,22 @@ func getSpecDeployment(
 	}
 
 	return deployment, nil
+}
+
+// Returns the ProgressDeadlineSeconds to use for workspace deployments as an int32.
+// The ProgressDeadLineSeconds returned is the same length of time as DWOC's workspace.ProgressTimeout.
+// Returns an error if ProgressTimeout could not be properly parsed,
+// or if the ProgressTimeout would exceed the maximum capacity of an int32
+func getProgressDeadlineSeconds(config *v1alpha1.OperatorConfiguration) (int32, error) {
+	timeout, err := time.ParseDuration(config.Workspace.ProgressTimeout)
+	if err != nil {
+		return 0, fmt.Errorf("invalid duration specified for workspace progress timeout: %w", err)
+	}
+	// Prevent overflow
+	if timeout.Seconds() > math.MaxInt32 {
+		return 0, fmt.Errorf("duration specified for workspace progress timeout is too long: %s", config.Workspace.ProgressTimeout)
+	}
+	return int32(timeout.Seconds()), nil
 }
 
 func mergePodAdditions(toMerge []v1alpha1.PodAdditions) (*v1alpha1.PodAdditions, error) {
--- a/pkg/provision/workspace/deployment.go
+++ b/pkg/provision/workspace/deployment.go
@@ -199,6 +199,16 @@ func getSpecDeployment(
 		return nil, err
 	}
 
+	deploymentStrategy := appsv1.DeploymentStrategy{
+		Type: workspace.Config.Workspace.DeploymentStrategy,
+	}
+	if workspace.Config.Workspace.DeploymentStrategy == appsv1.RollingUpdateDeploymentStrategyType {
+		deploymentStrategy.RollingUpdate = &appsv1.RollingUpdateDeployment{
+			MaxUnavailable: &constants.RollingUpdateMaxUnavailable,
+			MaxSurge:       &constants.RollingUpdateMaximumSurge,
+		}
+	}
+
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        common.DeploymentName(workspace.Status.DevWorkspaceId),
@@ -213,9 +223,7 @@ func getSpecDeployment(
 					constants.DevWorkspaceIDLabel: workspace.Status.DevWorkspaceId,
 				},
 			},
-			Strategy: appsv1.DeploymentStrategy{
-				Type: appsv1.RecreateDeploymentStrategyType,
-			},
+			Strategy: deploymentStrategy,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      workspace.Status.DevWorkspaceId,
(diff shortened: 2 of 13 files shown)


pkg/provision/workspace/deployment.go
```diff
@@ -214,9 +229,10 @@ func getSpecDeployment(
 					constants.DevWorkspaceIDLabel: workspace.Status.DevWorkspaceId,
 				},
 			},
-			Strategy: appsv1.DeploymentStrategy{
-				Type: appsv1.RecreateDeploymentStrategyType,
-			},
+			Strategy:                deploymentStrategy,
+			MinReadySeconds:         0,
+			RevisionHistoryLimit:    pointer.Int32(10),
+			ProgressDeadlineSeconds: pointer.Int32(600),
```
amisevsk, line 235: I get that these are defaults but if we're setting them directly it may be worth considering whether these values are appropriate for our use case. 10 minutes for progress is a very long time, and we don't necessarily need 10 revisions in the history (and I believe this setting would result in up to ten replicasets for each deployment)

We should also verify what a failed rollout looks like, and if that's caught as a workspace failure or handled in some way by DWO.

AObuchow: > if we're setting them directly it may be worth considering whether these values are appropriate for our use case. 10 minutes for progress is a very long time, and we don't necessarily need 10 revisions in the history (and I believe this setting would result in up to ten replicasets for each deployment)

Agreed. I initially added them because I thought they were necessary to prevent issues with the spec/cluster deployment diff, but they actually aren't. However, it's worth considering whether we should set (sane) values instead of letting them be the defaults set by Kubernetes. 

> We should also verify what a failed rollout looks like, and if that's caught as a workspace failure or handled in some way by DWO.

Agreed, will look into it.

AObuchow:
@amisevsk another issue I've encountered is fixing the bug W.R.T the workspace status saying it's ready before the rolling update has completed.  I don't think the [deployment status struct fields](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/status/check.go#L57) have enough granularity to determine when the rolling update has finished (I tried something like `deployment.Status.ReadyReplicas > 0 && deployment.Status.UnavailableReplicas == 0` and other variations). 

I think I'll have to add some extra logic for checking the RollingUpdate deployment status specifically when it's used. It's a bit tricky to debug this properly since the RollingDeployment update progresses quickly :P

AObuchow:
> @amisevsk another issue I've encountered is fixing the bug W.R.T the workspace status saying it's ready before the rolling update has completed. I don't think the [deployment status struct fields](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/status/check.go?rgh-link-date=2023-03-03T18%3A19%3A42Z#L57) have enough granularity to determine when the rolling update has finished (I tried something like `deployment.Status.ReadyReplicas > 0 && deployment.Status.UnavailableReplicas == 0` and other variations).
> 
> I think I'll have to add some extra logic for checking the RollingUpdate deployment status specifically when it's used. It's a bit tricky to debug this properly since the RollingDeployment update progresses quickly :P

This has now been addressed in [`checkRollingUpdateDeploymentStatus()`](https://github.com/devfile/devworkspace-operator/pull/1057/files#diff-2e34131f532ed47948e0f42496cfa1ef129c9f60d040ca0a80ce568cecd965a4R66)

pkg/library/status/check.go
```diff
@@ -51,12 +51,34 @@ var unrecoverablePodEventReasons = map[string]int32{
 
 var unrecoverableDeploymentConditionReasons = []string{
 	"FailedCreate",
+	"ProgressDeadlineExceeded",
```
AObuchow, line 54: @amisevsk if we catch the `ProgressDeadlineExceeded` condition as I'm doing here, we should probably also specify `ProgressDeadlineSeconds` explicitly in the deployment spec (as 10 minutes is a long time to wait to see if the deployment rollout failed). Maybe `config.workspace.progressTimeout` would be a reasonable parameter to use for this?

pkg/provision/workspace/deployment.go
AObuchow: > We should also verify what a failed rollout looks like, and if that's caught as a workspace failure or handled in some way by DWO.

Not sure if this is sufficient of a check, but I checked the case you mentioned [here](https://github.com/eclipse/che/issues/21974#issuecomment-1453826209), and after modifying the memoryRequest and memoryLimit to 9Gi, the DevWorkspace failed:

```
[aobuchow@fedora devworkspace-operator]$ kubectl get dw -n $NAMESPACE -w
NAME                 DEVWORKSPACE ID             PHASE      INFO
code-latest          workspacefcb80b330a494679   Starting   Initializing DevWorkspace
code-latest          workspacefcb80b330a494679   Starting   Storage ready
code-latest          workspacefcb80b330a494679   Starting   Preparing networking
code-latest          workspacefcb80b330a494679   Starting   Preparing services
code-latest          workspacefcb80b330a494679   Starting   Preparing services
code-latest          workspacefcb80b330a494679   Starting   Networking ready
code-latest          workspacefcb80b330a494679   Starting   Waiting for DevWorkspace ServiceAccount
code-latest          workspacefcb80b330a494679   Starting   Waiting for DevWorkspace pull secrets
code-latest          workspacefcb80b330a494679   Starting   Waiting for workspace deployment
code-latest          workspacefcb80b330a494679   Starting   Waiting for editor to start
code-latest          workspacefcb80b330a494679   Running    https://workspacefcb80b330a494679.apps.ci-ln-ch9m0rb-72292.origin-ci-int-gce.dev.rhcloud.com/che-code/

code-latest          workspacefcb80b330a494679   Starting   Networking ready
code-latest          workspacefcb80b330a494679   Starting   Waiting for workspace deployment
code-latest          workspacefcb80b330a494679   Failing    Detected unrecoverable event FailedScheduling: 0/6 nodes are available: 1 Insufficient memory, 2 node(s) had volume node affinity conflict, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }. preemption: 0/6 nodes are available: 1 No preemption victims found for incoming pod, 5 Preemption is not helpful for scheduling..
```

I still haven't tested getting the deployment rollout to not complete before  `ProgressDeadlineSeconds`

pkg/provision/storage/shared.go
```diff
@@ -198,11 +198,11 @@ func getWorkspaceVolumes(workspace *common.DevWorkspaceWithConfig) (persistent,
...
-// 1. The projects volume is not needed for the workspace (no component has mountSources: true)
-// 2. The projects volume is needed but not defined in the devfile. This is the usual case, as the projects volume
-//    is implicitly defined by mountSources
-// 3. The projects volume is explicitly defined in the workspace, as a regular volume.
-// 4. The projects volume is explicitly defined in the workspace as an ephemeral volume.
+//  1. The projects volume is not needed for the workspace (no component has mountSources: true)
+//  2. The projects volume is needed but not defined in the devfile. This is the usual case, as the projects volume
+//     is implicitly defined by mountSources
+//  3. The projects volume is explicitly defined in the workspace, as a regular volume.
+//  4. The projects volume is explicitly defined in the workspace as an ephemeral volume.
```
amisevsk, line 205: Not sure where these whitespace changes are coming from

apis/controller/v1alpha1/doc.go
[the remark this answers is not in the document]
AObuchow: Yeah, I gotta fix that, something weird is going on :facepalm: VSCode is changing the license headers, and the whitespace added to other files resulted from either `go fmt` or `make fmt`.

pkg/library/status/check.go
```diff
@@ -51,12 +51,34 @@ var unrecoverablePodEventReasons = map[string]int32{
...
+		if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
+			return false
+		}
+		if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
+			return false
+		}
+		return true
+	}
+	return false
+}
```
amisevsk, line 80: I believe this check can be simplified, as the kubectl implementation is designed to report specific information about the state of the rollout. Is it sufficient to check `updatedReplicas == availableReplicas == replicas == 1`?

```suggestion
func checkRollingUpdateDeploymentStatus(deployment *appsv1.Deployment) (ready bool) {
	if deployment.Generation > deployment.Status.ObservedGeneration {
		// Current deployment spec not observed by cluster
		return false
	}
	if deployment.Status.UpdatedReplicas < 1 {
		// Replica has not been updated
		return false
	}
	if deployment.Status.AvailableReplicas < 1 {
		// Updated replica is not available yet
		return false
	}
	if deployment.Status.Replicas != 1 {
		// Old replica is still terminating
		return false // Or true -- the new replica is indeed running and can be accessed
		             // why wait for the old one to terminate before letting the user see
								 // their workspace
	}
	return true
}
```

In particular, we don't really care if the deployment is completed from the cluster's perspective. All we care about is that the updated replica is available (we don't need to wait for termination of the old one necessarily, it _should_ be caught by events if it times out.

pkg/provision/workspace/deployment.go
```diff
@@ -300,6 +315,23 @@ func getSpecDeployment(
...
+// Returns the ProgressDeadlineSeconds to use for workspace deployments as an int32.
+// The ProgressDeadLineSeconds returned is the same length of time as DWOC's workspace.ProgressTimeout.
+// Returns an error if ProgressTimeout could not be properly parsed,
+// or if the ProgressTimeout would exceed the maximum capacity of an int32
+func getProgressDeadlineSeconds(config *v1alpha1.OperatorConfiguration) (int32, error) {
+	timeout, err := time.ParseDuration(config.Workspace.ProgressTimeout)
+	if err != nil {
+		return 0, fmt.Errorf("invalid duration specified for workspace progress timeout: %w", err)
+	}
+	timeoutRounded := timeout.Round(time.Second)
```
amisevsk, line 327: Do you need to round? I believe converting to an `int32` will drop the factional part.

amisevsk (commented):
Looking good, just a couple more changes.

pkg/provision/workspace/deployment.go
AObuchow: You're right, no need to round -- I was just being overly cautious :P

pkg/provision/workspace/deployment.go
AObuchow: > we don't necessarily need 10 revisions in the history (and I believe this setting would result in up to ten replicasets for each deployment)

Last thing I thought about for this PR: I checked the deployment spec and the default is indeed 10 for the revision history limit. Maybe this should be modified to 2? Or even 3? Or we can leave this for another PR/if an issue arises from this in the future.

pkg/library/status/check.go
AObuchow: +1 That's a lot better, thanks for the suggestion :)

amisevsk (approved):
LGTM

PR #1063, opened 2023-03-11
Ensure order of projected sources is sorted

amisevsk:
### What does this PR do?
Occasionally, the operator will get stuck updating a workspace deployment repeatedly, switching the order of sources in a projected volume. Sort projected volume sources (alphabetically, with configmaps before secrets) to ensure we always use the same order.

No issue created; cannot consistently reproduce and workspaces still start, but I hit it a few times while testing. Most likely to reproduce for new workspaces started shortly after operator is rolled out.

### Is it tested? How?
Create a workspace that uses projected automount volumes. For testing, the volumes below can be used
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    controller.devfile.io/mount-as: file
    controller.devfile.io/mount-path: /tmp/test-projected/
  labels:
  name: test-cm-1
  namespace: dw
data:
  configmap-2-key: hello
---
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    controller.devfile.io/mount-as: file
    controller.devfile.io/mount-path: /tmp/test-projected/
  labels:
  name: test-cm-2
  namespace: dw
data:
  configmap-2-key: hello
```
To test if reordering occurs, watch workspace startup via `kubectl get dw -w` and see if state switches from Running to Starting with no changes to the DevWorkspace object itself. In addition, if `enableExperimentalFeatures` is set in the config, the operator will log the diff of changes its making to the deployment (copy the long string and `echo -e` it)

### PR Checklist

### DIFF
--- a/pkg/provision/automount/projected.go
+++ b/pkg/provision/automount/projected.go
@@ -117,6 +117,22 @@ func generateProjectedVolume(mountPath string, volumeMounts []corev1.VolumeMount
 		projectedVolume.Projected.Sources = append(projectedVolume.Projected.Sources, projection)
 	}
 
+	// Order of resources here may be random; to avoid unnecessarily updating deployment
+	// we need to sort them somehow.
+	sort.Slice(projectedVolume.Projected.Sources, func(i, j int) bool {
+		iSource, jSource := projectedVolume.Projected.Sources[i], projectedVolume.Projected.Sources[j]
+		switch {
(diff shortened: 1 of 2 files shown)


AObuchow (approved):
Code looks good to me, tested on Minikube with the provided configmap and a devworkspace started right after DWO was deployed, and wasn't able to reproduce the issue (though it's not consistently reproducible).

PR #1070, opened 2023-03-21
allow to configure specific annotations

### LINKED ISSUE #1068: Parametrize IngressClass name
### Description

On a K3S kubernetes, the default ingress controller is traefik. It uses a "traefik" IngressClass.

Currently, the Ingress generated by devworkspace-controler is hardcoded to "nginx"
https://github.com/devfile/devworkspace-operator/blob/7e8aa722dfb4f2a168933bc51edb1aedb5b56b85/controllers/controller/devworkspacerouting/solvers/basic_solver.go#L34

Having a parameter to control this name would help to reuse in different K8S.

gbonnefille:
### What does this PR do?

It adds the ability to configure the annotations applied to routing object (Route or Ingress).

It fixes #1068 

### Is it tested? How?

Currently, only tested by unit tests

### PR Checklist


controllers/controller/devworkspacerouting/solvers/basic_solver.go
```diff
@@ -67,12 +80,12 @@ func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.DevWorkspaceRou
 	services := getServicesForEndpoints(spec.Endpoints, workspaceMeta)
 	services = append(services, GetDiscoverableServicesForEndpoints(spec.Endpoints, workspaceMeta)...)
 	routingObjects.Services = services
+	routingAnnotations := config.GetGlobalConfig().Routing.Annotations
```
AObuchow, line 83: Similar to [here](https://github.com/devfile/devworkspace-operator/pull/1070/files#diff-9bf8bfd60b6c1418e0e13699e8c3437689b65c8841a1bd728dd1506ed4866ae9R74), do you mind adding a note?:
`TODO: Use workspace-scoped routing annotatiosn to allow overriding`

Individual workspaces should be allowed to override the global config's routing settings, but this isin't currently done. It's best we make note of where we need to address this for the future.

controllers/controller/devworkspacerouting/solvers/common_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+		routingSuffix string
+		endpoint      v1alpha1.Endpoint
+		meta          DevWorkspaceMetadata
+		annotations   map[string]string
+
+		expectedAnnotationsKeys []string
+	}{
+		{
+			name: "nil",
```
AObuchow, line 22: Very happy to see you added test cases, great work :sunglasses: 
Do you mind using more descriptive test case names?

For example, this test could be called: "Gets default OpenShift route annotation when annotations aren't defined".

controllers/controller/devworkspacerouting/solvers/common_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+			annotations: nil,
+
+			expectedAnnotationsKeys: []string{
+				"controller.devfile.io/endpoint_name",
+				"haproxy.router.openshift.io/rewrite-target",
+			},
+		},
+		{
+			name: "empty",
```
AObuchow, line 32: This test could be called: "Gets default OpenShift route annotation when annotations are empty".

controllers/controller/devworkspacerouting/solvers/common_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+			annotations: map[string]string{},
+
+			expectedAnnotationsKeys: []string{
+				"controller.devfile.io/endpoint_name",
+				"haproxy.router.openshift.io/rewrite-target",
+			},
+		},
+		{
+			name: "defined",
```
AObuchow, line 42: This test could be called: "Gets default OpenShift route annotation when annotations are defined".

controllers/controller/devworkspacerouting/solvers/common_test.go
```diff
@@ -0,0 +1,125 @@
...
+func TestGetRouteForEndpointAnnotations(t *testing.T) {
+	tests := []struct {
+		name string
+
+		routingSuffix string
+		endpoint      v1alpha1.Endpoint
+		meta          DevWorkspaceMetadata
+		annotations   map[string]string
+
+		expectedAnnotationsKeys []string
```
AObuchow, line 19: Might almost feel redundant, but it might be nicer to also compare the resulting annotation values (in addition to the annotation keys) with the expected key/values. So this could just be `expectedAnnotations map[string]string`

controllers/controller/devworkspacerouting/solvers/common_test.go
```diff
@@ -0,0 +1,125 @@
...
+			for _, expected := range tt.expectedAnnotationsKeys {
+				_, ok := route.Annotations[expected]
+				assert.True(t, ok, "Key %s does not exist", expected)
+				assert.Equal(t, len(tt.expectedAnnotationsKeys), len(route.Annotations))
+			}
+		})
+	}
+}
+
+func TestGetIngressForEndpointAnnotations(t *testing.T) {
```
AObuchow, line 66: If you make the changes I suggested for the Route test, make sure they're also done for the ingress test: changing the test names (though the names should specify that they are for Kubernetes, not OpenShift) and comparing the keys and values of the annotations map.

controllers/controller/devworkspacerouting/solvers/common_test.go
```diff
@@ -0,0 +1,125 @@
...
+			expectedAnnotationsKeys: []string{
+				"controller.devfile.io/endpoint_name",
+				"example.com/extra"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			route := getRouteForEndpoint("routingSuffix", v1alpha1.Endpoint{Name: "Endpoint"}, DevWorkspaceMetadata{DevWorkspaceId: "WorkspaceTest"}, tt.annotations)
+			for _, expected := range tt.expectedAnnotationsKeys {
```
AObuchow, line 57: Instead of comparing each key (and value) individually, you could also do something like the following to compare the actual vs expected map and get a diff:

`assert.Equal(t, tt.expectedAnnotations, route.Annotations, "Annotations should match: Diff: %s", cmp.Diff( tt.expectedAnnotations, route.Annotations))`

AObuchow (changes requested):
@gbonnefille Thank you so much for the PR :) :pray: 

I left some initial review comments, though you'll have to wait for @amisevsk's final review and approval for this to get merged (assuming there are no blocking reasons why this couldn't be added to DWO). He's currently on PTO right now, but will be back soon enough.

3 other details I wanted to point out:

1. You'll have to modify `sync.go` to allow the new config.routing field you've added to be merged and logged.

You'll have to add something like the following [here](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go#L259):

```go
if from.Routing.Annotations != nil {
	to.Routing.Annotations = from.Routing.Annotations
}
```

And something like the following [here](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go#L390):


```go
if routing.Annotations != nil && len(routing.Annotations) > 0 {
	config = append(config, fmt.Sprintf("routing.Annotations=%s", routing.Annotations))
}
```

Make sure all tests complete successfully (`make test`), with [this one in particular for merging the config](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync_test.go#L346)

2. The CRDs will have to be re-generated since you added a new field to the operator configuration. Run the following, and add the resulting changes into a separate commit: `make generate manifests fmt generate_default_deployment generate_olm_bundle_yaml`

3. Once you've done all of the above, please manually test that your changes resolve [your issue](https://github.com/devfile/devworkspace-operator/issues/1068). If you need help deploying your custom build of DWO, just ping me :)

AObuchow:
Also @gbonnefille please sign-off your commits (you can fix this by doing `git commit --amend --signoff` for your existing commit, and `git commit -m "[commit message] --signoff` for future commits)

gbonnefille:
@AObuchow thanks a lot for this detailed review and all the suggestions.

Initially, I'm just interested in presenting a minimal change in order to clearly express my ideas and collecting feedback about this ideas. Now I understand it is a way to go, so I will improve my PR following your suggestions.

Thanks again.

controllers/controller/devworkspacerouting/solvers/common_test.go
gbonnefille: Yes. I blow my mind with the generated annotation for Endpoint name, but this value is controlled by the test runner. So, a simple Equal should be enough.

gbonnefille:
With the changes suggested, I was able to have my own version of devworkspace-operator running, with my changes. And it worked: traefik picked the Ingress and processed it.
I'm really happy.

gbonnefille:
I uploaded changes for generated code and CRDs. But I have to admit my workspace is not fully compliant with the requirements for developping on devworkspace-operator. The suggested make failed. I hope a maintainer will be able to terminate this PR if there is missing pieces of code.

AObuchow:
@gbonnefille Glad to hear you got DWO running with your changes and things worked correctly, :) I compiled the image based off of your PR successfully without any issues. We'll wait for @amisevsk's feedback before continuing with this PR for now.

controllers/controller/devworkspacerouting/solvers/common.go
```diff
@@ -220,7 +220,7 @@ func getIngressForEndpoint(routingSuffix string, endpoint controllerv1alpha1.End
 			Labels: map[string]string{
 				constants.DevWorkspaceIDLabel: meta.DevWorkspaceId,
 			},
-			Annotations: nginxIngressAnnotations(endpoint.Name),
+			Annotations: createAnnotations(endpoint.Name, annotations, nginxIngressAnnotations),
```
amisevsk, line 223: I'm not sure about this approach; there's a pretty distant link between where `annotations` is defined and where it's used, and how it's defined completely changes the code significantly further down the call stack. The function `createAnnotations` behaves completely differently based on whether `annotations` is an empty map here:

```go
	if routingAnnotations == nil || len(routingAnnotations) == 0 {
		appendMap(annotations, defaultAnnotations)
	} else {
		appendMap(annotations, routingAnnotations)
	}
```

controllers/controller/devworkspacerouting/solvers/basic_solver.go
```diff
@@ -22,20 +22,33 @@ import (
...
+	annotations := map[string]string{
 		constants.DevWorkspaceEndpointNameAnnotation: endpointName,
 	}
+
+	if routingAnnotations == nil || len(routingAnnotations) == 0 {
+		appendMap(annotations, defaultAnnotations)
+	} else {
+		appendMap(annotations, routingAnnotations)
+	}
+	return annotations
```
amisevsk, line 51: A better approach for setting a default value if the configuration is unspecified would be similar to what we do in the [config](https://github.com/devfile/devworkspace-operator/blob/e97ad530f84b6529fd85109a83d16aa0a600c158/pkg/config/defaults.go#L53-L78) package -- this avoids needing functions that take an optional value and a default.

controllers/controller/devworkspacerouting/solvers/basic_solver.go
```diff
@@ -67,12 +80,13 @@ func (s *BasicSolver) GetSpecObjects(routing *controllerv1alpha1.DevWorkspaceRou
 	services := getServicesForEndpoints(spec.Endpoints, workspaceMeta)
 	services = append(services, GetDiscoverableServicesForEndpoints(spec.Endpoints, workspaceMeta)...)
 	routingObjects.Services = services
+	// TODO: Use workspace-scoped routing annotations to allow overriding
+	routingAnnotations := config.GetGlobalConfig().Routing.Annotations
```
amisevsk, line 84: It will likely not be possible to get workspace-scoped annotations, as this is within the context of the DevWorkspaceRouting reconciler. In fact, we should avoid using `config` within `controllers/controller/devworkspacerouting/...` in general, as the routing reconciler may be run outside the DWO context (as it is in Eclipse Che, for example).

In an ideal world, the configuration of DWO is propagated into the DevWorkspaceRouting CR directly in some way. However, some choices made in the past make this very difficult -- likely the only way to do this is to serialize the map to JSON and add it as an annotation on the routing CR.

amisevsk (commented):
Thank you for the PR @gbonnefille! This is quite a large change for the DevWorkspace Operator. 

I think there a few improvements we should make before merging, though they're on the complicated side, so let me know if you would like my help/contribution:

1. If we're adding `Annotations` to the devworkspaceoperatorconfig CR, setting the default value should also occur the config package instead of in `basic_solver.go` (similar to how we manage the initial set up of `[Kubernetes|OpenShift]PodSecurityContext` -- see pkg/config/defaults.go)

2. Any routing reconciler related packages should avoid references to the global or workspace-specific configuration in general, for a few reasons
    * The `basic` routing class is not really intended for anything other than development purposes; it offers no authentication and should ***not*** be used in production, so writing functionality specifically for it isn't the best approach. 
    * The routing reconciler is generally assumed to not have access to the DevWorkspace, DevWorkspaceOperatorConfig, or related objects. This general mechanism was chosen to allow other operators to implement routing as they require (since there are too many various configurations). For an example of this, see the Eclipse Che Operator's implementation of a routing reconciler that provides single-host access and auth for workspaces. When working on the `dwr` controller, we have to assume all we can see is the `dwr` itself and objects it owns. 

For the second point above, I think the best we can do for now is to serialize the routing annotations config to a JSON string and put it in DWO-defined annotation on the DevWorkspaceRouting resource, e.g.
```yaml
kind: DevWorkspaceRouting
metadata:
  annotations:
    controller.devfile.io/routing-annotations: '{"kubernetes.io/ingress.class":"nginx","nginx.ingress.kubernetes.io/rewrite-target":"/","nginx.ingress.kubernetes.io/ssl-redirect":"false"}'
```

The solvers in the DevWorkspaceRouting reconciler could then deserialize this annotation and apply it to ingresses/routes created. The benefits of this approach are
* It defines a general way for configuring this routing information through the DevWorkspace Operator for any routingClass (should the implementer choose to use this configuration)
* It makes applying workspace-scoped configuration simple, as the resolved configuration is available at the moment the DevWorkspaceRouting is created

gbonnefille:
Thanks @amisevsk for your review. I understand now how naive my proposal was.

One more naive question: is it really mandatory to serialize the annotations in another annotation? What about simply setting annotations on the DevWorkspaceRouting and then having the controller to copy all the annotations of the DevWorkspaceRouting to the effective object (Route/Ingress)? In think this is already how other controllers do, for example Deployment's annotations are copied to the underlying Pod.

amisevsk:
@gbonnefille Your PR isn't naive -- I had completely forgotten these details until I went to review it :). It's complicated since DevWorkspaceRoutings are typically reconciled by _other_ operators so we have to be more careful.

> What about simply setting annotations on the DevWorkspaceRouting and then having the controller to copy all the annotations of the DevWorkspaceRouting to the effective object (Route/Ingress)?

I have some concerns with this approach, though it is _much_ cleaner. The main issues I see are
1. It constrains how DWO can use annotations on DevWorkspaceRouting objects, as it's assumed that _all_ annotations will be copied. This could cause issues down the line if we want to set an annotation that isn't included in routing objects. For example, DWO supports copying special annotations to the DWR from the DevWorkspace (see PR https://github.com/devfile/devworkspace-operator/pull/245)
2. It's less clear how annotations get applied (do they apply to services as well as ingresses/routes?). In the future, there may be a use case for specifying both "annotations for route/ingress" and "annotations for the service", and copying all annotations interferes with this. 

Basically, using one annotation to group everything is the safer (and uglier) option. I'm still open to doing it this way, but we would need to consider the implications more than a simple scoped annotation.

Issue #1080, opened 2023-04-05
Add configuration fields for project clone resource requirements


amisevsk:
### Description
Add fields to the DevWorkspaceOperatorConfig CRD for configuring the project clone init container:
* memory/cpu limit/request
* image?
* anything else that may be useful to configure

Currently, project-clone limits/requests are hard-coded constants in DWO, which can cause issues (e.g. OOMKilled for large repos).

### Additional context
https://github.com/eclipse/che/issues/22036

PR #1091, opened 2023-04-20
Add DWOC fields for configuring project clone init container

amisevsk:
### What does this PR do?
Adds fields to the DevWorkspaceOperatorConfig CR for configuring project clone:
```yaml
config:
  workspace:
    projectClone:
      image: <project-clone-image>
      imagePullPolicy: <pull-policy>
      env: []<env-var>
      resources:
        limits:
          memory: <quantity>
          cpu: <quantity>
        requests:
          memory: <quantity>
          cpu: <quantity>
```
with the following behavior:
* If image is specified, it is used for project clone. Otherwise, the image from the operator definition (relatedImages) is used
* If imagePullPolicy is unspecified, the value from `config.workspace.imagePullPolicy` is used
* Environment variables are appended to the container's default env vars
* Resources override default configurations, with quantities specified as "0" interpreted as "unset"
    * I.e. If a quantity is `"0"`, that field is omitted from the project clone container's resources field. If all fields are zero, no resources are specified.

Closes https://github.com/devfile/devworkspace-operator/issues/1080

### Is it tested? How?
Test by setting fields and verifying that they are used for the project clone container. Note that editing the dwoc CR does not trigger reconciles for workspaces by itself, so if your workspace is running you may need to poke it (e.g. by setting an annotation or restarting it)

### PR Checklist


apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -62,76 +62,10 @@ type RoutingConfig struct {
...
-	// plugin will proactively rotate the service account token. The kubelet will
-	// start trying to rotate the token if the token is older than 80 percent of
-	// its time to live or if the token is older than 24 hours. Defaults to 1 hour
-	// and must be at least 10 minutes.
-	// +kubebuilder:validation:Minimum=600
-	// +kubebuilder:default:=3600
-	// +kubebuilder:validation:Optional
-	ExpirationSeconds int64 `json:"expirationSeconds,omitempty"`
-}
-
```
amisevsk, line 133: Note: I reorganized this file to put the `WorkspaceConfig` struct first; these fields are unchanged, just moved.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -203,6 +137,89 @@ type WorkspaceConfig struct {
...
+
+type ProjectCloneConfig struct {
+	// Image is the container image to use for cloning projects
+	Image string `json:"image,omitempty"`
+	// ImagePullPolicy configures the imagePullPolicy for the project clone container.
+	// If undefined, the general setting .config.workspace.imagePullPolicy is used instead.
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+	// Resources defines the resource (cpu, memory) limits and requests for the project
+	// clone container. To explicitly not specify a limit or request, define the resource
+	// quantity as zero ('0')
```
AObuchow, line 217: Maybe overkill, but might be worth also mentioning that "To use default limit or request values, the corresponding field should be left blank"

AObuchow:
Also just wanted to show:
When the CPU or memory request can't be satisfied in the project-clone container, the devworkspace will fail in a similar manner to when the main devworkspace container's memory or CPU request can't be satisfied:


```
NAME          DEVWORKSPACE ID             PHASE    INFO
code-latest   workspace764fb09a01b9487d   Failed   Error creating DevWorkspace deployment: Detected unrecoverable event FailedScheduling: 0/6 nodes are available: 3 Insufficient cpu, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }. preemption: 0/6 nodes are available: 3 No preemption victims found for incoming pod, 3 Preemption is not helpful for scheduling..
```

There's no mention of the project-clone container failing specifically. Not sure if this is something we should address (or perhaps leave as an issue for a future improvement).

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
amisevsk: Eh, I think it should be assumed that not setting a value says "use the default". We can update the CRD documentation if there's confusion here in the future.

amisevsk:
@AObuchow I don't think it's feasible to report why scheduling failed; in your case it failed because you increased the project clone requirements, but from the cluster's perspective, it failed because the pod itself requested more CPU than any node had available.

Issue #1095, opened 2023-05-04
Operator can get caught in a loop continually updating all workspaces


amisevsk:
### Description
When under some load, the DevWorkspace Operator enter a situation where it gets caught in a loop updating the status for all stopped workspaces. This occurs because PR https://github.com/devfile/devworkspace-operator/pull/1064 reworked how workspace conditions are handled and introduced a bug where conditions with status `Unknown` get updated timestamps on every reconcile. See changes [here](https://github.com/devfile/devworkspace-operator/pull/1064/commits/64392c53cf9e8f8606c17e5c1f3dce59c719785c#diff-65fcf1aa32aa49850ef197edbe1ff894195f860d3ac58908dc8801757db95616R111-R117).

The loop manifests as
1. Operator notices that workspace is stopped, queues reconcile to set stopped status, which sets all conditions except `Started` to have status `Unknown`
2. Since the DevWorkspace was changed, a reconcile is queued
3. The operator sees that the workspace is still stopped, and so it sets the stopped status, which updates the timestamp on `Unknown` status conditions... triggering another reconcile

The reason this bug wasn't discovered sooner is because the resolution for timestamps on conditions is one second. If the DevWorkspace Operator is able to process the second reconcile within one second (which is normal when not under load), then the change in step 3 above is a no-op and so no new reconciles are queued. This is particularly a problem in clusters with a large number of workspaces that are in the stopped state, as the DevWorkspace Operator processes every workspace on startup; the workqueue can easily grow long enough that the queued reconcile isn't processed for a second or more.

### How To Reproduce
1. (helpful) Update the DWO controller deployment to have a small CPU limit (e.g. `25m`) to ensure reconciles cannot be processed quickly enough
2. Create 20 workspaces on the cluster and then stop them after they've progressed -- e.g. in OpenShift
    ```bash
    for i in {1..20}; do 
      kubectl create ns wksp-$i
      kubectl apply -f samples/plain.yaml -n wksp-$i
      sleep 10s 
      kubectl patch dw -n wksp-$i plain-devworkspace --type merge -p '{"spec": {"started": false}}'
    done
    ```
3. Restart DWO controller pod to trigger a reconcile for all workspaces: `make restart`

This should be enough to trigger the operator continually reconciling workspaces; if it doesn't, try creating even more workspaces.

### Expected behavior
Reconciles should only be triggered when additional work is required by DWO.

PR #1096, opened 2023-05-04
Don't update timestamps on conditions with status 'Unknown'

amisevsk:
### What does this PR do?
Fix an issue where DWO always attempts to update conditions with status 'Unknown' to have a current timestamp. This can trigger an issue in when the workqueue is longer than 1s long, as each reconcile updates timestamps, triggering another reconcile.

Closes https://github.com/devfile/devworkspace-operator/issues/1095

### Is it tested? How?
Test using reproducer on issue.

### PR Checklist


AObuchow (approved):
Code looks good to me and tested that I was no longer able to get a reconcile loop happening with the reproducer instructions (when I restarted the controller pod, all the stopped workspaces only got a single reconcile instead, which is  expected).

PR #1098, opened 2023-05-09
Ensure order of git credentials in merged git-credentials secret

amisevsk:
### What does this PR do?
In some cases, we've seen the DevWorkspace Operator perform unnecessary changes to the devworkspace-merged-git-credentials secret when multiple credentials secrets are used. This appears to be due to varying order of secrets when listing from the controller cache.

To work around this, we sort git credentials secrets (and git tls configmaps) by name, first, so that objects should always be in a defined order.

N/A 

### Is it tested? How?
Hard to test as issue arises inconsistently.

### PR Checklist

### DIFF
--- a/pkg/provision/sync/sync.go
+++ b/pkg/provision/sync/sync.go
@@ -200,17 +200,19 @@ func printDiff(specObj, clusterObj crclient.Object, log logr.Logger) {
 			diffOpts = podDiffOpts
 		case *corev1.ConfigMap:
 			diffOpts = configmapDiffOpts
-		case *corev1.Secret:
-			diffOpts = secretDiffOpts
 		case *v1alpha1.DevWorkspaceRouting:
 			diffOpts = routingDiffOpts
 		case *networkingv1.Ingress:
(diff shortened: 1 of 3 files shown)


ibuziuk (approved):

pkg/provision/sync/sync.go
```diff
@@ -200,17 +200,19 @@ func printDiff(specObj, clusterObj crclient.Object, log logr.Logger) {
...
-		case *corev1.Secret:
-			diffOpts = secretDiffOpts
 		case *v1alpha1.DevWorkspaceRouting:
 			diffOpts = routingDiffOpts
 		case *networkingv1.Ingress:
 			diffOpts = ingressDiffOpts
 		case *routev1.Route:
 			diffOpts = routeDiffOpts
+		case *corev1.Secret:
+			log.Info(fmt.Sprintf("Diff: secret %s data upated", specObj.GetName()))
```
AObuchow, line 210: Why are we no longer printing the secret diff? To reduce logging? Or to make it more obvious that it's a secret that was updated?

Issue #1103, opened 2023-05-19
Workspaces that define Kubernetes/OpenShift components with URIs always fail to start


amisevsk:
### Description
If a DevWorkspace includes a Kubernetes or OpenShift-type component that uses the `uri` field, that workspace will fail to start, even if the component has `deployByDefault: false`. 

For components with `deployByDefault: false`, starting the workspace should not be a problem, as the component is expected to be applied to the cluster at some later point (by e.g. running an `apply` command). 

For components with `deployByDefault: true`, we should still start the workspace, instead ignoring those components and adding a warning to the workspace.  

### How To Reproduce
Apply the DevWorkspace
```yaml
apiVersion: workspace.devfile.io/v1alpha2
kind: DevWorkspace
metadata:
  name: plain
spec:
  routingClass: basic
  started: true
  template:
    attributes:
      controller.devfile.io/storage-type: ephemeral
    components:
    - name: web-terminal
      container:
        command: ["tail", "-f", "/dev/null"]
        image: quay.io/amisevsk/web-terminal-tooling:dev
        memoryLimit: 512Mi
        mountSources: true
    - name: test-k8s
      kubernetes: 
        deployByDefault: false
        uri: https://raw.githubusercontent.com/devfile/devworkspace-operator/main/deploy/deployment/kubernetes/objects/devworkspace-controller-metrics.Service.yaml
```

### Expected behavior
If a workspace uses a kubernetes component with `deployByDefault: false`, we shouldn't care if DWO supports URIs or not

If a workspace instead has `deployByDefault: true`, it would still be useful to start the workspace, downgrading the failure (and webhook rejection) to a warning.

amisevsk:
cc: @l0rd we discussed this on Wednesday as an issue with https://registry.devfile.io/ compatibility.

Issue #993, opened 2022-12-12
Implicitly set `merge-contribution:true` on first template component


l0rd:
### The problem

Currently both a `contribution` and the attribute `controller.devfile.io/merge-contribution: true` are required to properly "inject" an IDE in workspace Pod:

```yaml
kind: DevWorkspace
(...)
spec:
  contributions:
    - name: editor
      kubernetes:
        name: che-code 
  template:
    components:
      - name: tooling-container
        attributes:
          controller.devfile.io/merge-contribution: true
```

Considered that:
- most of the time the template only has one container component or, if there are more than one, the developer wants to use the first container as the dev tooling container.
- the `merge-contribution` attribute makes sense in a Che scenario but not in odo or dev console scenarios (yet) so we should not specify it in registry.devfile.io.
- we want the devworkspace template to match the original devfile so it's easier to update it if the original devfile is updated.

we should define a convention to apply contributions when that's not explicitly specified in the template/devfile. 

### A proposal

- **opt-in to contribution is implicit:** when no component has a `merge-contribution` attribute, the contribution should be applied to the first container component.
- **out-out to contribution is explicit:** the author or a devfile should explicitly set `merge-contribution` to `false` to avoid contributions to that component.
- **a devfile with multiple `merge-contributions: true` is invalid:** the devworkspace controller should return an error if multiple components have `merge-contribution` explicitly set to `true`.

amisevsk:
One grey area with implicitly including the `merge-contributions: true` attribute is for devfiles that do not explicitly include a container. What will happen in this case is that DWO will attempt to inject into the first container in the resolved (flattened) devfile. 

For example, in our default ["theia-next" sample](https://github.com/devfile/devworkspace-operator/blob/main/samples/theia-next.yaml), DWO would attempt to inject into the `theia-ide` component, which may not work as expected (the behavior of DWO as it is now would be to run the theia-ide-contributions container using the default image specified there -- i.e. UDI).

Currently, I believe DWO internally flattens contributions + plugin components into one list of components to start, so this behavior would be present with the [theia-next-contributions sample](https://github.com/devfile/devworkspace-operator/blob/main/samples/theia-next-contributions.yaml)

One of the reasons we defaulted to explicit opt-in, implicit opt-out is to be sure a workspace can be started (assuming the default image for unmerged components is suitable for running the injected component). With implicit injection, adding e.g. a busybox container to the workspace will cause startup to likely fail -- though this may not be a use case we care to support.

[...]

PR #1104, opened 2023-05-19
Don't fail workspaces that use Kubernetes/OpenShift components with URIs

amisevsk:
### What does this PR do?
Update how Kubernetes/OpenShift components with URIs are handled:

* If a component has `deployByDefault: false`, it is ignored and workspace starts normally (the component is expected to be applied later, e.g. via an apply command
* If a component has `deployByDefault: true`, it is _not_ applied to the cluster if it uses a URI, and instead the workspace gets a warning condition. This effectively downgrades the webhook rejection to a warning condition

Closes https://github.com/devfile/devworkspace-operator/issues/1103

### Is it tested? How?
Apply the DevWorkspace 
```yaml
kind: DevWorkspace
metadata:
  name: plain
spec:
  routingClass: basic
  started: true
  template:
    attributes:
    components:
    - name: web-terminal
      container:
        command: ["tail", "-f", "/dev/null"]
        image: quay.io/amisevsk/web-terminal-tooling:dev
        memoryLimit: 512Mi
        mountSources: true
    - name: test-k8s
      kubernetes: 
        deployByDefault: false
        uri: https://raw.githubusercontent.com/devfile/devworkspace-operator/main/deploy/deployment/kubernetes/objects/devworkspace-controller-metrics.Service.yaml
```

Update the workspace to set `deployByDefault`:
* If false, workspace should start normally, without any issues or warnings
* If true, the workspace should still start, but have a warning condition:
    > Ignored components that use unsupported features: component test-k8s
    > defines a Kubernetes/OpenShift component via URI

### PR Checklist

### DIFF
--- a/pkg/library/kubernetes/provision.go
+++ b/pkg/library/kubernetes/provision.go
@@ -16,6 +16,7 @@ package kubernetes
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/devfile/devworkspace-operator/pkg/common"
 	"github.com/devfile/devworkspace-operator/pkg/constants"
@@ -32,13 +33,18 @@ import (
 // has the correct permissions to create/update/delete these objects and instead assumes the
 // workspace owner has all applicable RBAC permissions.
 // Only Kubernetes/OpenShift components that are inlined are supported; components that define
-// a URI will cause a FailError to be returned
+// a URI will cause a WarningError to be returned
 func HandleKubernetesComponents(workspace *common.DevWorkspaceWithConfig, api sync.ClusterAPI) error {
-	kubeComponents, err := filterForKubeLikeComponents(workspace.Spec.Template.Components)
+	kubeComponents, warnings, err := filterForKubeLikeComponents(workspace.Spec.Template.Components)
 	if err != nil {
 		return err
 	}
 	if len(kubeComponents) == 0 {
+		if len(warnings) > 0 {
+			return &dwerrors.WarningError{
+				Message: fmt.Sprintf("Ignored components that use unsupported features: %s", strings.Join(warnings, ", ")),
+			}
+		}
 		return nil
 	}
 	for _, component := range kubeComponents {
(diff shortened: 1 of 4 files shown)


pkg/library/kubernetes/provision.go
```diff
@@ -34,11 +35,16 @@ import (
 // Only Kubernetes/OpenShift components that are inlined are supported; components that define
 // a URI will cause a FailError to be returned
```
AObuchow, line 36: Documentatioon should be changed to `...a URI will cause a WarningError to be returned`

webhook/workspace/handler/kubernetes.go
```diff
@@ -42,7 +42,8 @@ func (h *WebhookHandler) validateKubernetesObjectPermissionsOnCreate(ctx context
 			continue
 		}
 		if component.Uri != "" {
-			return fmt.Errorf("kubenetes components specified via URI are unsupported")
+			// We're going to ignore URI components for now
```
AObuchow, line 45: A future improvement could be to return a webhook warning (so that the dashboard could display it) but for now, I don't think this is necessary.

AObuchow (approved):
Code looks good to me and works as expected on OpenShift 4.12.
Minor comment regarding documentation, but LGTM

AObuchow (approved):

l0rd:
@amisevsk do you have an already built image to test this PR?

ibuziuk (approved):

l0rd:
I have tested with https://github.com/devfile-samples/devfile-sample-go-basic.git and I was able to start the workspace successfully.

The weird thing is that my workspace pod has 3 containers (but it may be related to a dashboard issue that ignores the parent and adds a default component):

image

amisevsk:
> do you have an already built image to test this PR?

I didn't build a specific image for this PR, so it was only pushed to `quay.io/amisevsk/devworkspace-controller:dev` (which is the image I use for whatever I'm working on).

> The weird thing is that my workspace pod has 3 containers (but it may be related to a dashboard issue that ignores the parent and adds a default component):

This should be reported as an issue on the Eclipse Che side -- it looks like something is going wrong with merging components. The `runtime` component is coming from the parent, but since https://github.com/devfile/devworkspace-operator/issues/993 the `che-code-runtime-description` component should have been merged into it automatically (i.e. even without the explicit merge-components attribute). I tried briefly to reproduce the issue but would need to roll changes from this PR out to test properly.

PR #1111, opened 2023-05-29
feat: Don't set CPU limits when possible

tolusha:
### What does this PR do?
https://github.com/eclipse/che/issues/22198

feat: Don't set CPU limits when possible

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -95,6 +95,7 @@ type DevWorkspaceReconciler struct {
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;create
 // +kubebuilder:rbac:groups=config.openshift.io,resources=proxies,verbs=get,resourceNames=cluster
 // +kubebuilder:rbac:groups=apps,resourceNames=devworkspace-controller,resources=deployments/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=limitranges,verbs=list
 /////// Required permissions for workspace ServiceAccount
 // +kubebuilder:rbac:groups="",resources=pods/exec,verbs=create
 // +kubebuilder:rbac:groups=apps;extensions,resources=replicasets,verbs=get;list;watch
@@ -299,6 +300,8 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	devfilePodAdditions, err := containerlib.GetKubeContainersFromDevfile(
+		r.NonCachingClient,
+		workspace.Namespace,
 		&workspace.Spec.Template,
 		workspace.Config.Workspace.ContainerSecurityContext,
 		workspace.Config.Workspace.ImagePullPolicy)
@@ -322,7 +325,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	} else {
 		projectCloneOptions.PullPolicy = corev1.PullPolicy(config.Workspace.ImagePullPolicy)
 	}
-	if projectClone, err := projects.GetProjectCloneInitContainer(&workspace.Spec.Template, projectCloneOptions, workspace.Config.Routing.ProxyConfig); err != nil {
+	if projectClone, err := projects.GetProjectCloneInitContainer(r.NonCachingClient, workspace.Namespace, &workspace.Spec.Template, projectCloneOptions, workspace.Config.Routing.ProxyConfig); err != nil {
 		return r.failWorkspace(workspace, fmt.Sprintf("Failed to set up project-clone init container: %s", err), metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus), nil
 	} else if projectClone != nil {
 		devfilePodAdditions.InitContainers = append(devfilePodAdditions.InitContainers, *projectClone)
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -26,6 +26,8 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+var ProjectCloneContainerDefaultCpuLimit = resource.MustParse("1000m")
+
 // defaultConfig represents the default configuration for the DevWorkspace Operator.
 var defaultConfig = &v1alpha1.OperatorConfiguration{
 	Routing: &v1alpha1.RoutingConfig{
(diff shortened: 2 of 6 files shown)


tolusha:
@amisevsk 
I am not sure about command to fix resources. Could you help me?

amisevsk:
To regenerate all resources:
```
make generate manifests generate_default_deployment generate_olm_bundle_yaml
```
the steps are
* Generate deepcopy/crd resources
* Generate deployment manifests (update RBAC, etc. -- that's the change that's missing from this PR)
* Generate OpenShift/Kubernetes depoyments (run kustomize)
* Generate OLM bundle 

Feel free to add a `make generate-all` rule to make this simpler in the future -- I just haven't gotten around to it since the combination above is in my shell history :)

pkg/library/container/container.go
```diff
@@ -46,7 +48,7 @@ import (
 // rewritten as Volumes are added to PodAdditions, in order to support e.g. using one PVC to hold all volumes
 //
 // Note: Requires DevWorkspace to be flattened (i.e. the DevWorkspace contains no Parent or Components of type Plugin)
-func GetKubeContainersFromDevfile(workspace *dw.DevWorkspaceTemplateSpec, securityContext *corev1.SecurityContext, pullPolicy string) (*v1alpha1.PodAdditions, error) {
+func GetKubeContainersFromDevfile(k8sClient client.Client, namespace string, workspace *dw.DevWorkspaceTemplateSpec, securityContext *corev1.SecurityContext, pullPolicy string) (*v1alpha1.PodAdditions, error) {
```
amisevsk, line 51: I'm mildly concerned with the high-level changes to `pkg/library/container` -- this package was intended to provide a straight conversion from devfile resources to Kubernetes resources, so including a client and depending on cluster state is something I'd prefer to avoid if possible. 

However, I don't see a great way of supporting this functionality cleanly otherwise. Probably the best solution is some "convert-to-container config" struct that sets whether we set a CPU limit by default, but that's also a much bigger change.

pkg/config/defaults.go
```diff
@@ -26,6 +26,8 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+var ProjectCloneContainerDefaultCpuLimit = resource.MustParse("1000m")
+
```
amisevsk, line 30: Could you add a doc line on this variable to explain why it's not included in the defaultConfig below? e.g.

```suggestion
// By default, we do not set a CPU limit for the project clone container. However, to ensure the container 
// still works in clusters with limit ranges configured, this quantity is used when a) a limitrange may set 
// a default value for CPU limit and b) the Operator's configuration does not specify a non-default value.
var ProjectCloneContainerDefaultCpuLimit = resource.MustParse("1000m")
```

pkg/library/container/conversion.go
```diff
@@ -108,6 +111,17 @@ func devfileResourcesToContainerResources(devfileContainer *dw.ContainerComponen
...
+		// Set empty CPU limits when possible:
+		// 1. If there is no LimitRange in the namespace
+		// 2. CPU limits is not overridden
+		// See details at https://github.com/eclipse/che/issues/22198
+		limitRanges := &corev1.LimitRangeList{}
+		if err := k8sClient.List(context.TODO(), limitRanges, &client.ListOptions{Namespace: namespace}); err != nil {
+			return nil, err
+		} else if len(limitRanges.Items) == 0 {
+			cpuLimit = ""
+		}
```
amisevsk, line 124: Slightly cleaner (set cpuLimit var fewer times)
```suggestion

		// Set empty CPU limits when possible:
		// 1. If there is no LimitRange in the namespace
		// 2. CPU limits is not overridden
		// See details at https://github.com/eclipse/che/issues/22198
		limitRanges := &corev1.LimitRangeList{}
		if err := k8sClient.List(context.TODO(), limitRanges, &client.ListOptions{Namespace: namespace}); err != nil {
			return nil, err
		} else if len(limitRanges.Items) == 0 {
			cpuLimit = ""
		} else {
			cpuLimit = constants.SidecarDefaultCpuLimit
		}
```

amisevsk (commented):
I'm not sure about the approach in this PR -- it is including `sigs.k8s.io/controller-runtime/pkg/client` in a lot of packages that previously did not depend on it and makes a significant chunk of the codebase now depend on cluster state for behavior. 

Give me a bit of time to think on it; I'd prefer a cleaner approach if possible. If I can't come up with something I'll approve the PR.

pkg/library/container/conversion.go
amisevsk: No idea why the suggestion diffs are so broken today :thinking:

controllers/workspace/devworkspace_controller.go
```diff
@@ -299,6 +300,8 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	devfilePodAdditions, err := containerlib.GetKubeContainersFromDevfile(
+		r.NonCachingClient,
+		workspace.Namespace,
```
amisevsk, line 304: I think using the `NonCachingClient` for this purpose is a problem -- the purpose of the main controller's client is that it makes far fewer API requests (basically, reads are free due to watches/caching, and only writes require separate API calls). This approach will lead to two API requests per reconcile, regardless of whether the controller does any work, and I worry that this will impact stability.

amisevsk:
I think there are too many issues with this approach to enable this functionality. A better approach, though less convenient, may be to allow configuration of the default container resources, to allow for specifying empty CPU limits, similar to what we have for the project clone init container, or to not set a default CPU limit (by default) and require configuration on clusters with limit ranges.

pkg/library/container/conversion.go
```diff
@@ -107,7 +110,18 @@ func devfileResourcesToContainerResources(devfileContainer *dw.ContainerComponen
...
+		// 2. CPU limits is not overridden
+		// See details at https://github.com/eclipse/che/issues/22198
+		limitRanges := &corev1.LimitRangeList{}
+		if err := k8sClient.List(context.TODO(), limitRanges, &client.ListOptions{Namespace: namespace}); err != nil {
+			return nil, err
+		} else if len(limitRanges.Items) == 0 {
+			cpuLimit = ""
+		} else {
+			cpuLimit = constants.SidecarDefaultCpuLimit
+		}
```
amisevsk, line 124: The current behavior for DevWorkspace container components is to not set a limit -- `constants.SidecarDefaultCpuLimit` is `""`, which will lead to the conditional on line 126 not passing and so not setting a CPU limit.

amisevsk (commented):
@tolusha I've opened PR https://github.com/devfile/devworkspace-operator/pull/1127 to not set the CPU limit for the project clone container. This PR addresses some of my concerns with this PR (don't propagate client and namespace into library functions, don't use non-caching client).

Please take a look and review when you have a moment (I can't request your review directly)

amisevsk:
Closing PR as out of date.

Issue #1126, opened 2023-06-14
Don't set CPU limits by default unless they are required by a LimitRange


amisevsk:
### Description
We've gotten a number of requests to not set CPU limits by default on containers. By default there is no CPU limit or request for sidecar containers, however for project-clone we do set a default CPU Limit of `1000m` unless otherwise specified.

If a namespace does not include a limitRange that may require a CPU Limit to be set, we should not set one by default. If the DevWorkspaceOperatorConfig specifies a non-default CPU limit, we should use that value

### Additional context
Related: https://github.com/eclipse/che/issues/22198

Issue #1056, opened 2023-02-28
Don't add contributions' resource limits/requests to containers that do not define them


amisevsk:
### Description
When a DevWorkspace component does not specifiy a memory/cpu limit/request, but a contribution is merged that does, DWO will add the resource requirement from the contribution to zero, resulting in a potentially much lower request/limit than expected.

### How To Reproduce
The `code-latest.yaml` sample in this repo uses a `dev` container that does not specify limits/requests (for simplicity). When the workspace is started in a cluster, the dev container gets the limits/requests from the editor only (e.g. `500m` cpu limit), which may be unintentional.

In an extreme case, merging a contribution that has very low limits/requests could result in the dev container failing to run anything useful.

### Expected behavior
If the target for merging a contribution does not define a resource field, it should be undefined in merged component

PR #1127, opened 2023-06-15
[WIP] Don't set default project-clone CPU limit unless we have to

amisevsk:
### What does this PR do?
This PR changes how the default project-clone CPU limit is handled: if the configured value for the project-clone CPU limit is a) the default value (`1000m`) and b) there are no limit ranges present in the workspace's namespace, we no longer set a CPU limit for project-clone.

Closes https://github.com/devfile/devworkspace-operator/issues/1126
Related: https://github.com/eclipse/che/issues/22198, https://github.com/devfile/devworkspace-operator/pull/1111

### Is it tested? How?
Test:
1. By default, project-clone container does not specify a CPU limit
2. If the DWOC is configured with a non-default value (e.g. `1001m`), that value is used for the project-clone CPU limit
3. If a limitRange is created in the workspace's namespace, the default CPU limit of `1000m` is used for the project-clone container.

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -96,6 +96,7 @@ type DevWorkspaceReconciler struct {
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;create
 // +kubebuilder:rbac:groups=config.openshift.io,resources=proxies,verbs=get,resourceNames=cluster
 // +kubebuilder:rbac:groups=apps,resourceNames=devworkspace-controller,resources=deployments/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=limitranges,verbs=list;watch
 /////// Required permissions for workspace ServiceAccount
 // +kubebuilder:rbac:groups="",resources=pods/exec,verbs=create
 // +kubebuilder:rbac:groups=apps;extensions,resources=replicasets,verbs=get;list;watch
@@ -134,6 +135,12 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		reconcileStatus.addWarning(fmt.Sprint("Error applying external DevWorkspace-Operator configuration: ", err.Error()))
 		config = wkspConfig.GetGlobalConfig()
 	}
+	if updatedConfig, err := wkspConfig.UpdateConfigDefaultsForNamespace(rawWorkspace.Namespace, config, r.Client); err != nil {
+		reqLogger.Error(err, "Failed to update configuration defaults for namespace")
+	} else {
+		config = updatedConfig
+	}
+
 	configString := wkspConfig.GetCurrentConfigString(config)
 	workspace := &common.DevWorkspaceWithConfig{}
 	workspace.DevWorkspace = rawWorkspace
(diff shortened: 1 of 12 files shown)


tolusha:
You starts watching all LimitRange objects which can affect DWO memory usage.

amisevsk:
Yeah this is the tradeoff (and partly why I don't think considering limitranges is a good approach in general).

Either we cache all limitranges (which could potentially take a lot of memory) or make API requests on every reconcile (which leads to a variety of problems). 

With caching, we can expect an increase in memory usage on the order of ~200 MB depending on cluster size (small clusters shouldn't really have a problem)

Without caching, we have to worry about DWO making _a lot_ of requests against the API in unpredicable patterns, and also tie the reconcile loop to API latency, which could cause instability in starting workspaces.

IMO it's sufficient that the project-clone container can be configured to not have a CPU limit, and we shouldn't be looking at limitranges at all. We should do the safe, compatible thing in default DWO containers (set a CPU limit that is sufficient for the performance we expect) and allow a way to remove the CPU limit when desired.

cc: @l0rd any thoughts here?

l0rd:
I agree with you. Watching the LimitRanges is overkill and may have some major side effects.  

That's my proposal: when the operator is deployed, based on the information that we have, we should decide (in advance) if the default CPU limits for workspaces UDI, project-clone and gateway containers should be set or not. How can we decide that when the developer namespaces don't exist yet? Well my guess would be that, if the `CheCluster` namespace has a default container CPU limit, then developers namespaces will have it too. Of course this is an extremely weak assumption so maybe it's better to do what @amisevsk proposes. But not setting the CPU limits is so better, it provides such a better UX, that I feel bad that we set it by default on the workspaces containers. Of course admins should have a way to enforce the default behavior (that can be setting a specific CPU limit or not setting it at all).

Out of curiosity: do you know if Developer Sandbox has a default CPU limit in the `openshift-devspaces` namespace?

amisevsk:
> Out of curiosity: do you know if Developer Sandbox has a default CPU limit in the openshift-devspaces namespace?

I just double-checked, and sadly the sandbox clusters don't have limitranges in non-user namespaces, so this heuristic wouldn't work there. 

Currently, the places CPU limits are used in DWO are
* The Operator's controller/webhooks containers have CPU limits set -- I'd love to be able to remove these to see if it improves workspace startup time, but they're defined in the CSV itself so it's likely a no-go.
* The project-clone container uses a default CPU limit to be safe, but this can be disabled via the DevWorkspaceOperatorConfig CR (and this field is supported by Eclipse Che as of the last release). I'm not sure how much of an impact the CPU limit has here, as this container may be bound by network rather than CPU (we should test this).

In particular, we _don't_ set CPU limits on workspace containers by default, so if a DevWorkspace is created that e.g. doesn't set a limit on the `tools` container, that container will be run without a limit (the same applies for CPU requests, actually). 

Related to this issue I think is the issue @AObuchow is working on: https://github.com/devfile/devworkspace-operator/issues/1056. Currently, if we merge a contribution (e.g. the editor) into the default development container, we get a CPU limit from that container that we don't necessarily want. Once #1056 is resolved, we should be able to run any non-project-clone containers without a CPU limit by just not setting a CPU limit in devfiles.

amisevsk:
Closing this pull request as I don't think it's something we want to do in DWO; CPU limits can now be configured for every container so it's possible to remove them as required.

PR #1129, opened 2023-06-19
Only contribute to resources defined in merge target

### LINKED ISSUE #1056: Don't add contributions' resource limits/requests to containers that do not define them

AObuchow:
### What does this PR do?

This PR modifies the container contribution functionality for resource requirements. 

With this change, the merge target's resource requirements will be summed with contribution resource requirements **only for resource limits & requests that are already set in the merge target**. For example, if the merge target **does not** set a CPU request and a container contribution **does** set a CPU request, the resulting CPU request of the merge target will remain unset. 

This PR also moves functions related to `corev1.ResourceRequirements` into `/pkg/library/resources/`. 

Fix #1056

### Is it tested? How?
I added an automated test for all 4 types of container contributions (normal, spec, implicit and implicit-spec) which verifies that resource requirements are only summed if they are defined in the merge target component. 

Additionally, the Che-Code container contribution test cases had to be modified by removing the expected CPU limit, CPU request and Memory Request as they are no longer contributed by the container contributions (since the `tools` container only defines a Memoy Limit)

#### Verify #1056 is fixed
1. Install DWO with the changes from this PR
2. Apply the Che-Code sample `kubectl apply -f ./samples/code-latest.yaml -n $NAMESPACE`
3. Check the spec of the `code-latest` workspace (`kubectl describe code-latest -n $NAMESPACE`) and ensure the `dev` container component does not have its CPU Request, CPU Limit, Memory Request or Memory Limit set.

### DIFF
--- a/pkg/library/flatten/helper.go
+++ b/pkg/library/flatten/helper.go
@@ -75,50 +75,49 @@ func parseResourcesFromComponent(component *dw.Component) (*corev1.ResourceRequi
 	if component.Container == nil {
 		return nil, fmt.Errorf("attemped to parse resource requirements from a non-container component")
 	}
+
+	resources := &corev1.ResourceRequirements{
+		Limits:   corev1.ResourceList{},
+		Requests: corev1.ResourceList{},
+	}
+
 	memLimitStr := component.Container.MemoryLimit
-	if memLimitStr == "" {
-		memLimitStr = "0Mi"
+	if memLimitStr != "" {
+		memoryLimit, err := resource.ParseQuantity(memLimitStr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse memory limit for container component %s: %w", component.Name, err)
+		}
+		resources.Limits[corev1.ResourceMemory] = memoryLimit
 	}
+
 	memRequestStr := component.Container.MemoryRequest
-	if memRequestStr == "" {
-		memRequestStr = "0Mi"
+	if memRequestStr != "" {
+		memoryRequest, err := resource.ParseQuantity(memRequestStr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse memory request for container component %s: %w", component.Name, err)
+		}
+		resources.Requests[corev1.ResourceMemory] = memoryRequest
 	}
+
 	cpuLimitStr := component.Container.CpuLimit
-	if cpuLimitStr == "" {
-		cpuLimitStr = "0m"
+	if cpuLimitStr != "" {
+		cpuLimit, err := resource.ParseQuantity(cpuLimitStr)
+		if err != nil {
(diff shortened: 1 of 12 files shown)


amisevsk (approved):
LGTM :+1:

l0rd (approved):
I am thinking out loud here but I am trying to think about any risk that this PR could introduce.

Not setting requests can have severe consequences. The scheduler cannot figure out how much memory and CPU will be used by Pods and is going to provision workloads on nodes that are out of memory/cpu. 

The developer that writes the devfile is probably not going to specify a request at first. He will specify the image, maybe some env variables but the memory is something that I think we should not require.

This is a problem we have seen recently on a customer cluster. They had no clue why it was happening but Dev Spaces was unstable and developers had errors when trying to start workspaces.

That said, although this PR may exacerbate the problem, I think that we are doing the right thing from a DevWorkspace Operator point of view. "Fixing" devfiles that do not have requests set is more of a Che problem (devworkspace generator problem) and we should address it there.

amisevsk:
@l0rd you raise an important point, and I think there is room for improvement here in the future. Prior to this PR, there were two competing elements -- the default limits/requests used for containers in DWO:
```yaml
resources:
  limits:
    memory: 128M
  requests:
    memory: 64M
```
and the issue being fixed here, which is that we were adding any limits/requests in contributions even if there was nothing to add it to (i.e. we were interpreting an unset CPU limit as "0" CPU, and adding to that). With this PR, we've addressed one side of the issue (contributions can restrict resources unexpectedly) but we should still address the other side (defaults can restrict resources semi-unexpectedly)

I think one way to improve what we've got here would be

1. Make container default resources configurable (via DWOC), use these values for containers that do not set a field
2. Use this default as the base to add contributions' requests/limits to if a field is unset

I'll try to put together a quick PR to test this as a solution and we can discuss the approach there.

AObuchow:
> I think one way to improve what we've got here would be
> 
>     1. Make container default resources configurable (via DWOC), use these values for containers that do not set a field
> 
>     2. Use this default as the base to add contributions' requests/limits to if a field is unset
> 
> 
> I'll try to put together a quick PR to test this as a solution and we can discuss the approach there.

This sounds like a good idea to me :+1: 

If you get caught up with other tasks and want me to give a try at this, let me know :)

PR #1131, opened 2023-06-22
Ensure all samples have memory limit and request

amisevsk:
### What does this PR do?
Ensure all samples have a memory limit and request defined in order to ensure they do not use the default value.

Changes introduced in b62f3d0e mean that any container component that does not define requests/limits will use default values, which results in a too-low memory limit for running editors such as Code.

Issue with samples introduced in #1129 

### Is it tested? How?
Apply samples to cluster and ensure editor starts and is usable.

### PR Checklist

### DIFF
--- a/samples/theia-latest.yaml
+++ b/samples/theia-latest.yaml
@@ -14,6 +14,9 @@ spec:
       - name: dev
         container:
           image: quay.io/devfile/universal-developer-image:latest
+          memoryLimit: 512Mi
+          memoryRequest: 256Mi
+          cpuRequest: 1000m
     commands:
       - id: say-hello
         exec:
--- a/samples/code-latest.yaml
+++ b/samples/code-latest.yaml
@@ -14,6 +14,9 @@ spec:
       - name: dev
         container:
           image: quay.io/devfile/universal-developer-image:latest
+          memoryLimit: 512Mi
+          memoryRequest: 256Mi
+          cpuRequest: 1000m
     commands:
       - id: say-hello
         exec:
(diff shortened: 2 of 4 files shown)


RomanNikitenko (approved):
tested https://raw.githubusercontent.com/devfile/devworkspace-operator/502de29cb035866876916a8176805d3566a33db5/samples/code-latest.yaml
on the `dogfooding` instance - it fixes the problem!

samples/theia-latest.yaml
AObuchow, line 1: The `theia-latest` is currently failing because the contributing devfile https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/eclipse/che-theia/latest/devfile.yaml is no longer being hosted. I don't see Theia anywhere on https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/, but it might be hosted elsewhere.  

We might want to consider removing theia-latest as part of https://github.com/eclipse/che/issues/21973

AObuchow (approved):
Looks good to me though, it might be worth removing the theia-latest sample.

I was going to suggest adding the missing CPU & Memory limits/requests to the other samples that use container contributions (`ephemeral-storage.yaml`, `per-workspace-storage.yaml`, `git-clone-sample.yaml` and `pod-overrides.yaml`) but I realized these don't have merge target components.

samples/theia-latest.yaml
amisevsk: Good idea -- I guess it's fully no longer supported.

Issue #1133, opened 2023-06-22
Make default container resource requirements configurable


amisevsk:
### Description
The DevWorkspace Operator currently [hard-codes the default resources](https://github.com/devfile/devworkspace-operator/blob/9439f60425cef498646ef0c87544839cca99cab5/pkg/constants/constants.go#L46-L50) used for container components that do not explicitly define a memory/cpu limit/request.

It would be more useful to be able to configure these default values from the DevWorkspaceOperatorConfig.

### Additional context
Originally brought in PR review for https://github.com/devfile/devworkspace-operator/pull/1129#pullrequestreview-1487863143

PR #1134, opened 2023-06-22
Allow configuration of default container resources and improve resource handling when merging components

amisevsk:
Closes https://github.com/devfile/devworkspace-operator/issues/1133

### Is it tested? How?
1. Verify that default container resources can be configured via DWOC and that values set here are propagated to the workspace pod's containers (when a devworkspace does not specify limits/requests)
2. Verify that when merging container contributions, the total resource requirements are the sum of the default value and whatever values are defined in the contribution (i.e. container _does not_ have default values) -- see https://github.com/devfile/devworkspace-operator/pull/1131
   * For example, a workspace created from [this sample](https://raw.githubusercontent.com/devfile/devworkspace-operator/2094e37953d61664d3b23014ace2aa55f780acff/samples/code-latest.yaml) (without the PR above) should have resources equal to
      ```yaml
      memoryLimit: 1024Mi
      memoryRequest: 256Mi
      cpuLimit: 500m
      cpuRequest: 30m
      ```
        plus the default values.

Basic testing can be done using the following

* DevWorkspaceOperatorConfig:
  ```yaml
  metadata:
  config:
    workspace:
      defaultContainerResources:
        limits:
          memory: 200Mi
          cpu: 200m
        requests:
          memory: 100Mi
          cpu: 100m

  ```

* DevWorkspace:
  ```yaml
  kind: DevWorkspace
  metadata:
    name: plain
  spec:
    started: true
    routingClass: 'basic'
    template:
      attributes:
      components:
        - name: merge-target
          attributes:
            controller.devfile.io/merge-contribution: "true"
          container:
            command:
            - "tail"
            - "-f"
            - "/dev/null"
        - name: merge-contribution
          attributes:
            controller.devfile.io/container-contribution: "true"
          container:
            memoryLimit: 555Mi
            memoryRequest: 55Mi
            cpuLimit: 555m
            cpuRequest: 55m
            command:
            - "tail"
            - "-f"
            - "/dev/null"
        - name: all-defaults-needed
          container:
            command:
            - "tail"
            - "-f"
            - "/dev/null"
        - name: no-defaults-needed
          container:
            memoryLimit: 555Mi
            memoryRequest: 55Mi
            cpuLimit: 555m
            cpuRequest: 55m
            command:
            - "tail"
            - "-f"
            - "/dev/null"
  ```

The resulting pod should have 
* Specified resources for the `no-defaults-needed` container
* All default values for the `all-defaults-needed` container
* For the `merge-target` container, the resource requirements should be the sum of the default values (in the DWOC) and the values added by the `merge-contribution` component.

### PR Checklist


pkg/library/resources/helper.go
```diff
@@ -151,24 +149,29 @@ func ProcessResources(resources *corev1.ResourceRequirements) (*corev1.ResourceR
...
 	}
 	if len(result.Requests) == 0 {
 		result.Requests = nil
 	}
 
-	return result, nil
+	return result
+}
+
+// ValidateResources validates that a corev1.ResourceRequirements is valid, i.e. that (if specified), limits are greater than requests
```
AObuchow, line 223: Resource requirement requests are still valid if they are equal to the limit.
The implementation of this function follows this rule, however the documentation and error message should probably be modified.

i.e.

`// ValidateResources validates that a corev1.ResourceRequirements is valid, i.e. that (if specified), limits are greater than or equal to requests`

pkg/library/resources/helper.go
```diff
@@ -151,24 +149,29 @@ func ProcessResources(resources *corev1.ResourceRequirements) (*corev1.ResourceR
...
+func ValidateResources(resources *corev1.ResourceRequirements) error {
+	memLimit, hasMemLimit := resources.Limits[corev1.ResourceMemory]
+	memRequest, hasMemRequest := resources.Requests[corev1.ResourceMemory]
+	if hasMemLimit && hasMemRequest && memRequest.Cmp(memLimit) > 0 {
+		return fmt.Errorf("memory request (%s) must be less than limit (%s)", memRequest.String(), memLimit.String())
```
AObuchow, line 228: Similarly, here: `memory request (%s) must be less than or equal to limit (%s)`

pkg/library/resources/helper.go
```diff
@@ -151,24 +149,29 @@ func ProcessResources(resources *corev1.ResourceRequirements) (*corev1.ResourceR
...
+	}
+
+	cpuLimit, hasCPULimit := resources.Limits[corev1.ResourceCPU]
+	cpuRequest, hasCPURequest := resources.Requests[corev1.ResourceCPU]
+	if hasCPULimit && hasCPURequest && cpuRequest.Cmp(cpuLimit) > 0 {
+		return fmt.Errorf("CPU request (%s) must be less than limit (%s)", cpuRequest.String(), cpuLimit.String())
```
AObuchow, line 234: `CPU request (%s) must be less than or equal to limit (%s)`*

AObuchow (changes requested):
Minor comments on wording, but looks great to me. Very satisfying refactor as well :sunglasses:. Tested on OpenShift 4.12.

AObuchow (approved):
Approving this PR as it works as intended, though note my comments on wording.

pkg/library/resources/helper_test.go
```diff
@@ -0,0 +1,371 @@
...
+			expected:  getResourceRequirements("250Mi", "450Mi", "250m", "450m"),
+		},
+		{
+			name:      "Does not add resources if not defined in base",
+			resources: getResourceRequirements("", "", "", ""),
+			toAdd:     getResourceRequirements("150Mi", "250Mi", "150m", "250m"),
+			expected:  getResourceRequirements("", "", "", ""),
+		},
+		{
+			// Not sure if this is ultimately the desired behavior here (maybe depends on usage)
```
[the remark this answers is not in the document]
amisevsk: This makes sense.

My concern here is specifically about our overloading of the meaning of "0" in resources -- sometimes it means "do not set a limit/request", and sometimes it's interpreted as "zero". It makes using `AddResources` a little more confusing, as doing
```
FilterResources(resources)
AddResources(resources, toAdd)
```
will have a different result than
```
AddResources(resources, toAdd)
FilterResources(resources)
```

pkg/library/resources/helper.go
```diff
@@ -78,34 +79,30 @@ func ParseResourcesFromComponent(component *dw.Component) (*corev1.ResourceRequi
...
-// Adds the resource limits and requests that are set in the component "toAdd" to "resources".
+// Adds the resource limits and requests that are set in the component toAdd to resources.
 // Returns an error if the resources defined in toAdd could not be parsed.
 //
-// Note: only resources that are set in the function argument "resources" will be summed
-// with the resources set in "toAdd".
-// For example, if "resources" does not set a CPU limit but "toAdd" does set a CPU limit,
+// Note: only resources that are set in the argument resources will be summed with the resources set in toAdd.
+// For example, if resources does not set a CPU limit but toAdd does set a CPU limit,
 // the CPU limit of "resources" will remain unset.
```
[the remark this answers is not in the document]
amisevsk: Good catch, though the idea of unquoting resources here makes me realize why you quoted them in the first place :D 

I might just switch back to quoted everywhere -- it's an annoying thing about godoc:

>A func’s doc comment should explain what the function returns or, for functions called for side effects, what it does. **Named arguments or results can be referred to directly in the comment**, without any special syntax like backquotes. (A consequence of this convention is that names like a, which might be mistaken for ordinary words, are typically avoided.)

pkg/library/resources/helper.go
amisevsk: Good catch, will update :+1:

pkg/library/resources/helper.go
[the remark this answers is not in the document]
amisevsk: Yup, quote is easier to read since we use "resources" as both a variable name and generic term :/

PR #1143, opened 2023-07-07
fix: replace deprecated Ingress annotation

gbonnefille:
### What does this PR do?

Before the IngressClass resource and ingressClassName field were added in Kubernetes 1.18, Ingress classes were specified with a kubernetes.io/ingress.class annotation on the Ingress. This annotation was never formally defined, but was widely supported by Ingress controllers.

The newer ingressClassName field on Ingresses is a replacement for that annotation, but is not a direct equivalent. While the annotation was generally used to reference the name of the Ingress controller that should implement the Ingress, the field is a reference to an IngressClass resource that contains additional Ingress configuration, including the name of the Ingress controller.

Cf. https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

### PR Checklist

### DIFF
--- a/controllers/controller/devworkspacerouting/solvers/basic_solver.go
+++ b/controllers/controller/devworkspacerouting/solvers/basic_solver.go
@@ -31,7 +31,6 @@ var routeAnnotations = func(endpointName string) map[string]string {
 
 var nginxIngressAnnotations = func(endpointName string) map[string]string {
 	return map[string]string{
-		"kubernetes.io/ingress.class":                "nginx",
 		"nginx.ingress.kubernetes.io/rewrite-target": "/",
 		"nginx.ingress.kubernetes.io/ssl-redirect":   "false",
 		constants.DevWorkspaceEndpointNameAnnotation: endpointName,
--- a/controllers/controller/devworkspacerouting/solvers/common.go
+++ b/controllers/controller/devworkspacerouting/solvers/common.go
@@ -25,6 +25,7 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 )
 
 type DevWorkspaceMetadata struct {
@@ -223,6 +224,7 @@ func getIngressForEndpoint(routingSuffix string, endpoint controllerv1alpha1.End
 			Annotations: nginxIngressAnnotations(endpoint.Name),
 		},
 		Spec: networkingv1.IngressSpec{
+			IngressClassName: pointer.String("nginx"),
 			Rules: []networkingv1.IngressRule{
 				{
 					Host: hostname,


gbonnefille:
Yes, finding a solution for #1068 could be really interesting.

Having in mind that the availability of Kind=ingressClass could be a solution to discover the default IngressClassName instead of having the user to configure it, I started reviewing the code. And it is hard to find an elegant solution. Because I will need to search for the Kind=IngressClass from solver implementations, but in this context we do not have simple access to the K8S itself (for example, for requesting the list of Kind=IngressClass).

I also realized that a good implementation would need to request the K8S in order to evaluate if we are using `networking.k8s.io/v1` (and so `ingressClassName:`) or  a previous version (and thus `kubernetes.io/ingress.class` annotation). But the same here: no simple way to request the K8S from solver implementations.

And finally, as Ingress is a mess with many implementation specific annotations, having a basic_router without any hardcoded values, but a good way to configure any aspect could help a lot.

amisevsk:
Yeah, making manual configuration possible would be ideal, though as you say the variety of ingress implementations is kind of a mess, and anything we do around this would _also_ require work in whatever operators depend on us (and implement a routing reconcile -- e.g. Che Operator). 

It'd be great if there was a way to automatically detect e.g. traefik vs nginx and apply the appropriate annotations, though I don't know if it's possible on Kubernetes (on OpenShift, we use e.g. a test route to discover the cluster's base hostname).

gbonnefille:
> Yeah, making manual configuration possible would be ideal, though as you say the variety of ingress implementations is kind of a mess, and anything we do around this would _also_ require work in whatever operators depend on us (and implement a routing reconcile -- e.g. Che Operator).

Yes, sure.

> It'd be great if there was a way to automatically detect e.g. traefik vs nginx and apply the appropriate annotations, though I don't know if it's possible on Kubernetes (on OpenShift, we use e.g. a test route to discover the cluster's base hostname).

First, the IngressClass' name. As spotted by https://kubernetes.io/docs/concepts/services-networking/ingress/, the `spec.ingressClassName:` attribute is *mostly* mandatory. When present, it should be respected. When absent, a default IngressClass will be selected. Recently, I noticed that ingress-nginx really needs it. If absent, traefik is able to update the resource and set it.

With the current fix proposed by this PR, a K8S cluster admin is able to have basic routing solver works, as he only have to declare a Kind=IngressClass with name nginx and routing to the traefik.

Other ideas:
1. exclude the `spec.ingressClassName` attribute from the reconciling part, in order to let ingress controler fix it.
2. avoid an hardcoded value and pick one from the installed Kind=IngressClass resources, if there is a default.

But it could be tricky to auto-discover and if the automatic job is not the right one, there is no chance to manually adapt it.

One possible other solution is to use a dedicated IngressClass resource, with name `dwo-ingress` for example. The generated Ingress then have hardcoded `spec.ingressClassName: dwo-ingress`. The initial installation/start of the operator create the IngressClass with name `dwo-ingress` if absent, referencing nginx or trying to discover an ingress controler.
And then, if the K8S admin has a different opinion, he is able to update the IngressClass with name `dwo-ingress` to reflect his better choice.

Concerning the other annotations (`nginx.ingress.kubernetes.io/rewrite-target` and  `nginx.ingress.kubernetes.io/ssl-redirect`) I have no idea. Traefik has also a lot of annotations: https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/
If it is not possible to manually configure them, perhaps should we add (hardcoded) all known annotations for any known ingress controller (currently nginx and traefik). For example `traefik.ingress.kubernetes.io/router.tls: false`. But I fear it could be rapidly related to the effective configuration of the ingress controler.

gbonnefille:
Please, note that cert-manager, an other controller having to create Ingress in order to solve the ACME challenge, has a setting for `ingressClassName`: https://cert-manager.io/docs/configuration/acme/http01/#ingressclassname

amisevsk:
Merging this PR as it fixes a problem (thanks again!). I'll try to schedule some work on configuring ingress annotations/ingress class when I have some time.

Issue #1121, opened 2023-06-09
Git operations show warning when using credentials mounted by DWO


amisevsk:
### Description
When git credentials are added to a workspace ([doc](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#adding-git-credentials-to-a-workspace)), the credentials file is necessarily mounted read-only to the workspace containers. 

This causes git operations, both in the running workspace and in the project-clone init container, to log 
```
fatal: unable to get credential storage lock in 1000 ms: Read-only file system
```
This message does not prevent the action from succeeding (i.e. the clone/fetch/push succeeds) but the message is confusing and suggests an unexpected error.

### How To Reproduce
1. Configure git credentials via personal access token as [documented](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#adding-git-credentials-to-a-workspace)
2. Start a workspace that clones a git repository
3. Note project clone logs contain message listed above
4. Attempt to `git fetch` within the cloned project, note error printed as above.

### Expected behavior
We should find a way of either ignoring or suppressing this error. 

One approach would be to
1. Mount credentials to some other path, and
2. Copy file to the path git expects with read-write permissions on startup
However, this approach would prevent on-cluster updates to the git credentials from automatically propagating to the workspace, which is a requirement (see: https://github.com/devfile/devworkspace-operator/issues/915)

PR #1145, opened 2023-07-14
Wrap git credential helper to ignore 'store' and 'erase' operations

amisevsk:
### What does this PR do?
Since the credentials file is mounted read-only to the filesystem, we need to configure a git credential helper that ignores 'store' and 'erase' operations in order to avoid printing error messages around getting a lock on the credentials file:

```
fatal: unable to get credential storage lock in 1000 ms: Read-only file system
```

In order to do this, we wrap the existing 'store' credential helper to effectively be read-only.

Closes https://github.com/devfile/devworkspace-operator/issues/1121

### Is it tested? How?
To test:
1. Configure a personal access token secret in the test namespace
2. Start a workspace that clones a project that uses that token

Verify:
1. Project clone container logs do not contain error message above
2. Execing into the container and running git commands that require credentials (e.g. git pull) do not show the error message.

### PR Checklist

### DIFF
--- a/pkg/provision/automount/templates.go
+++ b/pkg/provision/automount/templates.go
@@ -41,8 +41,12 @@ const gitLFSConfig = `[filter "lfs"]
     required = true
 `
 
+// Since we're mounting the credentials file read-only, we need to ignore 'store'
+// and 'erase' commands to the credential helper (will print an error about being
+// unable to get a lock on the credentials file). This snippet effectively wraps
+// only 'git credential-store get' to make it read-only.
 const credentialTemplate = `[credential]
-    helper = store --file %s
+    helper = "!f() { test \"$1\" = get && git credential-store --file %s \"$@\"; }; f"
 `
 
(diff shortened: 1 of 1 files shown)


AObuchow (approved):
Looks good to me. I setup a PAT secret and created a devworkspace that was using a private repo with an HTTPS git URL.

Before the patch, this resulted in `fatal: unable to get credential storage lock in 1000 ms: Read-only file system` appearing in the project-clone logs, and pulling a branch from my repo also gave that error.

With the patch applied, the error no longer occurs in the project-clone logs or when doing a git pull on the repo.

lobanov:
I stumbled on this issue in a completely different context. I'm running JupyterHub on Kubernetes and mounting git credentials to pods running user notebooks, so the users could use git commands. The credentials were mounted as secrets, so the are read-only, and this error message was a nuisance. Thank you for neat workaround!

Issue #1146, opened 2023-07-17
It should be possible to configure operator to ignore OpenShift cluster-wide proxy config


amisevsk:
### Description
On OpenShift clusters, the DevWorkspace Operator reads the `cluster` proxy and uses values from there to configure `httpProxy`, `httpsProxy`, and `noProxy` for itself and workspace containers. The DevWorkspace Operator Config CR can additionally be used to override `httpProxy` and `httpsProxy`, and to add values to `noProxy`, but there is no convenient way to override these fields to be unset -- setting e.g. `httpProxy` to `""` will result in DWO treating the field as if it were unset, continuing to use the cluster-detected value.

Instead, setting any proxy field to the empty string (`""`) should result in DWO ignoring any values detected from the cluster.

Currently, there is a workaround: if the field is set to `" "` (i.e. whitespace only), DWO will use this value for proxy environment variables, effectively not propagating proxy configuration into workspaces.

### Additional context

N/A

PR #1147, opened 2023-07-17
Allow cluster-detected proxy settings to be ignored in OpenShift

amisevsk:
### What does this PR do?
Converts the `httpProxy`, `httpsProxy` and `noProxy` fields in the DevWorkspaceOperatorConfig CR to be string pointers (internally), which allows us to distinguish between "not set" and "set explicitly to be empty". When any field is the empty string, DWO now interprets this as explicitly not wanting to set proxy environment variables for workspaces, ignoring the value detected from the cluster-wide ("`cluster`") proxy.

Closes https://github.com/devfile/devworkspace-operator/issues/1146

### Is it tested? How?
Test that DWO can be configured to ignore cluster-wide proxy settings in a proxy-enabled cluster. Setting any field to the empty string in the DWOC should result in the corresponding environment variables being unset in workspace containers. Otherwise, functionality should be unchanged.

### PR Checklist


AObuchow:
Code looks good to me, will do final testing tomorrow :+1:

AObuchow (approved):
Looks good to me & works as expected on a proxy configured cluster :+1:

Issue #1158, opened 2023-08-03
Include devfile parent in potential implicit merge contribution targets


amisevsk:
### Description
The DevWorkspace Operator explicitly excludes imported components from targets for implicitly-merged contributions. However, in many cases it would be useful to merge contributions into components imported by a parent devfile, as the parent often holds the main development container.

### Additional context
Eclipse Che issue: https://github.com/eclipse/che/issues/22371

PR #1159, opened 2023-08-03
Include parent-imported components in implicit merge targets

amisevsk:
### What does this PR do?
Include components that were imported by a parent in potential targets for implicit merging of container contributions

Closes https://github.com/devfile/devworkspace-operator/issues/1158

### Is it tested? How?
I haven't had time to test this PR thoroughly; it should be ensured that including parents in merge targets still behaves as expected:
* What happens if a devfile contains both a component and a parent? Is the merge processed on the "expected" component?
  * E.g. if we have a parent devfile and add e.g. a 'database' container, is merging still processed as expected?

### PR Checklist

### DIFF
--- a/pkg/library/flatten/merge.go
+++ b/pkg/library/flatten/merge.go
@@ -128,10 +128,10 @@ func mergeVolume(into, from *dw.VolumeComponent) error {
 // needsContainerContributionMerge returns whether merging container contributions is necessary for this workspace. Merging
 // is necessary if the following two conditions are met:
 //
-// - At least one component has the container-contribution: true attribute.
+//   - At least one component has the container-contribution: true attribute.
 //
-// - At least one component has the merge-contribution: true attribute OR there exists a container component that was not imported by a
-// plugin or parent devworkspace.
+//   - At least one component has the merge-contribution: true attribute OR there exists a container component that was not imported by a
+//     plugin devworkspace.
 //
 // If the container-contribution or merge-contribution attribute are present but cannot be parsed as a bool, an error is returned.
 // If multiple components have the merge-contribution: true attribute, an error is returned.
@@ -165,7 +165,8 @@ func needsContainerContributionMerge(flattenedSpec *dw.DevWorkspaceTemplateSpec)
 				return false, fmt.Errorf("failed to parse %s attribute on component %s as true or false", constants.MergeContributionAttribute, component.Name)
 			}
 		} else {
-			if !component.Attributes.Exists(constants.PluginSourceAttribute) {
+			pluginSource := component.Attributes.GetString(constants.PluginSourceAttribute, nil)
+			if pluginSource == "" || pluginSource == "parent" {
 				// First, non-imported container component is implicitly selected as a contribution target
 				hasTarget = true
 			}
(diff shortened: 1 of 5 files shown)


AObuchow:
> What happens if a devfile contains both a component and a parent? Is the merge processed on the "expected" component?

I'm in the process of writing test cases right now, but it seems that the parent component will be selected as the implicit merge target, i.e.

```YAML
  spec:
    parent:
      uri: parent.yaml # Parent's container component will get selected as implcit merge target
    components:
      - name: first-child-component # Won't get selected as an implicit merge target
        container:
          memoryLimit: 2Gi
          mountSources: true
  
  contributions:
    - name: che-code
      uri: che-code.yaml
```

I imagine we want to prioritize implicitly merging into components defined in the "child" devworkspace, rather than the parent right?  I can try and fix this.

pkg/library/flatten/merge.go
```diff
@@ -165,7 +165,8 @@ func needsContainerContributionMerge(flattenedSpec *dw.DevWorkspaceTemplateSpec)
 				return false, fmt.Errorf("failed to parse %s attribute on component %s as true or false", constants.MergeContributionAttribute, component.Name)
 			}
 		} else {
-			if !component.Attributes.Exists(constants.PluginSourceAttribute) {
+			pluginSource := component.Attributes.GetString(constants.PluginSourceAttribute, nil)
+			if pluginSource == "" || pluginSource == "parent" {
```
AObuchow, line 169: It doesn't really sit well to me that the merge target found by `needsContainerContributionMerge()` and  `findMergeTarget` may be different, but I'm not sure if that affects things. Regardless, I'm going to fix this disparity shortly

amisevsk:
> I imagine we want to prioritize implicitly merging into components defined in the "child" devworkspace, rather than the parent right? I can try and fix this.

I think the only path forward that remains consistent + configurable would be to merge into parents first; thinking through three cases:

Let's call the approach where we always try to merge into the main devfile first "Plan A" and the approach where we merge into parent first "Plan B"

* Parent devfile + no components in main devfile: No confusion here, either plan is equivalent
* Parent devfile + a component you want to merge into inside the main devfile:
  * Plan A: this is the normal behavior
  * Plan B: you would need to add the explicit `merge-contributions` attribute to your devfile's component (easy to do)
* Parent devfile + a component in main devfile; you want to merge into the parent devfile's component
  * Plan A: the only way you can do it is by modifying the parent devfile to have the `merge-contributions` attribute (hard to do; fragile as parent cannot be used in devfiles that use the `merge-contributions` attr)
  * Plan B: normal behavior; parent component will be used as merge target

AObuchow:
@amisevsk I agree with your reasoning, good call. Will make the appropriate changes

AObuchow:
I verified that this patch fixes the issue in https://github.com/eclipse/che/issues/22371 by installing Che with patched DWO (from this PR) and creating a workspace from this [devfile](https://github.com/AObuchow/dw-parent-contribution/blob/main/devfile.yaml). I saw that the `che-code-runtime-description` container was now being contributed to the tools container, and not being run as a standalone container:

image

AObuchow (approved):

PR #1166, opened 2023-08-17
Handle failures in routing reconciler in recoverable way

### LINKED ISSUE #923: DevWorkspaceRouting can get stuck in `failed` state
### Description
The DevWorkspaceRouting reconciler can fail to start workspaces (if e.g. the devworkspacerouting is misconfigured), which results in the DevWorkspace entering a failed state. However, when the DevWorkspaceRouting has `.status.phase` set to `failed`, no further reconciles are performed and so it is not possible to restart the workspace.

This is not reproducible with only the DevWorkspace Operator running as DevWorkspaceRoutings for a given spec always either succeed or fail -- e.g. using the `cluster-tls` routingClass on Kubernetes will always fail, and switching to `basic` will result in the DevWorkspaceRouting being deleted and re-created. However, for other dwr reconcilers, this is not always the case, and transient failures are possible.

### How To Reproduce
1. Install DWO and the Eclipse Che operator in a cluster
2. Create two CheCluster CRs in the cluster and wait for them to be set up. This will cause all `routingClass: che` DWRs to fail during reconcile.
3. Create a DevWorkspace with `routingClass: che`. Workspace will fail to start
4. Delete one of the two CheClusters. While new `routingClass: che` workspaces will start successfully, it's not possible to restart the failed workspace.

### Expected behavior
Once issue with DevWorkspaceRouting is fixed, workspace can be started.

### Workarounds
1. Trigger any change to the DevWorkspaceRouting, either by modifying the devfile or the dwr itself. If DWO needs to update it, it will be re-created and the issue will be avoided.
2. Delete the DevWorkspaceRouting so that DWO will re-create it. This does not impact the DevWorkspace or any data stored within.
3. Enable the `config.workspace.cleanupOnStop` option in the DevWorkspaceOperatorConfig. **note:** this currently fails to workaround the issue in some cases due to https://github.com/devfile/devworkspace-operator/issues/922
4. Manually edit the status of the DevWorkspaceRouting to remove the failed phase

amisevsk:
### What does this PR do?
When a DevWorkspaceRouting fails to provision objects, it enters the 'Failed' phase. In order to prevent repeated unnecessary reconciles, the controller ignores DWRs that are already Failed, which leads to the 'Failed' phase being permanent.

To avoid this, we instead set a Stopped phase when the parent workspace is stopped. This will clear the Failed phase and any message that was set, so the DevWorkspace controller will need to propagate it into the DevWorkspace object to allow it to be shown.

Closes #923 

### Is it tested? How?
To test, follow the reproducer in #923 and verify this PR fixes the issue (allows failed workspaces to be restarted):

1. Install DWO and the Eclipse Che operator in a cluster
2. Create two CheCluster CRs in the cluster and wait for them to be set up. This will cause all routingClass: che DWRs to fail during reconcile.
3. Create a DevWorkspace with routingClass: che. Workspace will fail to start
4. Delete one of the two CheClusters. While new routingClass: che workspaces will start successfully, it's not possible to restart the failed workspace.

alternatively, this PR can be tested by patching the DevWorkspace controller pod to mark routing objects as failed instead of ready by replacing [this block](https://github.com/devfile/devworkspace-operator/blob/65bdb9e652d16e1f8d413ae77e92a01464f43c7c/controllers/controller/devworkspacerouting/devworkspacerouting_controller.go#L287-L291) with
```go
r.markRoutingFailed(instance, "testing failure")
```
in order to get a failed DevWorkspaceRouting. Then undo the patch and retest to ensure workspace can be restarted after failure.

In testing, when the DevWorkspaceRouting fails, the DevWorkspace should also be marked as failed, and the failure message on the DevWorkspace should include all applicable context from the DevWorkspaceRouting.

### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspacerouting_types.go
+++ b/apis/controller/v1alpha1/devworkspacerouting_types.go
@@ -62,6 +62,7 @@ const (
 	RoutingReady     DevWorkspaceRoutingPhase = "Ready"
 	RoutingPreparing DevWorkspaceRoutingPhase = "Preparing"
 	RoutingFailed    DevWorkspaceRoutingPhase = "Failed"
+	RoutingStopped   DevWorkspaceRoutingPhase = "Stopped"
 )
 
 type ExposedEndpoint struct {
--- a/controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
+++ b/controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
@@ -108,6 +108,9 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(ctx context.Context, req ctrl.
 	}
 
 	if instance.Annotations != nil && instance.Annotations[constants.DevWorkspaceStartedStatusAnnotation] == "false" {
+		if err := r.setStatusStopped(instance); err != nil {
+			return reconcile.Result{}, err
+		}
 		return reconcile.Result{}, nil
 	}
 
@@ -288,6 +291,14 @@ func (r *DevWorkspaceRoutingReconciler) reconcileStatus(
 	instance.Status.Message = "DevWorkspaceRouting prepared"
 	instance.Status.PodAdditions = routingObjects.PodAdditions
 	instance.Status.ExposedEndpoints = exposedEndpoints
+	return r.Status().Update(context.TODO(), instance)
+}
+
+func (r *DevWorkspaceRoutingReconciler) setStatusStopped(instance *controllerv1alpha1.DevWorkspaceRouting) error {
+	instance.Status.Phase = controllerv1alpha1.RoutingStopped
+	instance.Status.Message = "DevWorkspace is not started"
+	instance.Status.PodAdditions = nil
+	instance.Status.ExposedEndpoints = nil
 	return r.Status().Update(context.TODO(), instance)
 }
 


amisevsk:
I've also checked that this PR doesn't break `debug-start` functionality; if the `controller.devfile.io/debug-start: 'true'` annotation is applied to the DevWorkspace, the DevWorkspaceRouting stays in the `Failed` state with the old status until the workspace is stopped or the annotation is removed.

AObuchow (approved):
LGTM :+1: 
I was unable to reproduce the original bug following the [reproduer issue](https://github.com/devfile/devworkspace-operator/issues/923): Che's validating webhook prevents the creation of 2 Che Cluster CR's so I had to delete the webhook and quickly create another Che Cluster CR. Even then, the second Che Cluster CR wouldn't get fully set up (The dashboard deployment would endlessly be created and deleted), and creating a workspace with the che routing class wouldn't cause the workspace to fail.

However, I was able to patch the DWR as suggested in the second testing approach and got the workspace to fail with the appropriate message from the DWR controller:

```
plain-devworkspace   workspace56a1660c61624a42   Failed   Failed to set up networking for workspace: testing failure
```

I undid the patch, set the workspace to started and it deployed successfully as expected.

Issue #1093, opened 2023-04-27
Operator does not handle defaulted fields in readiness/liveness probes if unset


amisevsk:
### Description
If a container specifies readiness or liveness probes (via e.g. container-overrides, or potentially also in k8s components) and does not fully specify the probe, DWO can enter a loop of updating the deployment where it unsets defaulted fields (that are then automatically re-set by the kube API)

### How To Reproduce
Add container-overrides to a DevWorkspace container:
```yaml
        container-overrides:
          readinessProbe:
            httpGet:
              path: /healthz
              port: 3100
              scheme: HTTPS
            initialDelaySeconds: 1
            periodSeconds: 1
            failureThreshold: 300
```
DWO will repeatedly update the deployment, preventing it from starting:
1. DWO sets no `successThreshold` causing a deployment rollout
2. Cluster sets `successThreshold: 1` causing another rollout
3. DWO sees deployment doesn't match spec, updates it to remove `successThreshold`
4. etc.

PR #1169, opened 2023-08-24
Handle unset fields in probes set by container-overrides

amisevsk:
### What does this PR do?
Updates how probes (readiness, liveness, startup) are handled when defined in container-overrides. Since fields for success/failure threshold, period, and timeout are defaulted by the cluster, we have to make sure they're either set to user-provided values or the defaults to avoid repeatedly reconciling the deployment yaml

Closes https://github.com/devfile/devworkspace-operator/issues/1093

### Is it tested? How?
To test, you can use the workspace
```yaml
kind: DevWorkspace
metadata:
  name: test-dw
spec:
  started: true
  template:
    attributes:
    components:
      - name: terminal
        container:
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
        attributes:
          container-overrides:
            readinessProbe:
              exec:
                command: ["echo", "hello"]
            livenessProbe:
              exec:
                command: ["echo", "hello"]
            startupProbe:
              exec:
                command: ["echo", "hello"]
```
This workspace should start normally with defaulted fields for the probes.

### PR Checklist

### DIFF
--- a/pkg/library/overrides/containers.go
+++ b/pkg/library/overrides/containers.go
@@ -24,6 +24,16 @@ import (
 	"github.com/devfile/devworkspace-operator/pkg/constants"
 )
 
+// Constants representing the default values used by Kubernetes for container fields.
+// If a pod is created that does not set these fields in a Probe (but does create a probe)
+// these are the values that will be automatically applied to the pod spec.
+const (
+	defaultProbeSuccessThreshold = 1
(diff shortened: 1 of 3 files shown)


AObuchow (approved):
LGTM & works as expected :+1:

Issue #1170, opened 2023-09-01
The default SCC used by DWO should match OpenShift restricted-v2


amisevsk:
### Description
By default, DWO does not set the security context on workspace containers unless it is configured. This can lead to issues where a higher-priority and more-restrictive SCC on OpenShift can lead to unexpected restrictions on workspace containers (e.g. an SCC that sets `readOnlyRootFilesystem: true` can cause issues writing files outside of a persistent volume).

### Expected behavior
DWO should define the full security context it expects for workspace pods, and provide a way to update that if necessary.

PR #1171, opened 2023-09-05
Set default container security context on OpenShift

amisevsk:
### What does this PR do?
Set the default value for `ContainerSecurityContext` to match the security context required by OpenShift's `restricted-v2` SCC. This ensures that workspaces don't accidentally use a higher-priority, lower-permission SCC (e.g. one that requires `readOnlyRootFilesystem: true`)

Closes https://github.com/devfile/devworkspace-operator/issues/1170

### Is it tested? How?
I haven't had a chance to test it yet (to come, tomorrow)

To verify:
* On Kubernetes, behavior is unchanged, containers do not get anything set for container security context
* On OpenShift, verify that #1170 cannot be reproduced; pod/container spec should look the same as before as the restricted-v2 SCC sets fields matching the default
* Also necessary to verify that this change does not interfere with how Eclipse Che sets the default security context to allow for `podman build`

### PR Checklist

### DIFF
--- a/pkg/config/common_test.go
+++ b/pkg/config/common_test.go
@@ -55,6 +55,7 @@ func setupForTest(t *testing.T) {
 	}
 	infrastructure.InitializeForTesting(infrastructure.Kubernetes)
 	setDefaultPodSecurityContext()
+	setDefaultContainerSecurityContext()
 	configNamespace = testNamespace
 	originalDefaultConfig := defaultConfig.DeepCopy()
 	t.Cleanup(func() {
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -49,8 +49,8 @@ var defaultConfig = &v1alpha1.OperatorConfiguration{
 		IdleTimeout:              "15m",
 		ProgressTimeout:          "5m",
 		CleanupOnStop:            pointer.Bool(false),
-		PodSecurityContext:       nil,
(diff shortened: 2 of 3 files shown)


AObuchow (approved):
Works as expected :+1: 
## OpenShift 4.12
Testing out the plain-devworkspace sample on DWO, the workspace pod's space had the `openshift.io/scc: restricted-v2` as well as the default container security context:

```YAML
      securityContext:
        capabilities:
          drop:
            - ALL
        runAsUser: 1000680000
        runAsNonRoot: true
        readOnlyRootFilesystem: false
        allowPrivilegeEscalation: false
```

I also tested things out with a full Che install (creating the DWO catalog source wasn't working, so I had to install the DWO operator and modify the CSV to set the controller manager replicas to 0 and run the controller locally instead).
The SCC used was container-build (`openshift.io/scc: container-build`), and I was able to do a podman build. The container securityContext also had the appropriate fields set:

```YAML
      securityContext:
        capabilities:
          add:
            - SETGID
            - SETUID
          drop:
            - ALL
            - KILL
            - MKNOD
        runAsUser: 1000690000
        runAsNonRoot: true
        readOnlyRootFilesystem: false
        allowPrivilegeEscalation: true
```
## Minikube
On minikube, creating a devworkspace resulted in the workspace pod's container security context field being empty, as expected: `securityContext: {} `

PR #1181, opened 2023-10-04
Improve process for clearing 'started-at' annotations from workspaces

amisevsk:
### What does this PR do?
Avoid attempting to update stopped workspaces to remove started-at annotations if the annotation is not present on the workspace.

Minor fix to avoid sending unnecessary update requests.

### Is it tested? How?
Should be no functional change -- annotations should still be removed from stopped workspaces.

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -621,8 +621,14 @@ func (r *DevWorkspaceReconciler) syncStartedAtToCluster(
 func (r *DevWorkspaceReconciler) removeStartedAtFromCluster(
 	ctx context.Context, workspace *common.DevWorkspaceWithConfig, reqLogger logr.Logger) {
 	if workspace.Annotations == nil {
-		workspace.Annotations = map[string]string{}
+		// No annotations, nothing to do
+		return
 	}
+	if _, ok := workspace.Annotations[constants.DevWorkspaceStartedAtAnnotation]; !ok {
+		// Annotation has already been deleted
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


amisevsk:
In my brief testing, this change cuts 50-150ms from reconcile times for stopped workspaces (which, on most clusters, tends to be most of them). This especially helps during phases like startup, where it can take a while to process hundreds of stopped workspaces.

amisevsk:
I've added a second commit that goes a step further to also not make any API requests for updating the status if the controller thinks this would be a no-op. This slightly increases the chances of an out-of-date status (in the case where DWO misses an on-cluster update and the cache is out of sync) but also makes re-reconciling workspaces that have not seen any changes take practically no time.

Below is a graph of DWO metrics with the following configuration:
* 500 stopped workspaces, unchanging
* 1 minute resync interval for DWO (i.e. it re-reconciles all resources every minute)
    * Resyncing all resources is a case where DWO will hit internal rate limits for requests, causing each API call to take a lot longer than normal

The first half of the graph is for the current `main` branch. The second half is with only the first commit in this PR applied. 



The relevant metric to follow is reconcile time, which goes from ~500ms to ~250 during the resync. Additionally completing the full resync takes a lot less time (due to each reconcile taking half the time).

With the second commit, the graph is even less interesting; reconciles of stopped workspaces take >5ms since no API calls are made, and DWO can re-reconcile all 500 stopped workspaces in one second. 

This change should be safe for regular workspaces, as we still update status when the computed status is different from the on-cluster status.

AObuchow (approved):

Issue #1179, opened 2023-10-03
Kubernetes/OpenShift component and container component cannot use same endpoint target port


AObuchow:
### Description
If a kubernetes or openshift component define endpoints with the same target port, the devworkspace will be deemed invalid and will fail to start. Since Kubernetes & OpenShift components are intended to run in a separate pod from container components, it should be valid to have them use the same target port in their endpoints.

### How To Reproduce
Start a devworkspace that defines a container component and kubernetes (or openshift) component that both use the same target port in their endpoints, such as the following:

```YAML
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: kubernetes-deploy
        kubernetes:
          endpoints:
            - name: http-8080
              path: /
              targetPort: 8080
          uri: deploy.yaml
      - name: web-terminal
        container:
          image: quay.io/wto/web-terminal-tooling:next
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
          endpoints:
            - name: 'http'
              protocol: http
              targetPort: 8080
              exposure: public
```

The devworkspace will fail to start, with the following error message: `* devfile contains multiple containers with same endpoint targetPort: 8080`

### Expected behavior
The devworkspace should start up correctly (or rather, it should not fail to start due to the same endpoint target port being used)

### Additional context
Downstream issue: https://issues.redhat.com/browse/CRW-4856

Related Devfile API issue: https://github.com/devfile/api/issues/1049

The issue seems to be coming from the devfile validation in the devfile API. Upgrading the devfile API dependency used in DWO (similar to [what was done for the devfile library](https://github.com/devfile/library/pull/170)) should resolve the issue.

AObuchow:
Resolved in https://github.com/devfile/devworkspace-operator/pull/1186

PR #1184, opened 2023-10-05
chore: Update devfile/api dependency to upstream commit a6c32fc

AObuchow:
### What does this PR do?
Updates the devfile/api dependency to the upstream commit https://github.com/devfile/api/commit/a6c32fca0dbdcb61454e7bb7ff5b5ef525bbbe86. This is required for https://github.com/devfile/devworkspace-operator/issues/1179, as the devfile validation library was previously preventing a deworkspace from having a container component and openshift/kubernetes component from using the same target port on an endpoint. 

 https://github.com/devfile/api/commit/a6c32fca0dbdcb61454e7bb7ff5b5ef525bbbe86 was chosen as it is the same commit [used by the devfile library](https://github.com/devfile/library/blob/main/go.mod#L6), and all commits past a6c32fc were non-function-changing commits: https://github.com/devfile/api/compare/a6c32fca0dbdcb61454e7bb7ff5b5ef525bbbe86...e05a23590d0f820bb7f070d7e782a5f9caf301c5

I also added a controller env test to test the case from https://github.com/devfile/devworkspace-operator/issues/1179, but this might be redundant as this functionality should already be tested in the devfile API repo. 

https://github.com/devfile/devworkspace-operator/issues/1179

### Is it tested? How?
Install DWO with the changes from this PR, and create a devworkspace with a container component and kubernetes (or openshift) component that uses the same endpoint target port, such as the following:

```YAML
kind: DevWorkspace
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: kubernetes-deploy
        kubernetes:
          deployByDefault: true
          endpoints:
            - name: http-8080
              path: /
              targetPort: 8080
          inlined: |
            apiVersion: v1
            kind: Pod
            metadata:
              name: webserver-dwo-deployed
            spec:
              containers:
                - name: webserver
                  image: nginx:latest
                  ports:
                    - containerPort: 8080
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
          endpoints:
            - name: 'http'
              protocol: http
              targetPort: 8080
              exposure: public
```

Ensure the workspace starts up as expected. You should see a pod get deployed to the `devworkspace-controller` namespace called `webserver-dwo-deployed`. The main thing to ensure is the workspace does not fail to startup with the following reason: `* devfile contains multiple containers with same endpoint targetPort: 8080`

### PR Checklist

Signed-off-by: Andrew Obuchowicz <aobuchow@redhat.com>

### DIFF
--- a/controllers/workspace/suite_test.go
+++ b/controllers/workspace/suite_test.go
@@ -25,6 +25,7 @@ import (
 	dwv2 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	controllerv1alpha1 "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/controllers/workspace/internal/testutil"
+	kubesync "github.com/devfile/devworkspace-operator/pkg/library/kubernetes"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
@@ -108,6 +109,9 @@ var _ = BeforeSuite(func() {
 	Expect(k8sClient).NotTo(BeNil())
 
 	// Replicate controller setup similarly to how we do for main.go
+	err = kubesync.InitializeDeserializer(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
 	cacheFunc, err := cache.GetCacheFunc()
 	Expect(err).NotTo(HaveOccurred())
 
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ export PROJECT_CLONE_IMG ?= quay.io/devfile/project-clone:next
 export PULL_POLICY ?= Always
 export DEFAULT_ROUTING ?= basic
 export KUBECONFIG ?= ${HOME}/.kube/config
-export DEVWORKSPACE_API_VERSION ?= 2568944140c38aa941390d0f8e941534bf88f793
+export DEVWORKSPACE_API_VERSION ?= a6c32fca0dbdcb61454e7bb7ff5b5ef525bbbe86
 
 # Enable using Podman instead of Docker
 export DOCKER ?= docker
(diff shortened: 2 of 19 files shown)


controllers/workspace/suite_test.go
```diff
@@ -108,6 +109,9 @@ var _ = BeforeSuite(func() {
 	Expect(k8sClient).NotTo(BeNil())
 
 	// Replicate controller setup similarly to how we do for main.go
+	err = kubesync.InitializeDeserializer(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
```
amisevsk, line 114: Could you explain the purpose of these lines? I'm not clear why it's necessary.

amisevsk (approved):
I don't _love_ depending on a non-released version of devfile API, but it seems we have no choice until v2.2.1 is released.

controllers/workspace/suite_test.go
AObuchow: This is required for the Kubernetes component in `test-devworkspace-duplicate-ports.yaml` (from the test data) to be [deserialized](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/kubernetes/provision.go#L53). It mirrors the [setup](https://github.com/devfile/devworkspace-operator/blob/main/main.go#L109-L111) from `main.go`

AObuchow:
> I don't love depending on a non-released version of devfile API, but it seems we have no choice until v2.2.1 is released.

@amisevsk agreed. We could delay merging this if 2.2.1 is expected to be released soon enough.

amisevsk:
Devfile v2.2.1 is now released: https://github.com/devfile/api/releases/tag/v2.2.1

Just took some asking questions :)

AObuchow:
@amisevsk awesome 🥳 I will rework this PR to update to the latest release

AObuchow:
Closing in favour of https://github.com/devfile/devworkspace-operator/pull/1186

Issue #1189, opened 2023-10-13
$PROJECT_SOURCE incorrectly set if starterProjects are used


AObuchow:
### Description
If a devworkspace has a **selected** starter project, but no regular projects, the `$PROJECT_SOURCE` environment variable will be set to `/projects`, instead of `/projects/<selected-starterProject-name>`.

Currently, the list of projects that we search [when determining](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/container/mountSources.go#L90-L101) what `$PROJECT_SOURCE` should be set to only contains regular projects, and not starterProjects. 

### How To Reproduce
1. Apply the following devworkspace:

```YAML
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: plain-starter-projects
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
      controller.devfile.io/use-starter-project: community
    starterProjects:
    - name: community
      zip:
        location: https://code.quarkus.io/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-micrometer&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift&cn=devfile
    - name: redhat-product
      zip:
        location: https://code.quarkus.redhat.com/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift
    components:
      - name: web-terminal
        container:
          image: quay.io/wto/web-terminal-tooling:next
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```
2. Check the workspace pod's spec for the `$PROJECT_SOURCE` environment variable in the web-terminal container. 

```YAML
spec:
(...)
  containers:
    - name: web-terminal
      env:
        - name: PROJECTS_ROOT
          value: /projects
        - name: PROJECT_SOURCE
          value: /projects
(...)

```
### Expected behavior
`$PROJECT_SOURCE` should be set to the selected starterProject (in this case, `community`). In this case, `$PROJECT_SOURCE` should be `/projects/community`.

If our devworkspace had a regular project, I would expect it to take precedence over any starterProjects as well.
[...]

PR #1190, opened 2023-10-13
Fix: include starter projects for $PROJECT_SOURCE path selection

AObuchow:
### What does this PR do?

This PR adds all starterProjects to the list of projects used [when determining](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/container/mountSources.go#L90-L101) the value of the `$PROJECT_SOURCE` environment variable used in workspace deployments. 

Now, when determining the value of `$PROJECT_SOURCE`, the following rules are considered (in order of priority):

1. If the workspace has at least one regular project, the first one will be selected. If the first project has a clone path, it will be used, otherwise the project's name will be used for the project source path. This was the default behaviour prior to this PR.

2. If the workspace has a starter project that is selected using the `controller.devfile.io/use-starter-project` workspace attribute, its name will be used for project source path.

3. Otherwise, the returned project source path will be an empty string, i.e. `$PROJECT_SOURCE = /projects/`

I also removed a completed TODO comment about adding sparseCheckout support to the project clone container and moved the `GetClonePath()` function to be accessible from other packages. 

#1189

### Is it tested? How?
There are 4 cases to test, each is described below. 

#### devworkspace with only starter projects
1. Apply a devworkspace that has at least one starter project, and no regular projects, such as the following:

```YAML
kind: DevWorkspace
metadata:
  name: plain-starter-projects
spec:
  started: true
  routingClass: 'basic'
  template:
    starterProjects:
    - name: community
      zip:
        location: https://code.quarkus.io/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-micrometer&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift&cn=devfile
    - name: redhat-product
      zip:
        location: https://code.quarkus.redhat.com/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

2. Ensure the workspace pod's spec has the `$PROJECT_SOURCE` environment variable set to `/projects/` as no starter project has been selected. For the previously given devworkspace, we should have:

```YAML
spec:
(...)
  containers:
    - name: web-terminal
      env:
        - name: PROJECTS_ROOT
          value: /projects
        - name: PROJECT_SOURCE
          value: /projects/
(...)
```

#### devworkspace with starter projects and regular projects
1. Apply a devworkspace that has a regular project as well as starter projects. For example:

```YAML
kind: DevWorkspace
metadata:
  name: plain-starter-projects
spec:
  started: true
  routingClass: 'basic'
  template:
    starterProjects:
    - name: community
      zip:
        location: https://code.quarkus.io/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-micrometer&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift&cn=devfile
    - name: redhat-product
      zip:
        location: https://code.quarkus.redhat.com/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

2. Ensure the workspace pod's spec has the `$PROJECT_SOURCE` environment variable set to `/projects/<first-project-name>`. For the previously given devworkspace, we should have:

```YAML
spec:
(...)
  containers:
    - name: web-terminal
      env:
        - name: PROJECTS_ROOT
          value: /projects
        - name: PROJECT_SOURCE
          value: /projects/web-nodejs-sample
(...)
```

#### devworkspace with only starter projects, with a starter project being selected
1. Apply a devworkspace that has at least two starter projects, where one of the starter projects is selected using the  `controller.devfile.io/use-starter-project` attribute, such as the following:

```YAML
kind: DevWorkspace
metadata:
  name: plain-starter-projects
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
      controller.devfile.io/use-starter-project: redhat-product
    starterProjects:
    - name: community
      zip:
        location: https://code.quarkus.io/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-micrometer&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift&cn=devfile
    - name: redhat-product
      zip:
        location: https://code.quarkus.redhat.com/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

2. Ensure the workspace pod's spec has the `$PROJECT_SOURCE` environment variable set to the starter project that was selected. For the previously given devworkspace, we should have:

```YAML
spec:
(...)
  containers:
    - name: web-terminal
      env:
        - name: PROJECTS_ROOT
          value: /projects
        - name: PROJECT_SOURCE
          value: /projects/redhat-product
(...)
```

#### devworkspace with regular and starter projects, with a selected starter project
1. Apply a devworkspace that has a regular and starter project, and the starter project is selected using the  `controller.devfile.io/use-starter-project` attribute, such as the following:

```YAML
kind: DevWorkspace
metadata:
  name: plain-starter-projects
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
      controller.devfile.io/use-starter-project: redhat-product
    starterProjects:
    - name: community
      zip:
        location: https://code.quarkus.io/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-micrometer&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift&cn=devfile
    - name: redhat-product
      zip:
        location: https://code.quarkus.redhat.com/d?e=io.quarkus%3Aquarkus-resteasy&e=io.quarkus%3Aquarkus-smallrye-health&e=io.quarkus%3Aquarkus-openshift
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"

```

2. Ensure the workspace pod's spec has the `$PROJECT_SOURCE` environment variable set to `/projects/<first-project-name>`. The regular project takes priority over the selected starter project when setting `$PROJECT_SOURCE`. For the previously given devworkspace, we should have:

```YAML
spec:
(...)
  containers:
    - name: web-terminal
      env:
        - name: PROJECTS_ROOT
          value: /projects
        - name: PROJECT_SOURCE
          value: /projects/web-nodejs-sample
(...)
```

### PR Checklist

### DIFF
--- a/pkg/library/container/container.go
+++ b/pkg/library/container/container.go
@@ -27,12 +27,11 @@ package container
 import (
 	"fmt"
 
-	"github.com/devfile/devworkspace-operator/pkg/library/overrides"
-
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/pkg/library/flatten"
 	"github.com/devfile/devworkspace-operator/pkg/library/lifecycle"
+	"github.com/devfile/devworkspace-operator/pkg/library/overrides"
 	corev1 "k8s.io/api/core/v1"
 )
 
@@ -65,7 +64,9 @@ func GetKubeContainersFromDevfile(workspace *dw.DevWorkspaceTemplateSpec, securi
 		if err != nil {
 			return nil, err
 		}
-		handleMountSources(k8sContainer, component.Container, workspace.Projects)
+		if err := handleMountSources(k8sContainer, component.Container, workspace); err != nil {
+			return nil, err
+		}
 		if overrides.NeedsContainerOverride(&component) {
 			patchedContainer, err := overrides.ApplyContainerOverrides(&component, k8sContainer)
 			if err != nil {
@@ -89,7 +90,9 @@ func GetKubeContainersFromDevfile(workspace *dw.DevWorkspaceTemplateSpec, securi
 		if err != nil {
 			return nil, err
 		}
-		handleMountSources(k8sContainer, initComponent.Container, workspace.Projects)
+		if err := handleMountSources(k8sContainer, initComponent.Container, workspace); err != nil {
+			return nil, err
+		}
 		if overrides.NeedsContainerOverride(&initComponent) {
 			patchedContainer, err := overrides.ApplyContainerOverrides(&initComponent, k8sContainer)
 			if err != nil {
--- a/pkg/library/container/mountSources.go
+++ b/pkg/library/container/mountSources.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/devfile/devworkspace-operator/pkg/constants"
 	devfileConstants "github.com/devfile/devworkspace-operator/pkg/library/constants"
+	projectslib "github.com/devfile/devworkspace-operator/pkg/library/projects"
 )
 
 // HasMountSources evaluates whether project sources should be mounted in the given container component.
@@ -53,9 +54,9 @@ func AnyMountSources(devfileComponents []dw.Component) bool {
 
 // handleMountSources adds a volumeMount to a container if the corresponding devfile container has
 // mountSources enabled.
-func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.ContainerComponent, projects []dw.Project) {
+func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.ContainerComponent, workspace *dw.DevWorkspaceTemplateSpec) error {
 	if !HasMountSources(devfileContainer) {
-		return
+		return nil
 	}
 	var sourceMapping string
 	if vm := getProjectsVolumeMount(k8sContainer); vm != nil {
@@ -75,7 +76,10 @@ func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.Con
 		})
 	}
 
-	projectsSourcePath := getProjectSourcePath(projects)
+	projectsSourcePath, err := getProjectSourcePath(workspace)
+	if err != nil {
+		return err
+	}
 
 	k8sContainer.Env = append(k8sContainer.Env, corev1.EnvVar{
 		Name:  devfileConstants.ProjectsRootEnvVar,
@@ -84,20 +88,37 @@ func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.Con
 		Name:  devfileConstants.ProjectsSourceEnvVar,
 		Value: path.Join(sourceMapping, projectsSourcePath),
 	})
+
+	return nil
 }
 
-// getProjectSourcePath gets the path, relative to PROJECTS_ROOT, that should be used for the PROJECT_SOURCE env var
-func getProjectSourcePath(projects []dw.Project) string {
-	projectPath := ""
+// getProjectSourcePath gets the path, relative to PROJECTS_ROOT, that should be used for the PROJECT_SOURCE env var.
+// Returns an error if there was a problem retrieving the selected starter project.
+//
+// The project source path is determined based on the following priorities:
+//
+//  1. If the workspace has at least one regular project, the first one will be selected.
+//     If the first project has a clone path, it will be used, otherwise the project's name will be used as the project source path.
+//
+//  2. If the workspace has a starter project that is selected, its name will be used as the project source path.
+//
+//  3. Otherwise, the returned project source path will be an empty string.
+func getProjectSourcePath(workspace *dw.DevWorkspaceTemplateSpec) (string, error) {
+	projects := workspace.Projects
+	// If there are any projects, return the first one's clone path
 	if len(projects) > 0 {
-		firstProject := projects[0]
-		if firstProject.ClonePath != "" {
-			projectPath = firstProject.ClonePath
-		} else {
-			projectPath = firstProject.Name
-		}
+		return projectslib.GetClonePath(&projects[0]), nil
 	}
-	return projectPath
+
+	// No projects, check if we have a selected starter project
+	selectedStarterProject, err := projectslib.GetStarterProject(workspace)
+	if err != nil {
+		return "", err
+	} else if selectedStarterProject != nil {
+		// Starter projects do not allow specifying a clone path, so use the name
+		return selectedStarterProject.Name, nil
+	}
+	return "", nil
 }
 
 // getProjectsVolumeMount returns the projects volumeMount in a container, if it is defined; if it does not exist,
--- a/project-clone/main.go
+++ b/project-clone/main.go
@@ -38,7 +38,6 @@ const (
 	tmpLogFilePath = "/tmp/" + logFileName
 )
 
-// TODO: Handle sparse checkout
 func main() {
 	f, err := os.Create(tmpLogFilePath)
 	if err != nil {
--- a/pkg/library/projects/clone.go
+++ b/pkg/library/projects/clone.go
@@ -115,6 +115,14 @@ func GetStarterProject(workspace *dw.DevWorkspaceTemplateSpec) (*dw.StarterProje
 	return nil, fmt.Errorf("selected starter project %s not found in workspace starterProjects", selectedStarterProject)
 }
 
+// GetClonePath gets the correct clonePath for a project, given the semantics in devfile/api
+func GetClonePath(project *dw.Project) string {
+	if project.ClonePath != "" {
+		return project.ClonePath
+	}
+	return project.Name
+}
+
 func hasContainerComponents(workspace *dw.DevWorkspaceTemplateSpec) bool {
 	for _, component := range workspace.Components {
 		if component.Container != nil {
--- a/project-clone/internal/devfile.go
+++ b/project-clone/internal/devfile.go
@@ -32,14 +32,6 @@ const (
 	ProjectSubDir         = "subDir"
 )
 
-// GetClonePath gets the correct clonePath for a project, given the semantics in devfile/api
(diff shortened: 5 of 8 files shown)


pkg/library/projects/util.go
```diff
@@ -0,0 +1,37 @@
...
+
+		project := dw.Project{
+			Name:          starterProject.Name,
+			Attributes:    starterProject.Attributes,
+			ProjectSource: starterProject.ProjectSource,
+		}
+		projects = append(projects, project)
+	}
+	return projects
+}
```
amisevsk, line 37: I'm not sure this is correct behavior, in-context (haven't had a chance to verify though). 

If a resolved workspace ends up with e.g. no projects and two starterProjects, with the second one being the "selected" project, how does the caller know that e.g. the first project will not be cloned but the second one will?

AObuchow: You're correct - there's a conflict with the "selected" starter project and how [`getProjectSourcePath()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/container/mountSources.go#L90-L101) only selects the first available project to set $PROJECT_SOURCE. 

I wasn't sure if this rule of "the first available project is what $PROJECT_SOURCE maps to" was documented somewhere, and I couldn't find it, or if it was just a decision that made sense at the time. 

I'll rework things to respect the selected starter project attribute.

amisevsk: The env var is (poorly) documented as part of the Devfile schema, in `.commands.exec.commandline`:

> $PROJECT_SOURCE: A path to a project source ($PROJECTS_ROOT/<project-name>). If there are multiple projects, this will point to the directory of the first one.

AObuchow: Thank you for clarifying on the documentation of `$PROJECT_SOURCE` @amisevsk :pray: 

@l0rd do you have any input on this? For context:

l0rd: If the attribute `controller.devfile.io/use-starter-project` is set, then we should configure PROJECT_SOURCE accordingly, otherwise we should fall back on the first project.

[And Eclipse Che dashboard is currently setting the attribute at startup](https://github.com/eclipse/che/issues/22486) (without interaction).

AObuchow:
@amisevsk I've added 2 extra commits to address the cases where the `controller.devfile.io/use-starter-project` attribute is used (however, the second fixup commit was just some code style changes that are optional). 

I also updated the PR description, including 2 new test cases.

The commit description for [`755dcd8` (#1190)](https://github.com/devfile/devworkspace-operator/pull/1190/commits/755dcd884820349ccd6f58a6c3dc69a512d5fa74) needs to be updated before this PR is merged to mention the new rules used when determining the value of `$PROJECT_SOURCE`.

pkg/library/container/container.go
```diff
@@ -65,7 +64,10 @@ func GetKubeContainersFromDevfile(workspace *dw.DevWorkspaceTemplateSpec, securi
 		if err != nil {
 			return nil, err
 		}
-		handleMountSources(k8sContainer, component.Container, workspace.Projects)
+		err = handleMountSources(k8sContainer, component.Container, workspace)
+		if err != nil {
+			return nil, err
+		}
```
amisevsk, line 70: (Optional, feel free to just resolve)

```suggestion
		if err := handleMountSources(k8sContainer, component.Container, workspace); err != nil {
			return nil, err
		}
```

I always think it's a little weird that Go code goes around reassigning `err` over and over again

pkg/library/container/mountSources.go
```diff
@@ -84,20 +88,43 @@ func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.Con
...
+//
+// The project source path is determined based on the following priorities:
+//
+//  1. If the workspace has at least one regular project, the first one will be selected.
+//     If the first project has a clone path, it will be used, otherwise the project's name will be used as the project source path.
+//
+//  2. If the workspace has a starter project that is selected, its name will be used as the project source path.
+//
+//  3. If the workspace has at least one starter project, but none are selected,
+//     the name of the first starter project will be used as the project source path.
```
amisevsk, line 106: I think this may break more than it fixes. If there are starterProjects but none selected, then `$PROJECT_SOURCE` will point to a nonexistent directory since project-clone won't clone anything.

It might be better to just take 1, 2, and 4 from this list. From my understanding, what Che will do is automatically "select" the first starterProject anyways, so this won't break their use case either.

pkg/library/container/mountSources.go
```diff
@@ -84,20 +88,43 @@ func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.Con
...
 		if firstProject.ClonePath != "" {
-			projectPath = firstProject.ClonePath
-		} else {
-			projectPath = firstProject.Name
+			return firstProject.ClonePath, nil
 		}
+		return firstProject.Name, nil
 	}
-	return projectPath
+	return "", nil
```
amisevsk, line 127: This still looks strange to me -- I think there's some duplication of work here since both `projectslib.GetAllProjects()` and `projectslib.GetStarterProject()` check for the selected starter project, so we're searching through `workspace.StarterProjects` twice. We're also aliasing `projects` to refer to "all projects", which is confusing given that there's a `workspace.Projects`

Would it make more sense to do something like

```go
// If there are any projects, return the first one's clone path
if len(workspace.Projects) > 0 {
  return projectslib.GetClonePath(workspace.Projects[0])
}
// No projects, check for starter project
starterProject, err := projectslib.GetStarterProject(workspace)
if err != nil {
  return "", err
} else if starterProject == nil {
  return "", nil
}
return starterProject.Name
``` 

This approach also won't throw an error here (but maybe will elsewhere) if there's a regular project but the selected starter project is misconfigured.

project-clone/main.go
```diff
@@ -38,7 +38,6 @@ const (
 	tmpLogFilePath = "/tmp/" + logFileName
 )
 
-// TODO: Handle sparse checkout
 func main() {
```
amisevsk, line 41: You might consider also moving `GetClonePath` to the projects library, since we're doing it both in `container` and project-clone

pkg/library/container/mountSources.go
```diff
@@ -84,20 +88,43 @@ func handleMountSources(k8sContainer *corev1.Container, devfileContainer *dw.Con
...
+		selectedStarterProject, err := projectslib.GetStarterProject(workspace)
+		if err != nil {
+			return "", err
+		}
+		// Only use the selected starterProject if we have no regular projects
+		if selectedStarterProject != nil && len(workspace.Projects) == 0 {
+			return selectedStarterProject.Name, nil
+		}
+
 		firstProject := projects[0]
```
amisevsk, line 121: This relies on `GetAllProjects` returning the projects with starterProjects being in the list after regular projects, which should either be documented as a requirement on the function (and maybe have a test case to verify) or be handled differently.

IMO it's easier to just check `workspace.Projects` first, and if it's empty check `workspace.StarterProjects`.

amisevsk (commented):
A few comments -- I think we're going down the path of patching existing code to do something different when rewriting the logic makes things clearer.

project-clone/main.go
amisevsk: This comment is here because it's the only file in `project-clone` in this PR, this has nothing to do with removing the TODO (thanks!) :D

pkg/library/container/mountSources.go
AObuchow: > I think this may break more than it fixes. If there are starterProjects but none selected, then `$PROJECT_SOURCE` will point to a nonexistent directory since project-clone won't clone anything.

Good catch, I hadn't considered that. Keeping points 1,2 and 4 as suggested.

pkg/library/container/mountSources.go
AObuchow: I agree this is a better approach and is a lot more obvious as to what's going on (especially without requring the list of projects to be sorted with regular projects first, then starter projects second). This also eliminates the need for `projectslib.GetAllProjects()`, which felt a bit awkward of a function anyhow.

project-clone/main.go
AObuchow: I ended up moving `GetClonePath` since it made sense anyhow :P

AObuchow:
@amisevsk Thank you for the great feedback, this PR is looking a lot better now. I also updated the PR description & test cases, as well as the requirements on the [original issue](https://github.com/devfile/devworkspace-operator/issues/1189) to specify that the starterProject must be selected in order to be used for `$PROJECT_SOURCE`

amisevsk (approved):
LGTM :+1:

PR #1191, opened 2023-10-16
Add timeout to health checks to avoid very long reconciles

amisevsk:
### What does this PR do?
Adds a timeout (500ms) to the health check HTTP client. This hopefully avoids cases where the controller effectively idles for long periods while the health check server is being started (i.e. the server is not ready and is not sending a response)

Haven't created an issue yet, need to do a bit more diagnosing.

### Is it tested? How?
Also hard to test -- it's hard to trigger a slow response from the editor server manually. Regular workspace startup should not be impacted by this change, however.

### PR Checklist

### DIFF
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -17,6 +17,7 @@ import (
 	"crypto/tls"
 	"net/http"
 	"net/url"
+	"time"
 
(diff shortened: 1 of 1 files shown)


AObuchow (approved):
Seems reasonable to me :+1:

PR #1196, opened 2023-10-23
Switch from using hub to gh in GitHub actions

amisevsk:
### What does this PR do?
The `hub` CLI has been removed from the Ubuntu 20.04 runner and is intended to be replaced with the 'gh' CLI. This commit adapts the previous hub command to open a PR into a gh command that does the same.

A note on arguments:
* There's no `-f` (force, even if not all commits are pushed to remote) equivalent in `gh`. This shouldn't be an issue because the prior command is `git push origin ${PR_BRANCH}`
    * Instead, `-f` in `gh` means "create PR description from last commit", which should duplicate what the `-m` was doing in `hub`
* `-b` (base branch) and `-h` (head branch) are replaced with `-B` and `-H`

`release` GH action is failing.

### Is it tested? How?
I haven't directly tested, will probably have to wait for the next release to see if everything works appropriately. 

### PR Checklist

### DIFF
--- a/make-release.sh
+++ b/make-release.sh
@@ -194,8 +194,7 @@ git_commit_and_push() {
     git checkout "${PR_BRANCH}"
     git pull origin "${PR_BRANCH}" || true
     $DRY_RUN git push origin "${PR_BRANCH}"
-    lastCommitComment="$(git log -1 --pretty=%B)"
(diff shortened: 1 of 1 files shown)


AObuchow (approved):
Thanks for catching & fixing this @amisevsk :)

AObuchow:
For some reason the check sources CI failed, the PR branch maybe wasn't based off of https://github.com/devfile/devworkspace-operator/pull/1180. Re-running it now

PR #1223, opened 2024-01-22
Allow pod/container overrides for volumes/volumemounts

dkwon17:
### What does this PR do?
Allows pod overrides for `spec.volumes` and container overrides for `spec.containers[*].volumeMounts`.

https://github.com/eclipse/che/issues/22751

### Is it tested? How?

Checkout this PR and install DWO on a fresh cluster with [Run controller locally](https://github.com/devfile/devworkspace-operator?tab=readme-ov-file#run-controller-locally). Switch to the `devworkspace-controller` namespace by running `oc project devworkspace-controller`

<details><summary>Mounting secrets to workspace container using overrides
</summary>

1. Create secret to mount:
```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: dotfile-secret
data:
  .secret-file: dmFsdWUtMg0KDQo=
EOF
```

2. Create a devworkspace that uses pod-overrides and container-overrides to mount the secret:
```
curl https://gist.githubusercontent.com/dkwon17/4e698c3dcf6e36fc474eb73f5c20ac7c/raw/e7e4cf11d2081ced4b69e9cf9a230a2c8c30cf0c/che-code-secret-override.yaml | kubectl apply -f -
```

3. Wait until the DevWorkspace's status becomes 'Running'.
4. Check that the volume has been added in the workspace pod. In this case, we should see `secret-volume` in the list of volumes:
```
$ oc get pod <podname> -o=jsonpath='{.spec.volumes[*].name}'
secret-volume checode projects workspace-metadata kube-api-access-hcdjv
```
5. Check that the volumeMount has been added in the workspace pod's `tools` container:
```
$oc get pod <podname> -o=jsonpath='{.spec.containers[1].volumeMounts[*].name}'
secret-volume projects workspace-metadata kube-api-access-hcdjv
```

6. Run the following in the workspace `tools` container to check that the secret has been mounted:
```
$ cat /etc/secret-volume/.secret-file 
value-2
```
7. Stop the workspace and update the `storage-type` to `per-user`:
``` 
oc patch dw secret-volume-override --type merge -p '{"spec":{"template":{"att
ributes":{"controller.devfile.io/storage-type":"per-user"}}}}'
```
8. Start the workspace, and recheck steps 3-6.
9. Stop the workspace and update the `storage-type` to `per-workspace`, start the workspace and recheck steps 3-6 again.

</details>

The feature also works for adding PVC volumes:
```
      pod-overrides:
        spec:
          volumes:
            - name: my-pv
              persistentVolumeClaim:
                claimName: my-pvc
```

### Caveat of this feature
When introducing a new volumes using pod-overrides, the new volume must contain all fields that Kubernetes may add. For example, pods that specify configmap and secret volumes have a `defaultMode: 420` added to the yaml even if it's not originally provided in the pod's yaml.

Therefore when introducing a new configmap/secret volumes using pod-overrides, the `defaultMode` field must be provided  along with the configmap/secret details.

For example:
```
      pod-overrides:
        spec:
          volumes:
            - name: secret-volume
              secret:
                secretName: dotfile-secret
                defaultMode: 420
```

If the `defaultMode` is missing in the pod overrides, the DevWorkspace will never reach the `Running` state because the new volume in the pod will have the `defaultMode: 420` added by default. At the same time, the DWO will try to remove `defaultMode` if it doesn't exist in pod overrides. To clarify, from my testing this only happens when introducing new volumes through pod-overrides. This logic causes reconciliation until workspace start fails with this message:
```
The workspace status remains "Starting" in the last 300 seconds.
```
  

### PR Checklist

### DIFF
--- a/pkg/library/overrides/pods.go
+++ b/pkg/library/overrides/pods.go
@@ -76,24 +76,27 @@ func ApplyPodOverrides(workspace *common.DevWorkspaceWithConfig, deployment *app
 	return patched, nil
 }
 
-func GetPodVolumeOverrides(workspace *dw.DevWorkspaceTemplateSpec) (*[]corev1.Volume, error) {
-	var volumeOverrides []corev1.Volume
+func GetVolumesFromOverrides(workspace *dw.DevWorkspaceTemplateSpec) (map[string]bool, error) {
+	overrideVolumes := map[string]bool{}
+
 	overrides, err := getPodOverrides(workspace)
 	if err != nil {
-		return &volumeOverrides, err
+		return nil, err
 	}
 
 	for _, override := range overrides {
 		podSpecTemplate := corev1.PodTemplateSpec{}
 		if err := json.Unmarshal(override.Raw, &podSpecTemplate); err != nil {
-			return &volumeOverrides, fmt.Errorf("error unmarshalling: %w", err)
+			return nil, fmt.Errorf("error unmarshalling: %w", err)
 		}
 
 		if podSpecTemplate.Spec.Volumes != nil {
-			volumeOverrides = append(volumeOverrides, podSpecTemplate.Spec.Volumes...)
+			for _, volume := range podSpecTemplate.Spec.Volumes {
+				overrideVolumes[volume.Name] = true
+			}
 		}
 	}
-	return &volumeOverrides, nil
+	return overrideVolumes, nil
 }
 
 // getPodOverrides returns PodTemplateSpecOverrides for every instance of the pod overrides attribute
--- a/pkg/library/overrides/pods.go
+++ b/pkg/library/overrides/pods.go
@@ -16,9 +16,11 @@ package overrides
 import (
 	"fmt"
 
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 
@@ -41,7 +43,7 @@ func NeedsPodOverrides(workspace *common.DevWorkspaceWithConfig) bool {
 }
 
 func ApplyPodOverrides(workspace *common.DevWorkspaceWithConfig, deployment *appsv1.Deployment) (*appsv1.Deployment, error) {
-	overrides, err := getPodOverrides(workspace)
+	overrides, err := getPodOverrides(&workspace.Spec.Template)
 	if err != nil {
 		return nil, err
 	}
@@ -53,7 +55,6 @@ func ApplyPodOverrides(workspace *common.DevWorkspaceWithConfig, deployment *app
 	originalContainers := patched.Spec.Template.Spec.Containers
 	// Save fields we do not allow to be configured in pod-overrides
 	originalInitContainers := patched.Spec.Template.Spec.InitContainers
-	originalVolumes := patched.Spec.Template.Spec.Volumes
 	patchedTemplateBytes, err := json.Marshal(patched.Spec.Template)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal deployment to yaml: %w", err)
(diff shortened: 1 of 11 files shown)


pkg/library/overrides/pods.go
```diff
@@ -72,55 +73,68 @@ func ApplyPodOverrides(workspace *common.DevWorkspaceWithConfig, deployment *app
 	patched.Spec.Template = patchedPodSpecTemplate
 	patched.Spec.Template.Spec.Containers = originalContainers
 	patched.Spec.Template.Spec.InitContainers = originalInitContainers
-	patched.Spec.Template.Spec.Volumes = originalVolumes
 	return patched, nil
 }
 
+func GetPodVolumeOverrides(workspace *dw.DevWorkspaceTemplateSpec) (*[]corev1.Volume, error) {
```
amisevsk, line 79: We probably don't want to return a reference to a slice here (`*[]corev1.Volume`), `[]corev1.Volume` should be enough (since a slice is already a reference).

pkg/library/overrides/pods.go
```diff
@@ -72,55 +73,68 @@ func ApplyPodOverrides(workspace *common.DevWorkspaceWithConfig, deployment *app
...
+	var volumeOverrides []corev1.Volume
+	overrides, err := getPodOverrides(workspace)
+	if err != nil {
+		return &volumeOverrides, err
```
amisevsk, line 83: It's clearer to just have `return nil, err` here.

pkg/library/overrides/pods.go
```diff
@@ -72,55 +73,68 @@ func ApplyPodOverrides(workspace *common.DevWorkspaceWithConfig, deployment *app
...
+		podSpecTemplate := corev1.PodTemplateSpec{}
+		if err := json.Unmarshal(override.Raw, &podSpecTemplate); err != nil {
+			return &volumeOverrides, fmt.Errorf("error unmarshalling: %w", err)
+		}
+
+		if podSpecTemplate.Spec.Volumes != nil {
+			volumeOverrides = append(volumeOverrides, podSpecTemplate.Spec.Volumes...)
+		}
+	}
+	return &volumeOverrides, nil
```
amisevsk, line 96: I believe this can return an invalid list of volumes -- what if multiple overrides apply to the same volume?

pkg/provision/storage/commonStorage.go
```diff
@@ -144,6 +145,15 @@ func (p *CommonStorageProvisioner) rewriteContainerVolumeMounts(workspaceId, pvc
 		additionalVolumes[additionalVolume.Name] = true
 	}
 
+	// Containers in podAdditions may reference volumes defined in pod overrides, and this is not an error
+	overridesVolumes, err := overrides.GetPodVolumeOverrides(workspace)
```
amisevsk, line 149: Given the usage, it might be simpler to instead have the method just return a set of volume names. 

```go
func GetVolumesFromOverrides(workspace *dw.DevWorkspaceTemplateSpec) (map[string]bool, error) {
```

I was slightly confused reading the method in `pods.go` as it makes it seem like something might require just the volumes from overrides.

amisevsk (commented):
Generally looks good, I have a few comments

Nice work :+1:

dkwon17:
Thank you for the review @amisevsk , I've updated the PR

pkg/provision/storage/perWorkspaceStorage.go
```diff
@@ -111,12 +111,12 @@ func (p *PerWorkspaceStorageProvisioner) rewriteContainerVolumeMounts(workspaceI
...
 	// Containers in podAdditions may reference volumes defined in pod overrides, and this is not an error
-	overridesVolumes, err := overrides.GetPodVolumeOverrides(workspace)
+	overridesVolumes, err := overrides.GetVolumesFromOverrides(workspace)
 	if err != nil {
 		return err
 	}
-	for _, overridesVolume := range *overridesVolumes {
-		additionalVolumes[overridesVolume.Name] = true
+	for key, value := range overridesVolumes {
+		additionalVolumes[key] = value
```
amisevsk, line 119: Minor nit:
```suggestion
	for volumeName, present := range overridesVolumes {
		additionalVolumes[volumeName] = present
```

pkg/provision/storage/commonStorage.go
```diff
@@ -146,12 +146,12 @@ func (p *CommonStorageProvisioner) rewriteContainerVolumeMounts(workspaceId, pvc
...
 	// Containers in podAdditions may reference volumes defined in pod overrides, and this is not an error
-	overridesVolumes, err := overrides.GetPodVolumeOverrides(workspace)
+	overridesVolumes, err := overrides.GetVolumesFromOverrides(workspace)
 	if err != nil {
 		return err
 	}
-	for _, overridesVolume := range *overridesVolumes {
-		additionalVolumes[overridesVolume.Name] = true
+	for key, value := range overridesVolumes {
+		additionalVolumes[key] = value
```
amisevsk, line 154: Minor nit:
```suggestion
	for volumeName, present := range overridesVolumes {
		additionalVolumes[volumeName] = present
```

pkg/library/overrides/testdata/container-overrides/overrides-can-define-volumemount.yaml
```diff
@@ -1,4 +1,4 @@
-name: "Applies volumeMount overrides from container-overrides attribute"
+name: "Container overrides can add define volumeMounts"
```
amisevsk, line 1:
```suggestion
name: "Container overrides can define a volumeMount"
```

amisevsk (approved):
LGTM :+1:

Thanks @dkwon17!

amisevsk:
Regarding the merge key for volumeMounts being `mountPath` -- we may in the future want to add some guardrails for which volumes/mounts can be overridden (e.g. you can't override default ones? I'm not sure here).

amisevsk:
@dkwon17 One somewhat annoying thing in this repo: all commits need a signoff before merge (including fixup commits that would be squashed when merging). You'll need to squash fixups manually (or sign off on them) before we can merge this PR.

dkwon17:
@amisevsk I've squashed the GitHub fixup commits and signed it off.

> we may in the future want to add some guardrails for which volumes/mounts can be overridden (e.g. you can't override default ones? I'm not sure here)

I will make a PR to update the pod/container overrides [documentation](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#pod-overrides) removing the mention about volume/mounts being non-overridable. Maybe I can add a sentence about not overriding default volumes/mounts? Nothing at the moment prevents a user from doing this however.

amisevsk:
> Maybe I can add a sentence about not overriding default volumes/mounts? Nothing at the moment prevents a user from doing this however.

That would be great -- even just a warning that you _can_ override built-in volume mounts and should avoid doing it would be useful.

PR #1224, opened 2024-01-24
chore: update pod/container overrides docs

dkwon17:
### What does this PR do?
Updates pod/container overrides docs for volume/volumeMounts.

https://github.com/eclipse/che/issues/22751

### PR Checklist


docs/additional-configuration.adoc
```diff
@@ -302,7 +302,9 @@ spec:
           securityContext:
             runAsUser: 1234
 ----
-Note that the pod-overrides field does not allow configuring `containers`, `initContainers`, and `volumes`. In order to configure these elements, use existing DevWorkspace fields (container components, volume components)
+Note that the pod-overrides field does not allow configuring `containers` and `initContainers`. In order to configure these elements, use the existing DevWorkspace `spec.template.components` field.
```
AObuchow, line 305: Potential suggestion:

"Note that the pod-overrides field does not allow configuring **the** `containers` and `initContainers` **fields from the pod spec**".

docs/additional-configuration.adoc
```diff
@@ -302,7 +302,9 @@ spec:
...
+
+Note that extra precaution should be taken when overriding the `volumes` field. The DevWorkspace operator defines pod volumes by default for config files, metadata, and credentials. To avoid unexpected behaviour, these default volumes should not be overridden.
```
AObuchow, line 307: "Operator" in "DevWorkspace Operator" is capitalized elsewhere in this doc. 

Also slight re-wording suggestion:

"The DevWorkspace Operator sets the pod volumes field by default for config files, metadata, and credentials. To avoid unexpected behaviour, the volumes field should not be overridden."

docs/additional-configuration.adoc
```diff
@@ -325,7 +327,9 @@ spec:
         attributes:
           container-overrides: {"resources":{"limits":{"nvidia.com/gpu":"1"}}}
 ----
-Note that the container-overrides field does not allow configuring fields that are available through the regular DevWorkspace API. In particular, workspaces will fail to start if the container-overrides attribute attempts to override `name`, `image`, `command`, `args`, `ports`, `volumeMounts`, or `env`.
+Note that the container-overrides field does not allow configuring most fields that are available through the regular DevWorkspace API. In particular, workspaces will fail to start if the container-overrides attribute attempts to override `name`, `image`, `command`, `args`, `ports`, or `env`.
```
AObuchow, line 330: We could link the actual devfile Schema/API here, i.e. "Note that the container-overrides field does not allow configuring most fields that are available through the regular [DevWorkspace API](https://devfile.io/docs/2.2.2/devfile-schema#components-container)".

 Though, this is kind of awkard (since it's the DevFile API, not the DevWorkspace API...)

Also, that link will eventually become outdated, as it points to v2.2.2 of the DevFile API (and it doesn't seem there's a way to make the link point to "latest"). 

Take this suggestion with a grain of salt :P

docs/additional-configuration.adoc
```diff
@@ -325,7 +327,9 @@ spec:
...
+
+Note that extra precaution should be taken when overriding the `volumeMounts` field. The DevWorkspace operator defines volumeMounts by default for config files, metadata, and credentials. To avoid unexpected behaviour, these default volumeMounts should not be overridden.
```
AObuchow, line 332: The two last times "volumeMounts" is mentioned are missing backticks, e.g. volumeMounts should be `volumeMounts` for consistency.

My earlier re-wording suggestion could also be used here:

The DevWorkspace operator sets the `volumeMounts` by default for config files, metadata, and credentials. To avoid unexpected behaviour, the `volumeMounts` field should not be overridden.

AObuchow (commented):
Overall looks great to me :) Thanks David :)

AObuchow:
Also, just thinking out loud:

I think documenting the caveats of modifying volume/volumeMounts in pod/container overrides is good enough for now.

If users end up misusing this feature (and ignoring the warning given in this PR), we could add a webhook warning when validating a devworkspace to check if the volume or volumeMount fields are being set in pod/container overrides.

amisevsk (approved):
LGTM apart from @AObuchow's comments

dkwon17:
Thanks @AObuchow @amisevsk , I've updated the PR

docs/additional-configuration.adoc
dkwon17: Updated to:

> The DevWorkspace Operator sets the pod `spec.volumes` field by default for config files, metadata, and credentials. To avoid unexpected behaviour, the `spec.volumes` field should not be overridden.

WDYT?

docs/additional-configuration.adoc
dkwon17: Thanks, updated

AObuchow (approved):

Issue #920, opened 2022-09-06
Annotate DWO-created PVC's


AObuchow:
### Description
Currently, to trigger reconciles when the common PVC is updated, we are relying on [filtering PVC's by their name](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L680) (checking if the updated PVC's name is equal to the common PVC name). 

With https://github.com/devfile/devworkspace-operator/pull/918, the common PVC name which used to be only set in the global DWOC can now be set on a per-workspace basis. However, the PVC handler function does not have access to the common PVC name set for the workspace's configuration.

Instead, it would be better to add a DWO-specific annotation to the PVC spec and filter for this annotation when checking if a reconcile should be triggered when the common PVC is updated. 

### Additional context
Limitation number 3 of https://github.com/devfile/devworkspace-operator/pull/918#issue-1360658181:
> There is [a handler](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go?rgh-link-date=2022-09-02T22%3A30%3A39Z#L680) that checks if the common PVC was deleted. Currently, it filters for PVC's related to DWO by checking the PVC name against the global configs PVC name. In the future, this limitation can be worked around by applying a DWO-specific annotation to all PVC's created by DWO, and filtering by that annotation.

AObuchow:
Though this issue involves specifically annotating DWO-created PVC's so that changes to PVC with custom names (using an external DWOC) can trigger workspace reconciles, @mkuznyetsov has brought up a good point regarding the implications of setting the common PVC name in an external DWOC.

I wanted to document this topic here, for the time being:

Consider the case where two workspaces A and B exist in a namespace. Workspace A uses the common PVC strategy and uses the global DWOC (i.e. it does not use an external DWOC). Workspace B uses the common PVC strategy **and an external DWOC where the PVC name field is configured.**

Upon deletion of workspace B, the common PVC cleanup job will not actually delete the PVC which has a custom name. This is because the PVC [cleanup logic](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/storage/commonStorage.go#L102) will detect that there are other workspaces still using the common PVC strategy (in this case, workspace A).

To resolve this case, we'd have to find out exactly which workspaces are using the custom-named PVC. To do this, we'd have to add extra logic for the common PVC cleanup logic:
1. Check if the workspace being deleted defines a custom PVC name
[...]

PR #1174, opened 2023-09-07
Label per-workspace PVCs with DevWorkspace ID.

amisevsk:
### What does this PR do?
Objects owned by a single DevWorkspace should be labelled with that workspace's ID. This change adds the `controller.devfile.io/devworkspace_id` to per-workspace PVCs to make it easier to find them on the cluster via label selector.

### What issues does this PR fix or reference?
N/A

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [ ] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-che-happy-path`: Happy path for verification integration with Che

PR #1233, opened 2024-02-14
feat: Use labels on PVCs for better reconciling

mkuznyetsov:
### What does this PR do?
Add new labels to indicate PVCs that are in use by Devworkspace Operator.
This allows to trigger workspace reconcile, that must be triggered upon PVC deletion.
and if it's a common PVC that had a name configured by external operator config, only those workspaces that would have the same PVC name should be reconciled 

https://github.com/devfile/devworkspace-operator/issues/920

### Is it tested? How?
Steps to test the new & old behavior with sample DevWorkspaces (DW)

1) install & run a **current version** of DWO:

https://github.com/devfile/devworkspace-operator/?tab=readme-ov-file#run-controller-locally

2) add some Devworkspaces and a devworkspace:

- DW with per-user storage

```
cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: old-dw-per-user
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

- DW with per-workspace storage
```
cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: old-dw-per-workspace
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: basic-volume
        volume: 
          size: 5Gi
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

- External operator config, that is configured with custom PVC name

```
cat <<'EOF' | kubectl apply -f - 
config:
  routing:
    defaultRoutingClass: basic
  workspace:
    defaultStorageSize:
      common: 6Gi
      perWorkspace: 4Gi
    imagePullPolicy: Always
    pvcName: old-custom-pvc-name
metadata:
  name: external-config
  namespace: default
EOF

cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: old-dw-per-user-external
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
        name: external-config
        namespace: default
      controller.devfile.io/storage-type: common
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```
  
Inspect PVC objects, that are created for these repos. 

```
kubectl describe pvc -n $NAMESPACE | grep "Labels:" -A 3 -B 6
```
- They WILL NOT have the label "controller.devfile.io/devworkspace_pvc_type". 
- Only per-workspace PVC will have "controller.devfile.io/devworkspace_id" label

1) build and run a **new version** of operator from this pull-request:
export NAMESPACE="devworkspace-controler"
...
make run

1) Create DWs from step TODO:

- DW with per-user storage

```
cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: new-dw-per-user
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

- DW with per-workspace storage
```
cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: new-dw-per-workspace
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: basic-volume
        volume: 
          size: 5Gi
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

- Another external operator config, that is configured with custom PVC name

```
cat <<'EOF' | kubectl apply -f - 
config:
  routing:
    defaultRoutingClass: basic
  workspace:
    defaultStorageSize:
      common: 20Gi
      perWorkspace: 29Gi
    imagePullPolicy: Always
    pvcName: new-custom-pvc-name
metadata:
  name: new-external-config
  namespace: default
EOF


cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: new-dw-per-user-external
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
        name: new-external-config
        namespace: default
      controller.devfile.io/storage-type: common
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

```
kubectl describe pvc -n $NAMESPACE | grep "Labels:" -A 3 -B 6
```

Observe a label "controller.devfile.io/devworkspace_pvc_type" added to certain kubernetes PVC objects:

- `claim-devworkspace` PVC will not have this label, since this PVC was created with old DWO. Even new DWO
- `old-custom-pvc-name` PVC will not have this label, but `new-custom-pvc-name` will have it
- same for per-workspace, only the PVC for workspace started with new operator will have the type label

Now start deleting PVCs one after another with `kubectl delete pvc`
And see that workspaces are reconciled respectivelly.

### PR Checklist

### DIFF
--- a/controllers/workspace/eventhandlers.go
+++ b/controllers/workspace/eventhandlers.go
@@ -15,6 +15,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	wkspConfig "github.com/devfile/devworkspace-operator/pkg/config"
@@ -47,7 +48,14 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
 }
 
 func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Request {
+	if obj.GetDeletionTimestamp() == nil {
+		// Do not reconcile unless PVC is being deleted.
+		return []reconcile.Request{}
+	}
+
 	// Check if PVC is owned by a DevWorkspace (per-workspace storage case)
+	// TODO: Ensure all new and existing PVC's get the `controller.devfile.io/devworkspace_pvc_type` label.
+	// See: https://github.com/devfile/devworkspace-operator/issues/1250
 	for _, ownerref := range obj.GetOwnerReferences() {
 		if ownerref.Kind != "DevWorkspace" {
 			continue
@@ -62,26 +70,49 @@ func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Req
 		}
 	}
 
-	// TODO: Label PVCs used for workspace storage so that they can be cleaned up if non-default name is used.
-	// Otherwise, check if common PVC is deleted to make sure all DevWorkspaces see it happen
-	if obj.GetName() != wkspConfig.GetGlobalConfig().Workspace.PVCName || obj.GetDeletionTimestamp() == nil {
-		// We're looking for a deleted common PVC
-		return []reconcile.Request{}
+	pvcLabel := obj.GetLabels()[constants.DevWorkspacePVCTypeLabel]
+
+	// TODO: This check is for legacy reasons as existing PVCs might not have the `controller.devfile.io/devworkspace_pvc_type` label.
+	// Remove once https://github.com/devfile/devworkspace-operator/issues/1250 is resolved
+	if pvcLabel == "" {
+		if obj.GetName() != wkspConfig.GetGlobalConfig().Workspace.PVCName {
+			// No need to reconcile if PVC doesn't have a PVC type label
+			// and it doesn't have a name of PVC from global config.
+			return []reconcile.Request{}
+		}
 	}
+
 	dwList := &dw.DevWorkspaceList{}
 	if err := r.Client.List(context.Background(), dwList, &client.ListOptions{Namespace: obj.GetNamespace()}); err != nil {
 		return []reconcile.Request{}
 	}
 	var reconciles []reconcile.Request
+
 	for _, workspace := range dwList.Items {
 		storageType := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAttribute, nil)
 		if storageType == constants.CommonStorageClassType || storageType == constants.PerUserStorageClassType || storageType == "" {
-			reconciles = append(reconciles, reconcile.Request{
-				NamespacedName: types.NamespacedName{
-					Name:      workspace.GetName(),
-					Namespace: workspace.GetNamespace(),
-				},
-			})
+
+			// Determine workspaces to reconcile that use the current common PVC.
+			// Workspaces can either use the common PVC where the PVC name
+			// is coming from the global config, or from an external config the workspace might use
+			workspacePVCName := wkspConfig.GetGlobalConfig().Workspace.PVCName
+
+			if workspace.Spec.Template.Attributes.Exists(constants.ExternalDevWorkspaceConfiguration) {
+				externalConfig, err := wkspConfig.ResolveConfigForWorkspace(&workspace, r.Client)
+				if err != nil {
+					r.Log.Info(fmt.Sprintf("Couldn't resolve PVC name for workspace '%s' in namespace '%s', using PVC name '%s' from global config instead: %s.", workspace.Name, workspace.Namespace, workspacePVCName, err.Error()))
+				} else {
+					workspacePVCName = externalConfig.Workspace.PVCName
+				}
+			}
+			if obj.GetName() == workspacePVCName {
+				reconciles = append(reconciles, reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Name:      workspace.GetName(),
+						Namespace: workspace.GetNamespace(),
+					},
+				})
+			}
 		}
 	}
 	return reconciles
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -20,6 +20,9 @@ const (
 	// DevWorkspaceIDLabel is the label key to store workspace identifier
 	DevWorkspaceIDLabel = "controller.devfile.io/devworkspace_id"
 
+	// DevWorkspacePVCTypeLabel is the label key to identify PVCs used by DevWorkspaces and indicate their storage strategy.
+	DevWorkspacePVCTypeLabel = "controller.devfile.io/devworkspace_pvc_type"
+
 	// WorkspaceIdOverrideAnnotation is an annotation that can be applied to DevWorkspaces
 	// to override the default DevWorkspace ID assigned by the Operator. Is only respected
 	// when a DevWorkspace is created. Once a DevWorkspace has an ID set, it cannot be changed.
--- a/pkg/provision/storage/shared.go
+++ b/pkg/provision/storage/shared.go
@@ -103,6 +103,10 @@ func syncCommonPVC(namespace string, config *v1alpha1.OperatorConfiguration, clu
 	if err != nil {
 		return nil, err
 	}
+	if pvc.Labels == nil {
+		pvc.Labels = map[string]string{}
+	}
+	pvc.Labels[constants.DevWorkspacePVCTypeLabel] = constants.PerUserStorageClassType
 	currObject, err := sync.SyncObjectWithCluster(pvc, clusterAPI)
 	switch t := err.(type) {
 	case nil:
--- a/pkg/provision/storage/perWorkspaceStorage.go
+++ b/pkg/provision/storage/perWorkspaceStorage.go
@@ -227,6 +227,7 @@ func syncPerWorkspacePVC(workspace *common.DevWorkspaceWithConfig, clusterAPI sy
 		pvc.Labels = map[string]string{}
 	}
 	pvc.Labels[constants.DevWorkspaceIDLabel] = workspace.Status.DevWorkspaceId
+	pvc.Labels[constants.DevWorkspacePVCTypeLabel] = constants.PerWorkspaceStorageClassType
 
 	if err := controllerutil.SetControllerReference(workspace.DevWorkspace, pvc, clusterAPI.Scheme); err != nil {
 		return nil, err


controllers/workspace/eventhandlers.go
```diff
@@ -47,44 +47,63 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
...
-		if ownerref.Kind != "DevWorkspace" {
-			continue
-		}
-		return []reconcile.Request{
-			{
-				NamespacedName: types.NamespacedName{
-					Name:      ownerref.Name,
-					Namespace: obj.GetNamespace(),
+	// No need to reconcile PVC, if it doesn't have a label specifying its type
+	pvcLabel, ok := obj.GetLabels()[constants.DevWorkspacePVCTypeLabel]
```
dkwon17, line 51: If a user upgrades from an older version of DWO to the latest version, will the older common PVCs get the new label?

controllers/workspace/eventhandlers.go
```diff
@@ -47,44 +47,63 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
...
-	for _, ownerref := range obj.GetOwnerReferences() {
-		if ownerref.Kind != "DevWorkspace" {
-			continue
-		}
-		return []reconcile.Request{
-			{
-				NamespacedName: types.NamespacedName{
-					Name:      ownerref.Name,
-					Namespace: obj.GetNamespace(),
+	// No need to reconcile PVC, if it doesn't have a label specifying its type
```
dkwon17, line 50:
```suggestion
	// No need to reconcile, if PVC doesn't have a label specifying its type
```

controllers/workspace/eventhandlers.go
```diff
@@ -47,44 +47,63 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
...
+		if err := r.Client.List(context.Background(), dwList, &client.ListOptions{Namespace: obj.GetNamespace()}); err != nil {
+			return []reconcile.Request{}
 		}
+		var reconciles []reconcile.Request
+
+
+		for _, workspace := range dwList.Items {
+			// determin if PVC belongs to the workspace that has to be reconcile by its name
+			// which comes from the DW operator global, or external config that may be defined
+			// for this workspace
```
dkwon17, line 83: I suggest:
```
Determine workspaces to reconcile that use the current common PVC. Workspaces can either use the common PVC where the PVC name is coming from the global config, or from an external config the workspace might use
```

controllers/workspace/eventhandlers.go
```diff
@@ -47,44 +47,63 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
...
+			var workspacePvcName string;
+
+			if (workspace.Spec.Template.Attributes.Exists(constants.ExternalDevWorkspaceConfiguration)) {
```
dkwon17, line 86: Should we check if the workspace uses common/per-user storage by checking 

`storageType := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAttribute, nil)`

first?

controllers/workspace/eventhandlers.go
AObuchow: @mkuznyetsov this is also my concern. There was an older [PR](https://github.com/devfile/devworkspace-operator/pull/1174) that added labels per-workspace PVCs, but I'm almost certain this only impacted _new_ per-workspace PVCs.

My concern stems from 2 things:

## 1. Making sure existing PVCs won't be deleted, then re-created to apply the label.
This would cause user data loss. Looking at the sync code, it seems like PVCs are [not considered mutable](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/sync/sync.go#L183), so the new label would not get applied to existing PVCs. There is however a note that we could still allow for [updating labels](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/sync/sync.go#L59).

If we want to update existing PVCs with the label, this complicates this PR further: either the sync code needs to be updated to only allow modifying PVCs if the labels are modified, or we need to manually `diff` the PVC after calling `SyncObjectWithCluster()`, see that the labels are missing and call` sync.updateObjectGeneric()` to apply the labels. The former approach is much more clean, though a bit trickier as `SyncObjectWithCluster()` is a critical function called throughout DWO. 

## 2. If we decide to not update existing PVCs with the new label, then the `dwPVCHandler()` should still work on PVCs that are missing the new label. 
This is currently not the case with the following check:

```go
	// No need to reconcile, if PVC doesn't have a label specifying its type
	pvcLabel, ok := obj.GetLabels()[constants.DevWorkspacePVCTypeLabel]

	if !ok {
		return []reconcile.Request{}
	}
```

You'd have to change the `dwPVCHandler` code to not assume that the `constants.DevWorkspacePVCTypeLabel` is applied to the PVC in question. If it does not exist, the behaviour should match the behaviour prior to this PR (i.e. the global common PVC name is assumed to be used, and the original PR bug will be present). If the label does exist, then you can make use of the logic that you've implemented. 

---

In general, I think it's easier as a first step to have this PR _not_ update existing PVCs with the new label (i.e. consider my 2nd concern). A later PR could update the sync code to allow label updating on PVCs.

pkg/constants/metadata.go
```diff
@@ -20,6 +20,9 @@ const (
 	// DevWorkspaceIDLabel is the label key to store workspace identifier
 	DevWorkspaceIDLabel = "controller.devfile.io/devworkspace_id"
 
+	// DevWorkspacePVCTypeLabel is the label key to identify DeWorkspace common PVCs and their type
```
AObuchow, line 23: This comment is slightly inaccurate, as we are using this label for both per-workspace and common/per-user PVC's.

I would rewrite it as:
`// DevWorkspacePVCTypeLabel is the label key to identify PVCs used by DevWorkspaces and indicate their storage strategy.`

AObuchow (changes requested):
@mkuznyetsov Thank you for the great first PR :partying_face: Excited to have you contributing to the project. I've made a few initial remarks, but overall things are looking good :)

controllers/workspace/eventhandlers.go
mkuznyetsov: alright, I will rewrite to account for the use of PVC with missing label, and leave a reference to this PR why this has to be this way

controllers/workspace/eventhandlers.go
mkuznyetsov: yes, thanks

AObuchow:
@mkuznyetsov sorry about the failing CI tests, that's due to https://github.com/devfile/devworkspace-operator/issues/1242 (not related to your PR). I've opened https://github.com/devfile/devworkspace-operator/pull/1243 to resolve this.

controllers/workspace/eventhandlers.go
```diff
@@ -47,44 +47,63 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
...
+				externalConfig, _ := wkspConfig.ResolveConfigForWorkspace(&workspace, r.Client)
+				workspacePvcName = externalConfig.Workspace.PVCName
+			}
+			if len(workspacePvcName) == 0 {
```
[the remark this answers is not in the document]
mkuznyetsov: fixed

controllers/workspace/eventhandlers.go
```diff
@@ -47,44 +47,63 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
...
+
+
+		for _, workspace := range dwList.Items {
+			// determin if PVC belongs to the workspace that has to be reconcile by its name
+			// which comes from the DW operator global, or external config that may be defined
+			// for this workspace
+			var workspacePvcName string;
+
+			if (workspace.Spec.Template.Attributes.Exists(constants.ExternalDevWorkspaceConfiguration)) {
+				externalConfig, _ := wkspConfig.ResolveConfigForWorkspace(&workspace, r.Client)
```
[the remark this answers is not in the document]
mkuznyetsov: added logger

pkg/constants/metadata.go
mkuznyetsov: thx, comment applied

controllers/workspace/eventhandlers.go
mkuznyetsov: pr rewritten to support old usecase, when there is no new label

controllers/workspace/eventhandlers.go
mkuznyetsov: comment applied

controllers/workspace/eventhandlers.go
```diff
@@ -47,7 +47,13 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
 }
 
 func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Request {
-	// Check if PVC is owned by a DevWorkspace (per-workspace storage case)
+	if obj.GetDeletionTimestamp() == nil {
+		return []reconcile.Request{}
+	}
+	// we can wrap the code below for handling per-workspace PVCs
```
AObuchow, line 53: Thanks for reporting https://github.com/devfile/devworkspace-operator/issues/1250 -- I think this comment could be now simplified to something along the lines of:

```
// TODO: Ensure all new and existing PVC's get the `controller.devfile.io/devworkspace_pvc_type` label.
// See: https://github.com/devfile/devworkspace-operator/issues/1250
```

pkg/constants/metadata.go
```diff
@@ -20,6 +20,9 @@ const (
 	// DevWorkspaceIDLabel is the label key to store workspace identifier
 	DevWorkspaceIDLabel = "controller.devfile.io/devworkspace_id"
 
+	// DevWorkspacePVCTypeLabel is the label key to identify PVCs used by DevWorkspaces and indicate their storage strategy.
+	DevWorkspacePVCTypeLabel = "controller.devfile.io/devworkspace_pvc_type_label"
```
AObuchow, line 24: I think the actual string value of this label could just be `controller.devfile.io/devworkspace_pvc_type` (i.e. remove the `_label` part at the end), since it's implied by its usage as a label in the PVC spec, and its name `DevWorkspacePVCTypeLabel` in the code.

controllers/workspace/eventhandlers.go
```diff
@@ -47,7 +47,13 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
 }
 
 func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Request {
-	// Check if PVC is owned by a DevWorkspace (per-workspace storage case)
+	if obj.GetDeletionTimestamp() == nil {
+		return []reconcile.Request{}
```
AObuchow, line 51: I'm slightly hesitant to prevent reconciles unless the PVC was deleted (i.e. ignoring modifications to a per-workspace PVC), just because @amisevsk might have had some edge case in mind when he made https://github.com/devfile/devworkspace-operator/commit/2a44ff6da5aa079edd1f21b4032bb9aba10600c2:

> Update the watch for PVCs to also enqueue reconciles for per-workspace
> PVC events. Previously, the per-workspace PVC could be **modified**/deleted
> without the DevWorkspace that owns that PVC noticing.

However:
- We weren't queuing reconciles for common/per-user PVCs unless they were being deleted, so there's an inconsistency between per-workspace PVCs & per-user PVCs causing reconciles upon modification
- Users are **not** supposed to **modify** PVCs created by DWO. Attempting to do so in the OpenShift UI will present the user with a warning: `This resource is managed by DevWorkspaceDW
<DevWorkspace Name> and any modifications may be overwritten. Edit the managing resource to preserve changes.`. **However:** editing the PVC with kubectl/oc will **not give any warning.**
- As discussed in point 1. of my earlier comment https://github.com/devfile/devworkspace-operator/pull/1233#discussion_r1508110341: PVCs are currently considered immutable from DWO's perspective. Any modification made to a DWO-owned PVC cannot be automatically amended by DWO, so reconciling on modification doesn't currently yield any advantages. **However:** in the future, if we resolve https://github.com/devfile/devworkspace-operator/issues/1250 and allow for DWO to modify PVCs to add the `controller.devfile.io/devworkspace_pvc_type` label, then a user removing this label _should_ trigger a reconcile to allow DWO to re-apply the label. This is a problem for the future however...


In summary, I think this optimization is okay :man_shrugging: I just wanted to write out these thoughts beforehand.

controllers/workspace/eventhandlers.go
```diff
@@ -47,7 +47,13 @@ func dwRelatedPodsHandler(obj client.Object) []reconcile.Request {
 }
 
 func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Request {
-	// Check if PVC is owned by a DevWorkspace (per-workspace storage case)
```
AObuchow, line 50: Rather than removing this comment, I'd move it back on-top of the code it was [originally above](https://github.com/devfile/devworkspace-operator/pull/1233/files#diff-1c6fd18e74606af2c4edc2a26c6df20b2030af8b9f72e3c6bb28796ee11620bdR57)

controllers/workspace/eventhandlers.go
```diff
@@ -62,20 +68,41 @@ func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Req
...
-		if storageType == constants.CommonStorageClassType || storageType == constants.PerUserStorageClassType || storageType == "" {
+		//Determine workspaces to reconcile that use the current common PVC.
+		// Workspaces can either use the common PVC where the PVC name
+		// is coming from the global config, or from an external config the workspace might use
+		workspacePVCName := wkspConfig.GetGlobalConfig().Workspace.PVCName
+
+		if workspace.Spec.Template.Attributes.Exists(constants.ExternalDevWorkspaceConfiguration) {
+			externalConfig, err := wkspConfig.ResolveConfigForWorkspace(&workspace, r.Client)
+			if err != nil {
+				r.Log.Info("Couldn't fetch external config for workspace %s, using PVC Name from global config instead", err.Error())
```
AObuchow, line 97: We should test this logger works as expected by:
1. Creating an external DWOC that uses a custom PVC name
2. Using the external DWOC in a workspace that uses the common PVC storage strategy
3. Starting up the workspace that uses the external DWOC, wait for it to be in the READY state
4. Deleting the external DWOC
5. Deleting the common PVC
6. Ensure the log appears in DWO's logs

AObuchow:
@mkuznyetsov when you get a chance, please rebase your PR onto the main branch so that the Validate PR GH Action will pass btw :)

controllers/workspace/eventhandlers.go
mkuznyetsov: Example logger output, if devworkpace had it external config deleted, when attempting to reconcile
`
{"level":"info","ts":"2024-04-10T18:39:02+03:00","logger":"controllers.DevWorkspace","msg":"Couldn't resolve PVC name for workspace 'external-plain1' in namespace 'devworkspace-controller', using PVC name 'claim-devworkspace' from global config instead: could not fetch external DWOC with name ec1 in namespace default: DevWorkspaceOperatorConfig.controller.devfile.io \"ec1\" not found."}
`

AObuchow (approved):
@mkuznyetsov overall looks good to me :) Manual testing was complicated by https://github.com/devfile/devworkspace-operator/issues/1258, but your PR seems good to me. 

Let's wait until I get back from PTO on May 9th before merging this.

Amazing work & congrats on your first (soon to be finally merged!!) PR :D

AObuchow:
@mkuznyetsov I think this PR is good to merge :) Before we do so, can you squash your commits with a `git rebase --autosquash` please? I also think [`a705e11` (#1233)](https://github.com/devfile/devworkspace-operator/pull/1233/commits/a705e118fc784cd0e3274689412bf7d0050055e8) could be squashed into [`08e12bc` (#1233)](https://github.com/devfile/devworkspace-operator/pull/1233/commits/08e12bc0b5938b82766d2bef191e64af9f5a7873).

Additionally, the commit description of [`08e12bc` (#1233)](https://github.com/devfile/devworkspace-operator/pull/1233/commits/08e12bc0b5938b82766d2bef191e64af9f5a7873) could maybe be re-worded to something like: `feat: label PVCs by storage strategy`

controllers/workspace/eventhandlers.go
```diff
@@ -62,20 +70,42 @@ func (r *DevWorkspaceReconciler) dwPVCHandler(obj client.Object) []reconcile.Req
...
+		// is coming from the global config, or from an external config the workspace might use
+		workspacePVCName := wkspConfig.GetGlobalConfig().Workspace.PVCName
+
+		if workspace.Spec.Template.Attributes.Exists(constants.ExternalDevWorkspaceConfiguration) {
+			externalConfig, err := wkspConfig.ResolveConfigForWorkspace(&workspace, r.Client)
+			if err != nil {
+				r.Log.Info(fmt.Sprintf("Couldn't resolve PVC name for workspace '%s' in namespace '%s', using PVC name '%s' from global config instead: %s.", workspace.Name, workspace.Namespace, workspacePVCName, err.Error()))
+			} else {
+				storageType := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAttribute, nil)
+				if storageType == constants.CommonStorageClassType || storageType == constants.PerUserStorageClassType {
```
dkwon17, line 103: Do we need to also check `storageType == ""` here?

AObuchow: @dkwon17 good catch!! Yes, we should be checking this too @mkuznyetsov

dkwon17: Also, should we be doing this check in the first line of this for loop instead?

dkwon17: Because AFAIK we are looking for common/per-user workspaces to reconcile in this for loop

AObuchow: +1 @dkwon17 you're right, the order of checking for workspace storage strategy before iterating over a workspace should have not changed. @mkuznyetsov please restore the [2 lines](https://github.com/devfile/devworkspace-operator/pull/1233/files#diff-1c6fd18e74606af2c4edc2a26c6df20b2030af8b9f72e3c6bb28796ee11620bdL77-L78) that were moved for checking the workspace storage strategy, so that they occur before we retrieve the PVC name.

pkg/provision/storage/shared.go
```diff
@@ -61,7 +61,6 @@ func WorkspaceNeedsStorage(workspace *dw.DevWorkspaceTemplateSpec) bool {
 }
 
 func getPVCSpec(name, namespace string, storageClass *string, size resource.Quantity) (*corev1.PersistentVolumeClaim, error) {
-
```
AObuchow, line 64: @mkuznyetsov sorry for nit-pick, but could you please also restore this newline so that the diff is a bit cleaner?

AObuchow: If possible, please re-add this newline.

pkg/provision/storage/shared.go
```diff
@@ -61,7 +61,7 @@ func WorkspaceNeedsStorage(workspace *dw.DevWorkspaceTemplateSpec) bool {
...
+
```
AObuchow, line 64: If possible, please remove this tab (unless it's coming from go format)

AObuchow (commented):
Please run `make format` with goimports installed `go install golang.org/x/tools/cmd/goimports@latest`

pkg/provision/storage/shared.go
mkuznyetsov: oops, fixed it in the amended commit

AObuchow (approved):
LGTM :)

AObuchow:
Congrats on having your first PR merged to the project @mkuznyetsov :tada: Amazing work :)

PR #1247, opened 2024-04-08
fix: Ignore certificate check on fetch devworkspace template

vinokurig:
### What does this PR do?
In order to fix the `x509: certificate signed by unknown authority error` ignore certificate check on fetch devworkspace template.

https://github.com/devfile/devworkspace-operator/issues/1248

### Is it tested? How?

Apply a workspace, from a devfile that has a nested parent devfile which is hosted on a service with self signed certificate:
```YAML
schemaVersion: 2.2.0
metadata:
  name: sample-using-parent
parent:
  uri: <url to the parent devfile>
```
parent devfile:
```YAML
schemaVersion: 2.2.0
metadata:
  name: parent-devfile
```

### PR Checklist


AObuchow:
I've opened https://github.com/devfile/devworkspace-operator/issues/1248 as the related issue to this PR.

controllers/workspace/http.go
```diff
@@ -29,11 +29,13 @@ var (
 )
 
 func setupHttpClients() {
-	transport := http.DefaultTransport.(*http.Transport).Clone()
-	healthCheckTransport := http.DefaultTransport.(*http.Transport).Clone()
-	healthCheckTransport.TLSClientConfig = &tls.Config{
+	tlsConfig := &tls.Config{
 		InsecureSkipVerify: true,
```
ibuziuk, line 33: @vinokurig but this means that we are skipping verification of TLS for every single request, no?

AObuchow: @ibuziuk this is my understanding as well. Previously, we were only skipping TLS verification for the health check

AObuchow: It might make more sense to add a DevWorkspace Operator Config field to allow disabling the TLS verification which users on an air-gapped cluster would toggle on.

ibuziuk:
@tolusha could you please review?

controllers/workspace/http.go
vinokurig: Added the property, @AObuchow please take a look.

tolusha:
Why not to add configuration with a secret storing extra certificates?
This secret can be created by che-operator.
It is better than ignoring insecure connections.

AObuchow:
> Why not to add configuration with a secret storing extra certificates? This secret can be created by che-operator. It is better than ignoring insecure connections.

+1, maybe something along the lines of adding into the DWOC `config.routing.TLSCertificateSecret` (might not be the best name for the field..). The field would let you specify the name and namespace of the secret. The name and namespace could then be used to [get](https://github.com/devfile/devworkspace-operator/blob/04f536990ebd114b4173bc3ba396e969f7863ebd/pkg/provision/automount/gitconfig.go#L104-L109) the secret and use its contents for the http transport object configuration.

If this approach is taken, we'd have to also modify the DWO controller to watch for Secrets that were created by the Che-Operator. The secrets could maybe contain a label (e.g. `controller.devfile.io/tls-cert`) that allows us to uniquely identify them. For reference, here we setup a [watch](https://github.com/devfile/devworkspace-operator/blob/04f536990ebd114b4173bc3ba396e969f7863ebd/controllers/workspace/devworkspace_controller.go#L701) for secrets that have been [automounted](https://github.com/devfile/devworkspace-operator/blob/04f536990ebd114b4173bc3ba396e969f7863ebd/controllers/workspace/devworkspace_controller.go#L683) . When a secret is created, deleted or updated, we [check](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/predicates.go#L89-L114) which labels it contains, and if it contains an [automount-related label](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/predicates.go#L105-L109), a reconcile is [triggered for all running workspaces](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/eventhandlers.go#L90-L108). A similar approach would have to be taken, but for secrets containing the TLS cert label.

I can help out with this process further or even add extra commits if desired.

controllers/workspace/http.go
AObuchow: Overall looks good to me if we're going with the approach of enabling/disabling TLS verification (one small note in my next comment below about modifying the DevWorkspace Operator Config though).  I do think @tolusha's suggestion of using certificates retrieved from a secret is more optimal than my suggestion though. 

Additionally, we have to commit to any changes made to the DevWorkspace Operator Config since these are considered API changes, and we unfortunately can't easily remove any fields we add. So taking my simpler approach first, and then implementing @tolusha's suggestion later on would mean we'd have to keep `DisableTLSVerification` in the DevWorkspace Operator Config.

Thankfully, though both my suggestion and @tolusha's suggestion would resolve https://github.com/devfile/devworkspace-operator/issues/1248, they both provide different functionality and would not necessarily be redundantly overlapping each other if we were to support both fields in the DevWorkspace Operator Config. So implementing both approaches would not be terrible. @ibuziuk @tolusha @dkwon17 if any of you have any input on whether we should take a simpler approach first or not, please feel free to share.

AObuchow: Regarding adding a new DevWorkspace Operator Config field, two areas need to be addressed:
1. Modifications need to be made to `mergeConfig()` and `GetCurrentConfigString()` in [sync.go](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go).

In [`mergeConfig()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go#L256-L272) we need to copy over the values of the new `DisableTLSVerification` field, something like: 


```diff
	if from.Routing != nil {
		if to.Routing == nil {
			to.Routing = &controller.RoutingConfig{}
		}
		if from.Routing.DefaultRoutingClass != "" {
			to.Routing.DefaultRoutingClass = from.Routing.DefaultRoutingClass
		}
		if from.Routing.ClusterHostSuffix != "" {
			to.Routing.ClusterHostSuffix = from.Routing.ClusterHostSuffix
		}
		if from.Routing.ProxyConfig != nil {
			if to.Routing.ProxyConfig == nil {
				to.Routing.ProxyConfig = &controller.Proxy{}
			}
			to.Routing.ProxyConfig = proxy.MergeProxyConfigs(from.Routing.ProxyConfig, defaultConfig.Routing.ProxyConfig)
		}
+		if from.Routing.DisableTLSVerification != nil {
+			to.Routing.DisableTLSVerification = from.Routing.DisableTLSVerification
+		}
	}
```


And in [`GetCurrentConfigString()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go#L463-L470) we need to declare if DisableTLSVerification is enabled. Something like:

```diff
	if routing != nil {
		if routing.ClusterHostSuffix != "" && routing.ClusterHostSuffix != defaultConfig.Routing.ClusterHostSuffix {
			config = append(config, fmt.Sprintf("routing.clusterHostSuffix=%s", routing.ClusterHostSuffix))
		}
		if routing.DefaultRoutingClass != defaultConfig.Routing.DefaultRoutingClass {
			config = append(config, fmt.Sprintf("routing.defaultRoutingClass=%s", routing.DefaultRoutingClass))
		}
+		if routing.DisableTLSVerification != nil && *routing.DisableTLSVerification {
+			config = append(config, "routing.DisableTLSVerification=true")
+		}
	}
```

You can run `make test` to make sure everything's working as expected. In particular, [this test](https://github.com/devfile/devworkspace-operator/blob/04f536990ebd114b4173bc3ba396e969f7863ebd/pkg/config/sync_test.go#L428-L441) should be passing.

2. The CRDs need to be re-generated since a new field was added to the DevWorkspace Operator Config. Run the following, and add the resulting changes into a separate commit: `make generate manifests fmt generate_default_deployment generate_olm_bundle_yaml`

tolusha:
My suggestion:
```
TLSCertificateSecretRef *types.SecretReference
```

PR #1193, opened 2023-10-20
Add ability for project clone to 'bootstrap' devworkspaces

amisevsk:
### What does this PR do?
Adds DevWorkspace attribute `controller.devfile.io/bootstrap-devworkspace`. If set to `true`, this attribute will configure project clone to do the following:

1. Clone all projects in the DevWorkspace as normal
2. Search cloned projects for a devfile.yaml or .devfile.yaml
3. Merge contents of this devfile into the DevWorkspace
   * Projects are merged (i.e. both devfile and DevWorkspace projects
     are used)
   * Attributes from the DevWorkspace override attributes from the
     devfile
   * Components, commands, events, etc. from the Devfile are used.
4. Apply the updated DevWorkspace to the cluster, automatically
   restarting the workspace.

### What issues does this PR fix or reference?
Closes #1192 

### Is it tested? How?
Basic testing of this PR is to create a DevWorkspace that uses the attribute and imports a project with a devfile.yaml at its root:
```yaml
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: plain
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
      controller.devfile.io/bootstrap-devworkspace: true
    projects:
      - name: che-dashboard
        git:
          remotes:
            origin: "https://github.com/eclipse-che/che-dashboard.git"
    components:
      - name: web-terminal
        container:
          image: quay.io/wto/web-terminal-tooling:next
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```
After starting, the workspace should update so that the `.spec.template` of the workspace matches the devfile of the imported project, with the merge process above (attributes, projects are merged instead of overwritten)

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [ ] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-che-happy-path`: Happy path for verification integration with Che

PR #1251, opened 2024-04-16
Create an initcontainer for persistent home setup

dkwon17:
### The problem

Today, the persistent home setup occurs in the entrypoint [[1]](https://github.com/devfile/developer-images/blob/3d74f3745dc75fe20ca283d3425ca5dbce849f7f/base/ubi8/Dockerfile#L123) of the tooling container (assuming it is based on the UBI). Notably, the entrypoint runs stow [[2]](https://github.com/devfile/developer-images/blob/3d74f3745dc75fe20ca283d3425ca5dbce849f7f/universal/ubi8/entrypoint.sh#L62) to make symbolic links for the home directory to be saved on the PVC.

If a poststart event depends on files or edits files located in the home directory, there is a race condition because the poststart event may start running before stow has finished running. Stow takes about 6 seconds to complete.

This race condition causes https://github.com/eclipse-che/che/issues/22914, because the checode poststart event runs before the entrypoint script completes.

### Proposed solution
To solve the race condition altogether, the entrypoint script must be complete before the post start events run. To do this, an initcontainer that runs the entrypoint script can be used.

#### What is the container image for the initcontainer?
To have a minimal effect on workspace startup times, the container image for the initcontainer should be the same image as the tooling container's image to avoid pulling an extra image. Assuming that the image is already pulled, I've observed no significant startup impact (see demo video in the next section) when running an extra container running the tooling image, even if the tooling image is large (quay.io/devfile/universal-developer-image:ubi8-latest has a compressed size of 2.9GB).

When thinking about it in the Docker perspective, running a container is only creating a writable layer on top of the already-pulled image layers, therefore if the image is already pulled the image size would not have a significant impact on container startup. ~~I'm guessing~~ this is what also happens in OpenShift/CRI-O, because adding the initcontainer did not increase workspace startup time significantly.

https://github.com/eclipse-che/che/issues/22914
Fixes https://github.com/devfile/devworkspace-operator/issues/1260

Demo: https://youtu.be/RQeKKJgaogM

In the demo above, the init container named `init-persistent-home` runs for than 9 seconds at the 0:20 second mark. Stow is only run once (because of this if statement [[3]](https://github.com/devfile/developer-images/blob/3d74f3745dc75fe20ca283d3425ca5dbce849f7f/universal/ubi8/entrypoint.sh#L60)) to set up the new PVC.

So for subsequent workspace starts, the `init-persistent-home` init container takes less than a second to finish. Here is a second demo showing that:

https://youtu.be/hKnb1mgBSqA?si=yuJ3EKmj_dX0qHj0&t=20

### Is it tested? How?

To test this PR with Eclipse Che:

Install DWO on your cluster by adding this catalog source and installing DWO from OperatorHub:
```
kind: CatalogSource
metadata:
spec:
  sourceType: grpc
  image: quay.io/dkwon17/devworkspace-operator-index:init-container
  publisher: Red Hat
  updateStrategy:
    registryPoll:
      interval: 5m
```
After waiting a few minutes, the DWO from the new catalog should appear in OperatorHub. Install the one from `DevWorkspace Operator Catalog`.

image

Install Eclipse Che with chectl:
```
chectl server:deploy -p openshift --skip-devworkspace-operator
```
Enabled persistent user home by adding the following in the CheCluster CR:
```
spec:
  devEnvironments:
    persistUserHome:
      enabled: true
```

Create a workspace with the following repositories and confirm that the `podman info` command runs without errors after workspace startup (just like in the demo video):
* https://github.com/che-samples/web-nodejs-sample
* empty sample workspace
* https://github.com/dkwon17/repo-with-poststart
* https://gist.githubusercontent.com/dkwon17/9ecac5397bef1178f77f308565814996/raw/4801b3c07a2cababd53b3594754c628cde7b8947/child-devfile.yaml

### PR Checklist

### PR Checklist


pkg/library/home/persistentHome.go
```diff
@@ -95,3 +115,32 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+// This function assumes that the provided image contains the /entrypoint.sh script.
+// If the script is not present in the image, the init container will not fail the workspace from starting.
+func addPersistentStorageInitContainer(dwTemplateSpecCopy *v1alpha2.DevWorkspaceTemplateSpec, image string) {
+	initContainerComponent := v1alpha2.Component{
+		Name: "init-persistent-home",
+		ComponentUnion: v1alpha2.ComponentUnion{
+			Container: &v1alpha2.ContainerComponent{
+				Container: v1alpha2.Container{
+					Image:   image,
+					Command: []string{"/bin/sh", "-c", "/entrypoint.sh || true"},
```
dkwon17, line 129: Making this assumption creates coupling between DWO and the UDI

dkwon17: @AObuchow it could be possible to introduce the initcontainer in the dashboard, I am interested in investigating

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +101,64 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+		component.Attributes.Get(constants.MergedContributionsAttribute, &err)
+		if err == nil {
+			firstUnflattenedComponent = component
+			break
+		}
+	}
+
+	if firstUnflattenedComponent.Name != "" {
+		image := firstUnflattenedComponent.Container.Image
+		command := []string{"/bin/sh", "-c", "/entrypoint.sh || true"}
```
dkwon17, line 156: I'm not sure how we can infer the initcontainer image and command. Maybe we can introduce a new attribute which would be added by Che, and have the attribute be configurable in the DWOC?

Some cases to think about:
- User is using a parent devfile
- User's main container image does not have the /entrypoint.sh script
- User's main container image has `["tail", "-f", "/dev/null"]` in it's entry point, which would cause the init container to never terminate

dkwon17:
Some cases to think about in general:
- User is using a parent devfile
- User's main container image does not have the /entrypoint.sh script
- User's main container image has `["tail", "-f", "/dev/null"]` in it's entry point, which would cause the init container to never terminate

AObuchow:
Amazing work so far David. I think this is a good proposed solution to fixing the race condition between the UDI's entrypoint and any post-start commands. 

There are a lot of considerations to make so my thoughts are a bit all over the place, but here goes:

## Should we create the home persistence init container component in DWO or Che Dashboard?
This is something you mentioned in your earlier draft. I think it makes the most sense for the init container component to be added in DWO:

- We want to minimize the transformation of a devfile -> devworkspace, so that [DWO's devfile bootstrap feature](https://github.com/devfile/devworkspace-operator/pull/1193) continues to work. If the Che-Dashboard was adding an extra component to the devworkspace for home persistence, this extra component would be lost when DWO restarts the devworkspace from the cloned devfile (as the bootstrap feature will use components from the devfile, not the devworkspace)
- I'm not sure if the Che-Dashboard can currently "read" the Che Cluster CR in order to determine the home persistence settings? The dashboard would have to watch for changes in the Che Cluster CR or get re-deployed (with the configured home persistence settings) when the home persistence configuration changes.
- There's already some level of coupling between the UDI and DWO for home persistence: mounting `/home/user/` as a volume assumes that `$HOME == /home/user/`, which is the case for the UDI but not necessarily true for other user-provided images.
    - Since there's already coupling between DWO and the UDI, we should keep this the same rather than move the coupling to be between the UDI and Dashboard.


## Should we call the main/init containers entrypoint? Or just the code relevant to home persistence (stow, cp)?
I don't think we want to call the UDI's entrypoint directly, as there's other code in it that is irrelevant to home persistence. Instead, we'll want to move the [stow-related code](https://github.com/devfile/developer-images/blob/main/universal/ubi8/entrypoint.sh#L51-L70) into its own script and execute that.

This script could potentially reside in the UDI under a pre-determined name (e.g.` /home-persistence-init.sh`), or it could be injected by DWO as the initContainer's command. In the latter case, we could test whether stow is installed in the container before calling it. 

## Should we _hope_ that the main/tooling component has stow installed? Or consider shipping a dedicated init container with stow installed.
Instead of depending on the first container component in the devworkspace being the UDI & having stow installed, we could ship a minimal image that has stow installed (similar to how we ship the project clone container).

This would simplify the `inferInitContainer()` [step](https://github.com/devfile/devworkspace-operator/pull/1251/files#diff-f1f3498fd50bbc235d4d09f5f5485a91f5a4a5cfeb20b127c96fc849a30ba56bR137-R162), but require us to productize a new container downstream that we'd be shipping. 

For a first pass of your proposed change, I think it's fine to require users to use a main/tooling image based on the UDI (with stow installed, as well as tooling installed in /home/tooling/ and $HOME set to /home/user/). This is already the case for $HOME persistence to work as expected, anyhow.

pkg/library/home/persistentHome.go
AObuchow: > Maybe we can introduce a new attribute which would be added by Che, and have the attribute be configurable in the DWOC?

Can you expand upon this? 
Do you mean, having the[ initContainer image](https://github.com/devfile/devworkspace-operator/pull/1251/files#diff-1eda12737507a4bd1af2a4ea8bdaf278095ec16b208cfc2007487a507e65d36dR159-R160) and the command it runs be a Che Cluster CR field?

AObuchow:
> Some cases to think about in general:
> 
>     * User's main container image does not have the /entrypoint.sh script
> 
>     * User's main container image has `["tail", "-f", "/dev/null"]` in it's entry point, which would cause the init container to never terminate

Both of these can be resolved by having DWO call stow and cp directly (by default), rather than requiring them to be in the entrypoint. We could then let users configure the init container command through the DWOC  if they don't want to use the default we provide.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -152,6 +152,12 @@ type PersistentHomeConfig struct {
 	// Must be used with the 'per-user'/'common' or 'per-workspace' storage class in order to take effect.
 	// Disabled by default.
 	Enabled *bool `json:"enabled,omitempty"`
+	// Determines whether the init container that initializes the persistent home directory should be disabled.
+	// When the `/home/user` directory is persisted, the init container is used to initialize the directory before
+	// the workspace starts. If set to true, the init container will not be used.
+	DisableInitContainer *bool `json:"initContainerEnabled,omitempty"`
+	// Image is the container image to use for the init container that initializes the persistent home directory.
+	InitContainer *dw.Container `json:"initContainer,omitempty"`
```
AObuchow, line 160: Is this field supposed to be about specifying the image to use for the init container, or the devworkspace container component? It seems to be the latter in the way it's used. 

Also, small typo in the docs `// Image is the container image to use...`

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +99,77 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+
+		pluginSource := component.Attributes.GetString(constants.PluginSourceAttribute, nil)
+		if pluginSource == "" || pluginSource == "parent" {
+			// First, non-imported container component is selected
+			nonImportedComponent = component
+			break
+		}
+	}
+
+	createOutFile := "touch /home/user/.init-persistent-home"
```
AObuchow, line 152: Is there an underlying reason for putting the `.init-persistent-home` script in the /home/user/ directory? Maybe it'd be better to keep it in the root directory `/` (similar to the entrypoint, though we might not have write access there...) or in` /tmp/` so that users don't notice a shell script being added to their $HOME directory?

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +99,77 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+	runSetupScript := `([ -n "$HOME_SETUP_SCRIPT" ] && [ -f "$HOME_SETUP_SCRIPT" ] && echo "Running $HOME_SETUP_SCRIPT" && (source "$HOME_SETUP_SCRIPT" || true))`
```
AObuchow, line 153: This `$HOME_SETUP_SCRIPT` is supposed to be an environment variable that points to a user-configured script for configuring home-persistence, right? For cases where users may be using an image other than the UDI and the default `runStow` script for home-persistence won't work. 

And to make sure I understand the logic here: we first check if the user defined the `$HOME_SETUP_SCRIPT` environment variable, then ensure it points to an actual file, then we run it using `source` and ensure it does not cause workspace startup to fail with `|| true`. Correct?

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +99,77 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+	runStow := `
+	(command -v stow > /dev/null && \
+		STOW_COMPLETE=/home/user/.stow_completed && \
+		[ ! -f $STOW_COMPLETE ] && \
+		echo "Running/ stow" && \
+		stow . -t /home/user/ -d /home/tooling/ --no-folding --adopt -v 2 > /tmp/stow.log 2>&1 && \
```
AObuchow, line 159: We don't want to use stow's `--adopt` option, as this will move any file present in `/home/user/` but not in `/home/tooling/` to `/home/tooling/` if there are any [stow conflicts](https://www.gnu.org/software/stow/manual/html_node/Conflicts.html). We don't want to move anything from `/home/user/` (persistent storage) into `/home/tooling/` (ephemeral storage) since this can cause data loss.

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +99,77 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+		cp /home/tooling/.viminfo /home/user/.viminfo && \
```
AObuchow, line 160: It's probably safer to use `cp -n` for `cp --no-clobber`, which will cancel the cp operator if the destination file already exists. It's rare that this will ever be needed, since the home persistence setup is only supposed to happen the first time a workspace starts, but it shouldn't hurt.

I should have used `cp -n` in the UDI's entrypoint from the start, to be honest.

pkg/library/home/persistentHome.go
AObuchow: My mistake, I misinterpreted what we were doing here. I now see that `/home/user/.init-persistent-home` is supposed to be the command output of the user-provided `$HOME_SETUP_SCRIPT` or the default `runStow` script. 

I still think it might make more sense to put the `.init-persistent-home` into /tmp/ (since we put the pre and post-start command outputs as well as the stow command output there). It also would make sense to not make it hidden, i.e. the file should be `/tmp/init-persistent-home` instead of `/tmp/.init-persistent-home/`.

AObuchow (commented):
Left some more feedback, great improvements :)

dkwon17:
Sorry, the PR still had rough work, I've still updated the PR recently with some of your feedback

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
dkwon17: It's the latter, fixed

pkg/library/home/persistentHome.go
dkwon17: > This $HOME_SETUP_SCRIPT is supposed to be an environment variable that points to a user-configured script for configuring home-persistence, right? 

Right, it allows users to provide their own script in the image.

> and ensure it does not cause workspace startup to fail with || true. Correct?

Yes, the `|| true` is to ensure that the initcontainer command's exit code is 0 allowing the workspace to still start if there was a non-zero exit code in the initialization script.

pkg/library/home/persistentHome.go
dkwon17: Thanks, fixed

pkg/library/home/persistentHome.go
dkwon17: Sorry, this functionality was just a small idea, I've removed it in my latest commits

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +99,92 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+	echo 'No home setup script found at $HOME_SETUP_SCRIPT'
+fi
+
+echo "Checking for stow command"
+STOW_COMPLETE=/home/user/.stow_completed
+if command -v stow &> /dev/null; then
+
+	if  [ ! -f $STOW_COMPLETE ]; then
+		echo "Running stow command"
+		stow . -t /home/user/ -d /home/tooling/ --no-folding --adopt -v 2 > /tmp/stow.log 2>&1
```
AObuchow, line 167: `--adopt` should still be removed

AObuchow (commented):
Thanks for the updates @dkwon17 :)

pkg/library/home/persistentHome.go
dkwon17: You're right, I'll remove it,  I had `--adopt` because without it, stow would fail when we create a conflicting file using a mounted configmap or secret like `/home/user/.config/containers/storage.conf`, for example:  https://eclipse.dev/che/docs/stable/end-user-guide/enabling-overlay-with-a-configmap/

I'm wondering, have we investigated how https://eclipse.dev/che/docs/stable/end-user-guide/requesting-persistent-storage-for-workspaces/#requesting-persistent-storage-in-a-pvc works with persistent-home? If there is a user-defined PVC with a mount path in the home directory, and if that PVC already has some data in it, I think stow could fail if there are conflicts with the contents of /home/tooling. I would have to verify though, but I would expect this to be the case.

A possible fix can be, the initcontainer could maybe read mount paths from volumes/secrets/configmap that mount under /home/user, and add them to the `.stow-local-ignore` file before running stow

dkwon17: I can create a separate issue for this case

AObuchow: > You're right, I'll remove it, I had `--adopt` because without it, stow would fail when we create a conflicting file using a mounted configmap or secret like `/home/user/.config/containers/storage.conf`, for example: [eclipse.dev/che/docs/stable/end-user-guide/enabling-overlay-with-a-configmap](https://eclipse.dev/che/docs/stable/end-user-guide/enabling-overlay-with-a-configmap/)

It's worth seeing if you can mount a custom `storage.conf` to `/home/tooling/.config/containers/storage.conf` instead of `/home/user/...` with a configmap and let stow create the symbolic link in $HOME. It's possible that this won't work due to the permissions that OpenShift uses when mounting files.

If this does work, then we'll have to document this quirk. 


> I'm wondering, have we investigated how [eclipse.dev/che/docs/stable/end-user-guide/requesting-persistent-storage-for-workspaces/#requesting-persistent-storage-in-a-pvc](https://eclipse.dev/che/docs/stable/end-user-guide/requesting-persistent-storage-for-workspaces/#requesting-persistent-storage-in-a-pvc) works with persistent-home? If there is a user-defined PVC with a mount path in the home directory, and if that PVC already has some data in it, I think stow could fail if there are conflicts with the contents of /home/tooling. I would have to verify though, but I would expect this to be the case.

I also think that stow would fail in this case but haven't verified it myself. There's an underlying assumption that the PVC's `/home/user/` directory would be empty. 


> A possible fix can be, the initcontainer could maybe read mount paths from volumes/secrets/configmap that mount under /home/user, and add them to the `.stow-local-ignore` file before running stow

The .stow-local-ignore actually [ignores source files](https://www.gnu.org/software/stow/manual/stow.html#Ignore-Lists) that we do not want to create symbolic links for. For example, we're ignore the `.vimrc` and cp'ing it in the UDI's entrypoint since vim does not allow the `.vimrc` to be a symbolic link for security reasons.

Home persistence definitely complicates a lot of things.  If files that are auto-mounted by OpenShift can be stowed (i.e. symbolic links for these files can be created by stow) then a potential workaround is to mount them to `/home/tooling/` in addition to `/home/user/`. I believe stow would ignore these files (since they exist in both the source and target directory), and they would remain regular files rather than symbolic links.

We'll have to investigate the case of automount volumes/secrets/configmaps conflicts with stow and potential workarounds.

AObuchow: > I can create a separate issue for this case

That'd be great when you have a moment :) Thank you!!

AObuchow:
@dkwon17 since we're adding new fields to the DWOC, you'll have to modify `mergeConfig()` and `GetCurrentConfigString() `in [sync.go](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go?rgh-link-date=2024-04-23T18%3A39%3A58Z) to account for the new fields. 

Additionally, the CRDs will have to be re-generated by running the following and adding the resulting changes into a separate commit: `make generate manifests fmt generate_default_deployment generate_olm_bundle_yaml`.

Since we're still working through the details of this PR, these 2 steps can wait until we're set on the new DWOC fields.

pkg/library/home/persistentHome.go
dkwon17: I created the issue: https://github.com/devfile/devworkspace-operator/issues/1257

> The .stow-local-ignore actually [ignores source files](https://www.gnu.org/software/stow/manual/stow.html#Ignore-Lists) that we do not want to create symbolic links for.

Yes, this would still help us prevent conflicts since stow won't try to create a potentially conflicting link

dkwon17: ^ Unless I'm misunderstanding something 


Let's continue the discussion at https://github.com/devfile/devworkspace-operator/issues/1257, I think it's out of scope of this PR

AObuchow: Thanks for creating the issue :pray: 

> Yes, this would still help us prevent conflicts since stow won't try to create a potentially conflicting link

I see what you mean now (thinking of the `storage.conf` case where the same file with different contents are present in both `/home/tooling/` and `/home/user/`).

I think your solution should work as long as we ignore individual files rather than entire directories, as this might cause other files we want to create symbolic links for to be ignored by stow.

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +99,92 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+		cp -n /home/tooling/.viminfo /home/user/.viminfo
+		cp -n /home/tooling/.bashrc /home/user/.bashrc
+		cp -n /home/tooling/.bash_profile /home/user/.bash_profile
+		touch $STOW_COMPLETE
+	else
+		echo "Stow command already run"
```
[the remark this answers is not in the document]
dkwon17: Thank you updated to:

"Stow command already run. If you wish to re-run it, delete $STOW_COMPLETE from the persistent volume and restart the workspace."

AObuchow (commented):
Some minor observations but looking awesome, great work so far @dkwon17

pkg/library/home/testdata/persistent-home/no-home-vm-when-disabled.yaml
```diff
@@ -6,6 +6,7 @@ input:
     workspace:
       persistUserHome:
         enabled: false
+        disableInitContainer: true
```
AObuchow, line 9: Btw, was this added just for explicitness? I would imagine we don't need to set `disableInitContainer: false` if `persistUserHome.enabled` is set to `false`?

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +117,107 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
...
+
+func addInitContainer(dwTemplateSpec *v1alpha2.DevWorkspaceTemplateSpec) {
+
+	initContainer := inferInitContainer(dwTemplateSpec)
+	if initContainer == nil {
+		// if cannot infer initcontainer, fail quietly
+		return
+	}
+
+	if !initComponentExists(dwTemplateSpec) {
```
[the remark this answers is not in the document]
dkwon17: The checks are set like this to still give some kind of freedom for the user for the initcontainer. For example, if the devfile author sets a the `init-persistent-home` container component like: https://github.com/devfile/devworkspace-operator/blob/57b69c4baca6017f5a44b9e1f9eea926010f0821/pkg/library/home/testdata/persistent-home/creates-home-vm-initcontainer-already-defined.yaml#L19-L23

and not provide the `init-persistent-home` command or event, the DWO would just add the necessary command and event to make it an initcontainer for the workspace pod.

WDYT?

pkg/library/home/persistentHome.go
```diff
@@ -95,3 +117,107 @@ func storageStrategySupportsPersistentHome(workspace *common.DevWorkspaceWithCon
 	storageClass := workspace.Spec.Template.Attributes.GetString(constants.DevWorkspaceStorageTypeAttribute, nil)
 	return storageClass != constants.EphemeralStorageClassType
 }
+
+func addInitContainer(dwTemplateSpec *v1alpha2.DevWorkspaceTemplateSpec) {
+
+	initContainer := inferInitContainer(dwTemplateSpec)
+	if initContainer == nil {
+		// if cannot infer initcontainer, fail quietly
+		return
```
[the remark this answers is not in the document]
dkwon17: Updated, thanks :+1:

pkg/library/home/testdata/persistent-home/no-home-vm-when-disabled.yaml
dkwon17: That's right, I removed it actually since `disableInitContainer` isn't used at all if `persistUserHome.enabled` is false

pkg/library/home/persistentHome.go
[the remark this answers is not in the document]
dkwon17: If the devfile already has a container component with the name `init-persistent-home`, what should we do?

Two options are, we can either disable persistent home completely, or we can just keep it and have the DWO create the devfile command (not to be confused with a container's `command` field) and devfile event and still let the persistent home feature work. The code right now just does the latter.

> This seems like it still may allow for users (who have edit rights to the DWOC) to set a custom initContainer and command that would be added to all devworkspaces

Devfile authors can create a devfile with any container image anyways, and I don't see how this is related to DWOC anymore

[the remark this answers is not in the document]
dkwon17: Sorry, to clarify a bit more, users won't be able to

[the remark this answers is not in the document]
dkwon17: > This seems like it still may allow for users (who have edit rights to the DWOC) to set a custom initContainer and command that would be added to all devworkspaces

Sorry, to clarify a bit more, users won't be able to do that because of this update to this PR: https://github.com/devfile/devworkspace-operator/pull/1251/commits/76db7ab5e2d5bc0eed4bc5fe05cb74cea1f84e1e

I think https://github.com/devfile/devworkspace-operator/pull/1251#discussion_r1581323016 is simpler and clearer to the user, and I don't feel too strongly on the current approach, I will update my PR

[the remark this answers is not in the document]
dkwon17: Updated: https://github.com/devfile/devworkspace-operator/pull/1251/commits/3e55cc347d5dacfd0c38f6c25b3da6fdc91acb95

svor (approved):
Tested as described in the description with che-plugin registry repo, works as expected.
Great job @dkwon17 
 
screenshot-eclipse-che apps ci-ln-jt06z1b-76ef8 origin-ci-int-aws dev rhcloud com-2024 04 30-13_08_24

dkwon17:
@tolusha could you provide a reproducer? Does this happen when enabling `persistUserHome` and starting an existing workspace?

ibuziuk (approved):

AObuchow:
@dkwon17 your latest changes look good to me :) I still need to do some final manual testing on a cluster but I think this is just about ready to merge.

My 2 last thoughts on this PR:
- We need to verify this doesn't break the Podman wrapper in the UDI (i.e. kubedock) since now Stow is being invoked before the [podman wrapper is symlinked](https://github.com/devfile/developer-images/blob/main/universal/ubi8/entrypoint.sh#L3-L48). However, as we briefly discussed on call, I don't think this is an issue since `/home/tooling/` (where the Podman wrapper [exists](https://github.com/devfile/developer-images/blob/main/universal/ubi8/entrypoint.sh#L31)) is on PATH
- It'd be nice to squash some of the commits from this PR together to create an easier-to-follow git log. My suggestion is that you could probably just have ~3 commits: one for the feature implementation, one for changing the CRDs/resources and one for adding tests. I know some commits (like [this one](https://github.com/devfile/devworkspace-operator/pull/1251/commits/3e55cc347d5dacfd0c38f6c25b3da6fdc91acb95)) change both the implementation and test cases, so don't worry if these remain their own separate commits.

dkwon17:
Thanks for the feedback, I was able to verify the kubedock functionality [1] in the UDI for this PR

[1] https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.9/html/release_notes_and_known_issues/new-features#enhancement-crw-3367

AObuchow:
> Thanks for the feedback, I was able to verify the kubedock functionality [1] in the UDI for this PR
> 
> [1] [access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.9/html/release_notes_and_known_issues/new-features#enhancement-crw-3367](https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.9/html/release_notes_and_known_issues/new-features#enhancement-crw-3367)

@dkwon17 awesome, thank you for verifying this. I'll do some final testing and have this merged.

dkwon17:
Sounds good @AObuchow , do you want me to squash the commits and force push?

AObuchow:
@dkwon17 yes please squash your commits then force push if you get a chance. I know you're going to be on PTO however, so I can do this for you.

AObuchow:
@dkwon17 thanks for squashing your commits. Will do final review & have this merged by Wednesday morning :)

AObuchow:
Note to myself: need to create a DWO issue before this is merged for tracking in the 0.28 release miletstones

pkg/library/home/persistentHome.go
```diff
@@ -27,6 +27,24 @@ import (
...
+const initScript = `(echo "Checking for stow command"
+STOW_COMPLETE=/home/user/.stow_completed
+if command -v stow &> /dev/null; then
+  if  [ ! -f $STOW_COMPLETE ]; then
+    echo "Running stow command"
+    stow . -t /home/user/ -d /home/tooling/ --no-folding -v 2 > /home/user/.stow.log 2>&1
+    cp -n /home/tooling/.viminfo /home/user/.viminfo
+    cp -n /home/tooling/.bashrc /home/user/.bashrc
+    cp -n /home/tooling/.bash_profile /home/user/.bash_profile
+    touch $STOW_COMPLETE
```
AObuchow, line 39: It'd be nice to log here that the init script completed here, but that can be added in a future PR. We could also check whether stow completed successfully and report that here.

AObuchow (approved):
Tested in Che on an OpenShift 4.15 cluster with a few different devfiles and tried restarting workspaces as well -- works as expected :) 

Looks good to me :) Great job @dkwon17 :partying_face:

Issue #1248, opened 2024-04-08
Unable to resolve parent devworkspace when using self-signed certs on disconnected/airgap cluster


AObuchow:
### Description
When using self-signed certs to host a devworkspace on an air-gapped cluster, starting a devworkspace that references a parent devworkspace by URI will fail. The parent devworkspace will not be resolved, and the devworkspace will fail citing a `x509: certificate signed by unknown authority error`

### How To Reproduce
1. Host the following devworkspace and parent devworkspace on a service that uses a self-signed certificate on a an air-gapped cluster:

**DevWorkspace:**

```YAML
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: child-devworkspace
spec:
  started: true
  routingClass: 'basic'
  parent: <url to the parent devfile>
  template:
    components:
      - name: tools
        container:
          env:
            - name: DEVFILE_ENV_VAR
              value: 'true'
```

**Parent DevWorkspace:**

```YAML
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: parent-devworkspace
spec:
  template:
    components:
      - name: tools
        attributes:
           controller.devfile.io/merge-contribution: true
        container:
          image:  quay.io/devspaces/udi-rhel8:latest
          env:
            - name: PARENT_ENV_VAR
              value: 'true'
    commands:
      - id: parent-command
        exec:
          label: '2. This command from the parent'
          component: tools
          commandLine: echo "Hello from parent"
```
2. Apply/start the (child) DevWorkspace
3. The Devworkspace will fail to start up, citing an error similar to:

`failed to resolve component parent by URI: failed to fetch file from <URL>: Get <URL>: x509: certificate signed by unknown authority`

### Expected behavior
The DevWorkspace should start up successfully.

### Additional context
[Downstream DevSpaces issue](https://issues.redhat.com/browse/CRW-6001)

PR #1259, opened 2024-05-01
Inject certificate to http client from a configmap referenced in the …

vinokurig:
…config

### What does this PR do?
* Add a new config to be able to define a configmap with certificate content
* Add a watcher that will trigger certificate injection to http client on the configmap create / update events

fixes https://github.com/devfile/devworkspace-operator/issues/1248
### Is it tested? How?

1. Create a configmap with certificate:
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: <name>
  namespace: <namespace>
  labels:
data:
  custom-ca-certificates.pem: <certificate pem content>
```
2. Define the configmap in the config:
```yaml
config:
  routing:
    tlsCertificateConfigmapRef:
      name: <the configmap name>
      namespace: <the configmap namespace>
```
3. Check the flow from https://github.com/devfile/devworkspace-operator/issues/1248 issue.
### PR Checklist


pkg/config/sync.go
```diff
@@ -269,6 +269,12 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
 			}
 			to.Routing.ProxyConfig = proxy.MergeProxyConfigs(from.Routing.ProxyConfig, defaultConfig.Routing.ProxyConfig)
 		}
+		if from.Routing.TLSCertificateConfigmapRef != nil {
+			if to.Routing.TLSCertificateConfigmapRef == nil {
+				to.Routing.TLSCertificateConfigmapRef = &controller.ConfigmapReference{}
+			}
+			to.Routing.TLSCertificateConfigmapRef = proxy.MergeTLSCertificateConfigmapRef(from.Routing.TLSCertificateConfigmapRef, defaultConfig.Routing.TLSCertificateConfigmapRef)
```
tolusha, line 276: It doesn't look like from `proxy` package.

controllers/workspace/eventhandlers.go
```diff
@@ -106,3 +108,11 @@ func (r *DevWorkspaceReconciler) runningWorkspacesHandler(obj client.Object) []r
 	}
 	return reconciles
 }
+
+func (r *DevWorkspaceReconciler) certificateHandler(obj client.Object) []reconcile.Request {
+	certsPem, ok := obj.(*corev1.ConfigMap).Data["custom-ca-certificates.pem"]
```
tolusha, line 113: All configmap keys can contain certificates. 
We shouldn't stick to a single hardcoded key.

controllers/workspace/devworkspace_controller.go
```diff
@@ -667,9 +667,8 @@ func (r *DevWorkspaceReconciler) getWorkspaceId(ctx context.Context, workspace *
 	}
 }
 
-func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	setupHttpClients()
-
+func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager, k8s client.Client) error {
+	setupHttpClients(k8s)
```
tolusha, line 671: We can use `mgr.GetClient()` instead of passing another parameter.

pkg/config/sync.go
vinokurig: Right, moved the function to the same class as a private function.

pkg/config/predicates.go
```diff
@@ -25,8 +25,13 @@ import (
 func Predicates() predicate.Predicate {
 	return predicate.Funcs{
 		UpdateFunc: func(evt event.UpdateEvent) bool {
-			if config, ok := evt.ObjectNew.(*dw.DevWorkspaceOperatorConfig); ok {
-				syncConfigFrom(config)
+			if newConfig, ok := evt.ObjectNew.(*dw.DevWorkspaceOperatorConfig); ok {
+				syncConfigFrom(newConfig)
+				if oldConfig, ok := evt.ObjectOld.(*dw.DevWorkspaceOperatorConfig); ok {
```
tolusha, line 30: Why do we need this?

controllers/workspace/predicates.go
```diff
@@ -112,3 +126,12 @@ func objectIsAutomountResource(obj client.Object) bool {
 	}
 
 }
+
+func objectIsCertificateConfigmap(obj client.Object) bool {
+	configmapRef := config.GetGlobalConfig().Routing.TLSCertificateConfigmapRef
```
tolusha, line 131: Please check ` config.GetGlobalConfig().Routing` is not null as well

controllers/workspace/devworkspace_controller.go
```diff
@@ -668,19 +669,23 @@ func (r *DevWorkspaceReconciler) getWorkspaceId(ctx context.Context, workspace *
...
 	if err != nil {
 		return err
 	}
 
-	var emptyMapper = func(obj client.Object) []reconcile.Request {
+	var certificatesHandler = func(obj client.Object) []reconcile.Request {
+		certs, ok := readCertificates(mgr.GetClient())
+		if ok {
+			injectCertificates(certs, httpClient.Transport.(*http.Transport))
+		}
```
tolusha, line 682: Could we print logs if something went wrong?

controllers/workspace/eventhandlers.go
vinokurig: done

controllers/workspace/devworkspace_controller.go
vinokurig: done

pkg/config/predicates.go
vinokurig: To trigger certificate injection on the `TLSCertificateConfigmapRef` config change

controllers/workspace/predicates.go
vinokurig: done

controllers/workspace/devworkspace_controller.go
vinokurig: done

controllers/workspace/http.go
```diff
@@ -64,3 +76,30 @@ func setupHttpClients() {
...
+		return configMap.Data, true
+	} else {
+		logger.Error(err, "Failed to read configmap with certificates")
+	}
+	return map[string]string{}, false
+}
+
+func injectCertificates(certsPem []byte, transport *http.Transport) {
+	caCertPool := x509.NewCertPool()
+	ok := caCertPool.AppendCertsFromPEM(certsPem)
```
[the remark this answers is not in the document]
vinokurig: fixed by using the previous cert pool

tolusha:
Since httpsClient can be changed after configuring certificates, it is better to move initialization from `SetupWithManager` to `Reconcile`

tolusha:
Sounds good, need to be tested

vinokurig:
Tested the latest PR version with the steps to reproduce: works as expected.

dkwon17:
I was able to successfully test this feature on CRC, by adding a plugin contribution (`spec.contributions`) uri hosted from CRC which is signed by a self-signed certificate:
```
kind: DevWorkspace
metadata:
  name: code-latest
spec:
  started: true
  template:
    projects:
      - name: web-nodejs-sample
        git:
          remotes:
    components:
      - name: dev
        container:
          memoryLimit: 512Mi
          memoryRequest: 256Mi
          cpuRequest: 1000m
    commands:
      - id: say-hello
        exec:
          component: dev
          workingDir: ${PROJECT_SOURCE}/app
  contributions:
    - name: che-code
      uri: https://eclipse-che.apps-crc.testing/plugin-registry/v3/plugins/che-incubator/che-code/latest/devfile.yaml
      components:
        - name: che-code-runtime-description
          container:
            env:
              - name: CODE_HOST
                value: 0.0.0.0

```

controllers/workspace/http.go
```diff
@@ -64,3 +76,33 @@ func setupHttpClients() {
 		Timeout:   500 * time.Millisecond,
 	}
 }
+
+func readCertificates(k8s client.Client, logger logr.Logger) (map[string]string, bool) {
+	configmapRef := config.GetGlobalConfig().Routing.TLSCertificateConfigmapRef
```
AObuchow, line 81: I mentioned in another comment asking whether we want to be setting up the http client on a per-workspace basis (i.e. during every workspace reconcile). Doing so doesn't seem to have any advantage to me.

However, if we do want to setup the http client on a per-workspace basis, then here we might as well take advantage of the fact that any given workspace might use a [custom/external DevWorkspaceOperatorConfig](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#setting-an-alternate-configuration-for-a-workspace). This means, users could provide different certs for different workspaces.

I don't think this is actually necessary or that useful however, and in the case of Che, every workspace uses the Che-owned external DevWorkspaceOperator Config. This means Che users have to decide between having their workspaces have Che settings propagated automatically to their workspaces, and using their own user-provided external DevWorkspaceOperatorConfig.

AObuchow (commented):
@vinokurig awesome PR so far :) I left a few minor comments.

Also, could you please install goimports `go install golang.org/x/tools/cmd/goimports@latest` and then run `make fmt`, as the CI checks are currently failing?

controllers/workspace/devworkspace_controller.go
```diff
@@ -114,6 +114,8 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		Ctx:              ctx,
 	}
 
+	setupHttpClients(r.Client, r.Log)
```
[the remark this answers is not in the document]
AObuchow: My mistake, I just saw @tolusha's [comment](https://github.com/devfile/devworkspace-operator/pull/1259#issuecomment-2098231359). Ignore this.

controllers/workspace/http.go
AObuchow: After seeing @tolusha's [comment](https://github.com/devfile/devworkspace-operator/pull/1259#issuecomment-2098231359), I think we might as well take advantage of the external workspace DWOC here. 

Rather than getting the config from the global config `config.GetGlobalConfig().Routing.TLSCertificateConfigmapRef`, we can move [`setupHttpClients()`](https://github.com/devfile/devworkspace-operator/pull/1259/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R117) so that it is called immediately after the [external DWOC is resolved](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L132-L143). Then we can pass in the config to `setupHttpClients()`.

AObuchow:
@vinokurig thanks for the latest changes :) Unfortunately, your last commit seems to have misformated imports I believe, as the CI formatting check is still failing. Please run `make fmt` again when you have a moment.

pkg/config/sync.go
```diff
@@ -269,6 +269,12 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
 			}
 			to.Routing.ProxyConfig = proxy.MergeProxyConfigs(from.Routing.ProxyConfig, defaultConfig.Routing.ProxyConfig)
 		}
+		if from.Routing.TLSCertificateConfigmapRef != nil {
+			if to.Routing.TLSCertificateConfigmapRef == nil {
+				to.Routing.TLSCertificateConfigmapRef = &controller.ConfigmapReference{}
```
AObuchow, line 274: Now that we've removed `controller.ConfigmapReference{}` from the DWOC, you'll have to use `types.NamespacedName` here.

pkg/config/sync.go
```diff
@@ -426,6 +432,22 @@ func mergeContainerSecurityContext(base, patch *corev1.SecurityContext) *corev1.
 	return patched
 }
 
+func mergeTLSCertificateConfigmapRef(operatorConfig, clusterConfig *controller.ConfigmapReference) *controller.ConfigmapReference {
+	mergedConfigmapReference := &controller.ConfigmapReference{
```
AObuchow, line 436: Now that we've removed `controller.ConfigmapReference{}` from the DWOC, you'll have to use `types.NamespacedName` here as well, and change this function's parameters.

AObuchow (commented):
Small changes required now that we've changed the CRDs

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -240,6 +243,13 @@ type ProjectCloneConfig struct {
 	Env []corev1.EnvVar `json:"env,omitempty"`
 }
 
+type ConfigmapReference struct {
```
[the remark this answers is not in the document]
vinokurig: done

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -61,6 +61,9 @@ type RoutingConfig struct {
 	// DevWorkspaces. However, changing the proxy configuration for the DevWorkspace Operator itself
 	// requires restarting the controller deployment.
 	ProxyConfig *Proxy `json:"proxyConfig,omitempty"`
+	// TLSCertificateConfigmapRef defines the name and namespace of the configmap with a certificate to inject to http
```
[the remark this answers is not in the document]
vinokurig: done

controllers/workspace/http.go
```diff
@@ -64,3 +76,33 @@ func setupHttpClients() {
...
+	if configmapRef == nil {
+		return nil, false
+	}
+	configMap := &corev1.ConfigMap{}
+	namespacedName := &types.NamespacedName{
+		Name:      configmapRef.Name,
+		Namespace: configmapRef.Namespace,
+	}
+	err := k8s.Get(context.Background(), *namespacedName, configMap)
+	if err == nil {
```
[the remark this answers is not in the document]
vinokurig: I simplified the code, but regarding throwing an error I do not think this is the case here. If the certificate reference is not configured we should not see warning or errors about it.

controllers/workspace/http.go
```diff
@@ -28,8 +34,14 @@ var (
 	healthCheckHttpClient *http.Client
 )
 
-func setupHttpClients() {
+func setupHttpClients(k8s client.Client, logger logr.Logger) {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
+	certs, ok := readCertificates(k8s, logger)
+	if ok {
```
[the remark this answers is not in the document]
vinokurig: done

controllers/workspace/http.go
```diff
@@ -64,3 +76,33 @@ func setupHttpClients() {
...
+	}
+	return map[string]string{}, false
+}
+
+func injectCertificates(certsPem []byte, transport *http.Transport) {
+	caCertPool := transport.TLSClientConfig.RootCAs
+	if caCertPool == nil {
+		caCertPool = x509.NewCertPool()
+	}
+	ok := caCertPool.AppendCertsFromPEM(certsPem)
```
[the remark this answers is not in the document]
vinokurig: done

controllers/workspace/http.go
vinokurig: moved [setupHttpClients()](https://github.com/devfile/devworkspace-operator/pull/1259/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R117) and pass workspace config into the function.

pkg/config/sync.go
vinokurig: done

pkg/config/sync.go
vinokurig: done

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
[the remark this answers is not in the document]
vinokurig: Reverted back because the [types.NamespacedName](https://github.com/kubernetes/apimachinery/blob/c9c3e94f52f06444120a09bfbdb11bec38a5c8e1/pkg/types/namespacedname.go#L27-L30) fields do not have the json representation.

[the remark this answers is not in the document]
AObuchow: @vinokurig sounds good - my mistake about that.

AObuchow:
@vinokurig Thank you for the latest changes! Please run `make fmt` again when you have a moment.

AObuchow:
The controller env tests are currently failing because of the change from having the HTTP client be setup on [controller initialization basis](https://github.com/devfile/devworkspace-operator/pull/1259/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529L671) to a [per-reconcile basis](https://github.com/devfile/devworkspace-operator/pull/1259/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R145). Previously, we used to [mock the HTTP transport](https://github.com/devfile/devworkspace-operator/commit/71c70456d9f5d2746b062d46d3cbf9072643ef97). Before the tests which needed a mocked HTTP transport object, it would be [configured by the testing suite](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller_test.go#L344-L356). But now, the HTTP transport is being configured during the workspace reconcile, overriding the mocked value. 

I'm not sure yet of how to elegantly solve this problem. 

We might be able to modify the `setupHttpClients()` function to check whether a test environment is being used, and not override the mocked http.Transport object if so. Though, it seems a bit awkward to check for a testing environment in production code.

vinokurig:
@AObuchow What if we revert the healthcheck http to be initialised on operator start but leave the base http client to be initialised on Reconcile?

AObuchow:
> @AObuchow What if we revert the healthcheck http to be initialised on operator start but leave the base http client to be initialised on Reconcile?

This would work and is a potential option we could take (i.e. something like have separate `setupHealthCheckHttpClient()` and `setupTransportHttpClient()` functions). However, I'm not the biggest fan of making this change since it moves around code that used to be executed in a single place, though I'm open to it.


> Since httpsClient can be changed after configuring certificates, it is better to move initialization from SetupWithManager to Reconcile

After reconsideration, I don't think it's worth moving the HTTP client initialization logic from the operator startup to the reconciliation function, given the complications it brings to the test suite. It definitely is advantageous from a UX point of view to have the HTTP client re-initialized on reconcile, as it allows users to configure the certificates while the operator is running and have the operator pick up these changes more immediately. However, previously admins just had to restart the devworkspace-controller pod to pick up HTTP configuration changes from the DWOC, which did not seem to be a problem. 

Maybe there's another detail I'm missing here?

Additionally, in the PRs current state, the [global httpClient object](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/http.go#L27) is being modified when calling `setupHttpClients()` and then being [retrieved later](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L262), rather than being a [return value](https://github.com/devfile/devworkspace-operator/pull/1259/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R145) of the `setupHttpClients()` function. I believe this is somewhat of an anti-pattern, as we are relying on the side-effect of `setupHttpClients()` rather than making it clear that `setupHttpClients()` modifies the httpClient.

@vinokurig @tolusha @ibuziuk I am open to either keeping the HTTP client initialization logic as part of the operator initialization, or moving it to the reconcile function as proposed in this PR, but would appreciate any input given the complications that this change brings.

vinokurig:
As another possible solution I propose to just [update certificates](https://github.com/devfile/devworkspace-operator/blob/4a99251a9063b89ab1d851e8fa9be085b59b7c13/controllers/workspace/http.go#L42-L46) on Reconcile step instead of [reinitialise the http client](https://github.com/devfile/devworkspace-operator/blob/4a99251a9063b89ab1d851e8fa9be085b59b7c13/controllers/workspace/devworkspace_controller.go#L145). @AObuchow @tolusha @ibuziuk WDYT?

AObuchow:
> As another possible solution I propose to just [update certificates](https://github.com/devfile/devworkspace-operator/blob/4a99251a9063b89ab1d851e8fa9be085b59b7c13/controllers/workspace/http.go#L42-L46) on Reconcile step instead of [reinitialise the http client](https://github.com/devfile/devworkspace-operator/blob/4a99251a9063b89ab1d851e8fa9be085b59b7c13/controllers/workspace/devworkspace_controller.go#L145). @AObuchow @tolusha @ibuziuk WDYT?

This sounds like the best solution so far to me and should work, great idea :) 

Something to be careful with here: If we retrieve the SSL certificates from the [workspace's config](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L132) (and not the global config), we need to `Clone()` the httpClient's transport object before updating the certificates, and use that for the rest of the workspace's reconcile function instead of updating and using the global httpClient object. The reason we should clone/copy the global httpClient instead of modifying it directly during a reconcile, is because it will be shared between [concurrent reconciles ](https://github.com/devfile/devworkspace-operator/blob/715d43f8c7da2da2f906706e723ada9cbbb450fd/deploy/templates/components/manager/manager.yaml#L75-L76). If workspace A and B both configure an external DWOC that reference different SSL certificates, then workspace A and B will also need their own HTTP clients to be configured with different certs.

Since the above might add some weird complexity for a case that might never happen (are users really going to need different certs for individual workspaces?...), I think it's fine to just use the global config when updating the SSL certificates (i.e. `globalConfig := config.GetGlobalConfg()`)

tolusha (approved):

controllers/workspace/devworkspace_controller.go
```diff
@@ -142,6 +144,13 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reqLogger = reqLogger.WithValues(constants.DevWorkspaceIDLoggerKey, workspace.Status.DevWorkspaceId)
 	reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
 
+	// Inject ca certificates to the http clint if the certificates configmap is created and defined in the config.
+	if certs, ok := readCertificates(r.Client, config, r.Log); ok {
```
AObuchow, line 148: 2 observations:
1. We still want to use `config.GetGlobalConfig()` instead of `config` when calling `readCertificates()`, because the `config` struct might refer to an external DWOC that is used [only by a single workspace](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#setting-an-alternate-configuration-for-a-workspace), yet, the global httpClient that is used by _all_ workspaces is being modified. Since multiple reconciles can happen in parallel, this could lead to a race condition where workspaces that do not configure SSL certs may end up using them when they are mid-reconcile. 

The `config *controller.OperatorConfiguration` parameter could thus be removed since we'd always be using  `config.GetGlobalConfig()` as the input, or it can remain. 

2. I think it might be worth to refactor this block of code that calls `readCertificates()` and `injectCertificates()` into an exported function `InjectCertificates()`, since we are calling the same things in `setupHttpClients()` and it'd be nice to hide some of these implementation details from the reconcile function. However, this is just my personal opinion, and not mandatory.

controllers/workspace/devworkspace_controller.go
```diff
@@ -142,6 +144,13 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reqLogger = reqLogger.WithValues(constants.DevWorkspaceIDLoggerKey, workspace.Status.DevWorkspaceId)
 	reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
 
+	// Inject ca certificates to the http clint if the certificates configmap is created and defined in the config.
```
AObuchow, line 147: Correcting a typo (`clint` -> `client)` and some capitalization:

`// Inject CA certificates to the HTTP client if a certificates configmap is created and defined in the config.`

AObuchow (changes requested):
Some final remarks but I think this is nearly ready for merge after the final changes :+1:

AObuchow:
@dkwon17 please have a look at this PR when you have a moment & I encourage you to test it on CRC if possible (though if you are caught up with other tasks, don't worry -- having another set of eyes on the code is what matters most right now :) ).

controllers/workspace/devworkspace_controller.go
dkwon17: I just want to double check, for an Eclipse Che installation, is the che-managed DWOC considered the global config?

AObuchow: No, but very good point. The global DWOC is created in the namespace where DWO is installed whereas the Che-owned DWOC is created in the namespace where Che is installed.

My suggestion would limit us from letting Che configure the certs used from DWO. Instead, users will have to modify the global DWOC.

If we want to allow for external DWOC's, we should copy the http transport struct when injecting certs and use the copied one for the rest of the reconcile function to avoid modifying the global http transport object (and potentially causing a race condition).

controllers/workspace/http.go
```diff
@@ -63,4 +72,41 @@ func setupHttpClients() {
 		Transport: healthCheckTransport,
 		Timeout:   500 * time.Millisecond,
 	}
+	InjectCertificates(k8s, logger)
+}
+
+func InjectCertificates(k8s client.Client, logger logr.Logger) {
```
AObuchow, line 78: Exported functions are generally placed above non-exported functions (in the code). However, this is just a style nitpick that I forgot to mention and can be changed later on as I don't want to delay merging this PR due to time zone differences and the upcoming DWO 0.28.0 upstream release that we will hopefully start tomorrow.

AObuchow (approved):
LGTM, @dkwon17 please have a look when you have a moment.

ibuziuk (approved):

dkwon17 (approved):

controllers/workspace/http.go
```diff
@@ -63,4 +72,41 @@ func setupHttpClients() {
...
+		logger.Error(err, "Failed to read configmap with certificates")
+		return nil, false
+	}
+	return configMap.Data, true
+}
+
+func injectCertificates(certsPem []byte, transport *http.Transport) {
+	caCertPool := transport.TLSClientConfig.RootCAs
+	if caCertPool == nil {
+		caCertPool = x509.NewCertPool()
```
dkwon17, line 107: According to [1]
```
	// RootCAs defines the set of root certificate authorities
	// that clients use when verifying server certificates.
	// If RootCAs is nil, TLS uses the host's root CA set.
```

@vinokurig  should we use SystemCertPool instead? 
```suggestion
		systemCertPool, err := x509.SystemCertPool()
		caCertPool = systemCertPool
```


[1] https://pkg.go.dev/crypto/tls

AObuchow: Good catch @dkwon17.

I'm under the impression we should be using the SystemCertPool as well. If the DWOC is configured to use specific certs, a DevWorkspace that tries to resolve a URI  that uses another certificate will fail (when this should normally would work). For example, `https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml` can normally be resolved, but cannot when custom certs are configured with the functionality this PR introduces. 

To reproduce this, the following devworkspace can be used:

```YAML
kind: DevWorkspace
metadata:
  name: code-latest
spec:
  contributions:
    - components:
        - container:
            env:
              - name: CODE_HOST
                value: 0.0.0.0
          name: che-code-runtime-description
      name: che-code
      uri: 'https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml'
  started: false
  template:
    commands:
      - exec:
          component: dev
          workingDir: '${PROJECT_SOURCE}/app'
        id: say-hello
    components:
      - container:
          cpuRequest: 1000m
          image: 'quay.io/devfile/universal-developer-image:latest'
          memoryLimit: 512Mi
          memoryRequest: 256Mi
          sourceMapping: /projects
        name: dev
    projects:
      - git:
          remotes:
            origin: 'https://github.com/che-samples/web-nodejs-sample.git'
        name: web-nodejs-sample
```

The above devworkspace fails with:

```
status:
  conditions:
    - lastTransitionTime: '2024-05-22T19:45:49Z'
      message: 'Error processing devfile: failed to resolve component che-code by URI: failed to fetch file from https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml: Get "https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml": tls: failed to verify certificate: x509: certificate signed by unknown authority'
      reason: BadRequest
      status: 'True'
      type: FailedStart
```

If the call to `x509.SystemCertPool()` returns an error, we could either log the error and abandon trying to inject the certs, or we could create a new cert pool with  `x509.NewCertPool()` and log the error. The latter option is probably preferable.

AObuchow (changes requested):
@vinokurig please consider @dkwon17's request to use the System Cert Pool instead of a new Cert Pool. If you have any opposing thoughts to this, however, please bring them up .

Also, for reference, here's how I've been testing this (thanks to @dkwon17 for the help with figuring this out):
1. Install DWO with the changes from this PR [using olm](https://github.com/devfile/devworkspace-operator?tab=readme-ov-file#with-operator-lifecycle-manager-olm): Set the `DWO_IMG` environment variable to your quay repo, do a `make docker`, then set the `DWO_BUNDLE_IMG`, `DWO_INDEX_IMG` & `DEFAULT_DWO_IMG` variables to your quay repos. Note: `DEFAULT_DWO_IMG`  ==  `DWO_IMG`. Finally, do a `make generate_olm_bundle_yaml build_bundle_and_index register_catalogsource` while logged into an OpenShift cluster.
2. Install Che from OperatorHub, create a Che Cluster CR and wait for Che to be started (i.e. the Che Dashboard URL is ready and accessible)
3. Find the Che plugin registry URL (can be seen on the Che Cluster CR in OpenShift UI), something like: `https://eclipse-che.apps.ci-ln-f2dgggk-72292.origin-ci-int-gce.dev.rhcloud.com/plugin-registry/v3`
4. `oc apply -f ./samples/code-latest.yaml` (i.e. [this sample devworkspace](https://github.com/devfile/devworkspace-operator/blob/main/samples/code-latest.yaml))
5. Modify the code-latest devworkspace so that the contribution URI points to `<che-plugin-registry-url>/plugins/che-incubator/che-code/latest/devfile.yaml`. For example: `https://eclipse-che.apps.ci-ln-f2dgggk-72292.origin-ci-int-gce.dev.rhcloud.com/plugin-registry/v3/plugins/che-incubator/che-code/latest/devfile.yaml`. @dkwon17's [earlier comment](https://github.com/devfile/devworkspace-operator/pull/1259#issuecomment-2101849769) gives an example of this.
6. The DevWorkspace should fail to resolve due to `tls: failed to verify certificate: x509: certificate signed by unknown authority'`
7. Visit the Che Plugin registry URL that hosts `che-code/latest/devfile.yaml` (from step 5.) and [download](https://medium.com/@menakajain/export-download-ssl-certificate-from-server-site-url-bcfc41ea46a2) the .cert from your browser. 
8. Create a configmap using the downloaded cert data, and reference it in the DWOC
9. Set the `code-latest` DevWorkspace `spec.started` to `true`.
10. The DevWorkspace should now resolve the che-code component, and start up eventually.

To verify that the default system pool certs are working for other certificate authorities, repeat step 5. but use `https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml` for the uri of the DevWorkspace's che-code contribution:

```YAML
 spec:
  contributions:
    - components:
        - container:
            env:
              - name: CODE_HOST
                value: 0.0.0.0
          name: che-code-runtime-description
      name: che-code
      uri: 'https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml'
```

The DevWorkspace should start up successfully.

controllers/workspace/http.go
```diff
@@ -63,4 +72,46 @@ func setupHttpClients() {
...
+
+func injectCertificates(certsPem []byte, transport *http.Transport, logger logr.Logger) {
+	caCertPool := transport.TLSClientConfig.RootCAs
+	if caCertPool == nil {
+		systemCertPool, err := x509.SystemCertPool()
+		if err != nil {
+			logger.Error(err, "Failed to load system cert pool")
+			caCertPool = x509.NewCertPool()
+		}
+		caCertPool = systemCertPool
```
AObuchow, line 112: This line needs to be in an `else` block so that `caCertPool` will only be set to `systemCertPool` if there was not error in retrieving the SystemCertPool (`x509.SystemCertPool()` will return a nil value and an error if an error occurs).

```go
	systemCertPool, err := x509.SystemCertPool()
	if err != nil {
		logger.Error(err, "Failed to load system cert pool")
		caCertPool = x509.NewCertPool()
	} else {
	        caCertPool = systemCertPool
	} 

```

vinokurig: Yes, thank you for the catch, fixed.

AObuchow:
@vinokurig thanks for the latest updates :) Will begin testing shortly. If all seems well on both my end and @dkwon17's end, we'll squash your fixup commits and have this merged.

Issue #1270, opened 2024-06-04
Provide workspaces.annotations field in DWOC for all devworkspaces


dkwon17:
### Description

There is no way to add annotations to all workspaces in the DWOC

My proposal is to add `config.workspaces.annotations` field in DWOC that allows setting annotations to add to a devworkspace's pod.

### Additional context

This feature would help implement:
* https://issues.redhat.com/browse/CRW-6550
* https://github.com/eclipse-che/che/issues/22809

AObuchow:
Resolved in https://github.com/devfile/devworkspace-operator/pull/1271

PR #1267, opened 2024-05-27
chore: Add v0.28.0 bundle to main branch

AObuchow:
### What does this PR do?
Adds the v0.28.0 bundle & digest bundle to the main branch.

Based off of the following two commits:
- [Automated commit that added v0.28.0 bundle](https://github.com/devfile/devworkspace-operator/commit/1d2a10145bfaa7e9bad47b3a5f929c6ec5a4dc7f)
- [Automated commit that added v0.28.0 bundle digest](https://github.com/devfile/devworkspace-operator/commit/4f820bc18b3497114c97834a10bde929ce63a50c)
### What issues does this PR fix or reference?
n/a

### Is it tested? How?
n/a 

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [ ] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-che-happy-path`: Happy path for verification integration with Che
Signed-off-by: Andrew Obuchowicz <aobuchow@redhat.com>Signed-off-by: Andrew Obuchowicz <aobuchow@redhat.com>

PR #1271, opened 2024-06-04
Support specifying annotations for devworkspace pods in the DWOC

dkwon17:
### What does this PR do?
Provides `config.workspace.podAnnotations` field for DWOC to specify annotations for the devworkspace pods

https://github.com/devfile/devworkspace-operator/issues/1270

### Is it tested? How?

1. Deploy DWO by using this catalog source and installing DWO from OperatorHub:
```
kind: CatalogSource
metadata:
spec:
  sourceType: grpc
  image: quay.io/dkwon17/devworkspace-operator-index:annotations-dwoc
  publisher: Red Hat
  updateStrategy:
    registryPoll:
      interval: 5m
```
2. Update the `devworkspace-operator-config` DWOC by adding an annotation:
```
config:
  workspace:
    podAnnotations:
      test-annotation: test
```

3. Create a workspace:
```
oc apply -f ./samples/plain.yaml -n openshift-operators 
```
4. Once the workspace pod is created, check the pod annotations and verify that `test-annoation: test` exists:
```
metadata:
  annotations:
    test-annotation: test
    ...
```
### PR Checklist


pkg/provision/workspace/deployment.go
```diff
@@ -197,6 +197,7 @@ func getSpecDeployment(
 						constants.DevWorkspaceIDLabel:   workspace.Status.DevWorkspaceId,
 						constants.DevWorkspaceNameLabel: workspace.Name,
 					},
+					Annotations: workspace.Config.Workspace.PodAnnotations,
```
AObuchow, line 200: I noticed we are [collecting the pod additions annotations](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/workspace/deployment.go#L302-L304) but never actually using them. Currently, this is not a big deal, since annotations are never gathered while creating the pod additions list. I believe this logic was added to `mergePodAdditions` for completeness, but hasn't been used yet.

We could merge the `podAdditions.aAnnotations` with the `workspace.Config.Workspace.PodAnnotations`, or just use the workspace.Config.Workspace.PodAnnotations alone as there won't be any actual difference. Though in the future, this might have to make this change if the podAddtions bring in any pod annotations.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -147,6 +147,8 @@ type WorkspaceConfig struct {
 	// the value "0" should be used. By default, the memory limit is 128Mi and the memory request is 64Mi.
 	// No CPU limit or request is added by default.
 	DefaultContainerResources *corev1.ResourceRequirements `json:"defaultContainerResources,omitempty"`
+	// PodAnnotations defines the metadata.annotations for pods created by the DevWorkspace Operator.
```
AObuchow, line 150: We should clarify that these annotations are for DevWorkspace pods specifically, i.e.
`// PodAnnotations defines the metadata.annotations for DevWorkspace pods created by the DevWorkspace Operator.`

pkg/config/sync.go
```diff
@@ -384,6 +384,15 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
 			}
 			to.Workspace.DefaultContainerResources = mergeResources(from.Workspace.DefaultContainerResources, to.Workspace.DefaultContainerResources)
 		}
+
```
AObuchow, line 387: Changes also need to be made to [`GetCurrentConfigString()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/config/sync.go#L487) to either give the list of the config's podAnnotations, or indicate that they are being set.

AObuchow (commented):
Looks great to me so far. Some minor comments. Will do some testing today, but this looks nearly ready to me :)

AObuchow:
Tested on Minikube and works as expected :+1:

pkg/provision/workspace/deployment.go
dkwon17: Sounds good, I've updated to use `mergePodAdditions`

pkg/config/sync.go
dkwon17: updated

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
dkwon17: Updated to suggestion

AObuchow:
@dkwon17 I haven't debugged this to verify, but I think the CI is failing because we're now creating a workspace pod that has an empty `map[string]string` of annotations, which gets converted into `nil` when the object is created on the cluster. As a result, the [SyncObjectWithCluster](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/sync/sync.go#L67)'s diff function for pods is probably seeing that the pod on the cluster has `nil` annotations, when it expects it to be an empty `map[string]string` of annotations. This would be why we're seeing printed diffs in the CI logs:

```
    	... // 2 ignored fields
    	Spec: v1.DeploymentSpec{
    		Replicas: &1,
    		Selector: &{MatchLabels: {"controller.devfile.io/devworkspace_id": "workspace05b48af9f4d14dd4"}},
    		Template: v1.PodTemplateSpec{
    			ObjectMeta: v1.ObjectMeta{
    				... // 9 identical fields
    				DeletionGracePeriodSeconds: nil,
    				Labels:                     {"controller.devfile.io/creator": "", "controller.devfile.io/devworkspace_id": "workspace05b48af9f4d14dd4", "controller.devfile.io/devworkspace_name": "test-devworkspace"},
  - 				Annotations:                map[string]string{},
  + 				Annotations:                nil,
    				OwnerReferences:            nil,
    				Finalizers:                 nil,
    				ManagedFields:              nil,
    			},
```
I think this [function](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/sync/diff.go#L65-L70) would be worth checking to see how it's behaving. Looking at its implementation, it doesn't seem like this would cause the `diff` to conclude that the cluster and spec objects are different, but maybe something subtle is going on.

dkwon17:
Thanks for the feedback, I updated the PR, I learned that before, it was this check always caused the operator to try updating the deployment:
https://github.com/devfile/devworkspace-operator/blob/b49427cecf4ba036a7587be4e7319b4dfec42dcf/pkg/provision/sync/diff.go#L58

AObuchow (approved):
Looks good to me, awesome work @dkwon17 :grin:.
I recommend squashing some of your commits together (if possible) before we merge, but asides from that this looks good to merge :) 

We should also probably merge https://github.com/devfile/devworkspace-operator/pull/1267 before merging this PR btw.

dkwon17:
I've squashed some commits and reviewed: https://github.com/devfile/devworkspace-operator/pull/1267

AObuchow:
> I've squashed some commits and reviewed: #1267

Thank you :) 

Sorry, I should have specified earlier, but it's probably worth updating the commit titles (and maybe descriptions) just to clarify a bit more what these commits change.

For example, [`b7ccba1` (#1271)](https://github.com/devfile/devworkspace-operator/pull/1271/commits/b7ccba1872ca7a64fa1031a4fd15ff8a8098748f) could be titled as: "Add config.workspace.podAnnotations field to DWOC".

[`2bfdcd4` (#1271)](https://github.com/devfile/devworkspace-operator/pull/1271/commits/2bfdcd46cc86695bf48119db1d2d752292a2de5e) could be titled as "Add pod annotations from DWOC & podAdditions to devworkspace pod" and the description could include "Fix #1270" to link the original issue to the commit. 

Also.. I just had a second thought about a detail in this PR so please hold off on merging. Will write up my comment shortly.

controllers/workspace/devworkspace_controller.go
```diff
@@ -413,6 +413,9 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		allPodAdditions = append(allPodAdditions, *routingPodAdditions)
 	}
 
+	annotationAdditions := controllerv1alpha1.PodAdditions{Annotations: workspace.Config.Workspace.PodAnnotations}
```
AObuchow, line 416: I was on the fence about whether adding the `workspace.Config.Workspace.PodAnnotations` should be done here or somewhere closer to the devworkspace's [`getSpecDeployment()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/workspace/deployment.go#L131) function, given that we have a function [`getAdditionalAnnotations()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/workspace/deployment.go#L373-L389) which takes in a `DevWorkspaceWithConfig`.

However, `getAdditionalAnnotations()` is actually more like `getAdditionalDeploymentAnnotations()` since its return value is used for the [deployment annotations](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/workspace/deployment.go#L181). You would have to rename `getAdditionalAnnotations()` -> `getAdditionalDeploymentAnnotations()`, and add append the DWOC annotations to podAdditions before [`mergePodAdditions()`](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/workspace/deployment.go#L141) is called, the latter of which makes `getSpecDeployment()` messier.

Moreso, the area in the devworkspace controller where you're appending the annotation additions makes sense: `// Step four: Collect all workspace deployment contributions `

So I think this makes sense :+1: Just wanted to think this over outloud.

However, it _might_ be worth adding an extra commit to rename `getAdditionalAnnotations()` -> `getAdditionalDeploymentAnnotations()`. Though this is entirely optional.

dkwon17: IMO it's ok as it is since all pod additions are collected before SyncDeploymentToCluster() in the main reconcile function https://github.com/devfile/devworkspace-operator/blob/69c21bd4ee855c9bb37db21e0eb32fcfbe36add3/controllers/workspace/devworkspace_controller.go#L471

I've added a new commit for the rename

AObuchow: Agreed +1

AObuchow (approved):
LGTM :)

Issue #1092, opened 2023-04-24
Use consistent capitalization in error messages 


AObuchow:
https://github.com/devfile/devworkspace-operator/pull/1064 refactored the way errors were propagated in the devworkspace-controller, and highlighted the fact that some error messages in DWO [start with a lower case letter](https://github.com/devfile/devworkspace-operator/pull/1064/files#diff-bfb432d51dee16f88d16a491fd7b8f24fbe2d56bc53d1036fee90e6fa27e0354R127) while [others start with a capital letter](https://github.com/devfile/devworkspace-operator/pull/1064/files#diff-bfb432d51dee16f88d16a491fd7b8f24fbe2d56bc53d1036fee90e6fa27e0354R88).

It would be nice to make the capitalization of error messages consistent, as they can be shown to the user in cases where the workspace startup fails.

Some things to consider:
- Do we want DW error & warning status messages to always start with a lower case letter, or upper case?
- Some error messages get [concatenated with another error message](https://github.com/devfile/devworkspace-operator/pull/1064/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R133) to provide some context, so it's not just a matter of changing all error messages to start with an upper or lower case letter.
  -   Maybe the functions set the workspace status messages (e.g. [this one for failing workspaces](https://github.com/devfile/devworkspace-operator/pull/1064/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R549)) should automatically capitalize the first letter of the workspace status message. This way, we could more easily enforce the standard of all error messages starting with a lowercase letter while ensuring user-facing error messages start with an uppercase letter.

Horiodino:
hey @AObuchow If no one else is currently working on this issue, I would like to take it on. Please let me know if it's available for me to handle.

AObuchow:
@Horiodino no one is currently working on this issue, I'd be happy to have you work on it :)

PR #1276, opened 2024-06-21
capitalizeMsg function to correctly capitalize the first character an…

Horiodino:
handle error msf scenarios based

https://github.com/devfile/devworkspace-operator/issues/1092

### PR Checklist

### DIFF
--- a/controllers/workspace/condition.go
+++ b/controllers/workspace/condition.go
@@ -16,6 +16,8 @@
 package controllers
 
 import (
+	"unicode"
+
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	corev1 "k8s.io/api/core/v1"
 
@@ -65,7 +67,7 @@ func (c *workspaceConditions) setConditionTrueWithReason(conditionType dw.DevWor
 
 	c.conditions[conditionType] = dw.DevWorkspaceCondition{
 		Status:  corev1.ConditionTrue,
-		Message: msg,
+		Message: capitalizeMessage(msg),
 		Reason:  reason,
 	}
 }
@@ -129,4 +131,14 @@ func getConditionIndexInOrder(condType dw.DevWorkspaceConditionType) int {
 		}
 	}
 	return -1
+}
+
+// Capitalizes the first character in the given string.
+func capitalizeMessage(msg string) string {
+	if msg == "" {
+		return msg
+	}
+	runes := []rune(msg)
+	runes[0] = unicode.ToUpper(runes[0])
+	return string(runes)
 }


Horiodino:
i was thinking in the dw structure we can add a context of type error if its for user we can set it as true and then at the end we can dw is for error handling and representing we can directly add function for that.
but for that we need to set context like if the error is for user then we need to set it as true and after that we can again set it to false ?

controllers/workspace/condition.go
```diff
@@ -130,3 +132,12 @@ func getConditionIndexInOrder(condType dw.DevWorkspaceConditionType) int {
 	}
 	return -1
 }
+
+func capitalizeMsg(msg string) string {
```
AObuchow, line 136: The go standard library actually has [cases](https://pkg.go.dev/golang.org/x/text/cases) for this use-case (handling unicode string capitalization).
Rather than re-invent the wheel, you could do something like:

```go
import (
	"golang.org/x/text/cases"
	"golang.org/x/text/language"
)
...

return cases.Title(language.English, cases.NoLower).String(msg)
```

AObuchow (commented):
Thank you for your PR @Horiodino :)
It seems you're missing testing instructions. Could you please add some?
Have you tried building DevWorkspace Operator with your changes and [deploying it](https://github.com/devfile/devworkspace-operator/tree/main?tab=readme-ov-file#deploying-devworkspace-operator) to a testing cluster (like minikube for instance?)

I took a look through the codebase to try and find a user-facing error message that is not already being capitalized, and is simple enough to trigger. What I found was that finalize function (which gets called when a workspace is being terminated/deleted off the cluster) will try and [check](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/finalize.go#L102-L108) the workspace's storageProvisioner (i.e. storage type). The potential error that could happen at this step is [not being capitalized](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/storage/provisioner.go#L30).

You should be able to trigger this (normally not-capitalized error) by:
1. Creating a workspace with a valid storage strategy attribute set, e.g.:

```YAML
kind: DevWorkspace
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```
2. Let the workspace startup successfully
3. Modify the workspace's storage-type attribute to an invalid value
4. Delete the workspace object off the cluster, i.e. `kubectl delete dw plain-devworkspace -n <your-workspace's-namespace>`
5. Check that the workspace status error message is capitalized by checking the devworkspace object or doing a `kubectl get dw -n <your-workspace's-namespace>`

AObuchow:
> i was thinking in the dw structure we can add a context of type error if its for user we can set it as true and then at the end we can dw is for error handling and representing we can directly add function for that. but for that we need to set context like if the error is for user then we need to set it as true and after that we can again set it to false ?

Anytime we're setting the workspace status, such as [here](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/devworkspace_controller.go#L585), it is already implied that it is a user-facing message. As the workspace status will be [provided in the workspace object on the cluster](https://github.com/devfile/devworkspace-operator/blob/main/controllers/workspace/status.go#L66). So there should be no need for this additional context IMO.

Horiodino:
sorry for making so long , the `make docker` was taking so much time (might be isp problem) .
on running 
```
kubectl patch dw plain-devworkspace --type=merge -p '{"spec":{"template":{"attributes":{"controller.devfile.io/storage-type": "invalid-type"}}}}'
```
```
kubectl get dw -w
NAME                 DEVWORKSPACE ID             PHASE     INFO
plain-devworkspace   workspace147061c666194f78   Running   Workspace is running
plain-devworkspace   workspace147061c666194f78   Running   Workspace is running
plain-devworkspace   workspace147061c666194f78   Failing   Error Provisioning Storage: Configured Storage Type Not Supported
plain-devworkspace   workspace147061c666194f78   Failing   Error Provisioning Storage: Configured Storage Type Not Supported
plain-devworkspace   workspace147061c666194f78   Failing   Error Provisioning Storage: Configured Storage Type Not Supported
plain-devworkspace   workspace147061c666194f78   Failing   Error Provisioning Storage: Configured Storage Type Not Supported
plain-devworkspace   workspace147061c666194f78   Failed    Error Provisioning Storage: Configured Storage Type Not Supported```

**Comment by AObuchow** [comment]
@Horiodino thank you for the follow-up. Could you please run a `go mod tidy` on your fork and commit the changes? This should fix the failing GH Action.

You might also need to do a `go install golang.org/x/tools/cmd/goimports@latest` followed by a `make fmt`

**Comment by Horiodino** [comment]
@AObuchow

**Comment by AObuchow** [comment]
@Horiodino sorry for the delay, I'm planning to continue reviewing your PR early this week.

**Review by AObuchow — CHANGES_REQUESTED** [review]
Sorry for the delay in following up on this @Horiodino.

I made a silly mistake in my [original review](https://github.com/devfile/devworkspace-operator/pull/1276/commits/e75892616cdb50fc3679543052cb5c71ca17f472#r1651703541): `cases.Title(language.English, cases.NoLower)` actually sets every word in a string to start with an upper case character, which is not what we want here.

You can remove commits https://github.com/devfile/devworkspace-operator/pull/1276/commits/f56a79cfd16957ca7204402d47ab622c84980d49 & https://github.com/devfile/devworkspace-operator/pull/1276/commits/c431b0ea38ee5ef12ed554ad5016adccfd12766f from your PR and just keep https://github.com/devfile/devworkspace-operator/pull/1276/commits/e75892616cdb50fc3679543052cb5c71ca17f472. I'll add some extra comments on that commit. 

Also for [testing](https://github.com/devfile/devworkspace-operator/pull/1276#issuecomment-2210852153), you should be running 
```
kubectl patch dw plain-devworkspace -n $NAMESPACE --type=merge -p '{"spec":{"template":{"attributes":{"controller.devfile.io/storage-type": "invalid-type"}}}}' && kubectl delete dw plain-devworkspace -n $NAMESPACE
```
as the workspace needs to be deleted immediately after setting an invalid storage type.

The output of `kubectl get dw -n $NAMESPACE -w` should include something like:
```
plain-devworkspace   workspace3a2024c23f13435f   Error         Configured Storage Type Not Supported
```

File: controllers/workspace/condition.go [hunk 2]
```diff
@@ -130,3 +132,12 @@ func getConditionIndexInOrder(condType dw.DevWorkspaceConditionType) int {
...
+	if len(msg) == 0 {
```
AObuchow: We usually check for empty strings directly, rather than checking the length of the string. So this can be rewritten as:
```go
if msg == "" {
  return msg
}
```

controllers/workspace/condition.go
AObuchow: I think this function name could be unabbreviated as `capitalizeMessage(msg string)`.

Also, it's worth adding a simple function doc, something like:
`// Capitalizes the first character in the given string.`

Horiodino:
working as expected 
```plain-devworkspace   workspaced049889b5c554e91   Error         Configured storage type not supported```

**Review by AObuchow — APPROVED** [review]
LGTM. @dkwon17 any thoughts? I'll wait for you to give this a look before we merge.

**Review by dkwon17 — APPROVED** [review]
Sorry, just getting to this now,

LGTM, let's have it for DWO 0.31.x

**Comment by AObuchow** [comment]
@Horiodino This looks just about good to merge finally. @Horiodino could you please squashing both of your commits and perhaps rewording the commit message to "feat: ensure all DevWorkspace condition messages are capitalized" (or something similar)?

Thank you :)

**Comment by AObuchow** [comment]
@Horiodino Sorry for being so picky, but your commit now has the author sign-off twice in the description, and it's also lacking a reference to the bug it's resolving. Could you please modify your commit message to:

```
feat: ensure all DevWorkspace condition messages are capitalized

Fix #1092 

Signed-off-by: Horiodino <holiodin@gmail.com>
```

Horiodino:
> @Horiodino Sorry for being so picky, but your commit now has the author sign-off twice in the description, and it's also lacking a reference to the bug it's resolving. Could you please modify your commit message to:
> 
> ```
> feat: ensure all DevWorkspace condition messages are capitalized
> 
> Fix #1092 
> 
> Signed-off-by: Horiodino <holiodin@gmail.com>
> ```

np , done :+1:

AObuchow:
@Horiodino Thank you! Merging, great work :)

Issue #1286, opened 2024-07-09
Allow setting runtimeClass for devworkspace pods


dkwon17:
### Description

A new field in the DWOC to set runtime class for all devworkspace pods:
```
config:
  workspace:
    runtimeClassName: 'myruntimeclass'
```

If this is set in the DWOC, devworkspaces that use the DWOC should have their pod use the `myruntimeclass` runtime class
### Additional context

Related issue: https://github.com/eclipse-che/che/issues/23032

dkwon17:
Fixed by https://github.com/devfile/devworkspace-operator/pull/1303

PR #1303, opened 2024-08-12
Add new field for DWOC for DevWorkspace pod's runtimeClass

dkwon17:
### What does this PR do?
Introduces a new field for the DWOC for the DevWorkspace pod's runtimeClass

https://github.com/devfile/devworkspace-operator/issues/1286

### Is it tested? How?

To test this PR,

1. Create the plain-devworkspace DevWorkspace on the cluster
```
kind: DevWorkspace
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

2. Verify that the workspace pod starts and runs properly.

3. Create this DWOC:
```
config:
  workspace:
    runtimeClassName: test
metadata:
```

4. Restart the plain-devworkspace DevWorkspace.
The DevWorkspace startup should fail with:
```
'Error creating DevWorkspace deployment: Detected unrecoverable deployment condition: FailedCreate pods "workspace9b9d2d153f3747de-55bb88dd5b-" is forbidden: pod rejected: RuntimeClass "test" not found'
```

5. To verify that the runtimeClassName attribute takes precedence over the DWOC-defined runtimeClassName, create the following DevWorkspace:
```
kind: DevWorkspace
metadata:
  name: plain-devworkspace-2
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
      controller.devfile.io/runtime-class: dw-class
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

6. The DevWorkspace startup should fail with:
```
'Error creating DevWorkspace deployment: Detected unrecoverable deployment condition: FailedCreate pods "workspaceb61b62899c1549aa-55bcc6dbfc-" is forbidden: pod rejected: RuntimeClass "dw-class" not found'
```

### PR Checklist


controllers/workspace/util_test.go
```diff
@@ -264,6 +269,20 @@ func generateConfigMap(name string) *corev1.ConfigMap {
 	}
 }
 
+func generateDwoc(name string) *controllerv1alpha1.DevWorkspaceOperatorConfig {
```
AObuchow, line 272: This function takes the `name` but doesn't actually use it. Also, maybe we should capitalize the abbreviation `Dwoc` -> `DWOC`?

controllers/workspace/util_test.go
```diff
@@ -264,6 +269,20 @@ func generateConfigMap(name string) *corev1.ConfigMap {
...
+	return &controllerv1alpha1.DevWorkspaceOperatorConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "devworkspace-operator-config",
+			Namespace: "devworkspace-test",
```
AObuchow, line 276: You can use the `testNamespace` constant instead since it's already defined

controllers/workspace/testdata/test-devworkspace-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  labels:
+    controller.devfile.io/creator: ""
+spec:
+  started: true
+  routingClass: 'basic'
+  template:
+    attributes:
+      controller.devfile.io/devworkspace-config:
+        name: devworkspace-operator-config
```
AObuchow, line 12: Are you specifically trying to make sure this works with an external DWOC instead of the global DWOC? If so, it might be worthwhile to use a different name for the DWOC, since this is the name of the global DWOC.

controllers/workspace/devworkspace_controller_test.go
```diff
@@ -602,6 +602,85 @@ var _ = Describe("DevWorkspace Controller", func() {
...
+				},
+			}
+
+			createObject(dwoc)
+			defer deleteObject(dwoc)
+
+			createDevWorkspace(devWorkspaceName, "test-devworkspace-config.yaml")
+			devworkspace := getExistingDevWorkspace(devWorkspaceName)
+			devworkspace.Spec.Template.Attributes.PutString(constants.RuntimeClassNameAttribute, "test-runtime-class-attribute")
+			updateDevWorkspace(devworkspace)
```
AObuchow, line 669: I suggest using an `Eventually()` here so that we can set a timeout for the updateDevWorkspace function, similar to what we are [doing](https://github.com/devfile/devworkspace-operator/blob/8fc183f55f20e6854f220a2c4bb844190366f67a/controllers/workspace/devworkspace_controller_test.go#L634) elsewhere.

Additionally, since `updateDevWorkspace()` is not called anywhere else, you could just write it out inline rather than have a function for it. Though, having a function that you call after modifying the devworkspace seems to make sense :)

controllers/workspace/util_test.go
```diff
@@ -58,6 +58,11 @@ func createDevWorkspace(name, fromFile string) {
 	}, 10*time.Second, 250*time.Millisecond).Should(BeTrue())
 }
 
+func updateDevWorkspace(devWorkspace *dw.DevWorkspace) {
```
AObuchow, line 61: If you keep this function and update it to use a timeout, you might also want to consider taking in an error message string that will be given if the update to the devworkspace fails.

controllers/workspace/testdata/test-devworkspace-config.yaml
AObuchow, line 1: The name of this file is a bit surprising -- I would expect it to actually contain a DWOC, not a devworkspace.
If the intention of this DevWorkspace is to specifically use a DWOC, maybe we should call it something like `test-devworkspace-using-config`?

Though, shouldn't devworkspaces use the global config anyhow?  We configure the global DWOC for testing [here](https://github.com/devfile/devworkspace-operator/blob/8fc183f55f20e6854f220a2c4bb844190366f67a/controllers/workspace/devworkspace_controller_test.go#L745-L750).

If you're specifically trying to test using the external DWOC, it might be more appropriate to name this file something like `test-devworkspace-external-config`.

AObuchow (commented):
Looks great to me so far, glad to see you're adding DevWorkspace Controller tests :grin: Left some minor comments.

controllers/workspace/util_test.go
dkwon17: I ended up removing this function since there is the `config.SetGlobalConfigForTesting` function

controllers/workspace/testdata/test-devworkspace-config.yaml
dkwon17: I wanted to use the global DWOC, I removed this .yaml in my latest commit

controllers/workspace/devworkspace_controller_test.go
dkwon17: Updated, and removed `updateDevWorkspace()`

controllers/workspace/testdata/test-devworkspace-config.yaml
dkwon17: My latest commit removes this file

dkwon17:
Thank you @AObuchow , I've updated this PR

controllers/workspace/devworkspace_controller_test.go
```diff
@@ -653,20 +650,23 @@ var _ = Describe("DevWorkspace Controller", func() {
...
-			defer deleteObject(dwoc)
+			devworkspace := &dw.DevWorkspace{}
+			createDevWorkspace(devWorkspaceName, "test-devworkspace.yaml")
+
+			By("Adding runtime-class annotation to the DevWorkspace")
+			Eventually(func() error {
+				devworkspace = getExistingDevWorkspace(devWorkspaceName)
+				devworkspace.Spec.Template.Attributes.PutString(constants.RuntimeClassNameAttribute, "test-runtime-class-attribute")
+				return k8sClient.Update(ctx, devworkspace)
+			}, timeout, interval).Should(Succeed(), "DevWorkspace should get `runtime-class: test-runtime-class-attribute` annotation")
```
AObuchow, line 668: Since we're adding an attribute, should we make the error message say: "DevWorkspace should get `runtime-class: test-runtime-class-attribute` attribute" (instead of annotation)?

controllers/workspace/devworkspace_controller_test.go
```diff
@@ -653,20 +650,23 @@ var _ = Describe("DevWorkspace Controller", func() {
...
-			}
+			})
+			defer config.SetGlobalConfigForTesting(nil)
 
-			createObject(dwoc)
-			defer deleteObject(dwoc)
+			devworkspace := &dw.DevWorkspace{}
+			createDevWorkspace(devWorkspaceName, "test-devworkspace.yaml")
+
+			By("Adding runtime-class annotation to the DevWorkspace")
```
AObuchow, line 663: Here as well, we should probably use "attribute" instead of "annotation"

AObuchow (commented):
I still have to do a bit of manual testing, but this looks great to me :) I left some very minor comments.

AObuchow:
@musienko-maxim Any idea what's going wrong with the CI prow tests here?

AObuchow (approved):
Tested on OpenShift 4.16 and works as expected, great work @dkwon17 :grin: :pray: 
My last suggestion is to squash your feedback commits, and maybe rename the `"Add tests"` commit to something a bit more specific like `"Adds controller environment tests for pod runtimeClass"`

dkwon17:
@AObuchow thanks, I've squashed the commits

AObuchow:
@dkwon17 Awesome, this looks good to merge :) Once all the CI checks pass, feel free to merge :)

Issue #1295, opened 2024-08-06
DevWorkspace doesn't respect SSH key passphrase when git pushing changes from workspace


dmytro-ndp:
### Description

Related changes:
- https://github.com/devfile/devworkspace-operator/pull/1291, 
- https://github.com/eclipse-che/che-dashboard/pull/1157

#### Prerequisites: 
a) create SSH key with passphrase, using the command `ssh-keygen -b 2048 -t rsa -N "test"`
b) add it to the GitHub.com

#### Steps to reproduce:
1. Install Eclipse Che using User Dashboard https://quay.io/repository/eclipse/che-dashboard:pr-1157 with support of SSH key with passphrase
2. Open User Dashboard and add SSH key to the User Preferences.
3. Go to GitHub.com and get SSH link to any private repo using.
4. Go to "Eclipse Che -> Create workspace" page and create workspace using SSH link to the private repo.
5. Wait on workspace to be created, started, and project to be cloned.
6. Make changes in the project, commit them and push.
7. Go to workspace Terminal and try to push changes from there.

[Let Dev Spaces keep ssh passphrase for git.webm](https://github.com/user-attachments/assets/283bd203-e9b1-405f-abcb-673041559a8a)

### Expected behavior
Changes are pushed without errors or asking for passphrase.

### Actual behavior
[screencast-bpconcjcammlapcogcnnelfmaeghhagj-2024.08.06-14_39_32.webm](https://github.com/user-attachments/assets/3ad32dc3-25c2-49a4-bc3a-84b7d1da3ac3)

#### At the step 6
there was an error `Git: Warning: Permanently added the ECDSA host key for IP address '140.82.112.3' to the list of known hosts.` when tried to push changes to the repo using _Source Control_ view of _VS Code Editor_.
**_git-error log_**:
```
> git pull --tags origin main
Warning: Permanently added the ECDSA host key for IP address '140.82.112.3' to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
![Screenshot from 2024-08-06 14-42-49](https://github.com/user-attachments/assets/63400a06-0172-48a5-8676-77da128a0270)

#### At the step 7
there was a request to enter passphrase for key '/etc/ssh/dwo_ssh_key":
[...]

PR #1307, opened 2024-08-14
Apply ssh askpass flow for the workspace container

vinokurig:
### What does this PR do?
* Add `DISPLAY` and ` SSH_ASKPASS` environment variables to the workspace container for the ssh askpass flow.
* Add ssh askpass provision step to create a configmap with the ssh askpass script content and propagate it to the workspace container.
* Change the root of the `ssh-askpass.sh` file in the project clone Dockerfile to the one in the `provision/automount` package 

fixes https://github.com/devfile/devworkspace-operator/issues/1295

### Is it tested? How?

1. Start a workspace and check that the `DISPLAY` and ` SSH_ASKPASS` environment variables are set
2. Check that the `/.ssh-askpass/ssh-askpass.sh` file is present and have the right content.
### PR Checklist


pkg/library/env/workspaceenv.go
AObuchow, line 1: Since the project clone container [also gets environment variables](https://github.com/devfile/devworkspace-operator/blob/main/pkg/library/env/workspaceenv.go#L53) from the `commonEnvironmentVariables()` function, we no longer have to setup the SSH askpass related environment variables in the project clone container's [Dockerfile](https://github.com/devfile/devworkspace-operator/blob/main/project-clone/Dockerfile#L48-L49).

AObuchow (commented):
Looks awesome to me so far @vinokurig, great work :grin: 

I left some small suggestions as to how we inject the SSH ask pass script into the workspace deployment, as well as removing some functionality from https://github.com/devfile/devworkspace-operator/issues/1295 that would be rendered redundant from this PR.

pkg/library/env/workspaceenv.go
```diff
@@ -80,6 +81,13 @@ func commonEnvironmentVariables(workspaceWithConfig *common.DevWorkspaceWithConf
 			Name:  constants.DevWorkspaceIdleTimeout,
 			Value: workspaceWithConfig.Config.Workspace.IdleTimeout,
 		},
+		{
+			Name:  constants.DISPLAY,
+			Value: ":0",
+		}, {
+			Name:  constants.SSHAskPass,
+			Value: automount.SshAskPassMountPath + automount.SshAskPassScriptFileName,
```
AObuchow, line 89: In the rest of DWO's codebase, we tend to use `fmt.Sprintf()` instead of using `+` to concatenate strings. In my opinion, it makes it more clear that we are concatenating two strings and not summing two numbers.

Admittedly, `fmt.Sprintf("%s%s", automount.SshAskPassMountPath, automount.SshAskPassScriptFileName)` does look a bit weird however, since we're just concatenating two strings with no extra formatting.

AObuchow:
@vinokurig Also, please run `make fmt` as the format validation CI check is currently failing

vinokurig:
@AObuchow Thank you for the review, I've added another commit to add a post start event to initialise `ssh-agent` and add the workspace ssh-key. Please take a look.

pkg/library/env/workspaceenv.go
```diff
@@ -80,6 +81,13 @@ func commonEnvironmentVariables(workspaceWithConfig *common.DevWorkspaceWithConf
...
+		},
 	}
 
 	envvars = append(envvars, GetProxyEnvVars(workspaceWithConfig.Config.Routing.ProxyConfig)...)
```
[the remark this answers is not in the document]
vinokurig: done

pkg/provision/automount/common.go
```diff
@@ -109,7 +109,12 @@ func getAutomountResources(api sync.ClusterAPI, namespace string) (*Resources, e
 		return nil, err
 	}
 
-	return mergeAutomountResources(gitCMAutoMountResources, mergedResources, pvcAutoMountResources), nil
+	sshCMAutoMountResources, err := ProvisionSshAskPassScript(api, namespace)
```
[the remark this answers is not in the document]
vinokurig: done

pkg/provision/automount/common_persistenthome_test.go
```diff
@@ -58,7 +58,7 @@ func TestProvisionAutomountResourcesIntoPersistentHomeEnabled(t *testing.T) {
 
 			err := ProvisionAutoMountResourcesInto(podAdditions, testAPI, testNamespace, true)
 
-			if !assert.NoError(t, err, "Unexpected error") {
+			if assert.ErrorContains(t, err, "Created object") || !assert.NoError(t, err, "Unexpected error") {
```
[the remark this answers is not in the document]
vinokurig: Reworked to intercept the error when initialising the configmap

pkg/provision/automount/ssh.go
```diff
@@ -0,0 +1,61 @@
...
+//go:embed ssh-askpass.sh
+var data string
+
+func ProvisionSshAskPassScript(api sync.ClusterAPI, namespace string) (*Resources, error) {
+	sshAskPassConfigMap := constructSshAskPassCM(namespace)
+	if _, err := sync.SyncObjectWithCluster(sshAskPassConfigMap, api); err != nil {
+		return nil, dwerrors.WrapSyncError(err)
+	}
+	resources := flattenAutomountResources([]Resources{
+		getAutomountConfigmap(SshAskPassMountPath, constants.DevWorkspaceMountAsSubpath, pointer.Int32(0755), sshAskPassConfigMap),
```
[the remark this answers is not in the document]
vinokurig: reworked to initialise the resources without using the `automount` package

pkg/provision/automount/ssh.go
```diff
@@ -0,0 +1,61 @@
...
+	_ "embed"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
+	"github.com/devfile/devworkspace-operator/pkg/dwerrors"
+	"github.com/devfile/devworkspace-operator/pkg/provision/sync"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+const SshAskPassMountPath = "/.ssh-askpass/"
```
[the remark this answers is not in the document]
vinokurig: It does not matter where to keep the script file but I changed it to be inline with https://github.com/devfile/devworkspace-operator/blob/8fc183f55f20e6854f220a2c4bb844190366f67a/pkg/provision/automount/gitconfig.go#L26

project-clone/Dockerfile
```diff
@@ -49,7 +49,7 @@ ENV USER_UID=1001 \
     SSH_ASKPASS=/usr/local/bin/ssh-askpass.sh
 
 COPY build/bin /usr/local/bin
-COPY project-clone/ssh-askpass.sh /usr/local/bin
+COPY pkg/provision/automount/ssh-askpass.sh /usr/local/bin
```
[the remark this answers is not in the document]
vinokurig: done

pkg/library/env/workspaceenv.go
vinokurig: done

pkg/library/env/workspaceenv.go
vinokurig: done

pkg/library/ssh/event.go
```diff
@@ -0,0 +1,56 @@
...
+func AddSshAgentPostStartEvent(spec *v1alpha2.DevWorkspaceTemplateSpec) error {
+	if spec.Commands == nil {
+		spec.Commands = []v1alpha2.Command{}
+	}
+
+	if spec.Events == nil {
+		spec.Events = &v1alpha2.Events{}
+	}
+
+	var commandLine = `SSH_ENV_PATH=/home/user/ssh-environment \
```
AObuchow, line 33: A few comments:
- The `commandLine` variable could probably be a constant to make the rest of the code a little easier to read (like we did [here](https://github.com/devfile/devworkspace-operator/pull/1251/files#diff-f1f3498fd50bbc235d4d09f5f5485a91f5a4a5cfeb20b127c96fc849a30ba56bR30-R46))
- It might be better to use `$HOME` instead of hard-coding `/home/user/` since the users workspace might not be using the UDI. I'm not sure how common the `ssh-agent` binary is, but hopefully your implementation here could also work for other workspace images.
- This is very subjective: I find using `source` is more readable than `.`. This is relevant IMO because we are modifying the users `.bashrc`.


The only non-straightforward thing here is the  `ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH`. My understanding is that there's some echo commands as part of ssh-agent's output that we don't want to be executed everytime the `.bashrc` is sourced?

AObuchow (commented):
Thanks for addressing my earlier comments @vinokurig, this is looking great :)
I left some minor comments but I think this is about ready for some testing on my end.

pkg/library/ssh/event.go
vinokurig: >* The commandLine variable could probably be a constant to make the rest of the code a little easier to read (like we did [here](https://github.com/devfile/devworkspace-operator/pull/1251/files#diff-f1f3498fd50bbc235d4d09f5f5485a91f5a4a5cfeb20b127c96fc849a30ba56bR30-R46))
>* It might be better to use $HOME instead of hard-coding /home/user/ since the users workspace might not be using the UDI. I'm not sure how common the ssh-agent binary is, but hopefully your implementation here could also work for other workspace images.
>* This is very subjective: I find using source is more readable than .. This is relevant IMO because we are modifying the users .bashrc.

done
> The only non-straightforward thing here is the ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH. My understanding is that there's some echo commands as part of ssh-agent's output that we don't want to be executed everytime the .bashrc is sourced?

Right, we comment the `echo Agent pid ****;` line of the `ssh-agent` command output, so we have a clear terminal start. As an alternative we can add `>/dev/null` to the `source ${SSH_ENV_PATH}` bachrc command. WDYT?

AObuchow: I actually like your current solution of not echoing the ` Agent pid ****`  over doing a `>/dev/null` since this would probably let any errors related to sourcing the ssh-agent environment variables get logged in the user's terminal and let them know something is not working correctly.

AObuchow: Also @vinokurig it seems we're still referencing `/home/user/` in two liens of the postStart event: [here](https://github.com/devfile/devworkspace-operator/pull/1307/files#diff-12b01779c7cf67e18ef8d7bf46eb788bb6b640db0fa051804f2ef7b31a50d4d5R22) and [here](https://github.com/devfile/devworkspace-operator/pull/1307/files#diff-12b01779c7cf67e18ef8d7bf46eb788bb6b640db0fa051804f2ef7b31a50d4d5R27). Could you please change these to `$HOME`?

AObuchow (commented):
Posted some more feedback

vinokurig:
@AObuchow 
> it seems we're still referencing /home/user/ in two lines of the postStart event: [here](https://github.com/devfile/devworkspace-operator/pull/1307/files#diff-12b01779c7cf67e18ef8d7bf46eb788bb6b640db0fa051804f2ef7b31a50d4d5R22) and [here](https://github.com/devfile/devworkspace-operator/pull/1307/files#diff-12b01779c7cf67e18ef8d7bf46eb788bb6b640db0fa051804f2ef7b31a50d4d5R27). Could you please change these to $HOME or ~?

fixed

AObuchow (commented):
I re-built the project-clone image with the changes from this PR and verified this current PR resolves https://github.com/devfile/devworkspace-operator/issues/1295 without breaking the functionality of https://github.com/devfile/devworkspace-operator/pull/1291.

I still need to do one last pass-through of this PR, and @dkwon17 should take a look as well, but IMO this PR should be good to merge very soon (ideally, before the end of this week).

Great work @vinokurig :grin: And sorry for taking so long with my review.

dkwon17:
LGTMl, I just have one comment

pkg/library/ssh/event.go
```diff
@@ -0,0 +1,57 @@
...
+
+package ssh
+
+import (
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/pkg/constants"
+	"github.com/devfile/devworkspace-operator/pkg/library/lifecycle"
+)
+
+const commandLine = `SSH_ENV_PATH=$HOME/ssh-environment \
```
AObuchow, line 22: If we keep the current approach of modifying the .bashrc, it's probably worth making the ssh-environment file hidden with a `.` i.e. `SSH_ENV_PATH=$HOME/.ssh-environment`

pkg/library/ssh/event.go
```diff
@@ -0,0 +1,57 @@
...
+&& if [ -f /etc/ssh/passphrase ] && command -v ssh-add >/dev/null; \
+then ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
+&& chmod 600 $SSH_ENV_PATH && source $SSH_ENV_PATH \
+&& ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase \
+&& if [ -f $HOME/.bashrc ]; then echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc; fi; fi`
```
AObuchow, line 27: Modifying the `.bashrc` works, but if a user provides their own custom `.bashrc` through an [automount configmap](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#automatically-mounting-volumes-configmaps-and-secrets), this feature will actually break. Files that are mounted via the automount configmap mechanism cannot be written to.

Instead, we could place the SSH_ENV_PATH in `/etc/profile.d/`, which contains scripts that should be run when sourcing the `.bashrc`, e.g. `SSH_ENV_PATH=/etc/profile.d/.ssh-environment`. This way, any `.bashrc` will load the ssh-agent script.

```bash
$ cat ~/.bashrc
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
    . /etc/bashrc
fi
(...)


$ cat /etc/bashrc 
(...)
    SHELL=/bin/bash
    # Only display echos from profile.d scripts if we are no login shell
    # and interactive - otherwise just process them to set envvars
    for i in /etc/profile.d/*.sh; do
        if [ -r "$i" ]; then
            if [ "$PS1" ]; then
                . "$i" # Here is where all the scripts in /etc/profile.d/ get sourced
            else
                . "$i" >/dev/null
            fi
        fi
    done
```

A caveat is that the /etc/profile.d/ scripts only get run for interactive shells. But for https://github.com/devfile/devworkspace-operator/issues/1295 this is fine.

**I do understand that this is a rather big change to request from your PR @vinokurig, so I'm perfectly fine with keeping the current approach that modifies the .bashrc and making a new issue/PR in the future to implement my suggestions.** I haven't actually tested my suggestion, so it might not actually work (though I did [something similar](https://github.com/devfile/developer-images/pull/115/commits/7ac79710c16454500116b51c37e3bdffc30e66de) for the UDI a while back)

pkg/provision/workspace/ssh.go
```diff
@@ -0,0 +1,94 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package workspace
+
+import (
+	_ "embed"
```
AObuchow, line 19: Is there a reason why the blank identifier `_` is being used? Couldn't we just have 
```go
import (
	"embed"
...
)
```

I've never used the embed directive, but [examples of it](https://gobyexample.com/embed-directive) don't seem to be using the blank identifier.

pkg/provision/workspace/ssh.go
```diff
@@ -0,0 +1,94 @@
...
+	"k8s.io/utils/pointer"
+)
+
+const SshAskPassMountPath = "/.ssh-askpass/"
+const SshAskPassScriptFileName = "ssh-askpass.sh"
+
+//go:embed ssh-askpass.sh
+var data string
+
+func ProvisionSshSshAskPass(api sync.ClusterAPI, namespace string, podAdditions *v1alpha1.PodAdditions) error {
```
AObuchow, line 37: I believe there's a typo in this function name: `ProvisionSshSshAskPass` -> `ProvisionSshAskPass`

AObuchow (changes requested):
Some last minute remarks. I believe the typo and blank identifier are the only remarks that needs to be addressed before we merge this PR. Everything else could be left for a future PR.

pkg/library/ssh/event.go
dkwon17: > I do understand that this is a rather big change to request from your PR @vinokurig, so I'm perfectly fine with keeping the current approach that modifies the .bashrc and making a new issue/PR in the future to implement my suggestions. 

I think we should take the current approach for now since there isn't a straight forward alternative since the `/etc/profile.d` folder cannot be written by regular users by default in the UDI. I suggest we keep the functionality how it is for now, and document that if .bashrc is being mounted using a configmap, something like `[ -f /home/user/ssh-environment ] && source /home/user/ssh-environment` should be added to it.

Any thoughts?

dkwon17, line 27:
```suggestion
&& if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc; fi; fi`
```
Since @AObuchow mentioned that .bashrc might not be writable

dkwon17 (changes requested):

pkg/library/ssh/event.go
AObuchow: Good catch :) I agree that it's probably worth documenting this somewhere in the near future (maybe in the Che Docs as well as DWO additional docs)?

pkg/provision/workspace/ssh.go
vinokurig: Importing the `embed` explicitly causes an error: `pkg/provision/workspace/ssh.go:19:2: "embed" imported and not used`

pkg/provision/workspace/ssh.go
vinokurig: done

dkwon17 (approved):

AObuchow (approved):
LGTM, awesome work @vinokurig :grin: 
Please squash all your commits into a single commit (I think you could squash all your commits into your [first commit](https://github.com/devfile/devworkspace-operator/pull/1307/commits/405ec5373495cb098c56b143830aafc009be844f))

You could also add "Fixes #1295" to the commit description.

vinokurig:
@AObuchow 
>LGTM, awesome work @vinokurig 😁
Please squash all your commits into a single commit (I think you could squash all your commits into your [first commit](https://github.com/devfile/devworkspace-operator/pull/1307/commits/405ec5373495cb098c56b143830aafc009be844f))
>
>You could also add "Fixes https://github.com/devfile/devworkspace-operator/issues/1295" to the commit description.

thanks, done

AObuchow (approved):

PR #1306, opened 2024-08-14
feat: remove FailedScheduling event from list of unrecoverable worksp…

mkuznyetsov:
…ace pod events

### What issues does this PR fix or reference?
https://github.com/devfile/devworkspace-operator/issues/1280

### Is it tested? How?

create a workspace with exceeding resource requests/limits (modified samples/plain.yaml):
```
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          image: quay.io/wto/web-terminal-tooling:next
          memoryRequest: 1000Gi
          memoryLimit: 1000Gi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

check the workspace status, which will keep trying to start workspace, until it times out in 5 minutes:

```
[16:41:05] mykhailo@~/projects/devworkspace-operator$ kubectl get dw
plain-devworkspace   workspaceb6a633cbe27141fe   Failed   DevWorkspace failed to progress past step 'Waiting for workspace deployment' for longer than timeout (5m)
```

### PR Checklist

- [ ] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [ ] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [ ] `v8-che-happy-path`: Happy path for verification integration with Che

Issue #1280, opened 2024-06-26
Remove FailedScheduling event from list of unrecoverable workspace pod events


AObuchow:
There are many cases where causing the FailedScheduling event to result in workspace failure is problematic. For example, flaky cluster infrastructure can require multiple attempts to schedule a workspace pod on a cluster. Additionally, the [cluster auto-scaler](https://kubernetes.io/docs/concepts/cluster-administration/cluster-autoscaling/) can only kick in when a pod remains in the unschedulable state -- if we delete the deployment immediately after a pod is deteremiend to be unschedulable, the auto-scaler cannot kick in.

Thus we should remove the FailedScheduling event from the [list](https://github.com/devfile/devworkspace-operator/blob/1360f21b6ffd6a6a947301e1ab757d3891e219f0/pkg/library/status/check.go#L47) of unrecoverable workspace pod events. https://github.com/devfile/devworkspace-operator/issues/1279 is required to allow users the ability to re-add the FailedScheduling event to the list of unrecoverable workspace pod events.

AObuchow:
After further consideration, removing the FailedScheduling event from the hard coded list of unrecoverable workspace pod events might not be the best approach.

If the FailedScheduling event is ignored, it'll be [shown on workspace timeout](https://github.com/devfile/devworkspace-operator/pull/1051):
```
NAME                  DEVWORKSPACE ID             PHASE      INFO
theia-next-high-cpu   workspace656dfe6d86764967   Failed     devworkspace failed to progress past phase 'Starting' for longer than timeout (1m). Reason: Detected unrecoverable event FailedScheduling: 0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.. 
```

However, if we remove the FailedScheduling event from the hard coded list of unrecoverable pod events, then it will **not be shown on workspace timeout.**

Our goal is to have the FailedScheduling event not cause workspace timeouts by default. This would make it easier to use cluster autoscaling in Che, and prevent workspaces from failing immediately if there are transient cluster issues. 

However, we still would like the FailedScheduling event to be able to be caught (hence https://github.com/devfile/devworkspace-operator/issues/1279), and to let users know if their workspace timed out due to the FailedScheduling event.

[...]

PR #1310, opened 2024-08-22
feat: set defaults for ignoredUnrecoverableEvents operator config

mkuznyetsov:
### What does this PR do?
Add FailedScheduling event to the default list of ignoredUnrecoverableEvents list in operator config.

(this PR is an alternative to https://github.com/devfile/devworkspace-operator/pull/1306)

the relevant docs should also be updated:
https://eclipse.dev/che/docs/stable/administration-guide/configuring-machine-autoscaling/#_when_the_autoscaler_adds_a_new_node

https://github.com/devfile/devworkspace-operator/issues/1280
### Is it tested? How?

create a workspace with exceeding resource requests/limits (modified samples/plain.yaml):
```
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryRequest: 1000Gi
          memoryLimit: 1000Gi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

check the workspace status, which will keep trying to start workspace, until it times out in 5 minutes:
```
$ kdw get dw
plain-devworkspace   workspace8e15dba59ab04607   Failed   DevWorkspace failed to progress past step 'Waiting for workspace deployment' for longer than timeout (5m). Ignored events: Detected unrecoverable event FailedScheduling: 0/1 nodes are available: 1 Insufficient memory. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod...
```
### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -118,10 +118,11 @@ type WorkspaceConfig struct {
 	// "15m", "20s", "1h30m", etc. If not specified, the default value of "5m" is used.
 	ProgressTimeout string `json:"progressTimeout,omitempty"`
 	// IgnoredUnrecoverableEvents defines a list of Kubernetes event names that should
-	// be ignored when deciding to fail a DevWorkspace startup. This option should be used
-	// if a transient cluster issue is triggering false-positives (for example, if
-	// the cluster occasionally encounters FailedScheduling events). Events listed
-	// here will not trigger DevWorkspace failures.
+	// be ignored when deciding to fail a DevWorkspace startup.
+	// For example, a FailedScheduling event, that occurs when workspace cannot start
+	// due to exceeding available resources, should not fail the workspace startup, if there is
+	// an autoscaler configured on the cluster, and we want to wait until it provisions additional resources.
+	// FailedScheduling event can also occur as a false-positive, as a result of a transient cluster issue.
 	IgnoredUnrecoverableEvents []string `json:"ignoredUnrecoverableEvents,omitempty"`
 	// CleanupOnStop governs how the Operator handles stopped DevWorkspaces. If set to
 	// true, additional resources associated with a DevWorkspace (e.g. services, deployments,
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -76,6 +76,7 @@ var defaultConfig = &v1alpha1.OperatorConfiguration{
 				corev1.ResourceMemory: resource.MustParse("64Mi"),
 			},
 		},
+		IgnoredUnrecoverableEvents: []string {"FailedScheduling"},
 	},
 }
 


apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -118,10 +118,11 @@ type WorkspaceConfig struct {
 	// "15m", "20s", "1h30m", etc. If not specified, the default value of "5m" is used.
 	ProgressTimeout string `json:"progressTimeout,omitempty"`
 	// IgnoredUnrecoverableEvents defines a list of Kubernetes event names that should
-	// be ignored when deciding to fail a DevWorkspace startup. This option should be used
-	// if a transient cluster issue is triggering false-positives (for example, if
-	// the cluster occasionally encounters FailedScheduling events). Events listed
-	// here will not trigger DevWorkspace failures.
+	// be ignored when deciding to fail a DevWorkspace startup.
```
AObuchow, line 121: I'm not entirely sure we need to mention the cluster auto-scaler in DWO (or rewrite the docs here). It might be better to mention this in the Che Cluster CRD documentation, since the ignoredUnrecoverableEvents [can be configured](https://github.com/eclipse-che/che-operator/pull/1864) from the Che Cluster CRD. 

Instead, I would suggest:
- Mentioning "By default, the FailedScheduling is ignored"
- Removing the `"(for example, if the cluster occasionally encounters FailedScheduling events)"` since this example is no longer valid now that the FailedScheduling event is ignored by default

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -118,10 +118,11 @@ type WorkspaceConfig struct {
...
+	// For example, a FailedScheduling event, that occurs when workspace cannot start
+	// due to exceeding available resources, should not fail the workspace startup, if there is
+	// an autoscaler configured on the cluster, and we want to wait until it provisions additional resources.
+	// FailedScheduling event can also occur as a false-positive, as a result of a transient cluster issue.
```
AObuchow, line 125: I suggest experimenting with [kubebuilder annotations](https://book.kubebuilder.io/reference/markers/crd-validation) for the IgnoredUnrecoverableEvents field. 

We should try setting the default array value. I think this would be done with `+kubebuilder:default:={"FailedScheduling"}`

I believe that should be enough to populate the IgnoredUnrecoverableEvents list in the DWOC. Make sure you re-generate the CRD's in a seperate commit by running: `make update_devworkspace_api update_devworkspace_crds generate_all`

Something to note: This entire PR might be dropped and re-implemented in Che-Operator if we can get the kubebuilder approach working. We'd want Che admins to _see_ that the FailedSchedling event is ignored by default & there would be no advantages to duplicating this code change in both DWO & Che-Operator (unless users who use DWO in isolation want this feature, however, this is not the current reason why we're resolving https://github.com/devfile/devworkspace-operator/issues/1280).

AObuchow (commented):
Thanks for the PR @mkuznyetsov :) 
Please run `make fmt` but make sure you have goimports installed as well, as the format CI check is currently failing: `go install golang.org/x/tools/cmd/goimports@latest`

Some thoughts:

I think there's 3 important cases to test:
1. Is the FailedScheduling event ignored by default? Your current test case covers this.
2. Can users remove the FailedScheduling event from the ignoredUnrecoverableEvents list? In my testing, this is possible by setting ignoredUnrecoverableEvents to an empty array [] -- however, just adding ignoredUnrecoverableEvents:, won't work. To test this do a `kubectl edit dwoc -n $NAMESPACE`:

The following works:

```diff
apiVersion: controller.devfile.io/v1alpha1
config:
  routing:
    clusterHostSuffix: 192.168.49.2.nip.io
    defaultRoutingClass: basic
  workspace:
+    ignoredUnrecoverableEvents: []
    imagePullPolicy: Always
    progressTimeout: 60s
kind: DevWorkspaceOperatorConfig
```

The following will not work:

```diff
apiVersion: controller.devfile.io/v1alpha1
config:
  routing:
    clusterHostSuffix: 192.168.49.2.nip.io
    defaultRoutingClass: basic
  workspace:
+    ignoredUnrecoverableEvents:
    imagePullPolicy: Always
    progressTimeout: 60s
kind: DevWorkspaceOperatorConfig
```

IMO, this behaviour is acceptable.

3. What happens when we add an extra ignoredUnrecoverableEvent? Does it merge the user-provided event(s) with the default event list (that contains FailedScheduling)? Or does it overwrite the default list with the user-provided event(s) list.

Since the DWOC CR doesn't currently show that the FailedScheduling event is being ignored, I would expect it to overwrite the default list with the user-provided list.

However, merging the default event list with the user-provided list might make sense if we use Kubebuilder annotations to set the default value in the CR level as well.

PR #1321, opened 2024-09-16
feat: turn healthCheckHttpClient timeout from 500ms to 3s

batleforc:
### What does this PR do?

It change the timeout of the healthcheck client from 500ms to 3s

Linked to https://github.com/eclipse-che/che/issues/23067
Fixes https://github.com/devfile/devworkspace-operator/issues/1325

### Is it tested? How?

In progress

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -491,8 +491,10 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reconcileStatus.setConditionTrue(conditions.DeploymentReady, "DevWorkspace deployment ready")
 
 	serverReady, serverStatusCode, err := checkServerStatus(clusterWorkspace)
-	if err != nil {
-		return reconcile.Result{}, err
+	if shouldReturn, reconcileResult, reconcileErr := r.checkDWError(workspace, err, "Error checking server status", metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus); shouldReturn {
+		reqLogger.Info("Waiting for DevWorkspace health check endpoint to become available")
+		reconcileStatus.setConditionFalse(dw.DevWorkspaceReady, "Waiting for workspace health check to become available")
+		return reconcileResult, reconcileErr
 	}
 	if !serverReady {
 		reqLogger.Info("Main URL server not ready", "status-code", serverStatusCode)
--- a/controllers/workspace/status.go
+++ b/controllers/workspace/status.go
@@ -26,6 +26,7 @@ import (
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/devworkspace-operator/pkg/common"
+	"github.com/devfile/devworkspace-operator/pkg/dwerrors"
 	"github.com/devfile/devworkspace-operator/pkg/provision/sync"
 
 	"github.com/go-logr/logr"
@@ -211,7 +212,7 @@ func checkServerStatus(workspace *common.DevWorkspaceWithConfig) (ok bool, respo
 
 	resp, err := healthCheckHttpClient.Get(healthz.String())
 	if err != nil {
-		return false, nil, err
+		return false, nil, &dwerrors.RetryError{Err: err, Message: "Failed to check server status", RequeueAfter: 1 * time.Second}
 	}
 	if (resp.StatusCode / 100) == 4 {
 		// Assume endpoint is unimplemented and/or * is covered with authentication.


batleforc:
Setup in a working environment, with the linked PR work

controllers/workspace/http.go
```diff
@@ -70,7 +70,7 @@ func setupHttpClients(k8s client.Client, logger logr.Logger) {
 	}
 	healthCheckHttpClient = &http.Client{
 		Transport: healthCheckTransport,
-		Timeout:   500 * time.Millisecond,
+		Timeout:   3 * 500 * time.Millisecond,
```
AObuchow, line 73: This would actually be 1500 ms (1.5s) instead of 3s.

AObuchow (commented):
@batleforc Thank you for the PR :) 

I assume you're submitting this PR with the hopes of getting it into upstream DWO (and Che) rather than just making changes for your own testing, correct?

Rather than change the default timeout as part of your PR, my gut instinct is we should instead expose a configuration option in the DevWorkspaceOperatorConfig that would allow users to customize the healthCheckHttpClient timeout. Then you could configure the timeout to your desired value from there.

If you're okay with reworking your PR to do this, let me know and I can help guide you further.

batleforc:
HI @AObuchow,
This PR is part of an issue in the [Che Side](https://github.com/eclipse-che/che/issues/23067) where I have a problem with a slow CNI.
Your gut instinct are the same as mine, the end goal would be to have it merged with a possibility to set this value but i was conflicted on how to have it match between the Che Operator and the DevWorkspace Operator.
I'm totally okay on reworking this PR and if you have time I'm waiting for your guidance.

controllers/workspace/http.go
batleforc: Without a doubt a checkout of a stash to high, the test that I deployed was set to a hard coded 3s

dkwon17:
Instead of increasing the timeout, what about returning a `RetryError` when the health check fails [here](https://github.com/devfile/devworkspace-operator/blob/0055cb602028aed9a5ce86cbbf61d4f0c4377dd8/controllers/workspace/status.go#L214) , and handle it with [checkDWError](https://github.com/devfile/devworkspace-operator/blob/0055cb602028aed9a5ce86cbbf61d4f0c4377dd8/controllers/workspace/devworkspace_controller.go#L605)? This is so that another reconcile request would be created if the attempt to ping the health endpoint fails.

@batleforc does that work for your use case?

We do something similar when waiting for the workspace deployment to be ready: https://github.com/devfile/devworkspace-operator/blob/0055cb602028aed9a5ce86cbbf61d4f0c4377dd8/pkg/provision/workspace/deployment.go#L91

https://github.com/devfile/devworkspace-operator/blob/0055cb602028aed9a5ce86cbbf61d4f0c4377dd8/controllers/workspace/devworkspace_controller.go#L485-L489

batleforc:
I think that it could fix the problem, totally answer my case @dkwon17, and could remove the need of changing the Che operator source code. 
What I really need is to not wait for Five more minutes when the IDE is already up but the CNI took too long to broadcast the IP of the Pod (it annoys me and kind of irritate the user).

batleforc:
And your answer wouldn't lock the operator on this action and potentially unlock the process for future action

controllers/workspace/devworkspace_controller.go
```diff
@@ -491,8 +491,10 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reconcileStatus.setConditionTrue(conditions.DeploymentReady, "DevWorkspace deployment ready")
 
 	serverReady, serverStatusCode, err := checkServerStatus(clusterWorkspace)
-	if err != nil {
-		return reconcile.Result{}, err
+	if shouldReturn, reconcileResult, reconcileErr := r.checkDWError(workspace, err, "Error checking server status", metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus); shouldReturn {
+		reqLogger.Info("Waiting for DevWorkspace to be ready")
+		reconcileStatus.setConditionFalse(conditions.DeploymentReady, "Waiting for DevWorkspace to be ready") // No sure of the conditions.DeploymentReady is the right one
```
AObuchow, line 496: `dw.DevWorkspaceReady` may be a more appropriate condition (we use it when the health check fails)

dkwon17, line 496: WDYT @AObuchow ? After removing the comment, it looks good to me

AObuchow (commented):
Thank you for the update to your PR.
@batleforc Have you tried if your latest changes resolve the issue you were encountering?

In my experience, I've seen DWO continuously re-attempt the health check when getting a 502 bad gateway response. However, in these cases you'd see a `Main URL server not ready` -- are you not seeing that message?

dkwon17 (approved):
Thank you @batleforc !

dkwon17:
> In my experience, I've seen DWO continuously re-attempt the health check when getting a 502 bad gateway response. However, in these cases you'd see a Main URL server not ready -- are you not seeing that message?

I've experienced the opposite actually, after handling all the queued reconciles, a retry doesn't happen for the health fail check

dkwon17:
@AObuchow I think you're thinking about this case, where the health check endpoint is accessible, but a 502 is returned: https://github.com/devfile/devworkspace-operator/blob/533d1f0510b4dcfaee8682e5e0b8103812f79cfe/controllers/workspace/devworkspace_controller.go#L497-L501

This PR is to handle the case where there is an a timeout when trying to access to health check

AObuchow:
> @AObuchow I think you're thinking about this case, where the health check endpoint is accessible, but a 502 is returned:
> 
> https://github.com/devfile/devworkspace-operator/blob/533d1f0510b4dcfaee8682e5e0b8103812f79cfe/controllers/workspace/devworkspace_controller.go#L497-L501
> 
> This PR is to handle the case where there is a timeout when trying to access the health check endpoint

@dkwon17 thank you for the confirmation, yes you're right - I believe that's the case I was thinking about.

controllers/workspace/devworkspace_controller.go
AObuchow: After this change, this PR looks good to me :)

batleforc:
Hi @AObuchow  @dkwon17 ,
Didn't have time to try it yet (was pretty late), I will try to set it up in both instance that encounter the problem during the following day and will come back to you as fast as possible.

ibuziuk (approved):

AObuchow (commented):
@batleforc this looks just about good to me to merge :grin: Thank you so much for your contribution.

One last request: Could you please re-organize your commits from this PR?
Here are my suggestions:
- Squash all 3 of your commits into a single commit, as they are all iterations on the change introduced from this PR
- Change the final commit message to `feat: queue workspace reconcile if workspace health check encounters an error`

AObuchow:
> Hi @AObuchow @dkwon17 , Didn't have time to try it yet (was pretty late), I will try to set it up in both instance that encounter the problem during the following day and will come back to you as fast as possible.

Sounds great to me :) I hope this resolves your issue. FWIW: This PR is approved and can be merged (after my final request on cleaning up the commit log) :)

AObuchow:
@batleforc Thank you for updating your PR, sounds good to me. Keep us updated when you have a moment :) It's greatly appreciated.

controllers/workspace/devworkspace_controller.go
```diff
@@ -491,8 +491,10 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reconcileStatus.setConditionTrue(conditions.DeploymentReady, "DevWorkspace deployment ready")
 
 	serverReady, serverStatusCode, err := checkServerStatus(clusterWorkspace)
-	if err != nil {
-		return reconcile.Result{}, err
+	if shouldReturn, reconcileResult, reconcileErr := r.checkDWError(workspace, err, "Error checking server status", metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus); shouldReturn {
+		reqLogger.Info("Waiting for DevWorkspace to be ready")
+		reconcileStatus.setConditionFalse(dw.DevWorkspaceReady, "Waiting for DevWorkspace to be ready") // No sure of the conditions.DeploymentReady is the right one
```
AObuchow, line 496: My mistake, I should have caught this earlier: please remove the comment  `// No sure of the conditions.DeploymentReady is the right one`. You can remove the comment, then `git add controllers/workspace/devworkspace_controller.go && git commit --amend && git push -f`

AObuchow, line 496: @dkwon17 @batleforc  Any thoughts on if we should change the DevWorkspace status message & log here to something more specific to the health check timing out? `"Waiting for DevWorkspace to be ready"`

Maybe something like:
```go
reqLogger.Info("Waiting for DevWorkspace health check endpoint to become available")
reconcileStatus.setConditionFalse(dw.DevWorkspaceReady, "Waiting for workspace health check to become available") 
```

AObuchow (commented):
Some last minute comments, but looks great overall @batleforc  :)

AObuchow (approved):

batleforc:
After 5 day of testing, the problem didn't appear again, so it's a fix

AObuchow:
> After 5 day of testing, the problem didn't appear again, so it's a fix. And the user are more than happy about the 1 minute at most startup 🤣

I'm glad to hear this PR seems to resolve your issue :grin: 
Thank you so much for the great contribution @batleforc & thank you, @dkwon17, for steering this PR in the right direction with your review :) 

Merging this PR now.

PR #1326, opened 2024-10-10
fix: remove dummy caBundle from CRDs

dkwon17:
### What does this PR do?
Removes `webhook.clientConfig.caBundle: Cg==` from the DevWorkspace and DevWorkspaceTemplate CRDs.

For Kubernetes version 1.31 and later, CRDs with `clientConfig.caBundle: Cg==` cannot be applied anymore:
https://github.com/kubernetes/kubernetes/issues/125569#issuecomment-2261245120

This PR fixes https://github.com/redhat-developer/web-terminal-operator/issues/171, https://github.com/devfile/devworkspace-operator/issues/1327

### Is it tested? How?
Create a bundle and apply the catalog source to the cluster and install the DWO version provided from the catalog source:
```
export DWO_BUNDLE_IMG=quay.io/<username>/devworkspace-operator-bundle:removeCaBundle
export DWO_INDEX_IMG=quay.io/<username>/devworkspace-operator-index:removeCaBundle
make generate_olm_bundle_yaml build_bundle_and_index register_catalogsource
```

DWO installation should be successful.

To test even further, you can try creating a DevWorkspace and use conversion webhooks by retrieving the v1alpha1 version:
```
$ kubectl apply -f ./samples/code-latest.yaml 
$ kubectl get devworkspace.v1alpha1.workspace.devfile.io code-latest
```
### PR Checklist

### DIFF
--- a/deploy/templates/cert-manager/crd_webhooks_patch.yaml
+++ b/deploy/templates/cert-manager/crd_webhooks_patch.yaml
@@ -17,7 +17,6 @@ spec:
           path: /convert
           port: 443
         # caBundle will be filled by cert-manager on creation
-        # caBundle: Cg==
 ---
 # Add webhooks to the devfile/api CRDs
 apiVersion: apiextensions.k8s.io/v1
@@ -38,4 +37,3 @@ spec:
           path: /convert
           port: 443
         # caBundle will be filled by cert-manager on creation
-        # caBundle: Cg==
--- a/deploy/templates/cert-manager/crd_webhooks_patch.yaml
+++ b/deploy/templates/cert-manager/crd_webhooks_patch.yaml
@@ -17,7 +17,7 @@ spec:
           path: /convert
           port: 443
         # caBundle will be filled by cert-manager on creation
-        caBundle: Cg==
+        # caBundle: Cg==
 ---
 # Add webhooks to the devfile/api CRDs
 apiVersion: apiextensions.k8s.io/v1
@@ -38,4 +38,4 @@ spec:
           path: /convert
           port: 443
         # caBundle will be filled by cert-manager on creation
-        caBundle: Cg==
+        # caBundle: Cg==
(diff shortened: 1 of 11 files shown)


deploy/templates/cert-manager/crd_webhooks_patch.yaml
```diff
@@ -17,7 +17,7 @@ spec:
           path: /convert
           port: 443
         # caBundle will be filled by cert-manager on creation
-        caBundle: Cg==
+        # caBundle: Cg==
```
AObuchow, line 20: Maybe we should just remove this line (as well as the other mentions of `caBundle: Cg==` from the templates) rather than commenting it out?

AObuchow (commented):
Thanks for catching this issue & for the PR David :) I'm currently requesting an OCP 4.18 cluster to test this out.

Would you mind also creating an DWO issue about what this PR is for? Just something that briefly mentions that the caBundle field used to be required to be filled with a placeholder value for CRD validation to pass, but this is no longer required and having an (invalid) placeholder value actually causes CRD validation to fail. The issue description doesn't have to be thorough, and it could mention the Kubernetes slack thread discussion to provide more context.

This way, we can track this bug getting fixed in the DWO 0.32.0 milestones (for the downstream release changelog)

deploy/templates/cert-manager/crd_webhooks_patch.yaml
dkwon17: I removed them in my latest commit

AObuchow (approved):
Works as expected to me. I installed via OLM on an OpenShift 4.18 cluster & was able to retrieve a v2 devworkspace as a v1 devworkspace:

```
$ kubectl get devworkspace.v1alpha1.workspace.devfile.io plain-devworkspace -o yaml
Warning: workspace.devfile.io/v1alpha1 DevWorkspace is deprecated; use workspace.devfile.io/v1alpha2 DevWorkspace
kind: DevWorkspace
(...)
```

Awesome work @dkwon17, thank you so much for quickly catching and fixing this :partying_face: 

My final small suggestion before merging is to squash your second commit into your first one, and maybe add a `fix  #1327` in the commit description to directly link the GH issue.

Issue #1337, opened 2024-10-31
SSH agent initialization postStart event fails workspace if $HOME is not writable


AObuchow:
### Description
In container images where the $HOME directory is not writable (such as the [go-toolset image](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?q=go-toolset&architecture=amd64&image=671a510d619d7bdd5e9631a0), which has `$HOME=/opt/app-root/src/`) the `init-ssh-agent-command-...` postStart event will fail. 

This is due to the fact we assume `$HOME/ssh-environment` is writable when [doing](https://github.com/devfile/devworkspace-operator/blob/ff723835cc3ed38c9f8a5fbf36dcfddf1890ac9c/pkg/library/ssh/event.go#L26) `ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH`, however this is not always the case:
```bash
$ ssh-agent |  sed 's/^echo/#echo/' > $SSH_ENV_PATH
bash: /opt/app-root/src/ssh-environment: Permission denied
```

Maybe we should wrap the entire ssh-agent intialization command with a `(...) || true` so that regardless of wether a specific step of the ssh-agent initialization fails, the workspace will start up. This is the approach taken for the `init-persistent-home` [preStart even](https://github.com/devfile/devworkspace-operator/blob/ff723835cc3ed38c9f8a5fbf36dcfddf1890ac9c/pkg/library/home/persistentHome.go#L30-L46)t.

### How To Reproduce
1. [Set up an SSH key with a passphrase](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#configuring-devworkspaces-to-use-ssh-keys-for-git-operations) 
2. Create a workspace using the go-toolset image:

```YAML
kind: DevWorkspace
apiVersion: workspace.devfile.io/v1alpha2
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          image: registry.access.redhat.com/ubi9/go-toolset:1.19.13-4.1697647145
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

3. Check the workspace status, it should have failed: `oc get devworkspace -n $NAMESPACE`
 
### Expected behavior
[...]

PR #1338, opened 2024-11-01
Add conditions to the SSH agent post-start command

vinokurig:
### What does this PR do?
Add extra conditions to the SSH agent post-start command:
* Check the `$HOME` directory for the write permissions in order to create the `ssh-environment` file in the user home directory.
* Add `timeout` to the `ssh-add <key path> < <passphrase path>` command. If the passphrase is incorrect, then the `ssh-add` asks to input a new password:
```
Enter passphrase for <key path>: 
Bad passphrase, try again for <key path>:
```
The command will stuck to wait the user input, but the timeout will interrupt the command.
fixes https://github.com/devfile/devworkspace-operator/issues/1337 https://github.com/eclipse-che/che/issues/23213

### Is it tested? How?

see https://github.com/devfile/devworkspace-operator/issues/1337
### PR Checklist

### DIFF
--- a/pkg/library/ssh/event.go
+++ b/pkg/library/ssh/event.go
@@ -21,12 +21,17 @@ import (
 	"github.com/devfile/devworkspace-operator/pkg/library/lifecycle"
 )
 
-const commandLine = `SSH_ENV_PATH=$HOME/ssh-environment \
-&& if [ -f /etc/ssh/passphrase ] && command -v ssh-add >/dev/null; \
-then ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
-&& chmod 600 $SSH_ENV_PATH && source $SSH_ENV_PATH \
-&& ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase \
-&& if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc; fi; fi`
+const commandLine = `(
+SSH_ENV_PATH=$HOME/ssh-environment && \
+if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null && command -v ssh-agent >/dev/null; then
+    ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
+    && chmod 600 $SSH_ENV_PATH \
+    && source $SSH_ENV_PATH \
+    && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase && [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
+        echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
+    fi
+fi
+) || true`
 
 // AddSshAgentPostStartEvent Start ssh-agent and add the default ssh key to it, if the ssh key has a passphrase.
 // Initialise the ssh-agent session env variables in the user .bashrc file.


AObuchow:
@vinokurig Just to clarify: the intention of adding the timeout to the ssh-add is prevent the workspace from failing in the case where the user has provided an incorrect SSH passphrase?

AObuchow (commented):
@vinokurig Thank you for your quick PR in fixing this issue. I still have to do further testing, but here are my current thoughts & suggestions:

1. Maybe we should improve the formatting of the commands for clarity? Here's an idea of how this could look:

```bash
SSH_ENV_PATH=$HOME/ssh-environment && \
if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null; then
    ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
    && chmod 600 $SSH_ENV_PATH \
    && source $SSH_ENV_PATH \
    && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase && [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
        echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
    fi
fi
```
2. We could also add some `else` statements to clarify which steps might have failed:

```bash
SSH_ENV_PATH=$HOME/ssh-environment && \
if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null; then
    ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
    && chmod 600 $SSH_ENV_PATH \
    && source $SSH_ENV_PATH \
    && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase; then
        if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
            echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
        else
            echo "$HOME/.bashrc does not exist or is not writable."
        fi
    else
        echo "Provided SSH passphrase is invalid"
    fi
else
    if [ ! -f /etc/ssh/passphrase ]; then
        echo "/etc/ssh/passphrase file does not exist."
    fi
    if [ ! -w $HOME ]; then
        echo "$HOME directory is not writable."
    fi
    if ! command -v ssh-add >/dev/null; then
        echo "ssh-add command not found."
    fi
fi
```
3. We could wrap the entire command with a `(... || true)` as an added safety mechanism for commands that are unlikely to fail, but still could theoretically fail. For example, if the `ssh-agent` command is missing for some reason:

```bash
(
SSH_ENV_PATH=$HOME/ssh-environment && \
if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null; then
    ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
    && chmod 600 $SSH_ENV_PATH \
    && source $SSH_ENV_PATH \
    && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase; then
        if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
            echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
        else
            echo "$HOME/.bashrc does not exist or is not writable."
        fi
    else
        echo "Provided SSH passphrase is invalid"
    fi
else
    if [ ! -f /etc/ssh/passphrase ]; then
        echo "/etc/ssh/passphrase file does not exist."
    fi
    if [ ! -w $HOME ]; then
        echo "$HOME directory is not writable."
    fi
    if ! command -v ssh-add >/dev/null; then
        echo "ssh-add command not found."
    fi
fi
) || true
```


I realize with all these suggestions applied, the final command is extremely verbose (maybe overkill). A nice middle ground might be to apply some formatting and wrap the entire command in `(... || true)` to ensure this postStart will never fail a workspace that uses a container image with `sh` installed. For example:

```bash
(
SSH_ENV_PATH=$HOME/ssh-environment && \
if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null; then
    ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
    && chmod 600 $SSH_ENV_PATH \
    && source $SSH_ENV_PATH \
    && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase && [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
        echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
    fi
fi
) || true
```

When wrapping the entire command with `(... || true)`, any uncaught errors that occur will be outputted to `/tmp/poststart-stdout.txt`.

AObuchow:
@vinokurig I tested out your PR and it works as expected :) 
I set up an SSH key with a passphrase, and got the following (previously failing) devworkspace to start up:

```YAML
kind: DevWorkspace
metadata:
  name: plain-go-toolset
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          image: registry.access.redhat.com/ubi9/go-toolset:1.19.13-4.1697647145
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
```

pkg/library/ssh/event.go
```diff
@@ -22,11 +22,11 @@ import (
 )
 
 const commandLine = `SSH_ENV_PATH=$HOME/ssh-environment \
-&& if [ -f /etc/ssh/passphrase ] && command -v ssh-add >/dev/null; \
+&& if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null; \
 then ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
 && chmod 600 $SSH_ENV_PATH && source $SSH_ENV_PATH \
-&& ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase \
-&& if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc; fi; fi`
+&& if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase \
```
ibuziuk, line 28: @vinokurig could you clarify the timeout part, looks error prone

vinokurig: @ibuziuk This is need to skip the passphrase input in case if the stored passphrase is wrong. There are some elaborations in the pull request description.

vinokurig:
@AObuchow 
> Just to clarify: the intention of adding the timeout to the ssh-add is prevent the workspace from failing in the case where the user has provided an incorrect SSH passphrase?

yes

vinokurig:
@AObuchow Thank you for the review and for the proposals.
> Maybe we should improve the formatting of the commands for clarity? Here's an idea of how this could look|

done
> We could also add some else statements to clarify which steps might have failed

The post start event output is not exposed to the pod events or logs so echoing is useless at this point. We can redirect the output to a file but that my add some instability e.g. the path where the log file is going to be located may not have the write permissions, WDYT?

> We could wrap the entire command with a (... || true) as an added safety mechanism for commands that are unlikely to fail, but still could theoretically fail.

done

pkg/library/ssh/event.go
```diff
@@ -21,12 +21,15 @@ import (
 	"github.com/devfile/devworkspace-operator/pkg/library/lifecycle"
 )
 
-const commandLine = `SSH_ENV_PATH=$HOME/ssh-environment \
-&& if [ -f /etc/ssh/passphrase ] && command -v ssh-add >/dev/null; \
-then ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
-&& chmod 600 $SSH_ENV_PATH && source $SSH_ENV_PATH \
-&& ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase \
-&& if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc; fi; fi`
+const commandLine = `(SSH_ENV_PATH=$HOME/ssh-environment \
```
AObuchow, line 24: This looks good to me, but in my testing I was able to copy & paste my final [suggestion](https://github.com/devfile/devworkspace-operator/pull/1338#pullrequestreview-2410740468), which did not include a `\` for every line as well as `;` and had slightly different formatting that seemed a bit clearer to me:
```golang
const commandLine = `(
SSH_ENV_PATH=$HOME/ssh-environment && \
if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null; then
    ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
    && chmod 600 $SSH_ENV_PATH \
    && source $SSH_ENV_PATH \
    && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase && [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
        echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
    fi
fi
) || true`
```

WDYT? Do we really need the `\` line breaks and the `;`?

vinokurig: Thanks for the suggestion, applied.

vinokurig: BTW I added a new check for the `ssh-agent` availability.

AObuchow: Awesome, thank you so much @vinokurig & good call on the check for the `ssh-agent` command :+1:

AObuchow (approved):
LGTM, great work & thank you for fixing this so quickly @vinokurig :) 

In my testing, I tried the following devworkspace which also includes a succeeding and failing command, to ensure they would not be affected:

```YAML
kind: DevWorkspace
metadata:
  name: plain-go-toolset-with-event
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          image: registry.access.redhat.com/ubi9/go-toolset:1.19.13-4.1697647145
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
    commands:
      - id: say-hello
        exec:
          component: web-terminal
          workingDir: /
      - id: invalid
        exec:
          component: web-terminal
          commandLine: invalid-command
          workingDir: /
    events:
      postStart:
        - say-hello
        - invalid
```

The resulting postStart lifecycle hook in the pod:

```YAML
    lifecycle:
      postStart:
        exec:
          command:
          - /bin/sh
          - -c
          - |
            {
            cd /
            echo "Hello from $(pwd)"
            cd /
            invalid-command
            (
            SSH_ENV_PATH=$HOME/ssh-environment && \
            if [ -f /etc/ssh/passphrase ] && [ -w $HOME ] && command -v ssh-add >/dev/null && command -v ssh-agent >/dev/null; then
                ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \
                && chmod 600 $SSH_ENV_PATH \
                && source $SSH_ENV_PATH \
                && if timeout 3 ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase && [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then
                    echo "source ${SSH_ENV_PATH}" >> $HOME/.bashrc
                fi
            fi
            ) || true
            } 1>/tmp/poststart-stdout.txt 2>/tmp/poststart-stderr.txt
```

I saw that the valid command had it's stdout logged, and the invalid command had stderr logged correctly:

```YAML
bash-5.1$ cat /tmp/poststart-stderr.txt 
/bin/sh: line 5: invalid-command: command not found
bash-5.1$ cat /tmp/poststart-stdout.txt 
Hello from /
```

AObuchow:
@dkwon17 when you get a chance, could you please take a look at this PR?

dkwon17 (approved):

PR #1341, opened 2024-11-04
add SSH agent postStart event only if SSH key has a passphrase & experimental features enabled

### LINKED ISSUE #1340: Only add SSH Agent PostStart event if SSH key with passphrase is used
Currently, we are adding the `init-ssh-agent-command-...` postStart event unconditionally to all workspaces.
There are cases where this automatically-injected postStart event can actually cause the workspace to fail. For example, if a devworkspace contains a container component that uses a distro-less image where `sh` is missing, as described [here](https://github.com/devfile/devworkspace-operator/issues/1337#issuecomment-2450601646).

We should instead only inject the `init-ssh-agent-command-...` postStart event if the SSH secret that will be mounted to the workspace pod actually requires a passphrase.

## Short term solution
In the short term, we could check if there's an SSH secret with a predefined name `git-ssh-key` (that the Che Dashboard uses) in the workspace's namespace, and check if it has a passphrase. However, this solution creates a coupling between DevWorkspace Operator and the Eclipse Che Dashboard, as DWO does not actually enforce you to name your SSH secret `git-ssh-key`. Instead, it's only given as an example SSH secret name in the [docs](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#configuring-devworkspaces-to-use-ssh-keys-for-git-operations).

## Long term solution
In the long term, echo'ing @l0rd's [thoughts](https://github.com/devfile/devworkspace-operator/issues/1337#issuecomment-2452216834), we should revise how the automatic SSH passphrase injection is handled:

> In general I think that https://github.com/devfile/devworkspace-operator/pull/1307 was a bad idea because there is a high risk to break existing workspaces (we just found 2 container images where the command failed, there may be more). And no matter the value of the new feature implemented, avoiding breaking existing workspaces should be the priority. Also we should avoid, at the DWO level, to modify a DW applied by a user. The dashboard does an automatic generation of DW, so why not adding the post start event there?

Some ideas:
- Move all SSH agent functionality (creation of the SSH agent configmap script & the injection of the postStart event) to the Che Dashboard
- Keep the SSH agent functionality in DWO, but have it enabled only when a devworkspace attribute is provided, e.g. `controller.devfile.io/initialize-ssh-agent: true`. The Che Dashboard could then add this attribute to devworkspace's when the user provides an SSH key that has a passphrase.

AObuchow:
### What does this PR do?
The SSH agent initialization postStart event is now only injected under the following conditions:
- A secret named `git-ssh-key` exists in the workspace's namespace
- The `git-ssh-key` contains a data key called `passphrase`
- `config.enableExperimentalFeatures: true` is set in an [external](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#setting-an-alternate-configuration-for-a-workspace) DWOC used by the workspace, or in the global DWOC.

The intention of this PR is to ensure the SSH agent initialization postStart event is only injected if user's opt-in by configuring the DWOC accordingly, and provide a passphrase in their SSH key.

However, this is only a temporary workaround for DWO 0.31.2. After this PR, we should reconsider how this postStart event should be injected. I've mentioned 2 potential ideas in the long-term solution section of https://github.com/devfile/devworkspace-operator/issues/1340

#1340

### Is it tested? How?
First deploy DWO with the changes from this PR.

There are 4 scenarios to test: 
1. Starting a devworkspace with no SSH secret configured, and experimental features disabled in the DWOC. The SSH agent postStart event should not be injected, and this can be observed from the workspace pod's Kubernetes PostStart lifecycle hook.
2. Starting a devworkspace with a passphrase-less SSH secret configured, and experimental features disabled in the DWOC. The SSH agent postStart event should not be injected.
3. Starting a devworkspace with an SSH secret that contains a passphrase and experimental features disabled in the DWOC.  The SSH agent postStart event should not be injected.
4. Starting a devworkspace with an SSH secret that contains a passphrase and **experimental features enabled** in the DWOC. The SSH agent postStart event should be injected.

I recommend testing all 4 scenarios in order.

#### Scenario 1: no SSH secret configured; experimental features disabled
1. Create a devworkspace:

```YAML
cat <<'EOF' | kubectl apply -n $NAMESPACE -f - 
kind: DevWorkspace
metadata:
  name: plain-devworkspace
spec:
  started: true
  routingClass: 'basic'
  template:
    components:
      - name: web-terminal
        container:
          memoryRequest: 256Mi
          memoryLimit: 512Mi
          mountSources: true
          command:
           - "tail"
           - "-f"
           - "/dev/null"
EOF
```

2. Ensure the devworkspace starts up successfully.
3. Verify that the devworkspace's pod does not have the SSH agent command in its PostStart lifecycle hook. The result of  `oc get pod <workspace-pod-name> -n $NAMESPACE -o json | jq '.spec.containers[0].lifecycle.postStart'` should be `null`.
4. You can now delete your devworkspace: `oc delete dw plain-devworkspace -n $NAMESPACE`

#### Scenario 2: SSH secret configured with no passphrase; experimental features disabled
1. [Configure an SSH key](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#configuring-devworkspaces-to-use-ssh-keys-for-git-operations) in the same namespace that you will create your devworkspace. Make sure you do not set the `$PASSPHRASE` environment variable when creating the SSH secret.
2. Create a devworkspace.
3. Ensure the devworkspace starts up successfully.
5. Verify that the devworkspace's pod does not have the SSH agent command in its PostStart lifecycle hook.
6. You can now delete your devworkspace: `oc delete dw plain-devworkspace -n $NAMESPACE`

#### Scenario 3: SSH secret configured with a passphrase; experimental features disabled
1. Delete your SSH secret from Scenario 2:` oc delete secret git-ssh-key -n $NAMESPACE`
2. [Configure an SSH key](https://github.com/devfile/devworkspace-operator/blob/main/docs/additional-configuration.adoc#configuring-devworkspaces-to-use-ssh-keys-for-git-operations) in the same namespace that you will create your devworkspace. **Make sure you set the `$PASSPHRASE` environment variable when creating the SSH secret.**
3. Create a devworkspace.
5. Ensure the devworkspace starts up successfully.
7. Verify that the devworkspace's pod does not have the SSH agent command in its PostStart lifecycle hook.
8. You can now delete your devworkspace: `oc delete dw plain-devworkspace -n $NAMESPACE`

#### Scenario 4: SSH secret configured with a passphrase; experimental features enabled
1. After testing Scenario 3, enable experimental features in the DWOC: `oc edit dwoc -n $NAMESPACE`

```diff
kind: DevWorkspaceOperatorConfig
apiVersion: controller.devfile.io/v1alpha1
config:
+  enableExperimentalFeatures: true
  routing:
    clusterHostSuffix: 192.168.49.2.nip.io
    defaultRoutingClass: basic
  workspace:
    imagePullPolicy: Always
```
2. Create a devworkspace
3. Ensure the devworkspace starts up successfully.
4.  Verify that the devworkspace's pod **does have the SSH agent command in its PostStart lifecycle hook:** `oc get pod <workspace-pod-name> -n $NAMESPACE -o json | jq '.spec.containers[0].lifecycle.postStart'`

```json
{
  "exec": {
    "command": [
      "/bin/sh",
      "-c",
      "{\nSSH_ENV_PATH=$HOME/ssh-environment \\\n&& if [ -f /etc/ssh/passphrase ] && command -v ssh-add >/dev/null; \\\nthen ssh-agent | sed 's/^echo/#echo/' > $SSH_ENV_PATH \\\n&& chmod 600 $SSH_ENV_PATH && source $SSH_ENV_PATH \\\n&& ssh-add /etc/ssh/dwo_ssh_key < /etc/ssh/passphrase \\\n&& if [ -f $HOME/.bashrc ] && [ -w $HOME/.bashrc ]; then echo \"source ${SSH_ENV_PATH}\" >> $HOME/.bashrc; fi; fi\n} 1>/tmp/poststart-stdout.txt 2>/tmp/poststart-stderr.txt\n"
    ]
  }
}
```

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -281,9 +281,14 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	workspace.Spec.Template = *flattenedWorkspace
 
-	err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template)
-	if err != nil {
-		return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+	if workspace.Config.EnableExperimentalFeatures != nil && *workspace.Config.EnableExperimentalFeatures {
+		if needsSSHAgentPostStartEvent, err := ssh.NeedsSSHPostStartEvent(clusterAPI, workspace.Namespace); err != nil {
+			reqLogger.Error(err, "Error retrieving SSH secret")
+		} else if needsSSHAgentPostStartEvent {
+			if err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template); err != nil {
+				return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+			}
+		}
 	}
 
 	reconcileStatus.setConditionTrue(conditions.DevWorkspaceResolved, "Resolved plugins and parents from DevWorkspace")
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -286,7 +286,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			reqLogger.Error(err, "Error retrieving SSH secret")
 		} else if needsSSHAgentPostStartEvent {
 			if err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template); err != nil {
-				return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+				return r.failWorkspace(workspace, "Failed to add ssh-agent initialization postStart event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
 			}
 		}
 	}
--- a/docs/additional-configuration.adoc
+++ b/docs/additional-configuration.adoc
@@ -189,7 +189,7 @@ Prerequisites:
 ** The steps below assume the following environment variables are set:
 *** `$SSH_KEY`: path on disk to private key for SSH keypair (e.g. `~/.ssh/id_ed25519`)
 *** `$SSH_PUB_KEY`: path on disk to public key for SSH keypair (e.g. `~/.ssh/id_ed25519.pub`)
-*** `$PASSPHRASE`: SSH keypair passphrase (optional)
+*** `$PASSPHRASE`: SSH keypair passphrase (optional). *Note:* requires setting `config.enableExperimentalFeatures: true` in the DevWorkspaceOperatorConfig.
 *** `$NAMESPACE`: namespace where workspaces using the SSH keypair will be started.
 
 Process:
@@ -225,6 +225,8 @@ you must add the following in your `~/.bashrc`:
 ----
 [ -f $HOME/ssh-environment ] && source $HOME/ssh-environment
 ----
++
+*Note:*  Specifying a passphrase for an SSH key is an experimental feature and is controlled by the DevWorkspaceOperatorConfig's `config.enableExperimentalFeatures` field.
 
 3. Annotate the secret to configure automatic mounting to DevWorkspaces
 +
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -87,6 +87,15 @@ const (
 	// in a given namespace. It is used when e.g. adding Git credentials via secret
 	GitCredentialsConfigMapName = "devworkspace-gitconfig"
 
+	// SSHSecretName is the name used for the secret that stores the SSH key data for workspaces in a given namespace.
+	// TODO: This is a workaround for https://github.com/devfile/devworkspace-operator/issues/1340.
+	// We do not enforce the SSH secret to have this name, but it is used by the Che Dashboard and this allows us
+	// to detect if the user has provided an SSH key with a passhprase.
+	SSHSecretName = "git-ssh-key"
+
+	// SSHSecretPassphraseKey is the key used to retrieve the optional passphrase stored inside the SSH secret.
+	SSHSecretPassphraseKey = "passphrase"
+
 	SshAskPassConfigMapName = "devworkspace-ssh-askpass"
 
 	// GitCredentialsMergedSecretName is the name for the merged Git credentials secret that is mounted to workspaces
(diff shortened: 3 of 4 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -281,9 +281,15 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	workspace.Spec.Template = *flattenedWorkspace
 
-	err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template)
-	if err != nil {
-		return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+	if needsSSHAgentPostStartEvent, err := ssh.NeedsSSHPostStartEvent(clusterAPI, workspace.Namespace); err != nil {
```
ibuziuk, line 284: @AObuchow @dkwon17 to be on the safe side can we enable this feature based on DWOC property?

ibuziuk: based on the standup's discussion I like the idea of enabling it based on 

```
config:
```

AObuchow: Unfortunately, the additional commit I created for guarding the SSH agent initialization postStart event injection with `enableExperimentalFeatures` got merged into my original commit while I was cleaning up the git log. However, this change is [present](https://github.com/devfile/devworkspace-operator/pull/1341/files#diff-05e72230161a4d41c5a27fb453041c78839c193a9ddcb5066b1cb62c5a399529R284).

vinokurig (approved):

controllers/workspace/devworkspace_controller.go
```diff
@@ -281,9 +281,15 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	workspace.Spec.Template = *flattenedWorkspace
 
-	err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template)
-	if err != nil {
-		return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+	if workspace.Config.EnableExperimentalFeatures != nil && *workspace.Config.EnableExperimentalFeatures {
+		if needsSSHAgentPostStartEvent, err := ssh.NeedsSSHPostStartEvent(clusterAPI, workspace.Namespace); err != nil {
+			// TODO: Should we fail the workspace? Or log the error and continue on?
```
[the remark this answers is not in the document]
AObuchow: +1
Given the fact that this PR aims to prevent breaking any workspaces that would otherwise work fine prior to the SSH agent postStart event, we should probably just log an error rather than failing the workspace.

vinokurig:
It would be nice to add a documentation note e.g.
```
*Note:*  Specifying a passphrase for an SSH key is an experimental feature and is controlled by the `DevWorkspaceOperatorConfig.EnableExperimentalFeatures` option.
```

AObuchow:
> It would be nice to add a documentation note e.g.
> 
> ```
> *Note:*  Specifying a passphrase for an SSH key is an experimental feature and is controlled by the `DevWorkspaceOperatorConfig.EnableExperimentalFeatures` option.
> ```

+1 will add an extra commit for this

dkwon17 (approved):

AObuchow:
@dkwon17 thank you for the review :) will squash my fixup commits tomorrow & have this merged

ibuziuk (approved):

PR #1344, opened 2024-11-06
Mark the injection of the SSH agent postStart event as experimental feature

vinokurig:
### What does this PR do?
Wrap the `AddSshAgentPostStartEvent` function into the `EnableExperimentalFeatures` condition. 

https://github.com/devfile/devworkspace-operator/issues/1337

### Is it tested? How?

Create a workspace with a container without sh:
```yaml
kind: DevWorkspace
metadata:
  name: plain-devworkspace
spec:
  routingClass: che
  started: true
  template:
    components:
      - container:
          image: 'quay.io/devfile/universal-developer-image:ubi8-latest'
        name: universal-developer-image
      - container:
          image: 'ghcr.io/l0rd/outyet@sha256:3ab91b5801ab2e2a6147cab5d4a959838d9318921298e84cf7b6d19a3359e496'
        name: test
```
See: workspace starts successfully.
### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -281,9 +281,12 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	workspace.Spec.Template = *flattenedWorkspace
 
-	err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template)
-	if err != nil {
-		return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+	// Include to experimental features list because it is not clear how to handle post start events in containers without sh.
+	if *workspace.Config.EnableExperimentalFeatures {
+		err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template)
+		if err != nil {
+			return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+		}
 	}
 
 	reconcileStatus.setConditionTrue(conditions.DevWorkspaceResolved, "Resolved plugins and parents from DevWorkspace")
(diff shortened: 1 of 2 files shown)


controllers/workspace/devworkspace_controller.go
```diff
@@ -281,9 +281,12 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	workspace.Spec.Template = *flattenedWorkspace
 
-	err = ssh.AddSshAgentPostStartEvent(&workspace.Spec.Template)
-	if err != nil {
-		return r.failWorkspace(workspace, "Failed to add ssh-agent post start event", metrics.ReasonWorkspaceEngineFailure, reqLogger, &reconcileStatus), nil
+	// Include to experimental features list because it is not clear how to handle post start events in containers without sh.
+	if *workspace.Config.EnableExperimentalFeatures {
```
AObuchow, line 285: I believe this could result in a null pointer exception since `workspace.Config.EnableExperimentalFeatures` might be null. Instead we should check if it's null beforehand:
```go
if workspace.Config.EnableExperimentalFeatures != null && *workspace.Config.EnableExperimentalFeature
```

docs/additional-configuration.adoc
```diff
@@ -225,6 +225,7 @@ you must add the following in your `~/.bashrc`:
 ----
 [ -f $HOME/ssh-environment ] && source $HOME/ssh-environment
 ----
+*Note:*  This is an experimental feature and is controlled by the `DevWorkspaceOperatorConfig.EnableExperimentalFeatures` option.
```
AObuchow, line 228: Good call on documenting this! Maybe we should clarify that using an SSH passphrase is an experimental feature? Something like:
```adoc
*Note:*  Specifying a passphrase for an SSH key is an experimental feature and is controlled by the `DevWorkspaceOperatorConfig.EnableExperimentalFeatures` option.
```

AObuchow (commented):
Thank you for the PR @vinokurig :) I've left some comments, but it's worth noting that in https://github.com/devfile/devworkspace-operator/pull/1341 I've also guarded the SSH agent initialization postStart event as an experimental feature. 

Maybe we can take the documentation changes you've made here and add them to my existing PR (which I would appreciate your review of)?

Issue #1335, opened 2024-10-30
Document how to install the DWO on minikube/kind/ocp


l0rd:
### Description
We should document how to deploy DWO on Minikube, Kind, and OpenShift, with and without OLM, and without requiring `make` or `go` or any dev tools (`kubectl`, `oc`, `curl`, etc. are fine).

DWO on Kind may not work currently because admission webhooks are not supported there. If that's the case, we should discuss whether we can make admission webhooks optional (or explore other options for deploying DWO on Kind).

### Additional context
This issue is related to #1334

PR #1371, opened 2025-01-14
Add installation guides

akurinnoy:
### What does this PR do?

This PR adds installation guides for installing the DWO on Kind, minikube, and Openshift.

resolves https://github.com/devfile/devworkspace-operator/issues/1335

### Is it tested? How?

n/a

### PR Checklist

### DIFF
--- a/docs/installation/kind-without-olm-linux.md
+++ b/docs/installation/kind-without-olm-linux.md
@@ -180,3 +180,10 @@ You can also check the DevWorkspace status by running:
 ```sh
 kubectl get devworkspace -n default
 ```
+
+When the DevWorkspace is running according to the status, open the editor by accesssing the URL from the `INFO` column in a web browser. For example:
+
+```
+NAME                            DEVWORKSPACE ID             PHASE     INFO
+git-clone-sample-devworkspace   workspace0196ce197f0b4e90   Running   <URL>
+```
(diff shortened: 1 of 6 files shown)


docs/installation/kind-without-olm-linux.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+**Note:** The DevWorkspace creation may fail due to timeout if some container images are large and take longer than 5 minutes to pull. If the DevWorkspace fails, you can restart it by setting `spec.started` to `true`. Use the following command to re-trigger the DevWorkspace start:
+
+```bash
+kubectl patch devworkspace git-clone-sample-devworkspace -n default --type merge -p '{"spec": {"started": true}}'
+```
+You can also check the DevWorkspace status by running:
+```sh
+kubectl get devworkspace -n default
```
dkwon17, line 181: WDYT about adding something like:

dkwon17: When the DevWorkspace is running according to the status, open the editor by accesssing the URL from the `INFO` column in a web browser. For example:
```
NAME                            DEVWORKSPACE ID             PHASE     INFO
git-clone-sample-devworkspace   workspace0196ce197f0b4e90   Running   <URL>
```
?

dkwon17 (commented):
Looks great, I was able to follow all the steps for all the cases

docs/installation/kind-without-olm-linux.md
akurinnoy: Yes, this is an important step I forgot to mention, and it should be in this doc.

dkwon17 (approved):

Issue #1194, opened 2023-10-23
Improve handling of auto-mount path collisions


amisevsk:
### Description
When mounting configmaps/secrets as files to a workspace, DWO currently doesn't detect path collisions (e.g. mounting two files via subpath to the same path). When this occurs, container creation will fail with a message that doesn't make the issue clear, and the DevWorkspace will fail due to a CreateContainerError event. 

Instead, DWO should try to detect these issues before they occur and fail the workspace with an appropriate message.

PR #1378, opened 2025-02-10
fix: auto-mount path collisions

akurinnoy:
### What does this PR do?

This PR makes DWO handle auto-mount path collisions and fails the workspace with the error message.

fixes https://github.com/devfile/devworkspace-operator/issues/1194

### PR Checklist

### DIFF
--- a/pkg/provision/automount/projected.go
+++ b/pkg/provision/automount/projected.go
@@ -55,6 +55,7 @@ func mergeProjectedVolumes(resources *Resources) (*Resources, error) {
 	}
 
 	// Map of merged volume names -> bool, for not merging the same volume twice
+	// This can happen due to different subpath volume mounts, where the mount path is the same. In this case, there should be only one volume.
 	mergedVolumeNames := map[string]bool{}
 	for _, mountPath := range mountPathOrder {
 		volumeMounts := mountPathToVolumeMounts[mountPath]
--- a/pkg/provision/automount/projected.go
+++ b/pkg/provision/automount/projected.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	"github.com/devfile/devworkspace-operator/pkg/common"
+	"github.com/devfile/devworkspace-operator/pkg/dwerrors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


dkwon17:
Thank you for the PR @akurinnoy , 

Could we do the collision check here? 

https://github.com/devfile/devworkspace-operator/blob/5864a646dda8b2f92c91a77ec356679dea4ce060/pkg/provision/automount/projected.go#L42

Right now, if there is a collision, it fails silently because it overwrites the existing value of  the map

dkwon17:
This is the case I tested this PR with:

```
kind: ConfigMap
apiVersion: v1
metadata:
  name: my-configmap-mount-1
  labels:
    controller.devfile.io/mount-to-devworkspace: 'true'
  annotations:
    controller.devfile.io/mount-path: /home/user/hello-world
data:
  myfile: this is my file
```

and:

```
kind: ConfigMap
apiVersion: v1
metadata:
  name: my-configmap-mount-2
  labels:
    controller.devfile.io/mount-to-devworkspace: 'true'
  annotations:
    controller.devfile.io/mount-path: /home/user/hello-world
data:
  myfile: this is my file
```

However, I was able to get it working if I add this annotation:
```
controller.devfile.io/mount-as: subpath
```
image

dkwon17:
Hello @akurinnoy could you please rebase this PR?

pkg/provision/automount/projected.go
```diff
@@ -53,6 +54,8 @@ func mergeProjectedVolumes(resources *Resources) (*Resources, error) {
 		volumeNameToVolume[volume.Name] = volume
 	}
 
+	// Map of merged volume names -> bool, for not merging the same volume twice
```
dkwon17, line 57: Could we add an additional comment such as:

```
This can happen due to different subpath volume mounts, where the mount path is the same. In this case, there should be only one volume.
```

akurinnoy: Done!

dkwon17 (approved):

dkwon17:
Thank you @akurinnoy, please squash and merge

Issue #1376, opened 2025-02-07
Add a devworkspace pruner to the DevWorkspace Operator


cgruver:
### Description
A large scale deployment of Eclipse Che / OpenShift Dev Spaces can result in a lot of stale DevWorkspace objects that are really no longer necessary but continue to occupy space in `etcd`.

Over time the performance of `etcd` can be impacted resulting in the need to scale up the control plane nodes with more CPU/RAM.

### Additional context
Here is a prototype for implementing a devworkspace pruner based on the last time that a workspace was started:

```yaml
apiVersion: batch/v1
kind: CronJob
metadata:
  name: devworkspace-pruner
  namespace: openshift-operators
spec:
  schedule: "0 0 1 * *"
  successfulJobsHistoryLimit: 3
  failedJobsHistoryLimit: 3
  concurrencyPolicy: Forbid
  jobTemplate:
    spec:
      template:
        spec:
          volumes: 
          - name: script
            configMap:
              name: devworkspace-pruner
              defaultMode: 0555
              items:
              - key: devworkspace-pruner
                path: devworkspace-pruner.sh
          restartPolicy: OnFailure
          serviceAccount: devworkspace-controller-serviceaccount
          containers:
          - name: openshift-cli
            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
            env:
            - name: RETAIN_TIME
              # 30 days
              value: "2592000"
            command:
            - /script/devworkspace-pruner.sh
            resources:
              requests:
                cpu: 100m
                memory: 64Mi
              limits:
                cpu: 100m
                memory: 64Mi
            volumeMounts:
            - mountPath: /script
              name: script
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: devworkspace-pruner
  namespace: openshift-operators
data:
  devworkspace-pruner: |
    #!/usr/bin/env bash
    current_time=$(date +%s)
    for namespace in $(oc get namespaces -l app.kubernetes.io/component=workspaces-namespace -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
    do
      for workspace in $(oc get devworkspaces -n ${namespace} -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
      do
[...]

PR #1402, opened 2025-03-18
Add DevWorkspace Pruner (operator-lib/prune package)

akurinnoy:
### What does this PR do?

This pull request adds a new controller that prunes (removes) DevWorkspaces based on these rules:

- DevWorkspaces that haven't been started for a specific period are considered inactive. This period is configurable.
- The pruning behavior (enable/disable, retention time, cron schedule, dry-run mode) is configured via the `DevWorkspaceOperatorConfig` (DWOC) resource.
- The pruning process is scheduled using a `cron` library, allowing for flexible configuration of the pruning frequency.
- A dry-run mode is supported, allowing administrators to preview the pruning actions without actually deleting any resources.

resolves https://github.com/devfile/devworkspace-operator/issues/1376

### Is it tested? How?

1. Deploy the operator with this new code.
2. Ensure that the `DevWorksapceOperatorConfig` initially does not have pruning enabled by default
    - there is no `CleanupCronJob` section in the `config.workspace` section of the DWOC.
    - or, if there is the `CleanupCronJob` section, ensure that `enable: false`.
3. Patch the `DevWorkspaceOperatorConfig` to enable pruning and set parameters:
  ```sh
  kubectl patch DevWorkspaceOperatorConfig devworkspace-operator-config \
  -n devworkspace-controller \
  --type=merge \
  -p '{"config": {"workspace": {"cleanupCronJob": {"enable": true, "dryRun": true, "retainTime": 60, "schedule": "* * * * *"}}}}'
  ```
4. Create several DevWorkspaces. Some should be started, and some should not be started.
5. Wait for the cron schedule to trigger the pruning - at least 1 minute.
6. Check logs and verify that the correct DevWorkspaces are identified but not deleted in dry-run mode.
7. Disable dry-run mode and verify that all eligible DevWorkspaces are actually deleted:
  ```sh
  kubectl patch DevWorkspaceOperatorConfig devworkspace-operator-config \
  -n devworkspace-controller \
  --type=merge \
  -p '{"config": {"workspace": {"cleanupCronJob": {"dryRun": false }}}}'
  ```

### PR Checklist


controllers/pruner/pruner_controller.go
```diff
@@ -0,0 +1,338 @@
...
+func (r *DevWorkspacePrunerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	log := r.Log.WithName("setupWithManager")
+	log.Info("Setting up DevWorkspacePrunerReconciler")
+
+	configPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return shouldReconcileOnUpdate(e, log)
+		},
+		CreateFunc:  func(e event.CreateEvent) bool { return true },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
```
dkwon17, line 115: If the cronjob is enabled with a DWOC, and then if we delete the DWOC, will the cronjob still be enabled?

controllers/pruner/pruner_controller.go
```diff
@@ -0,0 +1,361 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconciles}).
+		Named("DevWorkspacePruner").
+		Watches(&source.Kind{Type: &controllerv1alpha1.DevWorkspaceOperatorConfig{}},
+			handler.EnqueueRequestsFromMapFunc(func(object client.Object) []ctrl.Request {
+				objectNamespace := object.GetNamespace()
+				operatorNamespace, err := infrastructure.GetNamespace()
+
+				// Ignore events from other namespaces
+				if err != nil || objectNamespace != operatorNamespace {
```
dkwon17, line 139: Could we also check name of the object to make sure it matches https://github.com/akurinnoy/devworkspace-operator/blob/f977dfbeea80f677aae7d997182234b83c41b7c0/pkg/config/sync.go#L44 ?

controllers/pruner/pruner_controller.go
```diff
@@ -0,0 +1,361 @@
...
+	configPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return shouldReconcileOnUpdate(e, log)
+		},
+		CreateFunc:  func(e event.CreateEvent) bool { return true },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}
+
+	maxConcurrentReconciles, err := config.GetMaxConcurrentReconciles()
```
dkwon17, line 122: It looks like `MaxConcurrentReconciles` creates new goroutines that run for the lifetime of the controller:
https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/internal/controller/controller.go#L197-L205

Maybe we should remove this so that the [default](https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/internal/controller/controller.go#L319) is used, since we are only reconciling one object now?

ibuziuk (approved):
@akurinnoy great job, could you please provide a separate PR with the docs to https://github.com/devfile/devworkspace-operator/blob/main/docs/dwo-configuration.md

Issue #1269, opened 2024-05-31
Common PVC cleanup job can be assigned to incorrect node in multi-node cluster


AObuchow:
### Description
In a multi-node cluster, it's possible when deleting a devworkspace that uses the per-user/common PVC strategy for the PVC cleanup pod to be scheduled on a node that is different than the node where the PVC is mounted. Since PVCs are created as ReadWriteOnce, only a single node can mount the PVC and thus the cleanup pod will fail to start with a PVC mount error. This causes the devworkspace to remain in a terminating state indefinitely. 

Since you cannot modify the node that a pod is scheduled on after the pod has been created, you need to delete the cleanup pod and have it automatically re-created until it is assigned to the node where the PVC is mounted in order for the workspace to be deleted.

What's odd is that we are already[ applying a node selector label ](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/storage/cleanup.go#L205) to the cleanup pod. Perhaps there are cases where the namespace is missing the [node selector annotation](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/config/config.go#L100)? CC: @musienko-maxim 

### How To Reproduce
Does not always occur, requires a multi-node cluster.
1. Create a devworkspace using the per-user/common storage strategy
2. Delete the devworkspace
3. If the cleanup-workspace<id> pod is scheduled on a different node than where the PVC is mounted, the pod will fail to be created and the devworkspace will remain in the terminating state

### Expected behavior
The cleanup-workspace<id> pod is scheduled on thesame node where the PVC is mounted and terminates successfully. The deworkspace gets terminated successfully. 

### Additional context
Encountered this while testing on @musienko-maxim 's OCP 4.15 test cluster.

AObuchow:
> What's odd is that we are already[ applying a node selector label ](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/storage/cleanup.go?rgh-link-date=2024-05-31T17%3A40%3A06Z#L205) to the cleanup pod. Perhaps there are cases where the namespace is missing the [node selector annotation](https://github.com/devfile/devworkspace-operator/blob/main/pkg/provision/config/config.go?rgh-link-date=2024-05-31T17%3A40%3A06Z#L100)? 

It seems like this annotation is supposed to be applied from Che, as configured from this [Che Cluster CR field](https://github.com/eclipse-che/che-operator/blob/main/api/v2/checluster_types.go#L97).

[...]

PR #1415, opened 2025-04-17
fix: common PVC cleanup job to be assigned to a correct node

akurinnoy:
### What does this PR do?

This pull request enhances the common PVC cleanup job by adding node affinity rule.

This change modifies the cleanup job creation logic to:
1.  Read the `volume.kubernetes.io/selected-node` annotation from the common PVC associated with the DevWorkspace.
2.  If the annotation exists, add a `requiredDuringSchedulingIgnoredDuringExecution` node affinity rule to the cleanup job's pod spec, targeting the specified node.

This ensures that the cleanup job pod runs on the same node as the volume. If the annotation is not present, no node affinity is added.

resolves https://github.com/devfile/devworkspace-operator/issues/1269

### Is it tested? How?

1. Create at least two DevWorkspaces; at least one should be running.
2. Verify that the PVC has the `volume.kubernetes.io/selected-node` annotation. Note the node name specified in the annotation value.
    ```bash
    kubectl get pvc -n <namespace> claim-devworkspace -o jsonpath='{.metadata.annotations.volume\.kubernetes\.io/selected-node}'
    ```  
3. Delete a DevWorkspace, and leave at least one running DevWorkspace untouched.
4. Inspect the cleanup job to verify that the node affinity rule was added:
    ```bash
    kubectl get job -n <namespace> <job-name> -o jsonpath='{.spec.template.spec.affinity.nodeAffinity}'
    ```
    Check that the `nodeSelectorTerms` target the node name noted in step 2 via the `kubernetes.io/hostname` key.
5. Ensure the cleanup Job completes successfully and the DevWorkspace is deleted.

### PR Checklist

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -96,3 +96,8 @@ const (
 	// ProjectCloneDisable specifies that project cloning should be disabled.
 	ProjectCloneDisable = "disable"
 )
+
+const (
+	// This annotation is added to a PVC that is triggered by a scheduler to be dynamically provisioned. Its value is the name of the selected node.
+	SelectedNodeAnnotation = "volume.kubernetes.io/selected-node"
+)
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -98,6 +98,6 @@ const (
 )
 
 const (
-	// This annotation is added to a PVC that is triggered by a scheduler to be dynamically provisioned. Its value is the name of the selected node.
+	// SelectedNodeAnnotation annotation is added to a PVC that is triggered by a scheduler to be dynamically provisioned. Its value is the name of the selected node.
 	SelectedNodeAnnotation = "volume.kubernetes.io/selected-node"
 )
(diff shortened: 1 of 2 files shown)


pkg/constants/constants.go
```diff
@@ -96,3 +96,8 @@ const (
 	// ProjectCloneDisable specifies that project cloning should be disabled.
 	ProjectCloneDisable = "disable"
 )
+
+const (
+	// This annotation is added to a PVC that is triggered by a scheduler to be dynamically provisioned. Its value is the name of the selected node.
```
rohanKanojia, line 101: nit, according to [godoc comments](https://go.dev/doc/comment#const) format. This comment Should start with variable name
```suggestion
	// SelectedNodeAnnotation annotation is added to a PVC that is triggered by a scheduler to be dynamically provisioned. Its value is the name of the selected node.
```

rohanKanojia (approved):

PR #1427, opened 2025-05-03
Update project to go 1.23

dkwon17:
### What does this PR do?
Updates project to use go 1.23

Update to go 1.23 is needed to fix CVEs:
[cve-2025-22869](https://access.redhat.com/security/cve/cve-2025-22869) which has a CVSS score of 7.5.
[cve-2025-22872](https://access.redhat.com/security/cve/cve-2025-22872) which has a CVSS score of 6.5.

### PR Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/devfile/devworkspace-operator
 
-go 1.22
-
-toolchain go1.22.0
+go 1.23.0
 
 require (
 	github.com/devfile/api/v2 v2.2.2
--- a/.ci/oci.Dockerfile
+++ b/.ci/oci.Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 
-FROM registry.ci.openshift.org/openshift/release:golang-1.22
+FROM registry.ci.openshift.org/openshift/release:golang-1.23
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/devfile/devworkspace-operator
 
-go 1.22
-
-toolchain go1.22.0
```
rohanKanojia, line 5: Why is the toolchain version omitted?

ibuziuk (approved):

go.mod
akurinnoy: I think this is a good point. Could you add the `toolchain` directive here so that we can merge this PR?

akurinnoy: @rohanKanojia I noticed this line in the [golang toolchain doc](https://tip.golang.org/doc/toolchain#get):

> When the toolchain matches the `go` line exactly, it can be omitted and implied, so this `go get` will delete the `toolchain` line.

I also checked the `go mod tidy` behaviour, and it works the same way. Hence, the pull request check will fail if we add the `toolchain` directive with the same version:

https://github.com/devfile/devworkspace-operator/blob/9ff4cee22e67cb71d7ed6ba6e473308fbcd8ae8e/.github/workflows/pr.yml#L62-L71

So, we can merge this PR as is.

rohanKanojia: Okay, thanks for the clarification!

rohanKanojia (approved):

Issue #1343, opened 2024-11-06
Allow configuring memory and CPU request / limits  


ibuziuk:
### Description
Currently, it is not possible to configure memory and CPU for DWO pod. The only workaround is patching CSV directly, but since the CSV gets replaced as part of an upgrade, the modified value gets reverted to the older default values e.g. for devworkspace-controller-manager
```
              spec:
                containers:
                  - resources:
                      limits:
                        cpu: '3'
                        memory: 3Gi
                      requests:
                        cpu: 250m
                        memory: 100Mi
```

### Additional context
Need to investigate if we can propagate RAM / CPU configuration from the subscription directly and if it is a good approach in general - https://developers.redhat.com/articles/2023/03/06/5-global-environment-variables-provided-openshift-gitops#

ibuziuk:
@dkwon17 @AObuchow please, review

dkwon17:
As discussed, it's possible to set the memory and CPU limits of the DWO pod in the `Subscription`: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md

Example:
```
spec:
  channel: fast
  installPlanApproval: Automatic
  name: devworkspace-operator
  source: redhat-operators
  sourceNamespace: openshift-marketplace
  startingCSV: devworkspace-operator.v0.31.1
  config:
    resources:
      limits:
        memory: "4Gi"
```

However, the resource limits are applied to all containers (2) in the pod, and not just the `devworkspace-controller` containers. Since there are two containers, the total memory limit for the configuration above is 8Gi

dkwon17:
> However, the resource limits are applied to all containers (2) in the pod, and not just the devworkspace-controller container. Since there are two containers (devworkspace-controller and kube-rbac-proxy), the total memory limit for the configuration above is 8Gi

If we remove the use of kube-rbac-proxy to use [WithAuthenticationAndAuthorization](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization) as described in https://github.com/devfile/devworkspace-operator/issues/1352, this won't be a problem anymore.

rohanKanojia:
In order to use [WithAuthenticationAndAuthorization](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization) , we need to have controller-runtime >=0.16.0
[...]

PR #1435, opened 2025-05-19
chore (deps) : bump controller-runtime to v0.18.7

rohanKanojia:
### What does this PR do?

Bump controller-runtime dependency to v0.18.7

There are lots of changes in controller-runtime since last upgrade. They're listed below:

- `Port` removed from controller Options, we now have to create a separate WebhookServer object and add it to manager

- Add new argument `ctx.Context` in Watch handler method

- `source.Kind` not available in latest versions anymore, use directly type instead in Watch()

- MetricsBindAddress no longer available in controller Options, use Metrics object instead.

- `cache.SelectorsByObject` no longer present in cache package, users are advised to use ByObject inside cache Options instead

- `cache.BuilderWithOptions` no longer present in controller-runtime, use inline method in return instead.

- `fake.NewFakeClientWithScheme` not available anymore, use `fake.NewClientBuilder()` instead

- Replace `corev1.ResourceRequirements` with `corev1.VolumeResourceRequirements`

- Specify finalizers in finalize_test wherever DeletionTimestamp is provided to avoid error thrown by test for missing finalizers

- `admission.Decoder` is an interface now, previously it was a concrete type.

Signed-off-by: Rohan Kumar <rohaan@redhat.com>
Related to #1343

### PR Checklist


pkg/cache/cache.go
```diff
@@ -94,39 +97,37 @@ func GetCacheFunc() (cache.NewCacheFunc, error) {
 	if infrastructure.IsOpenShift() {
 		// Annoying quirk: cache.SelectorsByObject uses an internal struct for values (internal.Selector)
 		// so we _can't_ just add Routes here since we cannot initialize the corresponding value.
```
dkwon17, line 99: Could we delete this comment?
```
		// Annoying quirk: cache.SelectorsByObject uses an internal struct for values (internal.Selector)
		// so we _can't_ just add Routes here since we cannot initialize the corresponding value.
```

rohanKanojia: This comment was added in this PR https://github.com/devfile/devworkspace-operator/pull/652

I'll remove it in next update.

dkwon17 (approved):

dkwon17:
I've tested that resource caching is working by creating many [automount configmaps](https://github.com/devfile/devworkspace-operator/blob/4823c6cf920e33d4d3101a53ef3b98517b3f8349/pkg/constants/metadata.go#L43), with and without the [watch-configmap label](https://github.com/devfile/devworkspace-operator/blob/4823c6cf920e33d4d3101a53ef3b98517b3f8349/pkg/constants/metadata.go#L43) and observing the memory usage of the `devworkspace-controller` pod.


Thank you @rohanKanojia , could you please squash and merge this PR?

PR #1440, opened 2025-05-29
postStart hook commands timeout

akurinnoy:
### What does this PR do?

This PR addresses the issue of `postStart` hook failures in DevWorkspaces when hook commands not exiting within the timeout period, so that the workspace pod gets stuck in `Terminating` state and never gets deleted.

This PR resolves the issue by:
- Introducing `timeout` for `postStart` hook. User-provided commands are now wrapped with the `timeout` utility. This ensures that `postStart` hook commands are terminated if they exceed a configurable duration. The timeout duration can be set in the `DevWorkspaceOperatorConfig` (a value of 0 means no timeout):
    ```yaml
    # DevWorkspaceOperatorConfig
    # ...
    config:
      workspace:
        postStartTimeout: 30 # Timeout in seconds
    ```
- Adding the parsing logic for interpreting various Kubelet messages to extract an exact reason or exit code for lifecycle hook failures.

https://issues.redhat.com/browse/CRW-8329

### Is it tested? How?

1. Install DWO from this PR:
```sh
oc apply -f - <<EOF
kind: CatalogSource
metadata:
spec:
  sourceType: grpc
  image: quay.io/okurinny/devworkspace-operator-index:postStartHookTimeout
  publisher: Red Hat
  updateStrategy:
    registryPoll:
      interval: 5m
EOF
```

```sh
oc apply -f - <<EOF
kind: Subscription
metadata:
  name: devworkspace-operator
spec:
  channel: next
  installPlanApproval: Automatic
  name: devworkspace-operator
  source: devworkspace-operator-catalog
  sourceNamespace: openshift-marketplace
EOF
```
2. Create `DevWorkspaceOperatorConfig` with the `postStart` hook timeout duration (in seconds):
```sh
oc apply -f - <<EOF
metadata:
config:
  workspace:
    postStartTimeout: 30
EOF
```
3. Create a problematic `DevWorkspace` designed to have its `postStart` hook time out:
```sh
oc apply -f - <<EOF
kind: DevWorkspace
metadata:
  name: problematic-workspace
spec:
  started: true
  template:
    components:
      - name: tools
        container:
          image: quay.io/devfile/universal-developer-image:ubi9-latest
          memoryLimit: "1Gi"
          memoryRequest: "512Mi"
          cpuRequest: "250m"
          cpuLimit: "1000m"
    commands:
      - id: sleep-infinity-cmd
        exec:
          component: tools
          commandLine: "echo 'PostStart: Starting infinite sleep...'; sleep infinity; echo 'PostStart: Sleep finished (should not be reached)'"
    events:
      postStart:
        - sleep-infinity-cmd
EOF
```
4. Watch the DevWorkspace:
```sh
oc get dw problematic-workspace -w
```
5. The DevWorkspace should eventually enter a `Failed` phase.
6. The `status.message` of the DevWorkspace should provide a reason for the failure, indicating a timeout. For example: `Error creating DevWorkspace deployment: Container tools has state [postStart hook] Commands terminated by SIGTERM (likely timed out after 30s). Exit code 143.`

### PR Checklist

### DIFF
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -192,12 +192,10 @@ type WorkspaceConfig struct {
 	// PostStartTimeout defines the maximum duration the PostStart hook can run
 	// before it is automatically failed. This timeout is used for the postStart lifecycle hook
 	// that is used to run commands in the workspace container. The timeout is specified in seconds.
-	// If not specified, the timeout is disabled (0 seconds).
-	// +kubebuilder:validation:Minimum=0
+	// Duration should be specified in a format parseable by Go's time package, e.g. "20s", "2m".
+	// If not specified or "0", the timeout is disabled.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Type=integer
-	// +kubebuilder:validation:Format=int32
-	PostStartTimeout *int32 `json:"postStartTimeout,omitempty"`
+	PostStartTimeout string `json:"postStartTimeout,omitempty"`
 }
 
 type WebhookConfig struct {
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -189,6 +189,16 @@ type WorkspaceConfig struct {
 	RuntimeClassName *string `json:"runtimeClassName,omitempty"`
 	// CleanupCronJobConfig defines configuration options for a cron job that automatically cleans up stale DevWorkspaces.
 	CleanupCronJob *CleanupCronJobConfig `json:"cleanupCronJob,omitempty"`
+	// PostStartTimeout defines the maximum duration the PostStart hook can run
+	// before it is automatically failed. This timeout is used for the postStart lifecycle hook
+	// that is used to run commands in the workspace container. The timeout is specified in seconds.
+	// If not specified, the timeout is disabled (0 seconds).
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default:=0
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Type=integer
+	// +kubebuilder:validation:Format=int32
+	PostStartTimeout *int32 `json:"postStartTimeout,omitempty"`
 }
 
 type WebhookConfig struct {
--- a/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
+++ b/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
@@ -194,7 +194,6 @@ type WorkspaceConfig struct {
 	// that is used to run commands in the workspace container. The timeout is specified in seconds.
 	// If not specified, the timeout is disabled (0 seconds).
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:default:=0
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Type=integer
 	// +kubebuilder:validation:Format=int32
--- a/pkg/library/lifecycle/poststart.go
+++ b/pkg/library/lifecycle/poststart.go
@@ -16,9 +16,11 @@ package lifecycle
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
@@ -39,7 +41,7 @@ const (
 `
 )
 
-func AddPostStartLifecycleHooks(wksp *dw.DevWorkspaceTemplateSpec, containers []corev1.Container, postStartTimeout *int32) error {
+func AddPostStartLifecycleHooks(wksp *dw.DevWorkspaceTemplateSpec, containers []corev1.Container, postStartTimeout string) error {
 	if wksp.Events == nil || len(wksp.Events.PostStart) == 0 {
 		return nil
 	}
@@ -83,8 +85,8 @@ func AddPostStartLifecycleHooks(wksp *dw.DevWorkspaceTemplateSpec, containers []
 
 // processCommandsForPostStart processes a list of DevWorkspace commands
 // and generates a corev1.LifecycleHandler for the PostStart lifecycle hook.
-func processCommandsForPostStart(commands []dw.Command, postStartTimeout *int32) (*corev1.LifecycleHandler, error) {
-	if postStartTimeout == nil || *postStartTimeout == 0 {
+func processCommandsForPostStart(commands []dw.Command, postStartTimeout string) (*corev1.LifecycleHandler, error) {
+	if postStartTimeout == "" {
 		// use the fallback if no timeout propagated
 		return processCommandsWithoutTimeoutFallback(commands)
 	}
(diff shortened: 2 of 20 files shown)


pkg/library/status/check.go
```diff
@@ -157,22 +163,110 @@ func CheckPodEvents(pod *corev1.Pod, workspaceID string, ignoredEvents []string,
...
+	reKilledSigkill := regexp.MustCompile(`(\[postStart hook\] Commands forcefully killed by SIGKILL \(likely after --kill-after [^)]+?\)\. Exit code 137\.)`)
+
+	// matches: "[postStart hook] Commands failed with exit code ..." (for any other script-reported non-zero exit code)
+	reGenericFailedExitCode := regexp.MustCompile(`(\[postStart hook\] Commands failed with exit code \d+\.)`)
+
+	// regex to capture Kubelet's explicit message field content if it exists
+	reKubeletInternalMessage := regexp.MustCompile(`message:\s*"([^"]*)"`)
+
+	// regex to capture Kubelet's reported exit code for the hook command
+	reKubeletExitCode := regexp.MustCompile(`exited with (\d+):`)
```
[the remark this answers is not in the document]
akurinnoy: sure, this makes sense!

rohanKanojia (approved):

pkg/library/lifecycle/poststart.go
```diff
@@ -68,38 +76,111 @@ func AddPostStartLifecycleHooks(wksp *dw.DevWorkspaceTemplateSpec, containers []
...
+// processCommandsForPostStart processes a list of DevWorkspace commands
+// and generates a corev1.LifecycleHandler for the PostStart lifecycle hook.
+func processCommandsForPostStart(commands []dw.Command, postStartTimeout *int32) (*corev1.LifecycleHandler, error) {
+	if postStartTimeout == nil {
+		// The 'timeout' command treats 0 as "no timeout", so it is disabled by default.
+		defaultTimeout := int32(0)
+		postStartTimeout = &defaultTimeout
+	}
+
+	originalUserScript, err := buildUserScript(commands)
```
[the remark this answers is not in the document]
akurinnoy: Sure, this makes sense. I'll implement this fallback.

dkwon17:
@akurinnoy thank you for the update, but I get the fallback status message when starting `problematic-workspace`:
```
'Error creating DevWorkspace deployment: Detected unrecoverable event FailedPostStartHook: [postStart hook] failed with an unknown error (see pod events or container logs for more details)'
```

Events:
image

Is it expected?

pkg/library/lifecycle/poststart.go
```diff
@@ -68,6 +81,116 @@ func AddPostStartLifecycleHooks(wksp *dw.DevWorkspaceTemplateSpec, containers []
...
+				"/bin/sh",
+				"-c",
+				finalScriptForHook,
+			},
+		},
+	}
+	return handler, nil
+}
+
 // processCommandsForPostStart builds a lifecycle handler that runs the provided command(s)
```
dkwon17, line 194:
```suggestion
// processCommandsWithoutTimeoutFallback builds a lifecycle handler that runs the provided command(s)
```

pkg/library/lifecycle/poststart.go
```diff
@@ -68,6 +81,116 @@ func AddPostStartLifecycleHooks(wksp *dw.DevWorkspaceTemplateSpec, containers []
...
+  fi
+fi
+
+exit $exit_code
+`, timeoutSeconds, escapedUserScript)
+}
+
+// processCommandsForPostStart processes a list of DevWorkspace commands
+// and generates a corev1.LifecycleHandler for the PostStart lifecycle hook.
+func processCommandsForPostStart(commands []dw.Command, postStartTimeout *int32) (*corev1.LifecycleHandler, error) {
```
dkwon17, line 162: In general, could we define the caller functions above the helper functions? ie,
```
processCommandsForPostStart(...)
processCommandsWithoutTimeoutFallback(...)
buildUserScripot(...)
generateScriptWithTimeout(...)
```

pkg/library/lifecycle/poststart.go
akurinnoy: Done

pkg/library/lifecycle/poststart.go
akurinnoy: Done

akurinnoy:
@dkwon17 Hi,
 
 > Is it expected?
 
I couldn't find any proof that this is expected behavior in the Kubernetes docs, but it seems to be the case. I also encountered this behavior while I was testing this PR. I ran the `problematic-workspace`, and for the first few runs, I got the message with the exact exit code, but for subsequent runs, it was "failed with an unknown error."

pkg/library/lifecycle/poststart.go
```diff
@@ -99,9 +146,85 @@ func processCommandsForPostStart(commands []dw.Command) (*corev1.LifecycleHandle
...
+func buildUserScript(commands []dw.Command) (string, error) {
+	var commandScriptLines []string
+	for _, command := range commands {
+		execCmd := command.Exec
+		if execCmd == nil {
+			// Should be caught by earlier validation, but good to be safe
+			return "", fmt.Errorf("exec command is nil for command ID %s", command.Id)
+		}
+		if len(execCmd.Env) > 0 {
+			return "", fmt.Errorf("env vars in postStart command %s are unsupported", command.Id)
```
dkwon17, line 167: Maybe I'm missing something obvious, but why are env vars in the postStart command not supported?

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -189,6 +189,15 @@ type WorkspaceConfig struct {
...
 	CleanupCronJob *CleanupCronJobConfig `json:"cleanupCronJob,omitempty"`
+	// PostStartTimeout defines the maximum duration the PostStart hook can run
+	// before it is automatically failed. This timeout is used for the postStart lifecycle hook
+	// that is used to run commands in the workspace container. The timeout is specified in seconds.
+	// If not specified, the timeout is disabled (0 seconds).
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Type=integer
+	// +kubebuilder:validation:Format=int32
+	PostStartTimeout *int32 `json:"postStartTimeout,omitempty"`
```
dkwon17, line 200: On second thought, could we change the type from `*int32` to `string` so that it is consistent with the other timeouts?

https://github.com/devfile/devworkspace-operator/blob/cf4c38da728e05a8a3dc4d3c35a62eb700ddadac/apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go#L144-L153

pkg/library/lifecycle/poststart.go
akurinnoy: You're right, there's nothing here that prevents us from passing env variables to the postStart command. I fixed this issue.

dkwon17:
After more testing, I noticed that this DW fails when the postStartTimeout is set, but succeeds when it is not set:
```
kind: DevWorkspace
metadata:
  name: problematic-workspace
spec:
  started: false
  template:
    components:
      - name: tools
        container:
          image: quay.io/dkwon17/test:test-dir
          memoryLimit: "1Gi"
          memoryRequest: "512Mi"
          cpuRequest: "250m"
          cpuLimit: "1000m"
    commands:
      - id: test
        exec:
          workingDir: '/projects/test dir'
          component: tools
          commandLine: "mkdir mydir"
    events:
      postStart:
        - test
```

@akurinnoy are you able to reproduce the problem?

akurinnoy:
@dkwon17 I can reproduce this problem. It occurs because the `cd` command in the fallback is not being quoted correctly. Do you think we should change this behavior?

If so, I see two ways we could fix it:
1. Correctly quote the `cd ...` command in the `processCommandsWithoutTimeoutFallback`;
2. Unify script generation by getting rid of `processCommandsWithoutTimeout`.

How would you like to proceed?

dkwon17:
As discussed with @akurinnoy , the post start workingDir with spaces in the directory should be quoted like so for the workspace to start and run the post start command successfully: `/projects/'test dir'`

PR #1522, opened 2025-10-08
feat (postStart) : Allow debugging poststart failures with sleep by trapping errors

rohanKanojia:
### What does this PR do?
In https://github.com/devfile/devworkspace-operator/pull/425 `controller.devfile.io/debug-start` annotation was added to aid in debugging failed devworkspaces: [Debugging a failing workspace](https://github.com/devfile/devworkspace-operator/blob/f62e1c38e28a66268e4028b6c5e43508b6bb763f/docs/additional-configuration.adoc#debugging-a-failing-workspace)

We extend the use case of this annotation so that any failure in a postStart command results in the container sleeping for a specified number of seconds, as per the configured `progressTimeout`, allowing developers time to inspect the container state.

- Added `enableDebugStart` parameter to poststart methods.
- Injects `trap ... sleep` into postStart scripts when debug mode is enabled.
- Includes support for both timeout-wrapped (`postStartTimeout`) and non-timeout lifecycle scripts.

This feature improves the debuggability of DevWorkspaces where postStart hooks fail and would otherwise cause container crashes/restarts.

https://github.com/eclipse-che/che/issues/23404

#### With Changes
1. Checkout code changes added in this PR
2. Deploy DevWorkspace Operator Kubernetes/OpenShift cluster  `make docker && make install`
3. Create DevWorkspace that has a failing poststart command
```shell
oc apply -f - <<EOF
kind: DevWorkspace
metadata:
  name: failing-poststart-debug-dw
  annotations:
    controller.devfile.io/debug-start: "true"
spec:
  started: true
  template:
    components:
      - name: tools
        container:
          sourceMapping: /projects
          command: [ "tail" ]
          args: [ "-f", "/dev/null" ]
    commands:
      - id: failing-command
        exec:
          commandLine: ls idontexist
          component: tools
    events:
      postStart:
        - failing-command
EOF
```
4. After creating the DevWorkspace, observe its pod status. It should stay in `ContainerCreating` phase
```shell
oc get dw                                                                           
NAME                         DEVWORKSPACE ID             PHASE      INFO
failing-poststart-debug-dw   workspace55bf350cfb754260   Starting   Waiting for workspace deployment
oc get pods                                                                          
NAME                                         READY   STATUS              RESTARTS   AGE
workspace55bf350cfb754260-54749bf7c5-288vt   0/1     ContainerCreating   0          10s
```
5. You should be able to exec into the pod and see `/tmp/poststart-stderr.txt` to see root cause of failure:
```shell
oc get pods                                                                         
NAME                                         READY   STATUS              RESTARTS   AGE
workspace55bf350cfb754260-54749bf7c5-288vt   0/1     ContainerCreating   0          14s
kubectl exec -it workspace55bf350cfb754260-54749bf7c5-288vt -- /bin/bash            
bash-4.4$ cat /tmp/poststart-stderr.txt 
ls: cannot access 'idontexist': No such file or directory
```
6. Verify the sleep process is active in the container:
```shell
ps -ax | grep sleep
      2 ?        Ss     0:00 /bin/sh -c { cat << 'EOF' > /tmp/poststart.sh #!/bin/sh set -e trap 'echo "[postStart] failure encountered, sleep for debugging"; sleep 3600' ERR ls idontexist EOF chmod +x /tmp/poststart.sh /tmp/poststart.sh  } 1>/tmp/poststart-stdout.txt 2>/tmp/poststart-stderr.txt 
      7 ?        S      0:00 /usr/bin/coreutils --coreutils-prog-shebang=sleep /usr/bin/sleep 3600
     19 pts/0    S+     0:00 grep sleep
```

##### With PostStartTimeout Enabled

PostStart commands are processed slightly differently when `postStartTimeout` field is enabled in DevWorkspaceOperatorConfig. You can verify the above flow after enabling it:

```
oc patch devworkspaceoperatorconfig devworkspace-operator-config -n openshift-operators --type=merge -p '{"config": {"workspace": {"postStartTimeout": "5m"}}}'

## Repeat steps 3-6
```

#### Without Changes
With the current changes in the main, when we create a DevWorkspace with a failing post-start event. The pod immediately goes into `PostStartHookFailed` error and then `CrashLoopbackOff` error. It doesn't allow execution into it to view failure:

```shell

# Create DevWorkspace with failing poststart
oc apply -f - <<EOF
kind: DevWorkspace
metadata:
  name: failing-poststart-debug-dw
  annotations:
    controller.devfile.io/debug-start: "true"
spec:
  started: true
  template:
    components:
      - name: tools
        container:
          sourceMapping: /projects
          command: [ "tail" ]
          args: [ "-f", "/dev/null" ]
    commands:
      - id: failing-command
        exec:
          commandLine: ls idontexist
          component: tools
    events:
      postStart:
        - failing-command
EOF

oc get dw                                                                            
NAME                         DEVWORKSPACE ID             PHASE     INFO
failing-poststart-debug-dw   workspace7ad9a94285b94f7c   Failing   Error creating DevWorkspace deployment: Container tools has state [postStart hook] Commands failed (Kubelet reported exit code 2)

oc get pods                                                                          
NAME                                         READY   STATUS             RESTARTS      AGE
workspace7ad9a94285b94f7c-579896cc48-wmtrj   0/1     CrashLoopBackOff   1 (20s ago)   50s
kubectl exec -it workspace7ad9a94285b94f7c-579896cc48-wmtrj -- /bin/bash             
error: unable to upgrade connection: container not found ("tools")
```

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -324,7 +324,10 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	postStartDebugTrapSleepDuration := ""
+	deploymentReadyConditionMessage := "Waiting for workspace deployment"
 	if workspace.Annotations[constants.DevWorkspaceDebugStartAnnotation] == "true" {
+		deploymentReadyConditionMessage = "Debug mode: failed postStart commands will be trapped; inspect logs/exec to debug"
+		reconcileStatus.setConditionTrue(conditions.Started, "DevWorkspace is starting in debug mode")
 		postStartDebugTrapSleepDuration = workspace.Config.Workspace.ProgressTimeout
 	}
 	devfilePodAdditions, err := containerlib.GetKubeContainersFromDevfile(
@@ -494,7 +497,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if err := wsprovision.SyncDeploymentToCluster(workspace, allPodAdditions, serviceAcctName, clusterAPI); err != nil {
 		if shouldReturn, reconcileResult, reconcileErr := r.checkDWError(workspace, err, "Error creating DevWorkspace deployment", metrics.DetermineProvisioningFailureReason(err.Error()), reqLogger, &reconcileStatus); shouldReturn {
 			reqLogger.Info("Waiting on deployment to be ready")
-			reconcileStatus.setConditionFalse(conditions.DeploymentReady, "Waiting for workspace deployment")
+			reconcileStatus.setConditionFalse(conditions.DeploymentReady, deploymentReadyConditionMessage)
 			return reconcileResult, reconcileErr
 		}
 	}
--- a/test/e2e/pkg/tests/devworkspace_debug_poststart_tests.go
+++ b/test/e2e/pkg/tests/devworkspace_debug_poststart_tests.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2019-2025 Red Hat, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"fmt"
+	"strings"
+
+	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/devworkspace-operator/test/e2e/pkg/config"
+	"github.com/onsi/ginkgo/v2"
+)
+
+var _ = ginkgo.Describe("[DevWorkspace Debug Start Mode]", func() {
+	defer ginkgo.GinkgoRecover()
+
+	ginkgo.It("Wait DevWorkspace Webhook Server Pod", func() {
+		controllerLabel := "app.kubernetes.io/name=devworkspace-webhook-server"
+
+		deploy, err := config.AdminK8sClient.WaitForPodRunningByLabel(config.OperatorNamespace, controllerLabel)
+		if err != nil {
+			ginkgo.Fail(fmt.Sprintf("cannot get the DevWorkspace Webhook Server Pod status with label %s: %s", controllerLabel, err.Error()))
+			return
+		}
+
+		if !deploy {
+			ginkgo.Fail("Devworkspace webhook didn't start properly")
+		}
+	})
+
+	ginkgo.It("Add Debug DevWorkspace to cluster and wait starting status", func() {
+		commandResult, err := config.DevK8sClient.OcApplyWorkspace(config.DevWorkspaceNamespace, "test/resources/simple-devworkspace-debug-start-annotation.yaml")
+		if err != nil {
+			ginkgo.Fail(fmt.Sprintf("Failed to create DevWorkspace: %s %s", err.Error(), commandResult))
+			return
+		}
+
+		deploy, err := config.DevK8sClient.WaitDevWsStatus("code-latest-with-debug-start", config.DevWorkspaceNamespace, dw.DevWorkspaceStatusStarting)
+		if !deploy {
+			ginkgo.Fail(fmt.Sprintf("DevWorkspace didn't start properly. Error: %s", err))
+		}
+	})
+
+	ginkgo.It("Check DevWorkspace Conditions for Debug Start message", func() {
+		devWorkspaceStatus, err := config.DevK8sClient.GetDevWsStatus("code-latest-with-debug-start", config.DevWorkspaceNamespace)
+		if err != nil {
+			ginkgo.Fail(fmt.Sprintf("Failure in fetching DevWorkspace status. Error: %s", err))
+		}
+
+		expectedSubstring := "Debug mode: failed postStart commands will be trapped; inspect logs/exec to debug"
+
+		found := false
+		for _, cond := range devWorkspaceStatus.Conditions {
+			if cond.Message != "" && strings.Contains(cond.Message, expectedSubstring) {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			ginkgo.Fail(fmt.Sprintf(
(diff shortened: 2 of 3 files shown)


pkg/library/lifecycle/poststart.go
```diff
@@ -142,6 +150,15 @@ func processCommandsWithoutTimeoutFallback(commands []dw.Command) (*corev1.Lifec
 	}
 
 	joinedCommands := strings.Join(dwCommands, "\n")
+	if postStartFailureDebugSleepSeconds > 0 {
+		joinedCommands = fmt.Sprintf(`cat << 'EOF' > /tmp/poststart.sh
+#!/bin/sh
+%s
+EOF
+chmod +x /tmp/poststart.sh
```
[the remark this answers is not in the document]
rohanKanojia: I really apologize for my mistake 🙏 . This seems to be a leftover from previous attempts. 

I'll remove it.

rohanKanojia:
@tolusha : I've created these videos based on OpenShift 4.20 via clusterbot

## Scenario 1 : No Poststart Timeout Configured


https://github.com/user-attachments/assets/da37a743-34fa-4d9b-9dc8-94aeb9c64cd2


## Scenario 2: PostStart Timeout Configured 


https://github.com/user-attachments/assets/7af376db-7144-46e7-8c4a-21bd7920498b

pkg/library/lifecycle/poststart.go
```diff
@@ -239,5 +266,18 @@ else
...
+}
+
+func parsePostStartFailureDebugSleepDurationToSeconds(durationStr string) int {
+	if durationStr == "" {
+		return 0
+	}
+
+	d, err := time.ParseDuration(durationStr)
+	if err != nil {
+		return 0
```
dkwon17: Could we also log the error in this case?

```
log.Log.Error(err, ...
```

dkwon17:
> ```
> oc get dw                                                                           
> NAME                         DEVWORKSPACE ID             PHASE      INFO
> failing-poststart-debug-dw   workspace55bf350cfb754260   Starting   Waiting for workspace deployment
> ```
> 

@rohanKanojia it would be great if the INFO can state something similar to `Post start hook failed, sleeping for <progressTimeout>` to let the user know that the failure has been detected, but I think this is a bit difficult to detect, WDYT?

rohanKanojia:
@dkwon17 : You're right about this. It would definitely improve user experience if DevWorkspace `INFO` field indicated that postStart hook failed and the container is sleeping for debugging.

However, as you mentioned, detecting this specific state from DevWorkspace Controller may be tricky since the DevWorkspace Controller doesn't have direct visibility into the postStart execution inside the container. 

During the PostStart hook execution, we inject a `sleep` between the following DevWorkspace states:

```
(Starting) ---> (if postStart hook failed, inject sleep) ---> (Failed)
```

When the postStart hook fails, the container enters a sleep state before transitioning to Failed. During this period, the DevWorkspace does not re-enter the reconciliation loop until the sleep duration completes.

I haven't dig deep into it but here are some ways we might be able to handle this:

### Option 1 :  Patch DevWorkspace from within the pod via curl:
It might also be technically possible for the workspace pod to curl the Kubernetes API directly and patch its own DevWorkspace resource to signal this failure state. However, it depends on how the workspace's ServiceAccount is configured. 

⚠️  I'm not sure standard DevWorkspaces would have the necessary RBAC permissions to patch or update their own DevWorkspace resource.

### Option 2 : Surface failure via inspecting container state

It might be possible to add logic in controller to periodically check DevWorkspaces with `debug-start` annotation
- For each workspace with annotation, the controller would inspect associated pod.
- if `exec` is available and a sleep process injected by postStart hook is detected, the controller can immediately update workspace status to indicate that the postStart hook failed and container is sleeping for debugging.

pkg/library/lifecycle/poststart.go
```diff
@@ -139,6 +146,21 @@ func processCommandsWithoutTimeoutFallback(commands []dw.Command) (*corev1.Lifec
...
+	}
+
+	if postStartFailureDebugSleepSeconds > 0 && !hasErrTrapInUserScript {
+		debugTrap := fmt.Sprintf(`
+trap 'echo "[postStart] failure encountered, sleep for debugging"; sleep %d' ERR
+`, postStartFailureDebugSleepSeconds)
+		debugTrapLine := strings.ReplaceAll(strings.TrimSpace(debugTrap), "\n", " ")
+
+		dwCommands = append([]string{
+			"set -e",
```
dkwon17: What is the purpose of the `set -e` here?

rohanKanojia: `set -e` makes the script exit immediately if any command fails, so the poststart hook stops on errors instead of continuing silently. This ensures failures are caught and trigger the debug trap properly. I had added it as a safeguard to make sure postStart script fails fast.

I checked if we can remove `set -e` and tested with this DevWorkspace:
```yaml
kind: DevWorkspace
metadata:
  name: dig-fail-debug
  annotations:
    controller.devfile.io/debug-start: "true"
spec:
  started: true
  template:
    components:
      - name: tools
        container:
          image: quay.io/devfile/universal-developer-image:ubi9-latest
          mountSources: false
          command: ["tail"]
          args: ["-f", "/dev/null"]
    commands:
      - id: poststart-wrapper
        exec:
          component: tools
          commandLine: |
            echo "Start"
            wget 'https://wrongexample.com'  # should fail 
            echo "After failure"
    events:
      postStart:
        - poststart-wrapper
```

When I tested with a version that had `set -e` removed, I observed that the `echo "After failure"` command ran after `trap sleep`, and Kubernetes treated the hook as successful even though wget failed.
```shell
$ oc get pods
NAME                                         READY   STATUS    RESTARTS   AGE
workspacef6c14b383c1240e6-7f98c9ffb7-t6m6g   1/1     Running   0          7m54s
$ kubectl exec -it pod/workspacef6c14b383c1240e6-7f98c9ffb7-t6m6g -- /bin/bash
projects $ cat /tmp/poststart-stderr.txt
--2025-11-05 09:53:47--  https://wrongexample.com/
Resolving wrongexample.com (wrongexample.com)... 172.67.177.31, 104.21.83.138, 2606:4700:3033::6815:538a, ...
Connecting to wrongexample.com (wrongexample.com)|172.67.177.31|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
index.html: Permission denied

Cannot write to 'index.html' (No such file or directory).
projects $ cat /tmp/poststart-stdout.txt
Start
[postStart] failure encountered, sleep for debugging
After failure
projects $ exit
exit
```

When I tested with `set -e` , the pod transitioned to `PostStartHookError` error after debug `trap` sleep:
```shell
oc get pods
NAME                                         READY   STATUS               RESTARTS   AGE
workspace50b3197366ca4c18-84cfdf596c-8rfz5   0/1     PostStartHookError   0          7m43s
```

Issue #1525, opened 2025-10-15
Implement restore feature for DevWorkspaces


dkwon17:
### Description

This issue tracks the DWO-side implementation of a restore feature that works alongside https://github.com/devfile/devworkspace-operator/issues/1524.

### Additional context

Related issue https://github.com/eclipse-che/che/issues/23570.

High level prototype workflow from the Eclipse Che perspective mentioned in https://github.com/eclipse-che/che/issues/23570:

> Prototype Restore flow -
> 
> 1. The user logs into a secondary OpenShift cluster that has Dev Spaces installed.
> 2. The user creates a new workspace from the Git URL of the workspace that needs to be restored.
> 3. The user indicates that they wish for the workspace to be restored from a backup. (Right now that is a manual flow using modifications to the Devfile to inject an init container). Desired flow is for a selection in the dashboard to request restore.)
> 4. The workspace is created via the normal flow except that an init-container is run after PVC creation that pulls the backup image and copies the contents to ${PROJECTS_ROOT} before starting the workspace.

PR #1572, opened 2026-01-14
Restore workspace from backup

Allda:
### What does this PR do?
Add init container for workspace restoration

A new init container is added to the workspace deployment in case user choose to restore the workspace from backup.

By setting the workspace attribute "controller.devfile.io/restore-workspace" the controller sets a new init container instead of cloning data from git repository.

By default an automated path to restore image is used based on cluster settings. However user is capable overwrite that value using another attribute "controller.devfile.io/restore-source-image".

The restore container runs a wokspace-recovery.sh script that pull an image using oras an extract files to a /project directory.

https://github.com/devfile/devworkspace-operator/issues/1525

### Is it tested? How?
No automated tests are available in the first phase. I will add tests once I get the first approval that the concept is ok.

How to test:

- Configure a cluster to enable backups
- Create a new workspace and make changes in any of its files and save it.
- Stop the workspace `kubectl patch devworkspace restore-workspace-2 --type=merge -p '{"spec": {"started": false}}'
- Wait till it is stopped, and backup is executed for the workspace (verify the backup image exists in the registry)
- Delete a workspace from cluster `kubectl delete devworkspace restore-workspace-2`
- Add an attribute to the workspace CRD as shown below (`controller.devfile.io/restore-workspace`)
- Create a workspace
- Wait till it boots up and verify the changed file is present

```yaml
kind: DevWorkspace
metadata:
  labels:
    controller.devfile.io/creator: ""
  name: restore-workspace-2
spec:
  started: true
  routingClass: 'basic'
  template:
    attributes:
      controller.devfile.io/storage-type: common
      controller.devfile.io/restore-workspace: 'true'
```

### PR Checklist

### What's missing:
- integration with registry authentication ✅
- integration with built-in OCP registry ✅

### DIFF
--- a/controllers/backupcronjob/rbac.go
+++ b/controllers/backupcronjob/rbac.go
@@ -99,7 +99,7 @@ func (r *BackupCronJobReconciler) ensureImagePushRoleBinding(ctx context.Context
 		},
 		RoleRef: rbacv1.RoleRef{
 			Kind:     constants.RbacClusterRoleKind,
-			Name:     common.RegistryImagePullerRoleName(),
+			Name:     common.RegistryImageBuilderRoleName(),
 			APIGroup: "rbac.authorization.k8s.io",
 		},
 	}
--- a/pkg/common/naming.go
+++ b/pkg/common/naming.go
@@ -160,7 +160,11 @@ func WorkspaceSCCRolebindingName(sccName string) string {
 }
 
 func RegistryImagePullerRoleName() string {
-	return fmt.Sprintf("system:image-puller")
+	return "system:image-puller"
+}
+
+func RegistryImageBuilderRoleName() string {
+	return "system:image-builder"
 }
 
 func RegistryImagePullerRolebindingName(namespace string) string {


pkg/library/env/workspaceenv.go
```diff
@@ -49,6 +49,23 @@ func AddCommonEnvironmentVariables(podAdditions *v1alpha1.PodAdditions, clusterD
 	return nil
 }
 
+func GetErnvinmentVariablesForProjectRestore(workspace *common.DevWorkspaceWithConfig) []corev1.EnvVar {
```
tolusha:
```suggestion
func GetEnvironmentVariablesForProjectRestore(workspace *common.DevWorkspaceWithConfig) []corev1.EnvVar {
```

pkg/library/restore/restore.go
```diff
@@ -0,0 +1,129 @@
...
+func GetWorkspaceRestoreInitContainer(
+	ctx context.Context,
+	workspace *common.DevWorkspaceWithConfig,
+	k8sClient client.Client,
+	options Options,
+	log logr.Logger,
+) (*corev1.Container, error) {
+	wokrspaceTempplate := &workspace.Spec.Template
+	// Check if restore is requested via workspace attribute
+	if !wokrspaceTempplate.Attributes.Exists(constants.WorkspaceRestoreAttribute) {
```
dkwon17: IMHO I think it would be more readable if this check was done by the caller in `devworkspace_controller.go`

pkg/library/restore/restore.go
```diff
@@ -0,0 +1,129 @@
+//
+// Copyright (c) 2019-2025 Red Hat, Inc.
```
dkwon17:
```suggestion
// Copyright (c) 2019-2026 Red Hat, Inc.
```

pkg/constants/attributes.go
```diff
@@ -151,4 +151,24 @@ const (
 	// of a cloned project. If the bootstrap process is successful, project-clone will automatically remove this attribute
 	// from the DevWorkspace
 	BootstrapDevWorkspaceAttribute = "controller.devfile.io/bootstrap-devworkspace"
+
+	// WorkspaceRestoreAttribute defines whether workspace restore should be performed when creating a DevWorkspace.
```
dkwon17:
```suggestion
	// WorkspaceRestoreAttribute defines whether workspace restore should be performed for a DevWorkspace.
```

Small suggestion

pkg/library/restore/restore.go
Allda: I extracted the condition and moved it to the controller itself to make the code path easier to read. I also enhanced the logic to recognize the attribute value. The new test verifies that setting a value to `false` skips the restore container.

pkg/constants/attributes.go
Allda: Fixed

pkg/library/restore/restore.go
Allda: Fixed

controllers/workspace/devworkspace_controller_test.go
```diff
@@ -1024,6 +1027,204 @@ var _ = Describe("DevWorkspace Controller", func() {
...
+			workspacecontroller.SetupHttpClientsForTesting(getBasicTestHttpClient())
+		})
+
+		It("Restores workspace from backup with common PVC", func() {
+			config.SetGlobalConfigForTesting(&controllerv1alpha1.OperatorConfiguration{
+				Workspace: &controllerv1alpha1.WorkspaceConfig{
+					BackupCronJob: &controllerv1alpha1.BackupCronJobConfig{
+						Enable: ptr.To[bool](true),
+						Registry: &controllerv1alpha1.RegistryConfig{
+							Path: "localhost:5000",
```
Allda: @dkwon17 For the purposes of the integration tests, I need a valid backup container somewhere in a registry (the container could be empty with vely low size) to be able start and successfully execute the restore container. What would be the best place to upload it?

To verify the tests works I used my local registry, but that's not an option for running it outside of the localhost.

pkg/library/restore/restore.go
```diff
@@ -0,0 +1,131 @@
...
+// GetWorkspaceRestoreInitContainer creates an init container that restores workspace data from a backup image.
+// The restore container uses the existing workspace-recovery.sh script to extract backup content.
+func GetWorkspaceRestoreInitContainer(
+	ctx context.Context,
+	workspace *common.DevWorkspaceWithConfig,
+	k8sClient client.Client,
+	options Options,
+	log logr.Logger,
+) (*corev1.Container, error) {
+	wokrspaceTempplate := &workspace.Spec.Template
```
rohanKanojia:
```suggestion
	workspaceTemplate := &workspace.Spec.Template
```

pkg/library/restore/restore.go
```diff
@@ -0,0 +1,131 @@
...
+	Resources  *corev1.ResourceRequirements
+	Env        []corev1.EnvVar
+}
+
+func IsWorkspaceRestoreRequested(workspace *dw.DevWorkspaceTemplateSpec) bool {
+	if !workspace.Attributes.Exists(constants.WorkspaceRestoreAttribute) {
+		return false
+	}
+	enableRecovery := workspace.Attributes.GetBoolean(constants.WorkspaceRestoreAttribute, nil)
+	return enableRecovery
```
rohanKanojia: perhaps it can be inlined?
```suggestion
	return  workspace.Attributes.GetBoolean(constants.WorkspaceRestoreAttribute, nil)
```

controllers/workspace/devworkspace_controller_test.go
```diff
@@ -1024,6 +1027,204 @@ var _ = Describe("DevWorkspace Controller", func() {
...
+						},
+					},
+				},
+			})
+			defer config.SetGlobalConfigForTesting(nil)
+			By("Reading DevWorkspace with restore configuration from testdata file")
+			createDevWorkspace(devWorkspaceName, "restore-workspace-disabled.yaml")
```
rohanKanojia: There is no file named `restore-workspace-disabled.yaml` in `testdata/`

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -376,6 +379,16 @@ type ProjectCloneConfig struct {
 	Env []corev1.EnvVar `json:"env,omitempty"`
 }
 
+type RestoreConfig struct {
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
```
rohanKanojia:
```suggestion
	// ImagePullPolicy configures the imagePullPolicy for the restore clone container.
	// If undefined, the general setting .config.workspace.imagePullPolicy is used instead.
	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
```

controllers/workspace/devworkspace_controller_test.go
ibuziuk: have just created https://quay.io/repository/devfile/project-backup-test?tab=info

pkg/library/restore/restore.go
```diff
@@ -0,0 +1,131 @@
...
+
+// GetWorkspaceRestoreInitContainer creates an init container that restores workspace data from a backup image.
+// The restore container uses the existing workspace-recovery.sh script to extract backup content.
+func GetWorkspaceRestoreInitContainer(
+	ctx context.Context,
+	workspace *common.DevWorkspaceWithConfig,
+	k8sClient client.Client,
+	options Options,
+	log logr.Logger,
+) (*corev1.Container, error) {
```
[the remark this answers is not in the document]
Allda: Fixed

project-backup/workspace-recovery.sh
```diff
@@ -92,12 +93,31 @@ backup() {
...
+    extra_args=( ${ORAS_EXTRA_ARGS} )
+    oras_args+=("${extra_args[@]}")
+  fi
+
+  # Pull the backup tarball from the OCI registry using oras and extract it
+  oras "${oras_args[@]}"
+  mkdir /tmp/extracted-backup
+  tar -xzvf /tmp/devworkspace-backup.tar.gz -C /tmp/extracted-backup
+
+  cp -r /tmp/extracted-backup/* "$PROJECTS_ROOT"
```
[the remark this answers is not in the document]
Allda: One clarifying question... How is the scenario you described handled for the clone container? Does it clone a repo again or does it skip the cloning?

pkg/library/restore/restore.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := dwResources.ValidateResources(resources); err != nil {
+		return nil, nil, fmt.Errorf("invalid resources for workspace restore container: %w", err)
+	}
+	volumeMounts := []corev1.VolumeMount{
+		{
+			Name:      devfileConstants.ProjectsVolumeName,
+			MountPath: constants.DefaultProjectsSourcesRoot,
+		},
+	}
+	registryAuthSecret, err := secrets.HandleRegistryAuthSecret(ctx, k8sClient, workspace.DevWorkspace, workspace.Config, "", scheme, log)
```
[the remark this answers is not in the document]
Allda: What do you suggest? `null` value or a different solution?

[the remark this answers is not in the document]
dkwon17: I notice for the restore init container, we are using `HandleRegistryAuthSecret` to only get the secret, and not copy. But in the backup cron job, we use it for both get and copy: https://github.com/devfile/devworkspace-operator/blob/8fcc0911070b908a0ecba52969f04aaee3949f86/controllers/backupcronjob/backupcronjob_controller.go#L348

I was thinking about creating:
```
func GetNamespaceRegistryAuthSecret(ctx context.Context, c client.Client, workspace *dw.DevWorkspace,
	dwOperatorConfig *controllerv1alpha1.OperatorConfiguration, operatorConfigNamespace string, scheme *runtime.Scheme, log logr.Logger,
) (*corev1.Secret, error) {
	return HandleRegistryAuthSecret(ctx, c, workspace, dwOperatorConfig, "", scheme, log)
}
```

inside pkg/secrets/backup.go, and use it here: 
```suggestion
	registryAuthSecret, err := secrets.GetNamespaceRegistryAuthSecret(ctx, k8sClient, workspace.DevWorkspace, workspace.Config, scheme, log)
```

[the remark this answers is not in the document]
Allda: I changed the functions based on the suggestion.

pkg/secrets/backup.go
```diff
@@ -0,0 +1,124 @@
...
+			Name:      constants.DevWorkspaceBackupAuthSecretName,
+			Namespace: workspace.Namespace,
+			Labels: map[string]string{
+				constants.DevWorkspaceIDLabel:          workspace.Status.DevWorkspaceId,
+				constants.DevWorkspaceWatchSecretLabel: "true",
+			},
+		},
+		Data: sourceSecret.Data,
+		Type: sourceSecret.Type,
+	}
```
akurinnoy: Here seems to be another race condition in secret copying. If multiple workspaces are restoring simultaneously in the same namespace, they will race to create/update the same secret name. Does this make sense?

```suggestion
	// Construct the desired secret state
	desiredSecret := &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      constants.DevWorkspaceBackupAuthSecretName + "=" + workspace.Status.DevWorkspaceId,
			Namespace: workspace.Namespace,
			Labels: map[string]string{
				constants.DevWorkspaceIDLabel:          workspace.Status.DevWorkspaceId,
				constants.DevWorkspaceWatchSecretLabel: "true",
			},
		},
		Data: sourceSecret.Data,
		Type: sourceSecret.Type,
	}
```

dkwon17:
> However, the DevWorkspace is not able to come out of Starting phase:

@rohanKanojia could you please share the DevWorkspace yaml that you used for the restore workspace?

pkg/secrets/backup.go
dkwon17: I don't think we need an auth secret for each workspace, especially since the secret has the same data, how about just removing the `constants.DevWorkspaceIDLabel: workspace.Status.DevWorkspaceId` label from the secret?

controllers/workspace/devworkspace_controller.go
```diff
@@ -353,21 +354,51 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	}
-	// Add init container to clone projects
-	projectCloneOptions := projects.Options{
-		Image:     workspace.Config.Workspace.ProjectCloneConfig.Image,
-		Env:       env.GetEnvironmentVariablesForProjectClone(workspace),
-		Resources: workspace.Config.Workspace.ProjectCloneConfig.Resources,
-	}
-	if workspace.Config.Workspace.ProjectCloneConfig.ImagePullPolicy != "" {
-		projectCloneOptions.PullPolicy = config.Workspace.ProjectCloneConfig.ImagePullPolicy
+	if restore.IsWorkspaceRestoreRequested(&workspace.Spec.Template) {
```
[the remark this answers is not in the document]
rohanKanojia: While I was testing yesterday, I keep bumping into https://github.com/devfile/devworkspace-operator/pull/1572#issuecomment-3835458677

DevWorkspace was able to come in Running state when I patched DevWorkspace to remove the restore attribute.

I'll check again if it is some problem with my setup.

controllers/workspace/devworkspace_controller.go
```diff
@@ -353,21 +354,51 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			return r.failWorkspace(workspace, fmt.Sprintf("Failed to set up workspace-restore init container: %s", err), metrics.ReasonInfrastructureFailure, reqLogger, &reconcileStatus), nil
+		} else if workspaceRestore != nil {
+			devfilePodAdditions.InitContainers = append([]corev1.Container{*workspaceRestore}, devfilePodAdditions.InitContainers...)
+			if registryAuthSecret != nil {
+				// Add the registry auth secret volume
+				devfilePodAdditions.Volumes = append(devfilePodAdditions.Volumes, corev1.Volume{
+					Name: constants.RegistryAuthVolumeName,
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							SecretName: registryAuthSecret.Name, // You may want to make this configurable
```
dkwon17:
```suggestion
							SecretName: registryAuthSecret.Name,
							DefaultMode: pointer.Int32(0640),
```

This should fix the issue described in https://github.com/devfile/devworkspace-operator/pull/1572#issuecomment-3835458677.

Basically I noticed that if the DefaultMode isn't set, OpenShift will automatically set it to a value when the deployment is created on the cluster. But on the next reconcile, the DefaultMode is still not being set in the registry auth secret.

Because of this, there will be another update to the deployment on the cluster to remove the DefaultMode.   OpenShift then adds it back, causing an endless loop. 

Setting the DefaultMode to a specific value here, fixes the issue.

@Allda could we apply this change?

pkg/secrets/backup.go
dkwon17: @Allda I suggest removing `constants.DevWorkspaceIDLabel` :
```suggestion
	// Construct the desired secret state
	desiredSecret := &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      constants.DevWorkspaceBackupAuthSecretName,
			Namespace: workspace.Namespace,
			Labels: map[string]string{
				constants.DevWorkspaceWatchSecretLabel: "true",
			},
		},
		Data: sourceSecret.Data,
		Type: sourceSecret.Type,
	}
```

dkwon17:
Added new commit to fix the rolebinding issue: [460cd25](https://github.com/devfile/devworkspace-operator/pull/1572/commits/460cd250eecd907c5d187869e45f5542d33d9556)

and rebased. 

I will merge this PR shortly

PR #1578, opened 2026-01-30
Sort automount configmaps and secrets to ensure deterministic ordering

dkwon17:
### What does this PR do?
Sorts automount configmaps and secrets so that they are ordered the same way for each devworkspace reconciliation. Sorting is done, since otherwise, there is no guarantee on the configmap/secret ordering returned by `api.Client.List`. This can cause the ordering to differ between DevWorkspace reconciliations, causing the problem described in https://github.com/devfile/devworkspace-operator/pull/1572#issuecomment-3808022169.

A similar change was made in the past with git credentials: https://github.com/devfile/devworkspace-operator/commit/e17e7999a13306225102c9cba981532d9507c335.

In this PR, we apply the same idea to the automount secrets and configmaps.

https://github.com/devfile/devworkspace-operator/pull/1572#issuecomment-3808022169

### Is it tested? How?

It can be tested by running:
```
git clone https://github.com/dkwon17/test-automount-ordering
cd test-automount-ordering
./create-secrets.sh
./test-ordering.sh
./cleanup.sh
```

The output should include:
```
✅ CONFIRMED: Volumes are sorted alphabetically (fix is working!)
```

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller_test.go
+++ b/controllers/workspace/devworkspace_controller_test.go
@@ -577,6 +577,264 @@ var _ = Describe("DevWorkspace Controller", func() {
 			}
 		})
 
+		It("Sorts automount secrets in consistent order", func() {
+			By("Creating automount secrets in non-sorted order")
+			secretZ := generateSecret("secret-z", corev1.SecretTypeOpaque)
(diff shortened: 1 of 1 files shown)


rohanKanojia:
I tested with the abovementioned steps and PR seems to be working as expected ✔️

PR #1614, opened 2026-04-10
fix: fall back to canonical backup auth secret name on restore

akurinnoy:
### What does this PR do?

When a workspace is backed up to a private registry, `CopySecret` copies the registry auth secret into the workspace namespace using a hardcoded name (`devworkspace-backup-registry-auth`), regardless of the name configured in the `DWOC`. During restore, `GetNamespaceRegistryAuthSecret` looks for the configured name, doesn't find it, and returns `nil` - so the `workspace-restore` init container has no credentials and crashes with `CrashLoopBackOff`.

This PR adds a fallback lookup in `HandleRegistryAuthSecret`: when called from the restore path (`operatorConfigNamespace == ""`), if the configured secret name is not found in the workspace namespace, it also
  checks for the canonical `devworkspace-backup-registry-auth` name that `CopySecret` always uses.

**Note**: This is a surgical workaround. See https://github.com/devfile/devworkspace-operator/pull/1615 for a root-cause fix that changes the secret naming in `CopySecret` instead.

[CRW-10591](https://redhat.atlassian.net/browse/CRW-10591)

### Is it tested? How?

New unit tests added.

### PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * During restore, the system now prefers the canonical workspace backup auth secret name when looking up registry credentials, with clearer logging and preserved fallback behavior.

* **Tests**
  * Added focused tests covering restore-path secret retrieval, the case of a missing workspace secret, and propagation of retrieval errors to ensure robust secret-handling.

### DIFF
--- a/pkg/secrets/backup.go
+++ b/pkg/secrets/backup.go
@@ -54,42 +54,23 @@ func HandleRegistryAuthSecret(ctx context.Context, c client.Client, workspace *d
 		return nil, nil
 	}
 
-	// First check the workspace namespace for the secret
+	// First check the workspace namespace for the secret.
+	// CopySecret always writes the copy under DevWorkspaceBackupAuthSecretName,
+	// so look it up by that canonical name.
 	registryAuthSecret := &corev1.Secret{}
 	err := c.Get(ctx, client.ObjectKey{
-		Name:      secretName,
+		Name:      constants.DevWorkspaceBackupAuthSecretName,
 		Namespace: workspace.Namespace}, registryAuthSecret)
 	if err == nil {
-		log.Info("Successfully retrieved registry auth secret for backup from workspace namespace", "secretName", secretName)
+		log.Info("Successfully retrieved registry auth secret for backup from workspace namespace",
+			"secretName", constants.DevWorkspaceBackupAuthSecretName)
 		return registryAuthSecret, nil
 	}
 	if client.IgnoreNotFound(err) != nil {
 		return nil, err
 	}
 	// If we don't provide an operator namespace, don't attempt to look there.
-	// However, CopySecret always writes the secret under DevWorkspaceBackupAuthSecretName,
-	// regardless of the configured name. If the configured name differs, attempt a fallback
-	// lookup under the canonical name so that the restore path can find it.
 	if operatorConfigNamespace == "" {
-		if secretName == constants.DevWorkspaceBackupAuthSecretName {
-			// The configured name IS the canonical name; it was already looked up and not
-			// found above, so there is nothing else to try.
-			return nil, nil
-		}
-		fallbackSecret := &corev1.Secret{}
-		fallbackErr := c.Get(ctx, client.ObjectKey{
-			Name:      constants.DevWorkspaceBackupAuthSecretName,
-			Namespace: workspace.Namespace,
-		}, fallbackSecret)
-		if fallbackErr == nil {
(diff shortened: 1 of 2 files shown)


pkg/secrets/backup.go
dkwon17: Since workspace namespace's secret name is always constants.DevWorkspaceBackupAuthSecretName, should we do the following? :

```suggestion
		Name:      constants.DevWorkspaceBackupAuthSecretName ,
```

rohanKanojia:
@akurinnoy : Do we need this now that #1615 seems to be ready?

dkwon17:
@rohanKanojia I'm leaning towards this PR, as opposed to https://github.com/devfile/devworkspace-operator/pull/1615, with one reason being, the secret name is predictable to the user. 

Another reason, is, for https://github.com/devfile/devworkspace-operator/pull/1615 if the admin repeatedly changes the secret name overtime, multiple secrets can get copied to user namespaces overtime (although, they will still get cleaned up if the owner DevWorkspaces is deleted). Although, this is a very unlikely an edge case.

Do you have a strong preference between https://github.com/devfile/devworkspace-operator/pull/1614 and https://github.com/devfile/devworkspace-operator/pull/1615 ?

rohanKanojia:
@dkwon17 : I don't have any objection with proceeding with this PR.

pkg/secrets/backup.go
```diff
@@ -66,8 +66,30 @@ func HandleRegistryAuthSecret(ctx context.Context, c client.Client, workspace *d
...
+		}
+		fallbackSecret := &corev1.Secret{}
+		fallbackErr := c.Get(ctx, client.ObjectKey{
+			Name:      constants.DevWorkspaceBackupAuthSecretName,
+			Namespace: workspace.Namespace,
+		}, fallbackSecret)
+		if fallbackErr == nil {
+			log.Info("Registry auth secret not found under configured name in workspace namespace; using canonical backup auth secret as fallback",
+				"configuredName", secretName,
+				"fallbackName", constants.DevWorkspaceBackupAuthSecretName)
```
rohanKanojia: Maybe in logs we can also specify the workspace namespace where secret wasn't found?
```suggestion
				"fallbackName", constants.DevWorkspaceBackupAuthSecretName,
				"namespace", workspace.Namespace)
```

pkg/secrets/backup_test.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+	}
+	return e.Client.Get(ctx, key, obj, opts...)
+}
+
+// Ensure errorOnNameClient satisfies client.Client at compile time.
+var _ client.Client = &errorOnNameClient{}
+
+// secretGR is the GroupResource for secrets, used when constructing test errors.
+var secretGR = schema.GroupResource{Group: "", Resource: "secrets"} //nolint:unused
```
rohanKanojia: Looks unused. Is there any reason to keep this? I see you've specifically added exclusion for this.
```suggestion
```

pkg/secrets/backup.go
akurinnoy: you're right, done

pkg/secrets/backup_test.go
akurinnoy: fixed, thanks!

rohanKanojia (approved):

dkwon17:
@akurinnoy thank you, a small suggestion, throughout this PR instead of using the term "canonical name", could we change it to "predefined name"?

PR #1615, opened 2026-04-10
fix: restore init container uses configured registry auth secret name

akurinnoy:
### What does this PR do?

When a workspace is backed up to a private registry, `CopySecret` hardcodes the name of the copied auth secret as `devworkspace-backup-registry-auth`, regardless of the name configured in the DWOC. During restore, `GetNamespaceRegistryAuthSecret` looks only in the workspace namespace for the configured name and passes an empty operator namespace - so it never finds anything, mounts no credentials, and the `workspace-restore` init container crashes with `CrashLoopBackOff`.

This PR fixes the root cause:
  - `CopySecret` now takes a `secretName `parameter and uses it instead of the hardcoded constant, so the copy in the
  workspace namespace gets the correct name
  - `HandleRegistryAuthSecret` passes the configured name through to `CopySecret`
  - `GetNamespaceRegistryAuthSecret` gains an `operatorConfigNamespace` parameter so the restore path can look up the operator namespace and copy the secret with the right name

Old copies named `devworkspace-backup-registry-auth` created by previous backups become orphaned but are owned by the workspace and will be garbage-collected when the workspace is deleted.

**Note**: This is the root-cause fix. See https://github.com/devfile/devworkspace-operator/pull/1614 for a more conservative fallback-only approach that avoids changing the `CopySecret` signature.

[CRW-10591](https://redhat.atlassian.net/browse/CRW-10591)

### Is it tested? How?

New unit tests added.

### PR Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved namespace handling for workspace restore and backup flows to reliably locate registry auth secrets.
  * Backup now uses the configured auth secret name instead of a legacy hardcoded name.

* **Tests**
  * Added comprehensive tests covering secret lookup, fallback behavior, and copy semantics.

* **Bug Fixes**
  * Fixed error handling when operator namespace lookup fails.


pkg/secrets/backup_test.go
```diff
@@ -0,0 +1,252 @@
...
+			legacySecret := &corev1.Secret{}
+			err = fakeClient.Get(ctx, types.NamespacedName{
+				Name:      "devworkspace-backup-registry-auth",
+				Namespace: workspaceNamespace,
+			}, legacySecret)
+			Expect(client.IgnoreNotFound(err)).To(Succeed())
+			Expect(err).To(HaveOccurred(), "legacy hardcoded name must not be created")
+		})
+
+		It("returns nil, nil when operator namespace is given but secret is in neither namespace", func() {
```
rohanKanojia, line 197: nit, test description a bit inaccurate.

rohanKanojia (approved):

akurinnoy:
closed in favour of https://github.com/devfile/devworkspace-operator/pull/1614

PR #1645, opened 2026-06-15
fix: Use merged DWOC config for HTTP client certificate initialization

tolusha:
### What does this PR do?
This PR changes the HTTP client initialization to use the merged DevWorkspaceOperatorConfig instead of only the global config.

https://github.com/eclipse-che/che/issues/23870
https://github.com/eclipse-che/che-operator/pull/2137

### Is it tested? How?

Followed https://github.com/eclipse-che/che-operator/pull/2137

### PR Checklist

## Release Notes

* **Refactor**
  * Improved HTTP client handling by replacing global, one-time setup with a routing-aware client holder that creates the appropriate HTTP and health-check clients as needed.
  * Updated health checks to use the routing-specific health-check client instead of a shared instance.
  * Enhanced TLS/certificate behavior for outbound HTTPS by loading certificates from the configured source when building clients, with error logging if certificate retrieval fails.

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -144,9 +144,6 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reqLogger = reqLogger.WithValues(constants.DevWorkspaceIDLoggerKey, workspace.Status.DevWorkspaceId)
 	reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
 
-	// Inject ca certificates to the http client, if the certificates configmap is created and defined in the config.
-	InjectCertificates(r.Client, r.Log)
-
 	// Check if the DevWorkspaceRouting instance is marked to be deleted, which is
 	// indicated by the deletion timestamp being set.
 	if workspace.GetDeletionTimestamp() != nil {
@@ -264,7 +261,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		WorkspaceNamespace:          workspace.Namespace,
 		Context:                     ctx,
 		K8sClient:                   r.Client,
-		HttpClient:                  httpClient,
+		HttpClient:                  httpClientsHolder.GetHttpClient(config.Routing),
 		DefaultResourceRequirements: workspace.Config.Workspace.DefaultContainerResources,
 	}
 
@@ -788,7 +785,7 @@ func (r *DevWorkspaceReconciler) getWorkspaceId(ctx context.Context, workspace *
 }
 
 func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	setupHttpClients(mgr.GetClient(), mgr.GetLogger())
+	httpClientsHolder = NewDefaultHttpsClientHolder(mgr.GetClient(), mgr.GetLogger())
 
 	maxConcurrentReconciles, err := wkspConfig.GetMaxConcurrentReconciles()
 	if err != nil {
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -141,10 +141,10 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	workspace.DevWorkspace = rawWorkspace
 	workspace.Config = config
 
+	httpClientsHolder.ConfigureHttpClients(config.Routing)
+
 	reqLogger = reqLogger.WithValues(constants.DevWorkspaceIDLoggerKey, workspace.Status.DevWorkspaceId)
 	reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
-
-	httpClientsHolder.ConfigureHttpClients(config.Routing)
 
 	// Check if the DevWorkspaceRouting instance is marked to be deleted, which is
 	// indicated by the deletion timestamp being set.
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -141,7 +141,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	workspace.DevWorkspace = rawWorkspace
 	workspace.Config = config
 
-	httpClientsHolder.ConfigureHttpClients(config.Routing)
+	httpClientsHolder.ConfigureHttpClients(ctx, config.Routing)
 
 	reqLogger = reqLogger.WithValues(constants.DevWorkspaceIDLoggerKey, workspace.Status.DevWorkspaceId)
 	reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -144,6 +144,8 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reqLogger = reqLogger.WithValues(constants.DevWorkspaceIDLoggerKey, workspace.Status.DevWorkspaceId)
 	reqLogger.Info("Reconciling Workspace", "resolvedConfig", configString)
 
+	httpClientsHolder.ConfigureHttpClients(config.Routing)
+
 	// Check if the DevWorkspaceRouting instance is marked to be deleted, which is
 	// indicated by the deletion timestamp being set.
 	if workspace.GetDeletionTimestamp() != nil {
@@ -261,7 +263,7 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		WorkspaceNamespace:          workspace.Namespace,
 		Context:                     ctx,
 		K8sClient:                   r.Client,
-		HttpClient:                  httpClientsHolder.GetHttpClient(config.Routing),
+		HttpClient:                  httpClientsHolder.GetHttpClient(),
 		DefaultResourceRequirements: workspace.Config.Workspace.DefaultContainerResources,
 	}
 
@@ -785,7 +787,7 @@ func (r *DevWorkspaceReconciler) getWorkspaceId(ctx context.Context, workspace *
 }
 
 func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	httpClientsHolder = NewDefaultHttpsClientHolder(mgr.GetClient(), mgr.GetLogger())
+	httpClientsHolder = NewDefaultHttpClientsHolder(mgr.GetClient(), mgr.GetLogger())
 
 	maxConcurrentReconciles, err := wkspConfig.GetMaxConcurrentReconciles()
 	if err != nil {
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -787,7 +787,10 @@ func (r *DevWorkspaceReconciler) getWorkspaceId(ctx context.Context, workspace *
 }
 
 func (r *DevWorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	httpClientsHolder = NewDefaultHttpClientsHolder(mgr.GetClient(), mgr.GetLogger())
+	err := SetupHttpClients(mgr.GetClient(), mgr.GetLogger())
+	if err != nil {
+		return err
+	}
 
 	maxConcurrentReconciles, err := wkspConfig.GetMaxConcurrentReconciles()
 	if err != nil {
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -40,9 +40,6 @@ type HttpClientsHolder interface {
 type DefaultHttpsClientHolder struct {
 	k8s    client.Client
 	logger logr.Logger
-
-	client                *http.Client
-	healthCheckHttpClient *http.Client
 }
 
 var httpClientsHolder HttpClientsHolder
@@ -110,7 +107,9 @@ func (h *DefaultHttpsClientHolder) getCaCertPool(routingConfig *controller.Routi
 		}
 
 		for _, certsPem := range certs {
-			caCertPool.AppendCertsFromPEM([]byte(certsPem))
+			if !caCertPool.AppendCertsFromPEM([]byte(certsPem)) {
+				h.logger.Info("Warning: failed to parse one or more certificates from ConfigMap")
+			}
 		}
 
 		return caCertPool
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -39,7 +39,12 @@ var httpClientsHolder HttpClientsHolder
 
 type HttpClientsHolder interface {
 	GetHttpClient() *http.Client
+
+	// GetHealthCheckHttpClient returns an HTTP client that skips TLS verification.
+	// This client MUST only be used for workspace health/readiness checks, not for
+	// fetching external content or making security-sensitive requests.
 	GetHealthCheckHttpClient() *http.Client
+
 	ConfigureHttpClients(context.Context, *controller.RoutingConfig)
 }
 
@@ -55,25 +60,23 @@ type DefaultHttpClientsHolder struct {
 	lastProxyConfig    *controller.Proxy
 	lastCertsCMVersion string
 
-	defaultCertPool *x509.CertPool
+	systemCertPool *x509.CertPool
 }
 
-func NewDefaultHttpClientsHolder(k8s client.Client, logger logr.Logger) *DefaultHttpClientsHolder {
-	defaultCertPool, err := x509.SystemCertPool()
+func SetupHttpClients(k8s client.Client, logger logr.Logger) error {
+	systemCertPool, err := x509.SystemCertPool()
 	if err != nil {
-		logger.Error(err, "Failed to load system cert pool")
-		defaultCertPool = x509.NewCertPool()
+		return fmt.Errorf("failed to load system cert pool: %w", err)
 	}
 
-	clientsHolder := &DefaultHttpClientsHolder{
-		k8s:             k8s,
-		logger:          logger,
-		defaultCertPool: defaultCertPool,
+	httpClientsHolder = &DefaultHttpClientsHolder{
+		k8s:            k8s,
+		logger:         logger,
+		systemCertPool: systemCertPool,
 	}
+	httpClientsHolder.ConfigureHttpClients(context.Background(), config.GetGlobalConfig().Routing)
 
-	clientsHolder.ConfigureHttpClients(context.Background(), config.GetGlobalConfig().Routing)
-
-	return clientsHolder
+	return nil
 }
 
 func (h *DefaultHttpClientsHolder) GetHttpClient() *http.Client {
@@ -98,10 +101,14 @@ func (h *DefaultHttpClientsHolder) ConfigureHttpClients(ctx context.Context, rou
 		if routingConfig.ProxyConfig != nil {
 			newProxyConfig = routingConfig.ProxyConfig
 		}
+
 		if routingConfig.TLSCertificateConfigmapRef != nil {
 			certsCM, err := h.readCertCM(ctx, routingConfig.TLSCertificateConfigmapRef)
 			if err != nil {
 				h.logger.Error(err, "Failed to read TLS certificate ConfigMap")
+
+				// certsCM == nil,
+				// http clients will be rebuilt with a system cert pool, not an issue at all
 			}
 
 			newCertsCM = certsCM
@@ -128,8 +135,8 @@ func (h *DefaultHttpClientsHolder) ConfigureHttpClients(ctx context.Context, rou
 }
 
 func (h *DefaultHttpClientsHolder) shouldRebuildClients(newProxyConfig *controller.Proxy, newCertsCM *corev1.ConfigMap) (bool, bool) {
-	defer h.mu.RUnlock()
 	h.mu.RLock()
+	defer h.mu.RUnlock()
 
 	// Always rebuild if clients haven't been initialized yet
 	if h.client == nil || h.healthCheckHttpClient == nil {
@@ -174,6 +181,7 @@ func (h *DefaultHttpClientsHolder) buildNewClients(
 
 		newClient = &http.Client{
 			Transport: transport,
+			Timeout:   5 * time.Second,
 		}
 	}
 
@@ -249,7 +257,7 @@ func (h *DefaultHttpClientsHolder) getCaCertPool(cm *corev1.ConfigMap) *x509.Cer
 		return nil
 	}
 
-	caCertPool := h.defaultCertPool.Clone()
+	caCertPool := h.systemCertPool.Clone()
 
 	for _, certsPem := range cm.Data {
 		if !caCertPool.AppendCertsFromPEM([]byte(certsPem)) {
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -102,7 +102,6 @@ func (h *DefaultHttpClientsHolder) ConfigureHttpClients(ctx context.Context, rou
 			certsCM, err := h.readCertCM(ctx, routingConfig.TLSCertificateConfigmapRef)
 			if err != nil {
 				h.logger.Error(err, "Failed to read TLS certificate ConfigMap")
-				return
 			}
 
 			newCertsCM = certsCM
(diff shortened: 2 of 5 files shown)


tolusha:

**Available commands**:

tolusha (commented):
## Review Summary

Comprehensive review completed across standard, deep, and impact analysis. The refactoring is well-executed and solves the linked P1 issue. The design improvement from global mutable state to interface-based factory is sound. Below are observations for consideration.

---

## 1. Consider caching HTTP clients to avoid per-reconcile overhead

**File:** `controllers/workspace/http.go:53-62`

Each call to `GetHttpClient()` creates a brand-new `http.Client` with a fresh `http.Transport`, which means:
- No TCP connection pooling across reconcile iterations
- System cert pool loaded from disk on every call (`x509.SystemCertPool()`)
- ConfigMap fetched from K8s API on every call (via `getCaCertPool` → `readCertificates`)

While the old code also fetched the ConfigMap on every reconcile, creating new transports means losing connection reuse. Consider caching clients keyed by routing config identity, or at minimum caching the cert pool to reduce I/O overhead under high concurrency.

**Context:** The impact review notes that since workspace provisioning is a state machine (not a tight loop), this is unlikely to be a practical performance issue. Each reconcile fetches at most a few URIs.

---

## 2. Thread context through readCertificates for proper cancellation

**File:** `controllers/workspace/http.go:131`

`readCertificates` uses `context.Background()` instead of the reconcile context. This means the ConfigMap API call won't be cancelled if the reconcile is cancelled or times out.

```suggestion
func (h *DefaultHttpsClientHolder) readCertificates(ctx context.Context, routingConfig *controller.RoutingConfig) (map[string]string, bool) {
	if routingConfig == nil || routingConfig.TLSCertificateConfigmapRef == nil {
		return nil, false
	}

	configmapRef := routingConfig.TLSCertificateConfigmapRef

	namespacedName := types.NamespacedName{
		Name:      configmapRef.Name,
		Namespace: configmapRef.Namespace,
	}

	configMap := &corev1.ConfigMap{}
	err := h.k8s.Get(ctx, namespacedName, configMap)
	if err != nil {
		h.logger.Error(err, "Failed to read configmap with certificates")
		return nil, false
	}

	return configMap.Data, true
}
```

This would require updating the signature and threading the context from `getCaCertPool` and `GetHttpClient`. The old code had the same issue, so this is not a regression - future improvement.

---

## 3. Add timeout to GetHttpClient to prevent indefinite hangs

**File:** `controllers/workspace/http.go:53-62`

`GetHttpClient()` creates an HTTP client with no timeout set. If the devfile parent URI endpoint hangs, the reconcile goroutine will block indefinitely (until controller-runtime's context timeout, if one is set upstream).

Consider adding a reasonable timeout:

```suggestion
return &http.Client{
	Transport: transport,
	Timeout:   30 * time.Second,  // or appropriate value based on expected devfile fetch time
}
```

**Note:** The old code had the same issue, so this is a pre-existing concern, not a regression.

---

## 4. Consider renaming "Holder" to "Factory" for clarity

**File:** `controllers/workspace/http.go:35-46`

`HttpClientsHolder` doesn't convey what it does - it's a factory that creates clients on demand, not a holder that stores them. The old code held clients; the new code creates them.

Consider renaming to better express the actual behavior:
- `HttpClientFactory` or `HttpClientBuilder` for the interface
- `DefaultHttpClientFactory` for the implementation

Similarly, `DefaultHttpsClientHolder` says "HTTPS" but `GetHealthCheckHttpClient` creates an HTTP client with `InsecureSkipVerify: true` - the "HTTPS" in the name is somewhat misleading.

**Also:** Go convention prefers `HTTPClientsHolder` (all caps for acronyms) over `HttpClientsHolder`, though this is a minor style point.

---

## 5. Add unit tests for DefaultHttpsClientHolder core logic

**General observation**

The test changes update the test infrastructure (`SetupHttpClientsForTesting`) but don't add new test cases for the production code. Given that TLS certificate handling is the core of this PR and the linked issue is P1, consider adding tests for:

- `getCaCertPool` with valid certificates from a mock ConfigMap
- `getCaCertPool` when ConfigMap read fails
- `getProxyFunc` with various proxy configurations
- `readCertificates` with nil routing config
- Interaction between proxy and TLS config

At minimum, a test verifying that `getCaCertPool` returns a pool containing certs from a mock ConfigMap would provide regression protection.

---

## 6. Certificate loading failures could be surfaced better for debugging

**General observation (from impact review)**

When certificate ConfigMap read fails or certs are invalid, the system degrades gracefully to system certs (correct behavior), but these failures are only logged, not surfaced as workspace events or status conditions.

If a workspace fails to start because of a TLS issue with a parent devfile URI, the operator logs the cert loading failure but the workspace status will show a generic HTTP error from the devfile fetch, not a "certificate configuration failed" message.

Consider recording an event or setting a status condition when cert loading fails to make debugging easier for cluster admins. Not blocking - the old code had the same behavior.

---

## Positive Feedback

- Clean interface extraction enables proper test mocking - `TestHttpsClientHolder` is well-structured
- Removing `InjectCertificates` from the reconcile hot path and replacing it with per-client certificate loading is a cleaner design
- The new code is more correct than the old - `getCaCertPool` is idempotent, whereas the old `injectCertificates` would accumulate certs across calls
- Nil-safety checks are consistent and correct
- Error logging uses structured `logr.Logger` with appropriate levels
- Removing the unnecessary pointer for `NamespacedName` shows attention to Go idioms

tolusha (commented):
## Code Review Summary

This PR successfully enables dynamic TLS certificate loading from DevWorkspaceOperatorConfig, addressing the core issue from eclipse-che/che#23870. The refactoring to use an interface-based holder pattern is a solid architectural improvement.

However, there are **two critical issues** that need to be addressed before merge:

### Critical Issues

1. **Proxy Regression**: The code disables environment-variable-based proxy (`HTTP_PROXY`, `HTTPS_PROXY`) when no explicit `ProxyConfig` is set in DWOC. This will break enterprise/air-gapped deployments.

2. **Stale Connections**: After TLS reconfiguration, idle connections in the connection pool continue using old TLS settings, delaying the very fix this PR delivers.

3. **Race Condition**: Concurrent reconciliations mutate shared HTTP transport fields without synchronization.

See inline comments for details and suggested fixes.

### Positive Notes

- Clean interface pattern for testability
- Proper cert pool cloning avoids mutation
- Resource leak fix in status.go
- Efficient system cert loading

Full review available: standard, deep, and impact analyses completed.

tolusha (commented):
## Review Summary

Comprehensive review completed across standard, deep, and impact analysis. The refactoring is well-executed and solves the linked P1 issue. The design improvement from global mutable state to routing-aware client management with smart caching is sound. Several operational and code quality improvements are recommended.

---

## 1. Add timeout to regular HTTP client

**File:** `controllers/workspace/http.go:138`

The regular HTTP client is created without a timeout field. Requests to external devfile servers could block indefinitely if the server is unresponsive, potentially tying up a reconciler thread.

The health check client correctly has a 500ms timeout. Consider adding a reasonable default timeout (e.g., 30s) to the regular client:

```suggestion
			newClient = &http.Client{
				Transport: transport,
				Timeout:   30 * time.Second,
			}
```

---

## 2. Propagate reconciliation context to ConfigMap read

**File:** `controllers/workspace/http.go:221`

The `readCertCM` method uses `context.Background()` instead of accepting and using the reconciliation context. This means:
- Cancellation signals from operator shutdown won't propagate
- OpenTelemetry traces won't connect to the reconciliation span  
- No deadline protection from the reconciliation context

Consider accepting a `context.Context` parameter:

```go
func (h *DefaultHttpClientsHolder) readCertCM(ctx context.Context, cmReference *controller.ConfigmapReference) *corev1.ConfigMap {
	// ...
	err := h.k8s.Get(ctx, namespacedName, configMap)
	// ...
}
```

And update the caller in `ConfigureHttpClients` to pass a context.

---

## 3. Add unit tests for ConfigureHttpClients logic

**File:** `controllers/workspace/http_test.go`

The test file only adds a mock implementation. The core caching/rebuild logic in `ConfigureHttpClients` has zero test coverage. At minimum, test:

- Client rebuild when cert ConfigMap ResourceVersion changes
- Client rebuild when proxy config changes  
- No rebuild when config is unchanged (caching works)
- Behavior when ConfigMap read fails
- Concurrent access patterns

This is the highest-value testing improvement for this PR.

---

## 4. Add observability for certificate loading failures

**File:** General

When the certificate ConfigMap cannot be read (deleted, RBAC issue, API timeout), the only signal is an operator log line. Operators troubleshooting "workspace can't fetch parent devfile" issues would need to check DWO logs.

Consider:
- Recording a Kubernetes event on the workspace or DevWorkspaceOperatorConfig
- Adding a Prometheus metric counter for cert loading failures
- Setting a condition on the operator's own status

This would significantly improve operational debuggability.

---

## 5. Document nil proxy function behavior

**File:** `controllers/workspace/http.go:180`

When no proxy config is set, `getProxyFunc` returns `nil`. Setting `transport.Proxy = nil` means `http.Transport` will use `http.ProxyFromEnvironment` (the Go default). This is correct but implicit.

Add a brief comment explaining this behavior:

```go
func (h *DefaultHttpClientsHolder) getProxyFunc(proxyConfig *controller.Proxy) func(*http.Request) (*url.URL, error) {
	if proxyConfig != nil {
		// ... existing code
	}
	
	// Returning nil preserves http.ProxyFromEnvironment behavior (Go default)
	return nil
}
```

---

## 6. Consider defer pattern for lock safety

**File:** `controllers/workspace/http.go:ConfigureHttpClients`

The method uses manual `RLock()/RUnlock()` and `Lock()/Unlock()` without defer. While safe in the current simple paths, future modifications adding early returns could introduce deadlocks.

The split-lock design makes full-method defer impractical, but consider using immediately-invoked functions with defer for each critical section:

```go
func() {
    h.mu.RLock()
    defer h.mu.RUnlock()
    // read-only operations
}()
```

This is defensive programming for future maintainability.

---

## 7. TOCTOU gap optimization (low priority)

**File:** `controllers/workspace/http.go:120-143`

Between `RUnlock()` (after comparing configs) and `Lock()` (before swapping clients), another concurrent reconciler could detect the same config change and build duplicate clients. Both would produce equivalent clients and the last write wins, so this is correct but causes redundant work under high concurrency.

Low-priority optimization: re-check the condition under the write lock before swapping.

---

## 8. Document reflect.DeepEqual safety

**File:** `controllers/workspace/http.go:120`

Using `reflect.DeepEqual` on the `Proxy` struct with pointer fields works correctly, but is fragile if the struct gains uncomparable types (maps, slices of interfaces) in the future.

Consider adding a comment explaining why `DeepEqual` is safe here and that `DeepCopy()` is used for storage to prevent aliasing.

---

## Positive Highlights

- **Clean interface design**: `HttpClientsHolder` with clear separation of concerns
- **Smart caching**: ResourceVersion tracking for certs, DeepCopy for proxy config
- **System cert pool optimization**: Pre-loaded once, cloned per-use  
- **Resource leak fix**: Response body cleanup in `status.go:checkServerStatus`
- **Thread-safe implementation**: Correct `sync.RWMutex` usage pattern

---

## Already Addressed

CodeRabbit's earlier feedback was addressed in commit de8aa1d:
- Unused struct fields removed
- `AppendCertsFromPEM` return value now checked with warning log

tolusha (commented):
## Review Summary

I've completed a comprehensive review (standard, deep, and impact analysis) of this PR. The core design is solid - the interface extraction and cached client approach with rebuild-on-change is well-implemented. However, there are a few important items to address:

**Critical Issues:**
- `context.Background()` usage bypasses reconcile context cancellation
- Silent certificate degradation when ConfigMap read fails

**Warnings:**
- Undocumented 5-second timeout behavioral change
- Missing unit tests for the dynamic reload logic

See inline comments for specific suggestions. Overall verdict: **Request Changes** - the issues are straightforward to fix.

---
*Generated by [ok-pr-review](https://github.com/ok-pr-review) via Claude Code*

tolusha (commented):
## Comprehensive Review Summary

I've completed a multi-stage review (summary, standard, deep, and impact analysis) of this PR. The changes correctly address the linked issue (eclipse-che/che#23870) by enabling per-workspace TLS certificate configuration. However, there are **2 critical issues** that must be fixed before merge:

### Critical Issues

1. **Nil HTTP client initialization bug** - When the operator starts without routing config (common default), clients remain nil, causing panics on first use
2. **Proxy config silently dropped on cert read failure** - If certificate ConfigMap read fails, proxy changes are also lost

### Additional Findings

- **Shared client across namespaces**: Concurrent reconcilers can interfere with each other's TLS trust (low probability but worth documenting)
- **Performance**: ConfigMap fetched on every reconcile even when rebuild not needed
- **Missing unit tests**: 120+ lines of new logic with zero test coverage

See inline comments for specific fixes. The core design (HttpClientsHolder interface, three-phase configure pattern) is sound - these are implementation bugs that can be addressed quickly.

---
📊 **Review Statistics**: 4 files changed (+253/-67), all reviewed
🔍 **Analysis performed**: Pre-flight summary, standard review, deep review, impact review
✅ **Solves issue**: Yes - enables merged DWOC config for HTTP client certs

controllers/workspace/http.go
```diff
@@ -32,87 +35,241 @@ import (
...
+	}
+
+	if certsCMVersion != h.lastCertsCMVersion {
+		return true, false
+	}
+
+	return false, false
+}
+
+func (h *DefaultHttpClientsHolder) buildNewClients(
```
tolusha, line 162: **Critical**: When cert ConfigMap read fails, the early return here prevents proxy config changes from being applied. The proxy update is silently lost.

Consider removing the early return and continuing with nil `newCertsCM` to allow proxy changes:
```go
if err != nil {
    h.logger.Error(err, "Failed to read TLS certificate ConfigMap, will use previous cert config")
    // Don't return - still apply proxy changes if they exist
} else {
    newCertsCM = certsCM
}
```

controllers/workspace/http.go
```diff
@@ -32,87 +35,241 @@ import (
...
+func (h *DefaultHttpClientsHolder) getProxyFunc(proxyConfig *controller.Proxy) func(*http.Request) (*url.URL, error) {
+	if proxyConfig != nil {
+		proxyConf := httpproxy.Config{}
+		if proxyConfig.HttpProxy != nil {
+			proxyConf.HTTPProxy = *proxyConfig.HttpProxy
+		}
+		if proxyConfig.HttpsProxy != nil {
+			proxyConf.HTTPSProxy = *proxyConfig.HttpsProxy
+		}
+		if proxyConfig.NoProxy != nil {
```
tolusha, line 243: Consider adding a comment explaining why TLS verification is skipped for health checks:
```go
// Health check endpoints may use self-signed certificates. We skip verification
// here since we only care about reachability, not identity verification.
InsecureSkipVerify: true,
```

controllers/workspace/http.go
```diff
@@ -32,87 +35,241 @@ import (
...
+
+	if !reflect.DeepEqual(newProxyConfig, h.lastProxyConfig) {
+		return true, true
+	}
+
+	if certsCMVersion != h.lastCertsCMVersion {
+		return true, false
+	}
+
+	return false, false
```
tolusha, line 159: **Performance**: `readCertCM` makes a K8s API call on every reconciliation, even when no rebuild is needed. At scale (hundreds of workspaces), this adds measurable API server load.

Consider checking proxy config first (no API call), then only fetching ConfigMap if cert check is actually required. Alternatively, watch the ConfigMap via an informer.

tolusha (commented):
## Review Summary

Thorough analysis conducted across standard code review, deep design review, and system impact review.

**Verdict**: 🔄 Request Changes

### What's Good ✅
- Clean interface extraction (`HttpClientsHolder`) enables test mocking - significant improvement over global variables
- Smart rebuild-on-change optimization using ResourceVersion tracking
- Response body leak fix in status.go
- Comprehensive test coverage (9 test cases covering edge cases)
- Defensive cert pool cloning prevents cross-contamination

### Must Address Before Merge 🔴
See inline comment #1: Transient ConfigMap failure silently strips custom certificates, which could cause workspace startup failures during brief K8s API server hiccups.

### Strongly Recommended ⚠️
- Add timeout to main HTTP client (see inline comment #2)
- Optimize ConfigMap reads to reduce API server load (see inline comment #3)
- Document health check client security contract (see inline comment #5)

### Test Coverage Gaps
- No test for proxy removal (ProxyConfig != nil → nil)
- No concurrent access test with -race detector

See inline comments for detailed findings.

controllers/workspace/http.go
```diff
@@ -32,87 +35,245 @@ import (
...
+	}
+
+	certsCMVersion := ""
+	if newCertsCM != nil {
+		certsCMVersion = newCertsCM.ResourceVersion
+	}
+
+	if certsCMVersion != h.lastCertsCMVersion {
+		return true, false
+	}
```
tolusha, line 157: 🔴 **Critical: Transient ConfigMap failure silently strips custom certificates**

When \`readCertCM\` fails (e.g., transient API server error), execution continues with \`newCertsCM = nil\`. This causes \`shouldRebuildClients\` to detect a version change (empty string vs stored version), triggering a rebuild with \`getCaCertPool(nil)\` which returns nil - effectively removing all custom certificates from the HTTP client until the next successful reconciliation.

**Operational impact**: A brief K8s API server hiccup during reconciliation could cause workspace startup failures for devfiles referencing custom-certificate endpoints.

Consider preserving the existing client when the ConfigMap read fails:

\`\`\`suggestion
if err != nil {
    h.logger.Error(err, "Failed to read TLS certificate ConfigMap")
    return // keep existing clients unchanged
}
\`\`\`

controllers/workspace/http.go
```diff
@@ -32,87 +35,245 @@ import (
...
+		h.lastCertsCMVersion = ""
 	}
-	InjectCertificates(k8s, logger)
 }
 
-func InjectCertificates(k8s client.Client, logger logr.Logger) {
-	if certs, ok := readCertificates(k8s, logger); ok {
-		for _, certsPem := range certs {
-			injectCertificates([]byte(certsPem), httpClient.Transport.(*http.Transport), logger)
+func (h *DefaultHttpClientsHolder) getProxyFunc(proxyConfig *controller.Proxy) func(*http.Request) (*url.URL, error) {
```
tolusha, line 234: ⚠️ **Warning: No timeout on main HTTP client**

The health check client has a 500ms timeout (line 240), but the main HTTP client has none. If a remote endpoint (e.g., serving a parent devfile) hangs, the reconciliation goroutine blocks indefinitely, consuming one of the `maxConcurrentReconciles` slots. Under pathological conditions (multiple hanging endpoints), this could exhaust all reconcile workers.

Consider adding a reasonable timeout:

```suggestion
        newClient = &http.Client{
            Transport: transport,
            Timeout:   30 * time.Second,
        }
```

controllers/workspace/http.go
```diff
@@ -32,87 +35,245 @@ import (
...
+	if !reflect.DeepEqual(newProxyConfig, h.lastProxyConfig) {
+		return true, true
+	}
+
+	certsCMVersion := ""
+	if newCertsCM != nil {
+		certsCMVersion = newCertsCM.ResourceVersion
+	}
+
+	if certsCMVersion != h.lastCertsCMVersion {
```
tolusha, line 155: ⚠️ **Performance: ConfigMap read on every reconciliation**

`readCertCM` makes a Kubernetes API call on every reconciliation when `TLSCertificateConfigmapRef` is set, even if the proxy config hasn't changed. At scale with hundreds of concurrent workspaces, this adds unnecessary API server load.

Consider checking proxy config first (cheap `reflect.DeepEqual`), and only fetching the ConfigMap if needed. Alternatively, use a ConfigMap informer/watch to cache the data.

controllers/workspace/http.go
```diff
@@ -32,87 +35,245 @@ import (
...
 		}
 
-		proxyFunc := func(req *http.Request) (*url.URL, error) {
-			return proxyConf.ProxyFunc()(req.URL)
+		newClient = &http.Client{
+			Transport: transport,
 		}
-		transport.Proxy = proxyFunc
+	}
+
```
tolusha, line 187: **Suggestion: Inconsistent mutex lock/defer ordering**

The defer statement appears before the lock acquisition here, while `GetHttpClient` and `GetHealthCheckHttpClient` use the conventional order (lock, then defer). Both are functionally correct, but the inconsistency could confuse future maintainers.

```suggestion
	h.mu.RLock()
	defer h.mu.RUnlock()
```

controllers/workspace/http.go
```diff
@@ -32,87 +35,245 @@ import (
...
+		logger:          logger,
+		defaultCertPool: defaultCertPool,
+	}
 
-	if globalConfig.Routing != nil && globalConfig.Routing.ProxyConfig != nil {
-		proxyConf := httpproxy.Config{}
-		if globalConfig.Routing.ProxyConfig.HttpProxy != nil {
-			proxyConf.HTTPProxy = *globalConfig.Routing.ProxyConfig.HttpProxy
+	clientsHolder.ConfigureHttpClients(context.Background(), config.GetGlobalConfig().Routing)
+
```
tolusha: **Suggestion: Document health check client security contract**

The health check client skips TLS verification (`InsecureSkipVerify: true`). While this is intentional for reachability checks, the interface doesn't document this security contract. If any code path accidentally uses the health check client for non-health-check requests, TLS verification would be bypassed.

Consider adding a comment to the interface method:

```suggestion
	GetHttpClient() *http.Client
	// GetHealthCheckHttpClient returns an HTTP client that skips TLS verification.
	// This client MUST only be used for workspace health/readiness checks, not for
	// fetching external content or making security-sensitive requests.
	GetHealthCheckHttpClient() *http.Client
```

tolusha:

Review is complete. Please check the review comments below.

tolusha (commented):
## Code Review Summary

Completed comprehensive review (standard + deep + impact analysis) of this PR.

### Does This PR Solve the Issue?
✅ Yes - enables DWO to use merged DevWorkspaceOperatorConfig for TLS certificate initialization.

### Critical Issues Identified

1. **Transient ConfigMap failure silently strips custom certificates** - When `readCertCM` fails, the client rebuilds without custom certs, breaking TLS trust cluster-wide
2. **TOCTOU race between shouldRebuildClients and setNewClients** - Concurrent reconciliations can both rebuild clients
3. **Shared HTTP client with per-workspace config** - Last reconciled workspace's config wins globally

### Positive Highlights
- Clean `HttpClientsHolder` interface design
- Excellent test coverage (8 test cases)
- Fixed pre-existing response body leak in `status.go`
- Smart rebuild optimization logic

### Recommendations
See inline comments for detailed findings and suggested fixes.

**Verdict**: 🔄 Request Changes - Critical #1 directly undermines the PR's purpose. Once fixed, this is a solid, well-tested change.

---
*Review generated by Claude Code using /ok-pr-review suite*

controllers/workspace/http.go
```diff
@@ -32,87 +35,253 @@ import (
...
+		certsCMVersion = newCertsCM.ResourceVersion
+	}
+
+	if certsCMVersion != h.lastCertsCMVersion {
+		return true, false
+	}
+
+	return false, false
+}
+
```
tolusha, line 161: **Critical: Transient ConfigMap failure silently strips custom certificates**

When `readCertCM` fails (transient API error, RBAC issue), `newCertsCM` becomes `nil`. The `shouldRebuildClients` method then detects a "change" and rebuilds the HTTP client with `getCaCertPool(nil)`, stripping all custom CA certificates.

A single transient Kubernetes API error will break TLS trust cluster-wide for all subsequent workspace operations.

**Suggested fix**: On `readCertCM` failure, return early without rebuilding clients. Only rebuild when we successfully read a ConfigMap with a different ResourceVersion.

```go
if routingConfig.TLSCertificateConfigmapRef != nil {
    certsCM, err := h.readCertCM(ctx, routingConfig.TLSCertificateConfigmapRef)
    if err != nil {
        h.logger.Error(err, "Failed to read TLS certificate ConfigMap")
        // Preserve existing HTTP clients on transient failure
        return
    }
    newCertsCM = certsCM
}
```

controllers/workspace/http.go
```diff
@@ -32,87 +35,253 @@ import (
...
+	}
+}
+
+func (h *DefaultHttpClientsHolder) shouldRebuildClients(newProxyConfig *controller.Proxy, newCertsCM *corev1.ConfigMap) (bool, bool) {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+
+	// Always rebuild if clients haven't been initialized yet
+	if h.client == nil || h.healthCheckHttpClient == nil {
+		return true, true
```
tolusha, line 143: **Race condition: TOCTOU gap between shouldRebuildClients and setNewClients**

`ConfigureHttpClients` calls `shouldRebuildClients` (RLock), then `buildNewClients` (no lock), then `setNewClients` (Lock). Between releasing the RLock and acquiring the Lock, another concurrent reconciliation could also pass the check and both would rebuild clients.

**Suggested fix**: Use a single Lock for the entire `ConfigureHttpClients` method to make check-and-set atomic.

controllers/workspace/http.go
```diff
@@ -32,87 +35,253 @@ import (
...
+	if newCertsCM != nil {
+		certsCMVersion = newCertsCM.ResourceVersion
+	}
+
+	if certsCMVersion != h.lastCertsCMVersion {
+		return true, false
+	}
+
+	return false, false
+}
```
tolusha, line 160: **Comment contradicts actual behavior**

The comment says "not an issue at all" but this IS an issue - the operator was configured to trust custom certificates. Rebuilding without them on transient failure breaks the trust chain.

Consider updating this comment after fixing the transient failure handling.

controllers/workspace/devworkspace_controller.go
```diff
@@ -141,12 +141,11 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	workspace.DevWorkspace = rawWorkspace
 	workspace.Config = config
 
+	httpClientsHolder.ConfigureHttpClients(ctx, config.Routing)
```
tolusha, line 144: **System-level concern: Shared HTTP client with per-workspace config**

The `httpClientsHolder` is shared across ALL concurrent reconciliations. If different workspaces have different merged `TLSCertificateConfigmapRef` values, the "last reconciled workspace wins" for HTTP client configuration.

**Needs verification**: Does DWOC merging always produce identical routing config regardless of workspace? Or can per-namespace overrides create different cert references?

If configs can differ, consider scoping the HTTP client holder per-namespace or per-config-hash.

PR #1651, opened 2026-06-17
fix(home): skip persistent home init container for ephemeral workspaces

tolusha:
When a DevWorkspaceOperatorConfig defines an init-persistent-home entry in workspace.initContainers, the config merge logic only checked whether persistent home was globally enabled (PersistUserHomeEnabled) but not whether the specific workspace actually needs it. Ephemeral workspaces passed this check, causing an init container to be injected that references a non-existent persistent-home volume and potentially has no image — resulting in an invalid Deployment.

Add a NeedsPersistentHomeDirectory check to also skip the config-level init container when the workspace storage strategy does not support home persistence (e.g. ephemeral storage).

Assisted-by: Claude Code

https://github.com/eclipse-che/che/issues/23884

### Is it tested? How?

TODO

### PR Checklist

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -417,6 +417,11 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 					reqLogger.Info("Skipping init-persistent-home container: persistent home is disabled")
 					continue
 				}
+				// Skip if workspace does not need persistent home (e.g. ephemeral storage)
+				if !home.NeedsPersistentHomeDirectory(workspace) {
+					reqLogger.Info("Skipping init-persistent-home container: workspace does not need persistent home directory")
+					continue
+				}
 				// Skip if init container is explicitly disabled
 				if disableHomeInit {
 					reqLogger.Info("Skipping init-persistent-home container: DisableInitContainer is true")


tolusha:

**Available commands**:

tolusha (commented):
## Review Summary

This is a well-crafted bug fix that correctly addresses the issue. The fix is minimal, surgical, and reuses existing abstractions appropriately. The main concern is test coverage, which you've already acknowledged as TODO.

### Positive Observations
- The fix correctly delegates to `NeedsPersistentHomeDirectory()` rather than duplicating logic
- The log message is distinct and diagnostic-friendly  
- Commit message thoroughly explains root cause and fix
- Design is sound - aligns the config-level guard logic with the Devfile-level path

### Key Findings
I've added inline comments with specific recommendations. The most important is adding test coverage for the ephemeral storage case.

---

**Review conducted using [ok-pr-review](https://github.com/ok-sh/ok-pr-review)**

controllers/workspace/devworkspace_controller.go
```diff
@@ -417,6 +417,11 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 					reqLogger.Info("Skipping init-persistent-home container: persistent home is disabled")
 					continue
 				}
+				// Skip if workspace does not need persistent home (e.g. ephemeral storage)
+				if !home.NeedsPersistentHomeDirectory(workspace) {
```
tolusha, line 421: ### Add test coverage for ephemeral storage case

The PR description marks testing as "TODO". This is a bug fix in the main reconciliation loop - the exact kind of change that benefits most from a unit test proving ephemeral workspaces skip the init container.

The existing `pkg/library/home/persistentHome_test.go` tests `NeedsPersistentHomeDirectory` itself, but there's no test verifying the controller's filtering logic applies it correctly. More importantly, none of the 10 test fixtures in `testdata/persistent-home/` test the ephemeral storage branch (`storageStrategySupportsPersistentHome` returning false).

**Recommendation:** Add a test fixture like `noop-if-ephemeral-storage.yaml` to `pkg/library/home/testdata/persistent-home/` with:
- Workspace that has `controller.devfile.io/storage-type: ephemeral` in its attributes
- Persistent home enabled in config
- Expected output: workspace unchanged (no home volume added)

This is low-effort (YAML file only) and closes the only real coverage gap.

tolusha, line 421: ### Minor note - redundant check

`home.NeedsPersistentHomeDirectory(workspace)` internally calls `home.PersistUserHomeEnabled(workspace)` (see `persistentHome.go:96`), which is already checked on line 418 above.

This is harmless - the function is cheap and the double-check doesn't affect correctness. The current approach preserves distinct log messages for each skip reason, which aids debugging. Not a defect, just worth noting for future maintainers.

PR #1650, opened 2026-06-17
fix: initialization HTTP clients with merged `DevWorkspaceOperatorConfig`

tolusha:
### What does this PR do?
This PR changes the HTTP client initialization to use the merged DevWorkspaceOperatorConfig instead of only the global config.

### What issues does this PR fix or reference?
https://github.com/eclipse-che/che/issues/23870
https://github.com/eclipse-che/che-operator/pull/2137

### Is it tested? How?

Followed https://github.com/eclipse-che/che-operator/pull/2137

### PR Checklist

- [x] E2E tests pass (when PR is ready, comment `/test v8-devworkspace-operator-e2e, v8-che-happy-path` to trigger)
    - [x] `v8-devworkspace-operator-e2e`: DevWorkspace e2e test
    - [x] `v8-che-happy-path`: Happy path for verification integration with Che

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health check reliability with proper response body cleanup.

* **Tests**
  * Added comprehensive unit tests for HTTP client creation, caching, and concurrent access.

* **Chores**
  * Refactored HTTP client initialization to use a thread-safe factory pattern with improved routing configuration support.

PR #1656, opened 2026-06-23
fix: preserve default proxy settings

tolusha:
### What does this PR do?
Preserve default proxy settings

```
{"level":"info","ts":"2026-06-23T07:31:51Z","logger":"controllers.DevWorkspace","msg":"Failed to check server status: Get \"https://eclipse-che.apps.che-airgap-v20.crw-qe.com/happypath-dev/quarkus-api-example-wgx6/3100/healthz\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)","Request.Namespace":"happypath-dev-che","Request.Name":"quarkus-api-example-wgx6","devworkspace_id":"workspace5e9f8620bd9f4773"}
```
fix https://github.com/devfile/devworkspace-operator/pull/1650

### Is it tested? How?
Manually on disconnected cluster

### PR Checklist

## Bug Fixes
* Fixed HTTP proxy handling so proxy behavior is computed once from global configuration and applied consistently, without overriding environment/default proxy behavior when no explicit proxy is set.
* Improved health-check reliability and performance by using a cached, stable health-check HTTP client that isn’t impacted by per-workspace routing changes.

## Tests
* Refactored and expanded HTTP client tests, including new health-check coverage, stronger caching assertions, and expanded concurrency/TLS validation.

### DIFF
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -122,7 +122,10 @@ func (h *DefaultHttpClientsFactory) GetHttpClient(ctx context.Context, routingCo
 
 func (h *DefaultHttpClientsFactory) createHttpClient(routingConfig *controller.RoutingConfig, certsCM *corev1.ConfigMap) *http.Client {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
-	transport.Proxy = h.getProxyFunc(routingConfig)
+	proxyFunc := h.getProxyFunc(routingConfig)
+	if proxyFunc != nil {
+		transport.Proxy = proxyFunc
+	}
 	transport.TLSClientConfig = &tls.Config{
 		RootCAs: h.getCaCertPool(certsCM),
 	}
@@ -199,7 +202,10 @@ func (h *DefaultHttpClientsFactory) shouldCreateHealthCheckHttpClient(routingCon
 
 func (h *DefaultHttpClientsFactory) createHealthCheckHttpClient(routingConfig *controller.RoutingConfig) *http.Client {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
-	transport.Proxy = h.getProxyFunc(routingConfig)
+	proxyFunc := h.getProxyFunc(routingConfig)
+	if proxyFunc != nil {
+		transport.Proxy = proxyFunc
+	}
 	transport.TLSClientConfig = &tls.Config{
 		InsecureSkipVerify: true,
 	}
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -124,6 +124,7 @@ func (h *DefaultHttpClientsFactory) createHttpClient(routingConfig *controller.R
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	proxyFunc := h.getProxyFunc(routingConfig)
 	if proxyFunc != nil {
+		// If Proxy is nil or returns a nil *URL, no proxy is used.
 		transport.Proxy = proxyFunc
 	}
 	transport.TLSClientConfig = &tls.Config{
@@ -204,6 +205,7 @@ func (h *DefaultHttpClientsFactory) createHealthCheckHttpClient(routingConfig *c
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	proxyFunc := h.getProxyFunc(routingConfig)
 	if proxyFunc != nil {
+		// If Proxy is nil or returns a nil *URL, no proxy is used.
 		transport.Proxy = proxyFunc
 	}
 	transport.TLSClientConfig = &tls.Config{
--- a/controllers/workspace/http.go
+++ b/controllers/workspace/http.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	controller "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
+	"github.com/devfile/devworkspace-operator/pkg/config"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/go-logr/logr"
@@ -44,7 +45,7 @@ type HttpClientsFactory interface {
 	// GetHealthCheckHttpClient returns an HTTP client that skips TLS verification.
 	// This client MUST only be used for workspace health/readiness checks, not for
 	// fetching external content or making security-sensitive requests.
-	GetHealthCheckHttpClient(*controller.RoutingConfig) *http.Client
+	GetHealthCheckHttpClient() *http.Client
 }
 
 // DefaultHttpClientsFactory is a thread-safe, caching implementation of HttpClientsFactory.
@@ -63,9 +64,8 @@ type DefaultHttpClientsFactory struct {
 	httpClientConfigmapRef *controller.ConfigmapReference
 	httpClientCertsVersion string
 
-	healthCheckHttpClientProxyConfig *controller.Proxy
-
 	systemCertPool *x509.CertPool
+	proxyFunc      func(*http.Request) (*url.URL, error)
 }
 
 func SetupHttpClientsFactory(k8s client.Client, logger logr.Logger) error {
@@ -74,10 +74,26 @@ func SetupHttpClientsFactory(k8s client.Client, logger logr.Logger) error {
 		return fmt.Errorf("failed to load system cert pool: %w", err)
 	}
 
+	proxyFunc := getProxyFunc()
+
+	healthCheckHttpClientTransport := http.DefaultTransport.(*http.Transport).Clone()
+	if proxyFunc != nil {
+		// Preserve the default proxy (from env vars) when no explicit proxy is configured.
+		healthCheckHttpClientTransport.Proxy = proxyFunc
+	}
+	healthCheckHttpClientTransport.TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: true,
+	}
+
 	httpClientsFactory = &DefaultHttpClientsFactory{
 		k8s:            k8s,
 		logger:         logger,
 		systemCertPool: systemCertPool,
+		proxyFunc:      proxyFunc,
+		healthCheckHttpClient: &http.Client{
+			Transport: healthCheckHttpClientTransport,
+			Timeout:   500 * time.Millisecond,
+		},
 	}
 
 	return nil
@@ -97,7 +113,7 @@ func (h *DefaultHttpClientsFactory) GetHttpClient(ctx context.Context, routingCo
 	defer h.mu.Unlock()
 
 	if h.shouldCreateHttpClient(routingConfig, certsCM) {
-		h.httpClient = h.createHttpClient(routingConfig, certsCM)
+		h.httpClient = h.createHttpClient(certsCM)
 
 		if routingConfig == nil {
 			h.httpClientProxyConfig = nil
@@ -120,12 +136,11 @@ func (h *DefaultHttpClientsFactory) GetHttpClient(ctx context.Context, routingCo
 	return h.httpClient
 }
 
-func (h *DefaultHttpClientsFactory) createHttpClient(routingConfig *controller.RoutingConfig, certsCM *corev1.ConfigMap) *http.Client {
+func (h *DefaultHttpClientsFactory) createHttpClient(certsCM *corev1.ConfigMap) *http.Client {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
-	proxyFunc := h.getProxyFunc(routingConfig)
-	if proxyFunc != nil {
-		// If Proxy is nil or returns a nil *URL, no proxy is used.
-		transport.Proxy = proxyFunc
+	if h.proxyFunc != nil {
+		// Preserve the default proxy (from env vars) when no explicit proxy is configured.
+		transport.Proxy = h.proxyFunc
 	}
 	transport.TLSClientConfig = &tls.Config{
 		RootCAs: h.getCaCertPool(certsCM),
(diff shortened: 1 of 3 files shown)


tolusha:

**Available commands**:

tolusha:



Review is complete. Please check the review comments below.

tolusha (commented):
## Comprehensive Review Summary

I've conducted a thorough multi-level review (standard, deep, and impact analysis) of this PR. While the core approach (nil-guard to preserve default proxy behavior) is sound, there are **critical issues** that need to be addressed before merge.

---

### 🔴 Critical Issue #1: Logic Bug in `controllers/workspace/http.go:238`

**`||` should be `&&` in empty-proxy guard**

The condition `if proxyConfig.HTTPProxy == "" || proxyConfig.HTTPSProxy == ""` returns nil when **either** proxy field is empty. This means a config with only `HTTPProxy` set (or only `HTTPSProxy`) is silently ignored.

Current behavior:
- `HTTPProxy="http://proxy:80"`, `HTTPSProxy=""` → returns nil → HTTP proxy is silently ignored
- `HTTPProxy=""`, `HTTPSProxy="https://proxy:443"` → returns nil → HTTPS proxy is silently ignored

**Fix**: Change to `&&` so nil is returned only when **both** are empty:
```go
if proxyConfig.HTTPProxy == "" && proxyConfig.HTTPSProxy == "" {
    return nil
}
```

This ensures proxy configs where only one protocol is configured are actually used instead of falling back to environment defaults.

---

### 🔴 Critical Issue #2: API Contract Semantic Conflict

**Location**: `controllers/workspace/http.go:237-240`

The Proxy struct documentation in `apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go` states:

> "To ignore automatically detected proxy settings for the cluster, set this field to an empty string ("")"

This means `""` is an **explicit opt-out**, semantically distinct from `nil` (unset/use defaults).

The new guard treats `""` as "use environment defaults," which is the **opposite** of what the API contract promises. If a user explicitly sets `httpProxy: ""` to disable the proxy, this fix would re-enable it by falling back to environment variables.

**Recommendation**: Consider distinguishing between:
- "Merged result was empty because no one configured it" 
- "User explicitly set empty string"

The API already supports this via nil vs empty-string semantics in the Proxy struct, but `getProxyFunc` doesn't leverage this distinction.

---

### ⚠️ Warning: Behavioral Change Needs Documentation

This PR changes the semantic meaning of "empty proxy config after merge" from "disable proxy" to "use environment defaults." This is a user-visible behavioral change.

**Impact on deployments:**
- **Air-gapped clusters (target fix)**: ✅ Environment proxy vars are set → health checks correctly route through proxy → workspaces start successfully
- **Standard clusters with proxy env vars**: ⚠️ If `HTTP_PROXY` is set cluster-wide but workspace endpoints are cluster-internal, health checks may route through proxy unnecessarily, adding latency or failing if proxy can't reach workspace endpoints. **Mitigation**: `NO_PROXY` must include workspace service URLs (new operational requirement)
- **Clusters without proxy env vars**: No behavioral change

**Recommendation**: Document this in the PR description and consider adding a note in `docs/dwo-configuration.md` or `docs/additional-configuration.adoc` explaining proxy resolution order: explicit config > environment variables > no proxy.

---

### ⚠️ Warning: Test Quality Gap

**Location**: `controllers/workspace/http_test.go`

**Critical gap**: Every proxy-related assertion checks `Proxy != nil` or `Proxy == nil`. No test verifies that a configured proxy actually intercepts requests.

With the `||` bug, `routingConfigWithProxy("http://proxy:80", "", "")` (used throughout tests) causes `getProxyFunc` to return nil (because HTTPSProxy is `""`). Tests assert `Proxy != nil`, which passes - but only because the nil-guard preserves the *default* proxy, not the *configured* proxy.

**Missing test cases:**
1. The exact merge scenario: `routingConfigWithProxy("", "", "")` - the bug the PR describes but no test exercises it
2. Both proxies configured: `routingConfigWithProxy("http://proxy:80", "https://proxy:443", "")` - the only case where `getProxyFunc` actually returns a proxy function with the `||` condition
3. Behavioral validation: Use `httptest.NewServer` as a proxy and verify requests are actually routed through it (would immediately catch the `||` vs `&&` bug)

---

### 💡 Suggestions

1. **Misleading comment** (`controllers/workspace/http.go:126` and line 209): The comment "If Proxy is nil or returns a nil *URL, no proxy is used" quotes Go stdlib docs, making it sound like the code is disabling proxies. The actual intent is to **preserve** the default. Consider: `// Preserve the default proxy (from env vars) when no explicit proxy is configured.`

2. **Reduced concurrency coverage** (`controllers/workspace/http_test.go:62-65`): The concurrent-access test removed `nil` routingConfig. With the `||` bug, both test configs produce the same nil proxyFunc → concurrency test no longer exercises distinct code paths. Consider restoring `nil` to test the nil-guard path under contention.

---

### ✅ Positive Feedback

- The nil-guard pattern in `createHttpClient` and `createHealthCheckHttpClient` is clean and correct
- Good comment explaining the merge scenario context (http.go:235-237)
- The fix correctly identifies that `http.DefaultTransport.Clone()` preserves the default proxy function
- Iterative refinement across commits (fixing nil-pointer dereference) shows careful development

---

### Verdict

🔄 **Request Changes** - The `||` vs `&&` logic bug silently drops proxy configs where only one protocol is configured, and the API contract tension needs resolution.

tolusha (commented):
## Comprehensive Review Summary

This PR successfully fixes the immediate issue (health check timeouts on disconnected clusters) by moving proxy resolution to global config. The architectural simplification is sound - health check clients don't need per-workspace proxy config.

### Critical Finding to Address

**Empty ProxyConfig edge case**: The main concern is that when `ProxyConfig` exists but all proxy fields are nil, the function builds a non-nil closure from empty config, which overrides env var defaults. This can reintroduce the bug being fixed. See inline comment #1.

### Testing Gaps

- No direct test for `getProxyFunc()` (the core change)
- Health check client test was removed without replacement
- Proxy test assertions are vacuous (see inline comments)

### Good Practices Observed

- Clean removal of double-checked locking
- Proper thread safety with read-lock-only accessor
- Concurrent access test properly refactored for TLS certs

### Operational Notes

- Proxy config is now static - runtime changes require pod restart
- Per-workspace proxy overrides no longer affect health checks (intentional)

**Verdict**: Approve with changes requested for the empty ProxyConfig guard.

---

### General Recommendations

**Missing test coverage for getProxyFunc()**

The core logic change - `getProxyFunc()` - has no direct test coverage. Consider adding a unit test that covers:
1. No routing/proxy config → returns nil
2. Proxy config with values → returns closure
3. Proxy config struct present but all fields nil → returns nil (validates the fix from comment #1)

**Missing test for health check client behavior**

The old `TestGetHealthCheckHttpClient` was removed entirely. Add at least a basic test confirming `GetHealthCheckHttpClient()` returns the pre-initialized client with `InsecureSkipVerify: true` and 500ms timeout.

**Document proxy resolution lifecycle**

The proxy configuration is now resolved once at startup and cached forever in `h.proxyFunc`. If operators update `DevWorkspaceOperatorConfig` proxy settings at runtime, the changes won't take effect until the operator pod restarts.

Consider adding a comment on `SetupHttpClientsFactory` noting that proxy config is resolved once at startup and not updated dynamically. This clarifies operational behavior and prevents future confusion.

controllers/workspace/http.go
```diff
@@ -159,79 +178,35 @@ func (h *DefaultHttpClientsFactory) shouldCreateHttpClient(routingConfig *contro
...
+		}
 
-	return func(req *http.Request) (*url.URL, error) {
-		return proxyConfig.ProxyFunc()(req.URL)
+		return func(req *http.Request) (*url.URL, error) {
+			return proxyConf.ProxyFunc()(req.URL)
+		}
 	}
+
+	return nil
```
tolusha, line 203: **⚠️ Empty ProxyConfig overrides environment variable defaults**

When `globalConfig.Routing.ProxyConfig` is non-nil but both `HttpProxy` and `HttpsProxy` fields are nil pointers, `getProxyFunc()` builds a closure from an empty `httpproxy.Config{}` and returns it as non-nil. This overrides `http.ProxyFromEnvironment` with a function that returns no proxy for any URL, **re-introducing the exact problem this PR fixes**.

**Recommendation**: Add a guard before building the proxy function:

```go
if globalConfig.Routing != nil && globalConfig.Routing.ProxyConfig != nil {
    proxyConfig := globalConfig.Routing.ProxyConfig
    // Only return a proxy function if at least one proxy field is explicitly configured.
    // nil fields mean "use default from environment", not "disable proxy"
    if proxyConfig.HttpProxy == nil && proxyConfig.HttpsProxy == nil {
        return nil
    }
    
    proxyConf := httpproxy.Config{}
    if proxyConfig.HttpProxy != nil {
        proxyConf.HTTPProxy = *proxyConfig.HttpProxy
    }
    if proxyConfig.HttpsProxy != nil {
        proxyConf.HTTPSProxy = *proxyConfig.HttpsProxy
    }
    if proxyConfig.NoProxy != nil {
        proxyConf.NoProxy = *proxyConfig.NoProxy
    }

    return func(req *http.Request) (*url.URL, error) {
        return proxyConf.ProxyFunc()(req.URL)
    }
}

return nil
```

controllers/workspace/http.go
```diff
@@ -159,79 +178,35 @@ func (h *DefaultHttpClientsFactory) shouldCreateHttpClient(routingConfig *contro
 		!reflect.DeepEqual(proxyConfig, h.httpClientProxyConfig)
 }
 
-func (h *DefaultHttpClientsFactory) GetHealthCheckHttpClient(routingConfig *controller.RoutingConfig) *http.Client {
+func (h *DefaultHttpClientsFactory) GetHealthCheckHttpClient() *http.Client {
```
tolusha, line 168: **Stale documentation**: The comment says "based on the proxy settings in routingConfig" but the function no longer takes `routingConfig` - it reads from global config. Consider updating to: `// getProxyFunc returns a proxy function based on the global operator configuration.`

tolusha (commented):
## Comprehensive Review Summary

I've completed a full review stack (summary, standard review, deep review, and impact analysis) of this PR. The core fix is **correct and well-targeted** for preserving proxy defaults in airgapped clusters.

### Verdict Summary
- **Standard Review**: ✅ Core fix is correct - preserves env proxy defaults  
- **Deep Review**: ⚠️ Design needs refinement - orphaned proxy tracking state contradicts new design  
- **Impact Review**: ⚠️ Dynamic proxy reconfiguration lost - should be documented or addressed

### Key Findings

**What's Good:**
- Clean simplification of `GetHealthCheckHttpClient()` - removing per-workspace routing dependency is architecturally sound
- The `if h.proxyFunc != nil` guard correctly preserves `http.ProxyFromEnvironment` behavior
- Health check client created once at startup eliminates unnecessary allocations - measurable performance improvement
- Properly addresses the regression from PR #1650 that was breaking health checks on disconnected clusters

**Main Concerns:**

1. **Dead proxy tracking logic** (`controllers/workspace/http.go:178`, lines 120-125) - `shouldCreateHttpClient` still tracks and compares `routingConfig.ProxyConfig`, but `createHttpClient` now ignores it and uses the global `h.proxyFunc`. This causes unnecessary client recreations with no behavioral effect.

2. **Test coverage gap** - The "rebuilds client when proxy changes" test (http_test.go:102-119) doesn't actually test the proxy logic because `newTestFactory` doesn't set up `proxyFunc`. The test passes due to `http.DefaultTransport.Clone()` behavior, not because the proxy feature works.

3. **Dynamic config changes require restart** - Proxy config is now read once at operator startup. If `DevWorkspaceOperatorConfig` CR is updated at runtime, changes won't take effect until the operator pod restarts (unlike PR #1650 which picked up changes dynamically).

See inline comments for specific recommendations and code suggestions.

---

**Review completed by che-ai-assistant using /ok-pr-review workflow**

controllers/workspace/http.go
```diff
@@ -159,79 +178,39 @@ func (h *DefaultHttpClientsFactory) shouldCreateHttpClient(routingConfig *contro
...
 	return h.healthCheckHttpClient
 }
 
-func (h *DefaultHttpClientsFactory) shouldCreateHealthCheckHttpClient(routingConfig *controller.RoutingConfig) bool {
-	if h.healthCheckHttpClient == nil {
-		return true
-	}
-
-	var proxyConfig *controller.Proxy
+// getProxyFunc returns a proxy function.
```
tolusha: **Stale comment needs update**

The comment says "based on the proxy settings in routingConfig" but `getProxyFunc()` no longer takes `routingConfig` - it reads from global config instead.

```suggestion
// getProxyFunc returns a proxy function based on the global operator configuration.
// Returns nil if no proxy is configured; a nil proxy func causes the HTTP transport to
// use the default proxy settings from environment variables.
func getProxyFunc() func(*http.Request) (*url.URL, error) {
```

controllers/workspace/http.go
```diff
@@ -159,79 +178,39 @@ func (h *DefaultHttpClientsFactory) shouldCreateHttpClient(routingConfig *contro
 		!reflect.DeepEqual(proxyConfig, h.httpClientProxyConfig)
```
tolusha: **Dead proxy tracking logic in shouldCreateHttpClient**

`shouldCreateHttpClient` still checks proxy config changes (line 178) and stores `routingConfig.ProxyConfig.DeepCopy()` into `h.httpClientProxyConfig` (lines 120-125), but `createHttpClient` no longer uses the routing config's proxy - it always uses `h.proxyFunc` from global config.

This means the client will be unnecessarily recreated when `routingConfig.ProxyConfig` changes, even though the resulting client will have identical proxy behavior.

Consider removing the proxy-related tracking from `shouldCreateHttpClient` since proxy is now determined at factory setup time. This would leave only cert-based rebuilding, which is the only dynamic config that remains.

controllers/workspace/http_test.go
```diff
@@ -140,24 +102,78 @@ func runCommonClientTests(t *testing.T, getClient getClientFunc) {
 		routingConfig1 := routingConfigWithProxy("http://proxy:80", "", "")
```
tolusha: **Test doesn't exercise the production proxy path**

The "rebuilds client when proxy changes" test calls `routingConfigWithProxy(...)` but the production `createHttpClient` ignores `routingConfig.ProxyConfig` and uses `h.proxyFunc` (which is nil in the test factory created by `newTestFactory`).

The `assert.NotNil(client.Transport.(*http.Transport).Proxy)` assertion passes because `http.DefaultTransport.Clone()` preserves the default proxy function - not because the factory's proxy logic works.

This test would pass even if the entire proxy feature were deleted. Consider adding a test that:
1. Sets up the factory with a mock global config containing proxy values
2. Verifies that `getProxyFunc()` returns the correct proxy function  
3. Verifies that clients created from the factory use this proxy

controllers/workspace/http_test.go
```diff
@@ -166,11 +182,14 @@ func runCommonClientTests(t *testing.T, getClient getClientFunc) {
 				defer wg.Done()
 
 				routingConfig := routingConfigs[idx%len(routingConfigs)]
-				client := getClient(factory, routingConfig)
+				client := factory.GetHttpClient(context.Background(), routingConfig)
+
+				require.NotNil(t, client)
```
tolusha, line 165: **require.NotNil in goroutines is unsafe**

`require.NotNil` calls `FailNow()` which invokes `runtime.Goexit()`. In a non-test goroutine, this can cause unpredictable behavior.

Use `assert.NotNil` (which calls `Fail()` instead) inside goroutines, or collect errors via a channel and assert after `wg.Wait()`.

controllers/workspace/http.go
```diff
@@ -74,10 +74,26 @@ func SetupHttpClientsFactory(k8s client.Client, logger logr.Logger) error {
 		return fmt.Errorf("failed to load system cert pool: %w", err)
 	}
 
+	proxyFunc := getProxyFunc()
```
tolusha, line 76: **Dynamic proxy reconfiguration lost**

The proxy function is now computed once at operator startup in `SetupHttpClientsFactory`. If the `DevWorkspaceOperatorConfig` CR's proxy settings are updated at runtime, the HTTP clients won't reflect the change until the operator pod restarts.

The old code (from PR #1650) picked up proxy changes dynamically via per-reconciliation routing config.

If dynamic proxy reconfiguration is needed, consider:
- Re-reading global config when proxy-related fields change  
- Refreshing `proxyFunc` and rebuilding the health check client transport

If this loss of dynamic reconfiguration is intentional (simplification for airgapped cluster fix), document it in a comment here.

controllers/workspace/http.go
```diff
@@ -159,79 +178,39 @@ func (h *DefaultHttpClientsFactory) shouldCreateHttpClient(routingConfig *contro
...
 
-// getProxyFunc returns a proxy function based on the proxy settings in routingConfig.
-// Returns nil if no proxy is configured; a nil proxy func causes the HTTP transport to
-// use the default proxy settings from environment variables.
-func (h *DefaultHttpClientsFactory) getProxyFunc(routingConfig *controller.RoutingConfig) func(*http.Request) (*url.URL, error) {
-	if routingConfig == nil || routingConfig.ProxyConfig == nil {
-		return nil
-	}
+		if proxyConf.HTTPProxy == "" && proxyConf.HTTPSProxy == "" {
+			return nil
```
tolusha, line 194: **Minor optimization: cache ProxyFunc result**

`httpproxy.Config.ProxyFunc()` creates a new closure each time it's called. Since the outer closure is stored once at factory setup, you could call `proxyConf.ProxyFunc()` once and store the result:

```suggestion
		proxyFn := proxyConf.ProxyFunc()
		return func(req *http.Request) (*url.URL, error) {
			return proxyFn(req.URL)
		}
```

This is a minor optimization but marginally cleaner.

tolusha (commented):
## Review Summary

This PR successfully fixes the proxy configuration bug on air-gapped clusters and significantly simplifies the HTTP client management code. The architectural decision to move from per-workspace to per-operator-lifecycle proxy configuration is sound.

### Key Findings

**Fix require.NotNil in concurrent test** ()
The concurrent test uses `require.NotNil(t, client)` inside goroutines. `require.NotNil` calls `t.FailNow()` which invokes `runtime.Goexit()`. From a non-test goroutine, this doesn't propagate correctly to the test runner - it silently kills the goroutine without failing the test, potentially masking real failures. Should use `assert.NotNil` instead.

**Document runtime proxy configuration behavior** ()
The proxy function is computed once at operator startup. If an admin updates the `DevWorkspaceOperatorConfig` CR's proxy settings at runtime, the change won't take effect until the operator pod is restarted. Consider adding a code comment or documenting this in the operator configuration docs.

**Clarify that routingConfig proxy fields are ignored** ()
`GetHttpClient` still accepts `*controller.RoutingConfig` but only uses it for TLS certs - the `ProxyConfig` field is now silently ignored. Consider adding a comment clarifying that proxy config from `routingConfig` is intentionally ignored.

**Add direct unit tests for getProxyFunc()**
The `getProxyFunc()` function has multiple code paths (nil config, empty proxy URLs, HTTP-only, HTTPS-only, NoProxy-only) that aren't tested directly. Consider adding tests using `config.SetGlobalConfigForTesting()`.

**Verify NoProxy-only configuration** ()
If only `NoProxy` is set (both `HttpProxy` and `HttpsProxy` are empty), `getProxyFunc()` returns nil, discarding the `NoProxy` setting. This is likely intentional but worth verifying against documented behavior.

### Positive Observations

Excellent work on this fix:
- **Significant simplification**: Removing ~40 lines of complex proxy-tracking logic
- **Correct nil-proxy semantics**: Properly preserves Go's default `http.ProxyFromEnvironment` behavior
- **Performance improvement**: Health check client uses only read locks, eliminating contention
- **Well-structured tests**: Concurrency test exercises realistic cert-based variations

The architectural decision to move proxy configuration from per-workspace to per-operator-lifecycle is sound for this problem domain.

---
*Generated by Claude Code PR Review*

tolusha (commented):
## Review Summary

This PR successfully fixes the proxy configuration bug on air-gapped clusters and significantly simplifies the HTTP client management code. The architectural decision to move from per-workspace to per-operator-lifecycle proxy configuration is sound.

### Key Findings

**Fix require.NotNil in concurrent test** (http_test.go:163-164)
The concurrent test uses require.NotNil(t, client) inside goroutines. require.NotNil calls t.FailNow() which invokes runtime.Goexit(). From a non-test goroutine, this doesn't propagate correctly to the test runner - it silently kills the goroutine without failing the test, potentially masking real failures. Should use assert.NotNil instead.

**Document runtime proxy configuration behavior** (http.go:73-88)
The proxy function is computed once at operator startup. If an admin updates the DevWorkspaceOperatorConfig CR's proxy settings at runtime, the change won't take effect until the operator pod is restarted. Consider adding a code comment or documenting this in the operator configuration docs.

**Clarify that routingConfig proxy fields are ignored** (http.go:112-129)
GetHttpClient still accepts *controller.RoutingConfig but only uses it for TLS certs - the ProxyConfig field is now silently ignored. Consider adding a comment clarifying that proxy config from routingConfig is intentionally ignored.

**Add direct unit tests for getProxyFunc()**
The getProxyFunc() function has multiple code paths (nil config, empty proxy URLs, HTTP-only, HTTPS-only, NoProxy-only) that aren't tested directly. Consider adding tests using config.SetGlobalConfigForTesting().

**Verify NoProxy-only configuration** (http.go:202-204)
If only NoProxy is set (both HttpProxy and HttpsProxy are empty), getProxyFunc() returns nil, discarding the NoProxy setting. This is likely intentional but worth verifying against documented behavior.

### Positive Observations

Excellent work on this fix:
- Significant simplification: Removing ~40 lines of complex proxy-tracking logic
- Correct nil-proxy semantics: Properly preserves Go's default http.ProxyFromEnvironment behavior  
- Performance improvement: Health check client uses only read locks, eliminating contention
- Well-structured tests: Concurrency test exercises realistic cert-based variations

The architectural decision to move proxy configuration from per-workspace to per-operator-lifecycle is sound for this problem domain.

---
Generated by Claude Code PR Review

dkwon17 (approved):

rohanKanojia:
LGTM, thanks 👍

Issue #1668, opened 2026-07-01
Add cert-manager CA injection annotation to DevWorkspace validating webhook


simxr:
## Problem

DevWorkspace has a known unresolved webhook certificate / `caBundle` drift issue.

The failure is documented in [devfile/devworkspace-operator#1157](https://github.com/devfile/devworkspace-operator/issues/1157). In that issue, GitLab Runner calls the DevWorkspace admission webhook:

```text
validate-exec.devworkspace-controller.svc
devworkspace-webhookserver.devworkspace-controller.svc:443/validate
```

and TLS verification fails. Restarting `devworkspace-webhook-server` temporarily fixes the issue, which suggests the webhook serving certificate and the Kubernetes webhook `caBundle` can drift out of sync.

## Related upstream discussion

A related Che issue, [eclipse-che/che#23184](https://github.com/eclipse-che/che/issues/23184), discusses the cleaner fix: add cert-manager CA injection metadata to CRDs and mutating/validating webhooks so cert-manager cainjector keeps webhook `caBundle` fields updated automatically.

That discussion explicitly mentions applying the same approach to DevWorkspace. A DWO maintainer commented that adding this annotation to the webhook should hopefully resolve the issue, but the issue later went stale/closed without the DWO fix being completed.

## Current gap

The dynamically generated DevWorkspace `ValidatingWebhookConfiguration` does not include the cert-manager CA injection annotation:

```yaml
cert-manager.io/inject-ca-from: <namespace>/devworkspace-controller-serving-cert
```

Without this annotation, cert-manager may renew the serving certificate Secret, but Kubernetes can continue using a stale or incorrect webhook `caBundle`.

## Proposed fix

Apply the same cainjector pattern used by Che: annotate the DevWorkspace validating webhook so cert-manager keeps the webhook `caBundle` synchronized with the cert-manager-managed serving certificate.

The generated DevWorkspace validating webhook should include:

```yaml
metadata:
  annotations:
    cert-manager.io/inject-ca-from: <namespace>/devworkspace-controller-serving-cert
```

This allows cert-manager cainjector to keep:

```yaml
webhooks:
  - clientConfig:
      caBundle: ...
```

updated automatically when the referenced certificate changes.

## Expected result

The Kubernetes API server should continue to trust the DevWorkspace webhook after cert-manager certificate renewal, without requiring manual deletion or restart of `devworkspace-webhook-server`.

## Related issues

[...]

PR #1669, opened 2026-07-02
Fix issue #1668 about auto renew certificate on mutating / validating

disaster37:
### What does this PR do?

Add cert-manager annotations on mutating and validating depoloyment to auto renew certificate and auto restart pod when it's done.

Fix issue #1668

### PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook certificate handling is improved by adding the required webhook annotations to both mutating and validating webhook configurations.
  * CA injection now adapts automatically: it uses cert-manager when available, otherwise it falls back to the OpenShift CABundle when running on OpenShift.
* **New Features**
  * The operator now detects whether cert-manager is present and uses that signal to select the correct webhook CA source.
* **Tests**
  * Added coverage for cert-manager detection behavior and for the exact webhook annotation sets across Kubernetes and OpenShift modes.


webhook/workspace/mutating_cfg.go
```diff
@@ -155,6 +157,9 @@ func BuildMutateWebhookCfg(namespace string) *admregv1.MutatingWebhookConfigurat
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   MutateWebhookCfgName,
 			Labels: server.WebhookServerAppLabels(),
+			Annotations: map[string]string{
```
tolusha: I think it fixes issue with Kubernetes infrastructure.
For OpenShift it has to be `service.beta.openshift.io/inject-cabundle: "true"`
Please correct me, if I am wrong.

disaster37: You are right when OpenShift's Native Service CA Operator is used. I have modified this PR to:
1. Check is OCP or regulare Kubernetes
2. If OCP, check if used OpenShift's Native Service CA Operator or cert-manager

rohanKanojia:
@disaster37 : Thank you for your PR!

Could you please amend your commit with `--signed-off` footer so that DCO check can pass? Also it would be nice to add some unit tests to validate that changes work as expected.

disaster37:
It's done (sign commit + test)

tolusha:
[DCO](https://github.com/devfile/devworkspace-operator/pull/1669/checks?check_run_id=86803987667)
still fails

rohanKanojia:
@disaster37 : Could you please squash your commits into one? I see you've added signed-off only in latest commit. DCO check is failing due to earlier two commits not aligned with commit policy

disaster37:
It's done

rohanKanojia:

Task completed.

rohanKanojia:

Task completed.

webhook/workspace/annotations.go
```diff
@@ -0,0 +1,32 @@
...
+package workspace
+
+import (
+	"fmt"
+
+	"github.com/devfile/devworkspace-operator/pkg/infrastructure"
+)
+
+func getWebhookAnnotations(namespace string) map[string]string {
+	annotations := map[string]string{}
```
tolusha, line 25: Consider adding a brief comment here explaining why cert-manager takes priority over the OpenShift Service CA annotation. The ordering is intentional and correct, but without a comment a future maintainer may read the `else if` as arbitrary ordering rather than a deliberate policy decision. Something like: `// cert-manager takes precedence when installed on OpenShift; mixing both annotations can cause conflicts.`

webhook/workspace/annotations.go
```diff
@@ -0,0 +1,32 @@
...
+	if infrastructure.CertManagerDetected() {
+		annotations["cert-manager.io/inject-ca-from"] = fmt.Sprintf("%s/devworkspace-controller-serving-cert", namespace)
+	} else if infrastructure.IsOpenShift() {
```
tolusha: Consider extracting `devworkspace-controller-serving-cert` to a named constant, similar to how `MutateWebhookCfgName` and `ValidateWebhookCfgName` are defined. This string is the post-kustomize composed name (from `namePrefix: devworkspace-controller-` applied to base name `serving-cert` in `deploy/templates/components/cert-manager/self-signed-certificates.yaml`). If the `namePrefix` changes, the kustomize-generated CRD webhook patches update automatically but this code silently breaks - cert-manager cainjector would find no Certificate and log errors without injecting the CA bundle. A constant (or at minimum a comment documenting the coupling and the kustomize file that must stay in sync) makes the contract explicit.

webhook/workspace/mutating_cfg.go
```diff
@@ -153,8 +153,9 @@ func BuildMutateWebhookCfg(namespace string) *admregv1.MutatingWebhookConfigurat
 
 	return &admregv1.MutatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   MutateWebhookCfgName,
-			Labels: server.WebhookServerAppLabels(),
+			Name:        MutateWebhookCfgName,
+			Labels:      server.WebhookServerAppLabels(),
+			Annotations: getWebhookAnnotations(namespace),
```
tolusha, line 158: When cert-manager or the OpenShift Service CA operator is active, the webhook config will have both an explicit `CABundle` (written from the cert file on disk by the caller) and a CA-injection annotation. During a cert rotation the operator may write a stale CABundle and the external controller will overwrite it - creating a brief window with a mismatched CA. Would it be possible to omit `CABundle` from `WebhookClientConfig` when `CertManagerDetected()` is true (or when using the OpenShift annotation), letting the external controller be the sole authority? The cert-manager CRD patch in `deploy/templates/cert-manager/crd_webhooks_patch.yaml` already follows this pattern, with a comment "caBundle will be filled by cert-manager on creation".

pkg/infrastructure/cluster_test.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func TestCertManagerDetectedWithOpenShift(t *testing.T) {
+	InitializeForTestingWithCertManager(OpenShiftv4)
+	assert.True(t, CertManagerDetected())
+}
+
+func TestCertManagerDetectedPanicsWhenNotInitialized(t *testing.T) {
+	initialized = false
+	defer func() {
```
tolusha: Consider using `t.Cleanup()` to restore the global state after this test instead of (or in addition to) the `defer`. The `defer` resets to `Kubernetes` after the panic test, but if the test panics in an unexpected way before the deferred function runs, subsequent tests in the same package may see `initialized = false`. A `t.Cleanup` or a helper that captures and restores all three package-level variables (`current`, `certManagerDetected`, `initialized`) would make the test order-independent. Also worth noting in the file or test that these tests cannot use `t.Parallel()` due to shared mutable package state.

pkg/infrastructure/cluster.go
```diff
@@ -85,6 +102,7 @@ func detect() (Type, error) {
 	if err != nil {
```
tolusha, line 85: Does it make sense to move the `certManagerDetected` assignment into `Initialize()` alongside the `detect()` call, rather than inside `detect()` as a side effect? The function's name and `(Type, error)` return type suggest it only determines the cluster type, so the hidden side effect is a bit surprising. It also makes `detect()` harder to test in isolation and is architecturally inconsistent with how OpenShift detection works (returned as a value, not a side effect). One option: return a richer struct from `detect()` (e.g., `DetectionResult{InfraType, CertManagerAvailable}`) or call a separate function from `Initialize()`.

tolusha (commented):
Thanks for addressing the stale caBundle problem. The annotation selection logic is well-structured and the test coverage is good. A few things worth discussing:

**PR title has typos.** The title reads "Fix issue #1668 about auto renew certificat on mutating / validatinf" - "certificat" should be "certificate" and "validatinf" should be "validating". Consider updating before merging, as it will appear in the merge commit message.

**SetupWebhooks TLS path inconsistency (pkg/webhook/create.go).** With `CertManagerDetected()` now in place, there are four infrastructure states but `SetupWebhooks` in `pkg/webhook/create.go` still branches only on `IsOpenShift()`. On an OpenShift cluster with cert-manager installed, the service gets annotated with `service.beta.openshift.io/serving-cert-secret-name` (OpenShift Service CA provisions the TLS secret), while the webhook gets the `cert-manager.io/inject-ca-from` annotation pointing at a cert-manager Certificate. These use different CAs - OpenShift Service CA vs. cert-manager's own Issuer - so the CA bundle cert-manager injects will not match the TLS certificate the server presents. Would it be possible to incorporate `CertManagerDetected()` into `SetupWebhooks` so that when cert-manager is present the service is not annotated for the OpenShift Service CA?

tolusha:
@disaster37 
Everything looks good to me.
The only thing I'd suggest is being consistent with the annotations:
For OpenShift, always use service.beta.openshift.io/inject-cabundle: "true" (independently of cert-manager existance)
For Kubernetes, use cert-manager.io/inject-ca-from.
I believe we don't need to detect cert-manager. Otherwise, we won't be able to deploy che-operator/DWO.

disaster37:
I now use static annotation depend on kubernetes or OCP. Remove the cert-manager detection. Is by design when deploy eclipse che.

tolusha (approved):

rohanKanojia (approved):

PR #1680, opened 2026-07-28
feat: implement Gateway API routing class for DevWorkspaceRouting

btjd:
### What does this PR do?

  Adds a new `gateway-api` routing class to the DevWorkspace Operator that exposes workspace endpoints via Kubernetes [Gateway API](https://gateway-api.sigs.k8s.io/)
  HTTPRoutes instead of Ingress/Route resources.

  - Introduces `GatewayAPISolver` that creates HTTPRoutes attached to a pre-provisioned Gateway referenced in the operator config (`config.routing.gatewayRef`)
  - Adds `GatewayReference` type to `DevWorkspaceOperatorConfig` API to configure the target Gateway name, namespace, and GatewayClass
  - Adds `DevWorkspaceRoutingGatewayAPI` (`"gateway-api"`) routing class constant
  - Implements HTTPRoute sync lifecycle in the routing controller (create/update/delete with owner references and watches)
  - Fixes a controller readiness stall by requeuing when endpoints are not yet ready
  - Adds unit tests covering HTTPRoute creation, endpoint exposure filtering, discoverable endpoints, and service generation
  - `sigs.k8s.io/gateway-api` added as a dependency; Gateway API HTTPRoute CRD bundled in `deploy/`

  https://issues.redhat.com/browse/CRW-23675

  ### Is it tested? How?

  Unit tests added for the `GatewayAPISolver` and routing controller covering:
  - HTTPRoute creation for exposed endpoints
  - Endpoint exposure filtering (only `public` endpoints get HTTPRoutes)
  - Discoverable endpoint labeled service generation
  - Service creation parity with the basic solver
  - Owner reference and annotation propagation

  Manual verification:
  - Configure `DevWorkspaceOperatorConfig` with `routing.gatewayRef` pointing to a pre-provisioned Gateway
  - Create a DevWorkspace with `routingClass: gateway-api`
  - Verify HTTPRoutes are created and endpoints are reachable through the Gateway
  - Verify cleanup on DevWorkspaceRouting deletion via owner references

  ### PR Checklist

  🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added Gateway API routing support for workspaces.
  * Gateway references can be configured by name and optional namespace.
  * Public endpoints receive HTTPS routes and HTTP-to-HTTPS redirects.
  * Discoverable endpoints are supported through Gateway API HTTPRoutes.
  * Gateway API can be selected as the workspace routing class.

* **Bug Fixes**
  * Improved endpoint readiness handling while HTTPRoutes are being provisioned.
  * HTTPRoutes are synchronized automatically as workspace routing changes.
  * Gateway reference names and namespaces are now validated as valid Kubernetes resource names.


apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -108,17 +108,36 @@ type BackupCronJobConfig struct {
...
+	// +kubebuilder:default:=3
 	// +kubebuilder:validation:Optional
 	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
 }
 
+// GatewayReference defines a reference to a Gateway API Gateway resource
+// that HTTPRoutes should attach to via parentRefs.
+type GatewayReference struct {
+	// Name is the name of the Gateway resource
+	// +kubebuilder:validation:Required
```
rohanKanojia, line 126: Required only checks presence; maybe add length bounds?
```suggestion
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
```

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -108,17 +108,36 @@ type BackupCronJobConfig struct {
...
+	Name string `json:"name"`
+	// Namespace is the namespace of the Gateway resource.
+	// If not specified, HTTPRoutes will reference a Gateway in the same namespace
+	// as the DevWorkspace.
+	// +kubebuilder:validation:Optional
```
rohanKanojia, line 133: nit, Perhaps add [DNS-1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) label bounds here too
```suggestion
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=63
```

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -108,17 +108,36 @@ type BackupCronJobConfig struct {
...
+	Namespace *string `json:"namespace,omitempty"`
+	// GatewayClassName is the name of the GatewayClass to use.
+	// This is used for validation and informational purposes.
+	// Defaults to "nginx" if not specified.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:="nginx"
+	GatewayClassName string `json:"gatewayClassName,omitempty"`
```
rohanKanojia: `+kubebuilder:default` always populates this, would `omitempty` be needed?

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -341,13 +374,11 @@ func (r *DevWorkspaceRoutingReconciler) SetupWithManager(mgr ctrl.Manager) error
...
-		WithOptions(controller.Options{
-			MaxConcurrentReconciles: maxConcurrentReconciles,
-			UsePriorityQueue:        ptr.To(false),
-		}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconciles}).
 		For(&controllerv1alpha1.DevWorkspaceRouting{}).
 		Owns(&corev1.Service{}).
-		Owns(&networkingv1.Ingress{})
+		Owns(&networkingv1.Ingress{}).
+		Owns(&gwapiv1.HTTPRoute{})
```
rohanKanojia: We currently register it unconditionally. Gateway API CRDs aren't installed by default on Kubernetes or OpenShift. Could we gate this the same way the Route watch is gated ?

deploy/templates/components/rbac/role.yaml
```diff
@@ -174,6 +174,12 @@ rules:
   - create
   - get
   - update
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - httproutes
+  verbs:
+  - '*'
```
rohanKanojia: nit, scope httproutes verbs instead of `*`

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -232,12 +243,34 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(ctx context.Context, req ctrl.
...
+
 	exposedEndpoints, endpointsAreReady, err := solver.GetExposedEndpoints(instance.Spec.Endpoints, clusterRoutingObj)
 	if err != nil {
 		reqLogger.Error(err, "Could not get exposed endpoints for devworkspace")
 		return reconcile.Result{}, r.markRoutingFailed(instance, fmt.Sprintf("Could not get exposed endpoints for DevWorkspace: %s", err))
 	}
 
+	if !endpointsAreReady {
+		reqLogger.Info("Endpoints not ready, requeuing to check HTTPRoute status")
+		return reconcile.Result{RequeueAfter: 3 * time.Second}, r.reconcileStatus(instance, nil, nil, false, "Waiting for HTTPRoute endpoints to be ready")
```
rohanKanojia: This message always says “HTTPRoute…” for all routing classes, perhaps we should keep it generic?

go.mod
```diff
@@ -1,132 +1,132 @@
...
-	k8s.io/kube-openapi v0.0.0-20260317180543-43fb72c5454a // indirect
-	k8s.io/streaming v0.36.0 // indirect
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.34.0 // indirect
-	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
+	k8s.io/apiserver v0.34.0 // indirect
+	k8s.io/component-base v0.34.0 // indirect
+	k8s.io/klog/v2 v2.130.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
+	sigs.k8s.io/gateway-api v1.2.1 // indirect
```
rohanKanojia: umm, why `sigs.k8s.io/gateway-api` marked `// indirect` despite direct import

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
```diff
@@ -0,0 +1,322 @@
...
+	endpointName string,
+	workspaceMeta DevWorkspaceMetadata,
+) gwapiv1.HTTPRoute {
+	routeName := common.RouteName(workspaceMeta.DevWorkspaceId, endpointName)
+
+	pathPrefix := gwapiv1.PathMatchPathPrefix
+	pathValue := "/"
+
+	// 10 hour timeout (matching ingress2gateway output and long-running workspace sessions)
+	requestTimeout := gwapiv1.Duration("10h")
```
rohanKanojia: nit, Perhaps introduce these timeouts as constants with links to ingress2gateway spec?

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
rohanKanojia: On second thought, I am not sure why this field is needed. It is not referenced in gateway solver. It only consumes Name and Namespace. So generated HTTPRoutes are identical regardless of what this is set to.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
```diff
@@ -0,0 +1,322 @@
...
+			// Find the HTTPRoute for this endpoint by matching the endpoint name annotation
+			var url string
+			for _, httpRoute := range routingObj.HTTPRoutes {
+				// Only check HTTPS routes (not HTTP redirects) - they're the actual backend routes
+				if httpRoute.Annotations[constants.DevWorkspaceEndpointNameAnnotation] == endpoint.Name {
+					if len(httpRoute.Spec.Rules) > 0 && len(httpRoute.Spec.Rules[0].BackendRefs) > 0 {
+						if len(httpRoute.Spec.Hostnames) > 0 {
+							hostname := string(httpRoute.Spec.Hostnames[0])
+							// Use HTTPS scheme
+							url = fmt.Sprintf("https://%s%s", hostname, endpoint.Path)
```
rohanKanojia: Could we reuse `getURLForEndpoint` (or mirror its logic) here so protocol, path-joining, and query/fragment handling stay consistent with the other solvers?

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
```diff
@@ -0,0 +1,322 @@
...
+	gatewayNamespace string,
+	hostname string,
+	endpoint controllerv1alpha1.Endpoint,
+	endpointName string,
+	workspaceMeta DevWorkspaceMetadata,
+) gwapiv1.HTTPRoute {
+	routeName := fmt.Sprintf("%s-http-redirect", common.RouteName(workspaceMeta.DevWorkspaceId, endpointName))
+
+	httpsScheme := "https"
+	statusCode := 301
```
rohanKanojia, line 137: nit, maybe use 308, which always preserves the request method?

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
```diff
@@ -0,0 +1,322 @@
...
+	gatewayRef *controllerv1alpha1.GatewayReference,
+	gatewayNamespace string,
+	hostname string,
+	endpoint controllerv1alpha1.Endpoint,
+	endpointName string,
+	workspaceMeta DevWorkspaceMetadata,
+) gwapiv1.HTTPRoute {
+	routeName := fmt.Sprintf("%s-http-redirect", common.RouteName(workspaceMeta.DevWorkspaceId, endpointName))
+
+	httpsScheme := "https"
```
rohanKanojia, line 136: A bit naive question: why are we always redirecting to HTTPS and routing over 443? I haven't checked but how would TLS be handled in this setup? Just want to make sure I understand the intended flow.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
```diff
@@ -0,0 +1,322 @@
...
+	// 10 hour timeout (matching ingress2gateway output and long-running workspace sessions)
+	requestTimeout := gwapiv1.Duration("10h")
+
+	group := gwapiv1.Group(gwapiv1.GroupVersion.Group)
+	kind := gwapiv1.Kind("Gateway")
+	namespace := gwapiv1.Namespace(gatewayNamespace)
+	httpsPort := gwapiv1.PortNumber(443)
+	servicePort := gwapiv1.PortNumber(endpoint.TargetPort)
+
+	return gwapiv1.HTTPRoute{
```
rohanKanojia, line 229: nit, this builder and the one is `createHTTPRedirectRoute` share very similar fields. Maybe we can create a shared helper to reduce duplication?

rohanKanojia:
I tested this PR on OpenShift with the built-in OpenShift Gateway API and can confirm the `gateway-api` routing class works as expected.

**Environment:** OCP 4.20 (AWS CI), DWO built from this PR.

### Steps I followed

#### 1. Enable OpenShift Gateway API

```bash
oc apply -f - <<'EOF'
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: openshift-default
spec:
  controllerName: openshift.io/gateway-controller/v1
EOF

oc get gatewayclass openshift-default -o jsonpath='{range .status.conditions[*]}{.type}={.status} {.message}{"\n"}{end}'
oc -n openshift-ingress get pods
```

Example output:

```text
Accepted=True Handled by Istio controller
ControllerInstalled=True istiod v1.26.8 installed
CRDsReady=True CRDs updated by CIO

NAME                                       READY   STATUS    RESTARTS   AGE
istiod-openshift-gateway-9dbdfcf6c-9gpz4   1/1     Running   0          92s
router-default-67bc5b596d-c4j2q            1/1     Running   0          33m
```

#### 2. Create a shared Gateway (80 + 443, cross-namespace routes)

```bash
oc -n openshift-ingress get secrets | grep tls
DOMAIN=$(oc get ingresses.config.openshift.io cluster -o jsonpath='{.spec.domain}')
echo "$DOMAIN"

oc apply -f - <<'EOF'
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: shared-gateway
  namespace: openshift-ingress
spec:
  gatewayClassName: openshift-default
  listeners:
  - name: http
    protocol: HTTP
    port: 80
    allowedRoutes:
      namespaces:
        from: All
  - name: https
    protocol: HTTPS
    port: 443
    tls:
      mode: Terminate
      certificateRefs:
      - name: router-certs-default
    allowedRoutes:
      namespaces:
        from: All
EOF

oc -n openshift-ingress get gateway shared-gateway
oc -n openshift-ingress get svc | grep shared-gateway
```

#### 3. Install DWO from this PR and configure DWOC

```bash
DOMAIN=$(oc get ingresses.config.openshift.io cluster -o jsonpath='{.spec.domain}')

oc apply -f - <<EOF
metadata:
config:
  routing:
    clusterHostSuffix: ${DOMAIN}
    defaultRoutingClass: gateway-api
    gatewayRef:
      name: shared-gateway
      namespace: openshift-ingress
      gatewayClassName: openshift-default
EOF

oc -n openshift-operators get dwoc devworkspace-operator-config -o yaml
oc -n openshift-operators get deploy | grep -i workspace
```

#### 4. Create a test DevWorkspace

Use `traefik/whoami` on port **8080** (OpenShift blocks non-root bind on :80):

```bash
oc new-project gw-api-test 2>/dev/null || oc project gw-api-test

oc apply -f - <<'EOF'
kind: DevWorkspace
metadata:
  name: gw-test
  namespace: gw-api-test
spec:
  started: true
  routingClass: gateway-api
  template:
    components:
      - name: tools
        container:
          image: traefik/whoami:v1.10.3
          args: ["--port", "8080"]
          memoryLimit: 128Mi
          endpoints:
            - name: http-server
              targetPort: 8080
              exposure: public
              protocol: http
EOF

oc -n gw-api-test get dw,pod
```

Example output:

```text
NAMESPACE     NAME      DEVWORKSPACE ID             PHASE     INFO
gw-api-test   gw-test   workspace1f219d6e267842ba   Running   Workspace is running

NAME                                         READY   STATUS    RESTARTS   AGE
workspace1f219d6e267842ba-66bb5d6d8d-2h5wh   1/1     Running   0          13s
```

#### 5. Verify routing objects

```bash
oc -n gw-api-test get dw,dwr,svc,httproute
oc -n gw-api-test get dwr -o jsonpath='{.items[0].status.phase}{"\n"}{.items[0].status.exposedEndpoints.tools[0].url}{"\n"}'
oc -n gw-api-test get httproute -o yaml
```

Example output:

```text
NAME                                        DEVWORKSPACE ID             PHASE     INFO
devworkspace.workspace.devfile.io/gw-test   workspace1f219d6e267842ba   Running   Workspace is running

NAME                                                                          DEVWORKSPACE ID             PHASE   INFO
devworkspacerouting.controller.devfile.io/routing-workspace1f219d6e267842ba   workspace1f219d6e267842ba   Ready   DevWorkspaceRouting prepared

NAME                                        TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
service/workspace1f219d6e267842ba-service   ClusterIP   172.30.74.209   <none>        8080/TCP   31s

NAME                                                                                      HOSTNAMES                                                                                        AGE
httproute.../workspace1f219d6e267842ba-http-server                 ["workspace1f219d6e267842ba-http-server-8080.apps.ci-ln-….aws-4.ci.openshift.org"]   30s
httproute.../workspace1f219d6e267842ba-http-server-http-redirect   ["workspace1f219d6e267842ba-http-server-8080.apps.ci-ln-….aws-4.ci.openshift.org"]   30s

Ready
https://workspace1f219d6e267842ba-http-server-8080.apps.ci-ln-bipwf72-76ef8.aws-4.ci.openshift.org
```

HTTPRoute (HTTPS) — key fields:

```yaml
spec:
  parentRefs:
  - name: shared-gateway
    namespace: openshift-ingress
    port: 443
  rules:
  - backendRefs:
    - name: workspace1f219d6e267842ba-service
      port: 8080
status:
  parents:
  - controllerName: openshift.io/gateway-controller/v1
    conditions:
    - type: Accepted
      status: "True"
    - type: ResolvedRefs
      status: "True"
```

#### 6. Verify traffic through the Gateway

```bash
HOST=$(oc -n gw-api-test get dwr -o jsonpath='{.items[0].status.exposedEndpoints.tools[0].url}' | sed 's|https://||')
GATEWAY_LB=$(oc -n openshift-ingress get svc shared-gateway-openshift-default -o jsonpath='{.status.loadBalancer.ingress[0].hostname}')
GW_IP=$(dig +short "$GATEWAY_LB" | head -1)

echo "HOST=$HOST"
echo "GATEWAY_LB=$GATEWAY_LB"
echo "GW_IP=$GW_IP"

curl -vk --resolve "${HOST}:443:${GW_IP}" "https://${HOST}/"
```

Example output:

```text
HOST=workspace1f219d6e267842ba-http-server-8080.apps.ci-ln-********s-4.ci.openshift.org
GATEWAY_LB=a164c792794794586a2894a79******.us-west-1.elb.amazonaws.com
GW_IP=54.177.216.135

HTTP/2 200
…
Hostname: workspace1f219d6e267842ba-66bb5d6d8d-2h5wh
IP: 10.131.0.25
RemoteAddr: 10.128.2.15:41966
GET / HTTP/1.1
Host: workspace1f219d6e267842ba-http-server-8080.apps.ci-ln-bipwf72-76ef8.aws-4.ci.openshift.org
X-Forwarded-Proto: https
…
```

#### 7. Cleanup

```bash
oc delete dw gw-test -n gw-api-test
oc get httproute -n gw-api-test
oc delete project gw-api-test
```

### Result

- GatewayClass / Gateway: OpenShift built-in controller, Gateway `Programmed=True`
- DWO creates HTTPRoutes with correct `parentRefs`; controller `Accepted=True`
- Traffic through Gateway LoadBalancer with `curl --resolve` → **HTTP 200** (`whoami`)

I can confirm the `gateway-api` routing class from this PR works as expected
with OpenShift’s built-in Gateway API.

rohanKanojia:
Also tested this setup on minikube with traefic following these steps https://gist.github.com/btjd/28e07c0f3b07ef0b35fa85ce7bc4f17c 

I can confirm these steps also work :heavy_check_mark:

tolusha:

I check for new comments every **10m0s**, so there may be a short delay before I respond.

**Available commands**:
- `/che-ai-assistant ok-pr-readiness` — Ensure PR has validation steps
- `/che-ai-assistant check-pr-test-failures` — Analyze failing CI checks, identify root causes, and suggest fixes
- `/che-ai-assistant update-che-e2e-tests` — Update Eclipse Che e2e tests
- `/che-ai-assistant claude` — Run a free-form instruction on this PR

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
btjd: Good catch, I will add the length bounds.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
btjd: Agreed, adding bounds.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
btjd: You're right, this field isn't used anywhere in the solver. I'll remove it to keep the API surface minimal — we can always add it back if we find a concrete use case.

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
btjd: Good point, this will cause informer failures on clusters without Gateway API CRDs installed. I'll gate this similarly to how Route watches are gated. I'm thinking we can check for the HTTPRoute CRD presence at startup and only register the watch if it exists.

deploy/templates/components/rbac/role.yaml
btjd: Agreed, will scope to the specific verbs needed: get, list, watch, create, update, patch, delete.

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
btjd: Good catch, this should be generic. Will fix.

go.mod
btjd: This was fixed during the rebase I did to fix pre merge checks failures, it's now listed as a direct dependency.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
btjd: Sure, will extract to a constant.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
btjd: Good call, I will refactor to use getURLForEndpoint to get consistent path/query/fragment handling.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
btjd: 301 is the conventional choice for HTTP→HTTPS redirects (matches ingress-nginx, Traefik defaults, and ingress2gateway output). Since workspace endpoints are primarily browser-accessed, 301 is appropriate. Happy to discuss further if there's a specific use case for method-preserving redirects.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
btjd: TLS termination is handled at the Gateway level, not in the solver. The Gateway is configured with an HTTPS listener (port 443) that has a TLS certificate. The solver creates two HTTPRoutes per endpoint: one on port 80 that redirects to HTTPS, and one on port 443 that routes to the workspace service. Traffic is encrypted between the client and Gateway; traffic from Gateway to the pod is cluster-internal plaintext (same model as Ingress with TLS termination).

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
btjd: The two builders share ObjectMeta/ParentRefs setup but diverge significantly in their Rules (redirect filter vs backendRef + timeout). I  lean toward keeping them separate for readability, but happy to extract the common ObjectMeta/ParentRef setup into a helper if you feel strongly.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
```diff
@@ -144,6 +161,12 @@ type RoutingConfig struct {
 	// TLSCertificateConfigmapRef defines the name and namespace of the configmap with a certificate to inject into the
 	// HTTP client.
 	TLSCertificateConfigmapRef *ConfigmapReference `json:"tlsCertificateConfigmapRef,omitempty"`
+	// GatewayRef defines a reference to a Gateway API Gateway resource that HTTPRoutes
+	// should attach to when using the 'gateway-api' routing class. This field is required
+	// when routingClass is set to 'gateway-api'. The referenced Gateway must be provisioned
+	// by the cluster administrator or Che Operator before workspaces can use Gateway API routing.
```
dkwon17, line 167:
```suggestion
	// by the cluster administrator before workspaces can use Gateway API routing.
```

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -232,12 +244,34 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	exposedEndpoints, endpointsAreReady, err := solver.GetExposedEndpoints(instance.Spec.Endpoints, clusterRoutingObj)
 	if err != nil {
 		reqLogger.Error(err, "Could not get exposed endpoints for devworkspace")
 		return reconcile.Result{}, r.markRoutingFailed(instance, fmt.Sprintf("Could not get exposed endpoints for DevWorkspace: %s", err))
 	}
 
+	if !endpointsAreReady {
+		reqLogger.Info("Endpoints not ready, requeuing")
+		return reconcile.Result{RequeueAfter: 3 * time.Second}, r.reconcileStatus(instance, nil, nil, false, "Waiting for endpoints to be ready")
+	}
```
dkwon17, line 273: I'm not sure if a requeue is necessary here, since there nothing in the reconciliation loop that should happen after endpointsAreReady goes from false to true?

I could be missing something but is there a reason to requeue?

pkg/config/sync.go
```diff
@@ -281,6 +281,9 @@ func mergeConfig(from, to *controller.OperatorConfiguration) {
 		if from.Routing.ClusterHostSuffix != "" {
 			to.Routing.ClusterHostSuffix = from.Routing.ClusterHostSuffix
 		}
+		if from.Routing.GatewayRef != nil {
+			to.Routing.GatewayRef = from.Routing.GatewayRef.DeepCopy()
+		}
```
dkwon17, line 286: Could you please also add the Routing.GatewayRef details in the GetCurrentConfigString function in this same file?

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
dkwon17: To provide context, in DWO we try to minimize any mention of Eclipse Che, since technically, Che is not a dependency of DWO

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
```diff
@@ -0,0 +1,339 @@
...
+	kind := gwapiv1.Kind("Gateway")
+	namespace := gwapiv1.Namespace(gatewayNamespace)
+	port := gwapiv1.PortNumber(80)
+
+	annotations := map[string]string{
+		constants.DevWorkspaceEndpointNameAnnotation: endpoint.Name,
+	}
+	for k, v := range endpoint.Annotations {
+		annotations[k] = v
+	}
```
rohanKanojia, line 152: Nit: The controller annotation is set before iterating user annotations, which means a user-provided annotation with the same key (controller.devfile.io/endpoint-name) could overwrite it. The basic solver does this in the opposite order (user annotations first, controller annotation last) to prevent that.

https://github.com/devfile/devworkspace-operator/blob/5684f196bf4f3e4e9c3ecc361d0f9268fa6ab364/controllers/controller/devworkspacerouting/solvers/basic_solver.go#L35

controllers/controller/devworkspacerouting/devworkspacerouting_controller_test.go
```diff
@@ -354,6 +355,237 @@ var _ = Describe("DevWorkspaceRouting Controller", func() {
...
+				dwrNamespacedName := namespacedName(devWorkspaceRoutingName, testNamespace)
+				createdDWR := &controllerv1alpha1.DevWorkspaceRouting{}
+				Eventually(func() (phase controllerv1alpha1.DevWorkspaceRoutingPhase, err error) {
+					if err := k8sClient.Get(ctx, dwrNamespacedName, createdDWR); err != nil {
+						return "", err
+					}
+					return createdDWR.Status.Phase, nil
+				}, timeout, interval).Should(Equal(controllerv1alpha1.RoutingReady), "DevWorkspaceRouting should have Ready phase")
+				Expect(createdDWR.Status.Message).ShouldNot(BeNil(), "Status message should be set for preparing DevWorkspaceRoutings")
+				Expect(createdDWR.Status.Message).Should(Equal("DevWorkspaceRouting prepared"), "Status message should indicate that the DevWorkspaceRouting is prepared")
```
rohanKanojia, line 451: Would it be worth adding an assertion on `Status.ExposedEndpoints` here? The test verifies Phase and Message, but the URL construction in `getExposedEndpointsFromHTTPRoutes` has some interesting matching logic (annotation-based, hostname extraction) that isn't covered elsewhere. Even a basic non-empty check might catch regressions.

controllers/controller/devworkspacerouting/devworkspacerouting_controller_test.go
```diff
@@ -354,6 +355,237 @@ var _ = Describe("DevWorkspaceRouting Controller", func() {
...
+			It("Creates HTTPRoutes", func() {
+				createdDWR := getExistingDevWorkspaceRouting(devWorkspaceRoutingName)
+
+				By("Checking HTTPS HTTPRoute is created for exposed endpoint")
+				createdHTTPRoute := gwapiv1.HTTPRoute{}
+				httpRouteNamespacedName := namespacedName(common.RouteName(testWorkspaceID, exposedEndPointName), testNamespace)
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, httpRouteNamespacedName, &createdHTTPRoute)
+					return err == nil
+				}, timeout, interval).Should(BeTrue(), "HTTPRoute should exist in cluster")
```
rohanKanojia, line 510: Might be good to assert the expected `Spec.Hostnames` is set correctly?

controllers/controller/devworkspacerouting/devworkspacerouting_controller_test.go
```diff
@@ -354,6 +355,237 @@ var _ = Describe("DevWorkspaceRouting Controller", func() {
...
+
+				By("Checking HTTPRoute is not created for non-exposed endpoint")
+				nonExposedHTTPRoute := gwapiv1.HTTPRoute{}
+				nonExposedHTTPRouteNamespacedName := namespacedName(common.RouteName(testWorkspaceID, nonExposedEndpointName), testNamespace)
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, nonExposedHTTPRouteNamespacedName, &nonExposedHTTPRoute)
+					return k8sErrors.IsNotFound(err)
+				}, timeout, interval).Should(BeTrue(), "HTTPRoute for non-exposed endpoint should not exist in cluster")
+			})
+		})
```
rohanKanojia, line 588: I see the failure cases cover missing host suffix and invalid routing class for the basic solver.

Would it make sense to add a similar  failure scenarios cases for Gateway API as well?

controllers/controller/devworkspacerouting/solvers/solver.go
```diff
@@ -112,6 +115,8 @@ func (_ *SolverGetter) GetSolver(_ client.Client, routingClass controllerv1alpha
 			return nil, fmt.Errorf("routing class %s only supported on OpenShift", routingClass)
 		}
 		return &ClusterSolver{TLS: true}, nil
+	case controllerv1alpha1.DevWorkspaceRoutingGatewayAPI:
+		return &GatewayAPISolver{}, nil
```
rohanKanojia, line 119: Should we also check `IsGatewayAPIInstalled()` before returning the solver? That would match the pattern for cluster-tls/web-terminal which check `IsOpenShift()` here, and would give users a clear error message upfront rather than failing later during sync.

controllers/controller/devworkspacerouting/sync_httproutes.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	var updatedClusterHTTPRoutes []gwapiv1.HTTPRoute
+	for _, specHTTPRoute := range specHTTPRoutes {
+		clusterObj, err := sync.SyncObjectWithCluster(&specHTTPRoute, clusterAPI)
+		switch t := err.(type) {
+		case nil:
+			break
+		case *sync.NotInSyncError:
+			clusterAPI.Logger.Info(t.Error())
```
rohanKanojia, line 62: Other three sync files (`sync_services.go`, `sync_ingresses.go`, `sync_routes.go`) silently set the flag and continue. Was the logging intentional here? 
https://github.com/devfile/devworkspace-operator/blob/5684f196bf4f3e4e9c3ecc361d0f9268fa6ab364/controllers/controller/devworkspacerouting/sync_ingresses.go#L62-L64

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
```diff
@@ -231,12 +243,34 @@ func (r *DevWorkspaceRoutingReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			}
+			reqLogger.Error(err, "Error syncing HTTPRoutes")
+			return reconcile.Result{Requeue: true}, r.reconcileStatus(instance, nil, nil, false, "Preparing HTTPRoutes")
+		} else if !httpRoutesInSync {
+			reqLogger.Info("HTTPRoutes not in sync")
+			return reconcile.Result{Requeue: true}, r.reconcileStatus(instance, nil, nil, false, "Preparing HTTPRoutes")
+		}
+		clusterRoutingObj.HTTPRoutes = clusterHTTPRoutes
+	}
+
```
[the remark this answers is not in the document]
btjd: Good catch, fixed. syncHTTPRoutes is now called unconditionally, matching the routes/ingresses pattern so stale HTTPRoutes get cleaned up when the desired list becomes empty.

pkg/provision/sync/diffopts.go
```diff
@@ -92,6 +93,13 @@ var ingressDiffOpts = cmp.Options{
 	cmpopts.IgnoreFields(networkingv1.HTTPIngressPath{}, "PathType"),
 }
 
+var httpRouteDiffOpts = cmp.Options{
+	cmpopts.IgnoreFields(gwapiv1.HTTPRoute{}, "TypeMeta", "ObjectMeta", "Status"),
+	cmpopts.IgnoreFields(gwapiv1.BackendRef{}, "Weight"),
+	cmpopts.IgnoreFields(gwapiv1.BackendObjectReference{}, "Group", "Kind", "Namespace"),
+	cmpopts.IgnoreFields(gwapiv1.ParentReference{}, "Group", "Kind", "Namespace"),
+}
```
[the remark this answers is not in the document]
btjd: Agreed, Namespace on ParentReference is meaningful since it comes from gatewayRef.namespace. Removed it from the ignore list so config changes propagate to existing HTTPRoutes.

controllers/controller/devworkspacerouting/solvers/solver.go
btjd: Good suggestion, added an IsGatewayAPIInstalled() check before returning the Gateway API solver, matching the IsOpenShift() pattern used for cluster-tls/web-terminal. Users get a clear error upfront instead of failing during sync.

controllers/controller/devworkspacerouting/solvers/gateway_api_solver.go
btjd: Good catch, reordered to match the basic solver: user annotations first, then controller annotation. This prevents user-provided annotations from overwriting controller.devfile.io/endpoint-name.

pkg/config/sync.go
btjd: Done, added GatewayRef name and namespace to the config string output.

main.go
```diff
@@ -79,6 +80,7 @@ func init() {
 	utilruntime.Must(controllerv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(dwv1.AddToScheme(scheme))
 	utilruntime.Must(dwv2.AddToScheme(scheme))
+	utilruntime.Must(gwapiv1.AddToScheme(scheme))
```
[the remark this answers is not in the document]
btjd: Fixed, switched to gwapiv1.Install in both main.go and suite_test.go.

controllers/controller/devworkspacerouting/devworkspacerouting_controller.go
btjd: You're right, the Owns(&gwapiv1.HTTPRoute{}) watch already triggers reconciliation when HTTPRoute status changes, so the explicit requeue is redundant. Removed the RequeueAfter while keeping the status update so the workspace still shows a "Waiting for endpoints" message during provisioning.

controllers/controller/devworkspacerouting/sync_httproutes.go
btjd: The logging was unintentional, I'll align with the other sync files and remove the log.Error call, just setting the flag and continuing silently.

apis/controller/v1alpha1/devworkspaceoperatorconfig_types.go
btjd: Good point, updated the comment to remove the Che reference. Thanks for the context.

controllers/controller/devworkspacerouting/devworkspacerouting_controller_test.go
btjd: Good idea, added an assertion on Status.ExposedEndpoints to verify the URL construction logic.

controllers/controller/devworkspacerouting/devworkspacerouting_controller_test.go
btjd: Agreed, added hostname assertions to verify the ClusterHostSuffix is applied correctly.

controllers/controller/devworkspacerouting/devworkspacerouting_controller_test.go
btjd: Added gateway-api specific failure test cases for parity with the basic solver.

Issue #1674, opened 2026-07-17
Set the default imagePullPolicy for projectClone to ifNotPresent


dkwon17:
### Description

Consider changing the default imagePullPolicy for project clone container to ifNotPresent: https://github.com/devfile/devworkspace-operator/blob/5413a9806e41e212483f6a235c12a8d80cc579cb/pkg/config/defaults.go#L39

### Additional context

When the imageRegistry for the project clone container is unavailable, this causes an error on workspace startup:
```
Failed to pull image "[registry.redhat.io/devworkspace/devworkspace-project-clone-rhel9@sha256:c3c279ce6e0c85f983ec5ece1b9c9e5a7773dd25611f0c15fe2849ce0666deb6](http://registry.redhat.io/devworkspace/devworkspace-project-clone-rhel9@sha256:c3c279ce6e0c85f983ec5ece1b9c9e5a7773dd25611f0c15fe2849ce0666deb6)": unable to pull image or OCI artifact: pull image err: initializing source docker://registry.redhat.io/devworkspace/devworkspace-project-clone-rhel9@sha256:c3c279ce6e0c85f983ec5ece1b9c9e5a7773dd25611f0c15fe2849ce0666deb6: Requesting bearer token: received unexpected HTTP status: 503 Service Unavailable; artifact err: get manifest: build image source: Requesting bearer token: received unexpected HTTP status: 503 Service Unavailable
```

ibuziuk:
@dkwon17 by any chance do we have granularity for defaults? e.g. set IfNotPresent only for init containers. 
I think that "Always" for workspace / devfile images makes sense (e.g. user updates the image but still wants to use the same tag in devfile). However, for DWO internals (init containers / traefik) it makes sense to default to "ifNotPresent"

ibuziuk:
Related JIRA that we need to plan for the next sprint - https://redhat.atlassian.net/browse/CRW-10867

PR #1692, opened 2026-08-12
fix: default init container imagePullPolicy to IfNotPresent

dkwon17:
Assisted-by: Claude Opus 4.6

### What does this PR do?
Changes the default `imagePullPolicy` for DWO-managed init containers (project-clone, and DWOC-defined init containers) from `Always` to `IfNotPresent`. The global `workspace.imagePullPolicy` for user/devfile containers remains Always.

Admins can still override via DWOC (`projectClone.imagePullPolicy`, or per init container).

https://github.com/devfile/devworkspace-operator/issues/1674

### Is it tested? How?

Tested with this test plan: https://gist.github.com/dkwon17/9f21c98fae6e9eeecf985cf7c818eaf4

### PR Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Init containers now default to the `IfNotPresent` image pull policy when none is specified, including containers defined in workspace configuration.
  - Explicit image pull policies remain unchanged.
  - Workspace containers continue using the `Always` policy.
  - Project-clone containers now use `IfNotPresent` by default.

- **Documentation**
  - Clarified image pull policy defaults for project-clone and restore containers in configuration references and deployment schemas.

- **Tests**
  - Added coverage for default, configured, and merged image pull policies, including configuration merge scenarios.

### DIFF
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -424,6 +424,9 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 					continue
 				}
 			}
+			if container.ImagePullPolicy == "" {
+				container.ImagePullPolicy = corev1.PullIfNotPresent
+			}
 			patches = append(patches, container)
 		}
 
--- a/controllers/workspace/devworkspace_controller.go
+++ b/controllers/workspace/devworkspace_controller.go
@@ -424,9 +424,6 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 					continue
 				}
 			}
-			if container.ImagePullPolicy == "" {
-				container.ImagePullPolicy = corev1.PullIfNotPresent
-			}
 			patches = append(patches, container)
 		}
 
@@ -436,12 +433,15 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			return r.failWorkspace(workspace, fmt.Sprintf("Failed to merge init containers: %s", err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus), nil
 		}
 
-		// Ensure init-persistent-home container has correct fields after merge
+		// Post-merge fixups on init containers
 		for i := range merged {
 			if merged[i].Name == constants.HomeInitComponentName {
 				if err := home.EnsureHomeInitContainerFields(&merged[i]); err != nil {
 					return r.failWorkspace(workspace, fmt.Sprintf("Failed to configure %s container: %s", constants.HomeInitComponentName, err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus), nil
 				}
+			}
+			if merged[i].ImagePullPolicy == "" {
+				merged[i].ImagePullPolicy = corev1.PullIfNotPresent
 			}
 		}
 
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -57,6 +57,7 @@ var defaultConfig = &v1alpha1.OperatorConfiguration{
 		ContainerSecurityContext: nil, // Set per-platform in setDefaultContainerSecurityContext()
 		DefaultTemplate:          nil,
 		ProjectCloneConfig: &v1alpha1.ProjectCloneConfig{
+			ImagePullPolicy: corev1.PullIfNotPresent,
 			Resources: &corev1.ResourceRequirements{
 				Limits: corev1.ResourceList{
 					corev1.ResourceMemory: resource.MustParse("1Gi"),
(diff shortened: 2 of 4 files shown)


pkg/config/defaults.go
```diff
@@ -57,6 +57,7 @@ var defaultConfig = &v1alpha1.OperatorConfiguration{
 		ContainerSecurityContext: nil, // Set per-platform in setDefaultContainerSecurityContext()
 		DefaultTemplate:          nil,
 		ProjectCloneConfig: &v1alpha1.ProjectCloneConfig{
+			ImagePullPolicy: corev1.PullIfNotPresent,
```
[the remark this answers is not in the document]
rohanKanojia: `devworkspace_controller.go:396` : this else is unreachable after the defaults.go change. `workspace.Config` comes from  ResolveConfigForWorkspace → mergeConfig, which only overwrites
 non-empty values, so ProjectCloneConfig.ImagePullPolicy is never "".

rohanKanojia:
I tested all scenarios in test plan and can confirm they work as expected :heavy_check_mark: 

Claude pointed out this scenario, do you think it's valid?

## DWOC init container default overrides admin-configured imagePullPolicy during strategic merge

The new default at line 427 sets `ImagePullPolicy = IfNotPresent` on DWOC init containers **before** they enter `MergeInitContainers()`. Because `corev1.Container.ImagePullPolicy` has `json:"imagePullPolicy,omitempty"`, a previously-empty value was omitted from the strategic merge patch JSON, preserving whatever the base init container had. Now `IfNotPresent` is always serialized into the patch and overrides the base value when the containers share a name.

### Steps to reproduce

**Step 1.** Set `projectClone.imagePullPolicy` to `Always` in DWOC, and add a DWOC init container named `project-clone` (to inject env vars) **without** explicit imagePullPolicy:

```bash
oc patch devworkspaceoperatorconfig devworkspace-operator-config -n "$DWO_NS" \
  --type=merge -p '{
    "config":{
      "workspace":{
        "projectClone":{"imagePullPolicy":"Always"},
        "initContainers":[
          {"name":"project-clone","env":[{"name":"EXTRA_VAR","value":"injected"}]}
        ]
      }
    }
  }'
```

**Step 2.** Create a DevWorkspace with a project:

```yaml
cat <<'EOF' | oc apply -n "$TEST_NS" -f -
kind: DevWorkspace
metadata:
  name: test-merge-conflict
spec:
  started: true
  template:
    projects:
      - name: sample
        git:
          remotes:
    components:
      - name: dev
        container:
          memoryLimit: 512Mi
EOF
```

**Step 3.** Wait for the deployment to be created:

```bash
while ! oc get deployment -l controller.devfile.io/devworkspace_name=test-merge-conflict \
  -n "$TEST_NS" 2>/dev/null | grep -q .; do sleep 5; done
```

**Step 4.** Check the project-clone init container's imagePullPolicy:

```bash
oc get deployment -l controller.devfile.io/devworkspace_name=test-merge-conflict -n "$TEST_NS" \
  -o jsonpath='{.items[0].spec.template.spec.initContainers[?(@.name=="project-clone")].imagePullPolicy}'
```

- **Expected:** `Always` (admin explicitly set `projectClone.imagePullPolicy: Always` — the DWOC patch only adds env vars)
- **Actual:** `IfNotPresent` (DWOC patch gets defaulted to `IfNotPresent` at line 427, which overrides the base during strategic merge)

dkwon17:
Thank you @rohanKanojia , that's a valid point. My latest commit now sets the default imagepullpolicy after the `merged, err := initcontainers.MergeInitContainers(...` function call

controllers/workspace/devworkspace_controller.go
```diff
@@ -433,13 +433,16 @@ func (r *DevWorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-		// Ensure init-persistent-home container has correct fields after merge
+		// Post-merge fixups on init containers
 		for i := range merged {
 			if merged[i].Name == constants.HomeInitComponentName {
 				if err := home.EnsureHomeInitContainerFields(&merged[i]); err != nil {
 					return r.failWorkspace(workspace, fmt.Sprintf("Failed to configure %s container: %s", constants.HomeInitComponentName, err), metrics.ReasonBadRequest, reqLogger, &reconcileStatus), nil
 				}
 			}
+			if merged[i].ImagePullPolicy == "" {
```
btjd, line 443: Based on my understanding, the linked issue (#1674) specifically targets the project-clone container, but this post-merge fixup applies IfNotPresent to all init containers with an empty policy, including the restore container and any future init containers. Was this intentional broadening, or should this be scoped to just project-clone (and DWOC-defined init containers without an explicit policy)?
If the broader scope is intentional, it might be worth noting that in the PR description, and the RestoreConfig.ImagePullPolicy doc comment would need the same update as ProjectCloneConfig.

dkwon17: Yes, it was intentional for the default to apply to all init container to prevent workspace startup failures if for example the image registry is down. In my latest commit, I've updated the imagePullPolicy fields' documentation comments to mention this.

PR #1695, opened 2026-08-20
Retrieve TLS config from apiserver for OpenShift

dkwon17:
### What does this PR do?
* Retrieve TLS config from api server for Openshift (no changes for Kubernetes).
  * Example PR: https://github.com/openshift/cluster-machine-approver/pull/286/changes#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261.
* `openshift/api` dependency bump was needed to use the `configv1.TLSAdherencePolicy` field 

### Is it tested? How?

To test this PR, I used this catalog source to install DWO on an OCP 5.0 cluster:
```
kind: CatalogSource
metadata:
  name: dwo-tls-adherence
spec:
  sourceType: grpc
  image: quay.io/dkwon17/devworkspace-operator-index:tls-adherence
  displayName: "DWO TLS Adherence"
  publisher: "Test"
  updateStrategy:
    registryPoll:
      interval: 15m
```
For verification, I followed these testing steps: https://gist.github.com/dkwon17/92211bd5cf6a7a101d1e27c3179acb7a

### PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added OpenShift-aware TLS configuration for metrics and webhook servers.
  * Servers now gracefully restart when relevant TLS security profiles or policies change.
  * Applied cluster TLS profiles while preserving safe defaults on other platforms.

* **Bug Fixes**
  * Updated permissions to support cluster-wide API server discovery while restricting detailed access appropriately.
  * Improved handling of TLS configuration errors during startup.

* **Tests**
  * Added coverage for TLS profiles, adherence policies, platform differences, restart behavior, and configuration edge cases.


TLS_ADHERENCE_TEST_PLAN.md
```diff
@@ -0,0 +1,139 @@
+# TLS Adherence Feature Test Plan
```
rohanKanojia: Shall we use test plan in https://gist.github.com/dkwon17/92211bd5cf6a7a101d1e27c3179acb7a or this one?

pkg/tlssetup/server_tls.go
```diff
@@ -0,0 +1,141 @@
...
+
+// RegisterSecurityProfileWatcher watches the APIServer TLS profile and adherence policy.
+// Calls onCancel to trigger restart when either changes. No-op on non-OpenShift.
+func RegisterSecurityProfileWatcher(mgr manager.Manager, serverTLS ServerTLS, onCancel context.CancelFunc, log logr.Logger) error {
+	if !infrastructure.IsOpenShift() {
+		return nil
+	}
+
+	// Only set up the watcher if we successfully fetched the initial profile
+	if len(serverTLS.TLSOpts) == 0 {
```
rohanKanojia: `TLSOpts` ends up empty in two cases: when a fetch failed, and when the fetch succeeded but the policy was "" / `LegacyAdheringComponentsOnly`. Skipping the watcher makes sense in the first case since there's no valid baseline to diff against. In the second, though, I think we'd still want to watch, so we'd catch an admin later setting `StrictAllComponents` ; otherwise that change wouldn't take effect until the pod happens to restart.

Is that scenario already handled somewhere upstream that I'm missing?

controllers/workspace/devworkspace_controller.go
```diff
@@ -96,6 +96,7 @@ type DevWorkspaceReconciler struct {
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=oauth.openshift.io,resources=oauthclients,verbs=get;list;watch;create;update;patch;delete;deletecollection
+// +kubebuilder:rbac:groups=config.openshift.io,resources=apiservers,verbs=get;list;watch
```
rohanKanojia: Since only the singleton cluster APIServer is ever read, consider scoping with `resourceNames=cluster` (similar to the existing proxies marker on line 101) to follow least-privilege

pkg/tlssetup/server_tls.go
```diff
@@ -0,0 +1,141 @@
...
+		return false
+	default:
+		// StrictAllComponents or unknown future value → honor the profile
+		return true
+	}
+}
+
+// BuildServerTLSOptions fetches TLS settings from the OpenShift API server.
+// Only applies the cluster profile when the tlsAdherence policy requires it.
+func BuildServerTLSOptions(ctx context.Context, cfg *rest.Config, scheme *k8sruntime.Scheme, log logr.Logger) (ServerTLS, error) {
```
rohanKanojia: `BuildServerTLSOptions()` currently logs errors and falls back to Go defaults, but always returns `nil` error. Could we either propagate these errors or remove the error return value?

webhook/main.go
```diff
@@ -89,6 +88,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	serverTLS, err := tlssetup.BuildServerTLSOptions(
```
rohanKanojia, line 94: Could we ensure `configv1.AddToScheme(scheme)` is registered in the webhook entrypoint as well?

pkg/tlssetup/server_tls_test.go
```diff
@@ -0,0 +1,88 @@
...
+	defer infrastructure.InitializeForTesting(infrastructure.Kubernetes)
+
+	log := zap.New(zap.UseDevMode(true))
+
+	// When TLSOpts is empty (profile not applied), should skip watcher setup and return nil
+	err := RegisterSecurityProfileWatcher(nil, ServerTLS{}, nil, log)
+	if err != nil {
+		t.Errorf("RegisterSecurityProfileWatcher() with empty TLSOpts should skip setup, got error = %v", err)
+	}
+}
```
rohanKanojia, line 91: Could we also add tests for `BuildServerTLSOptions` covering the main OpenShift paths?

controllers/workspace/devworkspace_controller.go
rohanKanojia: the OLM bundle CSV (`deploy/bundle/manifests/devworkspace-operator.clusterserviceversion.yaml`) wasn't regenerated, so it still grants apiservers only verbs: [`get`] (grouped with proxies).

Could you please run `make generate_all` again?

TLS_ADHERENCE_TEST_PLAN.md
dkwon17: Oops, I did not mean to commit this, I will remove it. I would follow what's mentioned in https://gist.github.com/dkwon17/92211bd5cf6a7a101d1e27c3179acb7a

pkg/tlssetup/server_tls.go
rohanKanojia: Is manual oc delete pod in Tests 2 and 5 intended? Or is this a workaround for `RegisterSecurityProfileWatcher` skipping setup when TLSOpts is empty? Asking because it means the watcher can detect leaving strict adherence but not entering it, so an admin turning this on doesn't get it until someone restarts the operator.

controllers/workspace/devworkspace_controller.go
dkwon17: Good catch, fixed both issues in my latest commits.

pkg/tlssetup/server_tls.go
dkwon17: I opted to remove the return value and fallback on go defaults.

pkg/tlssetup/server_tls.go
dkwon17: With the latest changes, the watcher is not skipped anymore when the policy is an empty string or `LegacyAdheringComponentsOnly`.

The watcher is still skipped if the initial profile fetch fails however, which acts as a fallback to how DWO functioned before this PR

pkg/tlssetup/server_tls_test.go
dkwon17: Sounds good, updated

webhook/main.go
dkwon17: Updated in my new commits

pkg/tlssetup/server_tls_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	log := zap.New(zap.UseDevMode(true))
+
+	err := RegisterSecurityProfileWatcher(nil, ServerTLS{}, nil, log)
+	if err != nil {
+		t.Errorf("RegisterSecurityProfileWatcher() on Kubernetes should be no-op, got error = %v", err)
+	}
+}
+
+func TestRegisterSecurityProfileWatcher_NoTLSOpts(t *testing.T) {
```
rohanKanojia:
```suggestion
func TestRegisterSecurityProfileWatcher_ProfileNotFetched(t *testing.T) {
```

rohanKanojia:
I tested provided test plan and can confirm it works as expected ✅

rohanKanojia (approved):

pkg/tlssetup/server_tls_test.go
dkwon17: Updated in my latest commit

btjd (approved):
In webhook/main.go, the old signal handling via os/signal and syscall.SIGTERM is removed and replaced with the cancellable context pattern. The old code created a shutdownChan that was never actually wired to stop the manager, so this looks like a bug fix on top of the TLS work. Do you think that this should be called out explicitly in the PR description?
Other than that it looks good to me.

PR #1699, opened 2026-08-26
fix: sanitize automount volume names to be DNS-1123 compliant

oskutka:
## What does this PR do?

Auto-mounted Secrets, ConfigMaps, and PVCs derive their pod volume name from the object's name. Kubernetes object names and volume names have different validation rules: a name that is legal for a Secret (e.g. `test.pullsecret`) is invalid as a volume name, which must be a DNS-1123 label. Previously the object name was used verbatim, so mounting such an object produced an invalid Deployment and the workspace failed to start.

This sanitizes the derived volume name (lowercase, replace invalid characters with `-`, trim edges, truncate to 63 chars) via a shared helper in `pkg/common/naming.go`, used by `AutoMountSecretVolumeName`, `AutoMountConfigMapVolumeName`, and `AutoMountPVCVolumeName`. The volume still references the underlying object by its original, unmodified name.

**Known trade-off:** sanitization is not injective — e.g. test.pullsecret and test-pullsecret both map to test-pullsecret. Before this change, such a collision was not caught: checkAutomountVolumesForCollision only detected DevWorkspace-vs-automount name collisions and mount-path collisions, so two automounted objects resolving to the same name silently produced a pod spec with duplicate volume names (rejected later by the API server). This PR extends that check to detect automount-vs-automount name collisions and fail with a clear error naming both source objects. The collision case is rare; see the included ADR for details.

## What issues does this PR fix or reference?

[CRW-9800](https://redhat.atlassian.net/browse/CRW-9800)

## Is it tested? How?

- **Unit** — `pkg/common/naming_test.go` covers the sanitization helper (dots, mixed invalid chars, trimming, 63-char truncation). `pkg/provision/automount/common_test.go` covers the new automount-vs-automount volume name collision.
- **Integration** — a fixture in the existing automount test framework (`pkg/provision/automount/testdata/testSanitizesInvalidVolumeNames.yaml`) verifies a `test.pullsecret` secret and `test.configmap` configmap produce sanitized volume names while referencing the original objects; `testdata/errorDuplicateVolumeNameAfterSanitization.yaml` verifies that two objects resolving to the same sanitized volume name fail with a clear error.
- **E2E** — `test/e2e/pkg/tests/automount_volume_sanitization_tests.go` mounts a secret named `test.pullsecret` into a running workspace and asserts the sanitized volume name, mount, and readable data. Verified on a live cluster: 6/6 specs passing.
- An ADR is included documenting the sanitize-vs-reject design decision and the volume-name collision trade-off.

## Bug Fixes
- Auto-mounted Secret, ConfigMap, and PVC volumes now use valid DNS-1123-compliant names.
- Names are normalized by lowercasing, replacing invalid characters, trimming separators, and limiting length.
- Original resource references remain unchanged, ensuring resources continue to mount correctly.
- Clear errors are reported when sanitization causes duplicate volume names.

## Tests
- Added coverage for sanitization, collision handling, and successful end-to-end mounting.

## Documentation
- Documented volume-name sanitization rules, collision handling, and expected behavior.


pkg/common/naming.go
```diff
@@ -116,19 +116,38 @@ func MetadataConfigMapName(workspaceId string) string {
...
+func sanitizeVolumeName(name string) string {
+	// Convert to lowercase
+	sanitized := strings.ToLower(name)
+	// Replace any non-alphanumeric characters (including dots) with hyphens
+	sanitized = NonAlphaNumRegexp.ReplaceAllString(sanitized, "-")
+	// Trim leading and trailing hyphens
+	sanitized = strings.Trim(sanitized, "-")
+	// Ensure length is at most 63 characters (DNS-1123 label limit)
+	if len(sanitized) > 63 {
+		sanitized = strings.TrimSuffix(sanitized[:63], "-")
```
rohanKanojia: Nit: the four inline comments describe what each line does, which the code already says clearly.  We could drop the inline ones.
```suggestion
	sanitized := strings.ToLower(name)
	sanitized = NonAlphaNumRegexp.ReplaceAllString(sanitized, "-")
	sanitized = strings.Trim(sanitized, "-")
	if len(sanitized) > 63 {
		sanitized = strings.TrimSuffix(sanitized[:63], "-")
```

adr/2026-08-26-sanitize-automount-volume-names.md
```diff
@@ -0,0 +1,92 @@
...
+   previous reliance on never adding characters to the name.
+
+### Negative
+
+1. Sanitization is not injective: two distinct object names can map to the same
+   volume name (e.g. `test.pullsecret` and `test-pullsecret`). This is an
+   accepted trade-off — such a collision is caught by the existing
+   `checkAutoMountVolumesForCollision` check, which surfaces a clear error rather
+   than producing a silently broken spec. Previously these names were distinct;
+   the collision case is new but rare and fails loudly.
```
rohanKanojia: Optional: It would be nice to add a test to confirm that collision surfaces a clear error.

rohanKanojia (approved):

pkg/common/naming.go
```diff
@@ -116,19 +116,38 @@ func MetadataConfigMapName(workspaceId string) string {
...
+}
+
+// sanitizeVolumeName converts a name to be DNS-1123 label compliant for use as a Kubernetes volume name.
+// Volume names must:
+// - contain at most 63 characters
+// - contain only lowercase alphanumeric characters or '-'
+// - start with an alphanumeric character
+// - end with an alphanumeric character
+//
+// Generated by Claude
```
rohanKanojia, line 138: This would probably show up in function's godoc 
```suggestion
```

oskutka: I thought this comment was necessary as per https://github.com/devfile/devworkspace-operator/blob/main/redhat-compliance-and-responsible-ai.md section 2...

rohanKanojia: Thanks for sharing!

adr/2026-08-26-sanitize-automount-volume-names.md
oskutka: Good point. I'll add the test.

oskutka: Great intuition here, @rohanKanojia! Writing the test revealed that the collision wasn't actually caught: before adding detection, two secrets test.pullsecret and test-pullsecret (both sanitizing to test-pullsecret) silently produced a pod spec with duplicate volume names and no error. The existing checkAutomountVolumesForCollision only handled DevWorkspace-vs-automount and mount-path collisions. So the test both surfaced the gap and now guards the fix. Thanks for pushing on this!

pkg/common/naming.go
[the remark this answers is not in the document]
dkwon17: Correct me if I'm wrong, but this is not necessary because:
```
sanitized = NonAlphaNumRegexp.ReplaceAllString(sanitized, "-")
```

in line 141 would eliminate consecutive interior hyphens anyways?

dkwon17:
The e2e test failure seems to be unrelated to the current code changes:
```
{  /logs/artifacts/.openshift_install-1787757617.log:time="2026-08-26T15:20:17Z" level=error msg="Error: creating EC2 VPC Endpoint (com.amazonaws.us-east-1.s3): VpcEndpointLimitExceeded: The maximum number of VPC endpoints has been reached."
```

pkg/common/naming.go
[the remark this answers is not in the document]
oskutka: @dkwon17 is right. The regex is `NonAlphaNumRegexp = regexp.MustCompile([^a-z0-9]+)`. It collapses any run of **non-alpha-numeric** (including a dash) characters into a single `-`, so after ReplaceAllString the string can never contain two or more consecutive hyphens.

Should I add more unit tests to prove the correctness?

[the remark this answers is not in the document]
oskutka: I added the tests in dfdfdda.

